PR #5, opened 2015-11-24
refactor rcl into a functional library

wjwwood:
Currently `rcl` is just an idea and not being used. This pull request will refactor it into a working library which can be used by client libraries and which is a super set of the `rmw` functionality. The scope of this pull request is just for initialization, nodes, publishers, subscriptions, guard conditions, and waiting. This will allow prototyping of the Python API implementation while we work on services (client and server), execution logic, and parameters separately in a follow up pull request.

This is a work in progress still, I need to:

- [x] redesign the headers; removing functions for things out scope (services, etc...)
 - [x] I need to go over all the functions and think about alternative return code (not just ok and error)
 - [x] I need to make sure the use of `[in]`, `[out]`, and `[inout]` is consistent
 - [x] I need to consider `const`-ness of parameters and return types.
 - [x] I need to address as many of the TODO's in the code as possible (maybe after feedback)
 - [x] I need to add macros for visibility.
 - [x] I need to add warnings about not checking return codes.
- [x] implement the headers in a shared library ~~which is rmw implementation agnostic~~
- [x] implement tests which use the rcl functions:
 - [x] allocator
 - [x] guard_condition
 - [x] node
 - [x] ~~publisher~~
 - [x] rcl
 - [x] ~~subscription~~
 - [x] time
 - [x] ~~timer~~
 - [x] ~~wait~~
 - Some tests will be deferred to separate pr's.

And then obviously iterate on feedback from review.

---

I'm using C11 atomics for their semantic value, on unsupported OS's there are other options:

- https://gist.github.com/nhatminhle/5181506 (for clang and gcc versions that don't implement)
- https://github.com/mintomic/mintomic (potential general solution when C11 atomics are not available)


wjwwood:
I'm going to try and push the implementation of the headers and the corresponding tests tonight, but they are currently broken, but I figured people could go ahead and start looking at the functions and their documentation to make comments while I work on that.

rcl/CMakeLists.txt
```diff
@@ -3,11 +3,36 @@ cmake_minimum_required(VERSION 2.8.3)
 project(rcl)
 
 find_package(ament_cmake REQUIRED)
+find_package(rcl_interfaces REQUIRED)
+find_package(rmw REQUIRED)
+find_package(rosidl_generator_cpp REQUIRED)
```
jacquelinekay: I assume this will become rosidl_generator_cpp once Dirk finishes C typesupport?

wjwwood: Good catch, I can do it now. 483ddfd

jacquelinekay:
Looking at the Allocator interface here, I'm considering of changing the user allocator interface in rclcpp to work in a similar way.

That is, the user provides allocate and deallocate functions that could be bound to some state, then rclcpp constructs an allocator compliant with the C++ stdlib requirements behind the scenes for when it needs to pass the allocator to stdlib structures.

Then rclcpp could pass the allocate and deallocate functions from the user directly to rcl.

wjwwood:
That would probably work fine. The downside of the C approach is that you loose type safety (state object is `void *`). When the allocator class (which the user defined) is a template argument then the type safety can be preserved. Also you have to consider how to declare STL containers which are templated on an allocator as members of classes which take allocators in the rclcpp interface.

rcl/include/rcl/node.h
```diff
@@ -0,0 +1,228 @@
...
+  // bool anonymous_name;
+  // rmw_qos_profile_t parameter_qos;
+  /// If true, no parameter infrastructure will be setup.
+  bool no_parameters;
+  /// If set, then this value overrides the ROS_DOMAIN_ID environment variable.
+  /* It defaults to RCL_NODE_OPTIONS_DEFAULT_DOMAIN_ID, which will cause the
+   * node to use the ROS domain ID set in the ROS_DOMAIN_ID environment
+   * variable, or on some systems 0 if the environment variable is not set.
+   *
+   * \TODO(wjwwood): Should we put a limit on the ROS_DOMAIN_ID value, that way
```
jacquelinekay: What are the limits on domain ID enforced by Opensplice and Connext? If this is something that varies by middleware implementation, I don't think it needs to be enforced at the rcl layer.

rcl/include/rcl/publisher.h
```diff
@@ -0,0 +1,266 @@
...
+ * \param[inout] publisher preallocated publisher structure
+ * \param[in] node valid rcl node handle
+ * \param[in] type_support type support object for the topic's type
+ * \param[in] topic_name the name of the topic to publish on
+ * \param[in] options publisher options, including quality of service settings
+ * \return RCL_RET_OK if the publisher was initialized successfully, or
+ *         RCL_RET_NODE_INVALID if the node is invalid, or
+ *         RCL_RET_ALREADY_INIT if the publisher is already initialized, or
+ *         RCL_RET_INVALID_ARGUMENT if any arugments are invalid, or
+ *         RCL_RET_BAD_ALLOC if any arugments are invalid, or
```
jacquelinekay: The comment here looks wrong (copy pasted from above).

rcl/include/rcl/publisher.h
```diff
@@ -0,0 +1,266 @@
...
+ * This function can fail, and therefore return NULL, if the:
+ *   - publisher is NULL
+ *   - publisher is invalid (never called init, called fini, or invalid node)
+ *
+ * The returned string is only valid as long as the rcl_publisher_t is valid.
+ * The value of the string may change if the topic name changes, and therefore
+ * copying the string is recommended if this is a concern.
+ *
+ * This function is not thread-safe, and copying the result is not thread-safe.
+ * \TODO(wjwwood): should we provide a thread-safe copy_topic_name?
```
jacquelinekay: On the topic of thread safety, I would say that wrapping this in synchronization is the responsibility of the language client library, because of how synchronization varies between libraries.

rcl/include/rcl/subscription.h
```diff
@@ -0,0 +1,266 @@
...
+ * messages of the given type to the given topic using rcl_take().
+ *
+ * The given rcl_node_t must be valid and the resulting rcl_subscription_t is
+ * only valid as long as the given rcl_node_t remains valid.
+ *
+ * The rosidl_message_type_support_t is obtained on a per .msg type basis.
+ * When the user defines a ROS message, code is generated which provides the
+ * required rosidl_message_type_support_t object.
+ * This object can be obtained using a language appropriate mechanism.
+ * \TODO(wjwwood) probably should talk about this once and link to it instead
```
jacquelinekay: agreed; you can use Doxygen \sa to reference a different function and file

https://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdsa

rcl/include/rcl/wait.h
```diff
@@ -0,0 +1,306 @@
...
+
+#include "rcl/subscription.h"
+#include "rcl/guard_condition.h"
+#include "rcl/time.h"
+#include "rcl/types.h"
+
+struct rcl_wait_set_impl_t;
+
+/// Container for subscription's, guard condition's, etc to be waited on.
+typedef struct rcl_wait_set_t {
```
jacquelinekay: Do you think a `max_conditions` or `max_entities` value would be appropriate here, given the changes I'm making in the `waitset_handle` PRs (ros2/rmw#49 and others)?

rcl/include/rcl/wait.h
```diff
@@ -0,0 +1,306 @@
...
+  size_t __current_subscription_offset;
+  /// Storage for guard condition pointers.
+  const rcl_guard_condition_t ** guard_conditions;
+  size_t size_of_guard_conditions;
+  size_t __current_guard_condition_offset;
+  /// Allocator for storage.
+  rcl_allocator_t allocator;
+  /// If set to true, actions like add_subscription will fail until cleared.
+  bool pruned;
+  /// Implementation specific storage.
```
jacquelinekay: Does implementation specific storage here and elsewhere refer to implementation layer above or below rcl (language or middleware)?

wjwwood: It is only forward declared, so it cannot be used from above (the calling code does not know what its size is or what members it has. It is used in the implementation of `rcl`, and so does not have to have anything in it from "below" in `rmw`, but it could. This is simply the pointer to implementation pattern (similar to: https://en.wikipedia.org/wiki/Opaque_pointer) which I'm using here to help with ABI compatibility. Basically it serves as the C equivalent to the private section of a class.

jacquelinekay: Got it, I reading the .c files now and this makes more sense.

rcl/include/rcl/wait.h
wjwwood: Maybe, though I think it would be passed as an option to the `rcl_wait_set_init` function and the result would be hidden behind an accessor and stored in the `impl` pointer.

jacquelinekay: That sounds fine with me. It should be an immutable value once specified,
so the method you described works as the way to enforce it.

rcl/include/rcl/publisher.h
wjwwood: Thanks, I fixed this and few other typos: 77406de

wjwwood:
@ernestmc

rcl/src/rcl/stdatomic_helper.h
```diff
@@ -0,0 +1,111 @@
...
+
+#if defined(__GNUC__) && __GNUC__ <= 4 && __GNUC_MINOR__ <= 9
+// If GCC and below GCC-4.9, use the compatability header.
+#include "stdatomic_helper/gcc/stdatomic.h"
+#elif defined(__clang__) && defined(__has_feature) && !__has_feature(c_atomic)
+// If Clang and no c_atomics (true for some older versions), use the compatability header.
+#include "stdatomic_helper/gcc/stdatomic.h"
+#else
+#include <stdatomic.h>
+#endif
```
jacquelinekay: Can you add an inline comment to the endifs in this file (and in other places in this library) to denote which condition it's ending? It makes the macros a lot easier to read.

rcl/src/rcl/stdatomic_helper/win32/stdatomic.h
```diff
@@ -0,0 +1,361 @@
...
+
+/*
+ * 7.17.8 Atomic flag type and operations. (disabled for now)
+ */
+
+// typedef atomic_bool     atomic_flag;
+
+// #define ATOMIC_FLAG_INIT    ATOMIC_VAR_INIT(0)
+
+// #define atomic_flag_clear_explicit(object, order) \
```
jacquelinekay: I know you didn't write this, but why are these functions commented out in the Windows implementation?

wjwwood: Well I sort of wrote this (I used the FreeBSD header as a template, but I replaced the guts). They are commented out because I didn't need them :smile:. I can remove them completely if you like, but I figured I might uncomment them and update them for Windows at some point so I didn't yet.

jacquelinekay: No strong feelings here.

rcl/src/rcl/stdatomic_helper.h
wjwwood: I usually omit them unless the endif and if are far apart or part of a nested logic. I can go through and add them though. Always add them is a simpler rule to follow anyways, but I'll still avoid some trivial cases like wrapping `extern "C" {` and some boilerplate that is common like the visibility macros: aa83632

jacquelinekay: Yeah, I agree that they aren't needed everywhere. I think the case that really confuses me is nested #if statements. Thanks.

rcl/src/rcl/time_win32.c
```diff
@@ -0,0 +1,72 @@
...
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(now, RCL_RET_INVALID_ARGUMENT);
+  FILETIME ft;
+  GetSystemTimeAsFileTime(&ft);
+  ULARGE_INTEGER uli;
+  uli.LowPart = ft.dwLowDateTime;
+  uli.HighPart = ft.dwHighDateTime;
+  // Adjust for January 1st, 1970, see:
+  //   https://support.microsoft.com/en-us/kb/167296
+  uli.QuadPart -= 116444736000000000;
```
jacquelinekay: wow

rcl/src/rcl/timer.c
```diff
@@ -0,0 +1,239 @@
...
+  RCL_CHECK_ARGUMENT_FOR_NULL(timer, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(time_since_last_call, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_FOR_NULL_WITH_MSG(timer->impl, "timer is invalid", return RCL_RET_TIMER_INVALID);
+  rcl_steady_time_point_t now;
+  rcl_ret_t ret = rcl_steady_time_point_now(&now);
+  if (ret != RCL_RET_OK) {
+    return ret;  // rcl error state should already be set.
+  }
+  *time_since_last_call =
+    (now.nanoseconds - rcl_atomic_load_uint64_t(&timer->impl->last_call_time));
```
jacquelinekay: Style nitpick: why is this parenthesized?

rcl/src/rcl/timer.c
```diff
@@ -0,0 +1,239 @@
...
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(timer, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(period, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_FOR_NULL_WITH_MSG(timer->impl, "timer is invalid", return RCL_RET_TIMER_INVALID);
+  *period = rcl_atomic_load_uint64_t(&timer->impl->period);
+  return RCL_RET_OK;
+}
+
+rcl_ret_t
+rcl_timer_exchange_period(const rcl_timer_t * timer, uint64_t new_period, uint64_t * old_period)
```
jacquelinekay: It should be noted somewhere--and this is obvious to someone who is familiar with atomics, but maybe not to the average user--that this function and ones like it give the appearance of atomicity, but are in fact not completely safe from races. For example, if a context switch between lines 180 and 181 and the `timer->impl` became null, then the next line would segfault.

rcl/src/rcl/wait.c
```diff
@@ -0,0 +1,448 @@
...
+{
+  rcl_ret_t fail_ret = RCL_RET_ERROR;
+  RCL_CHECK_ARGUMENT_FOR_NULL(wait_set, RCL_RET_INVALID_ARGUMENT);
+  if (__wait_set_is_valid(wait_set)) {
+    RCL_SET_ERROR_MSG("wait_set already initialized, or memory was uninitialized.");
+    return RCL_RET_ALREADY_INIT;
+  }
+  RCL_CHECK_FOR_NULL_WITH_MSG(
+    number_of_subscriptions,
+    "number_of_subscriptions cannot be 0", return RCL_RET_INVALID_ARGUMENT);
```
jacquelinekay: Why can't the number of subscriptions be zero? We have examples of code without any subscriptions.

rcl/src/rcl/timer.c
wjwwood: They are atomic, just not thread-safe with the destructor. We need a better way to describe which functions are thread safe with which other functions. For now I've only mentioned that in one place: https://github.com/ros2/rcl/pull/5/files#diff-ebe52df03e12f75a7709b16908335c3aR127

Unfortunately I don't think we can do anything but document it better (maybe even to the point of redundancy).

wjwwood: I should say, the access to the information is atomic (but not the function). You are right about that.

rcl/src/rcl/wait.c
```diff
@@ -0,0 +1,448 @@
...
+}
+
+#define SET_ADD(Type) \
+  RCL_CHECK_ARGUMENT_FOR_NULL(wait_set, RCL_RET_INVALID_ARGUMENT); \
+  RCL_CHECK_ARGUMENT_FOR_NULL(Type, RCL_RET_INVALID_ARGUMENT); \
+  if (!__wait_set_is_valid(wait_set)) { \
+    RCL_SET_ERROR_MSG("wait set is invalid"); \
+    return RCL_RET_WAIT_SET_INVALID; \
+  } \
+  if (!(wait_set->impl->Type ## _index < wait_set->size_of_ ## Type ## s)) { \
```
jacquelinekay: what is going on here. what are these double octothorpes.

jacquelinekay: http://stackoverflow.com/questions/22975073/what-does-double-hash-do-in-a-preprocessor-directive

rcl/src/rcl/timer.c
wjwwood: But to be fair, I don't mention that this function is atomic (in the docstring), and so it seems that your familiarity with atomics may have lead you to assume this function was atomic, not the other way around. The word exchange in the name means exactly that, you give me a new value and I'll give the old, but doesn't mean to imply atomic behavior, though I understand what you mean that it invokes that idea based on the name.

wjwwood: Another option is that we can avoid error checking here and call directly through to the atomic function call, but that might not be thread-safe either.

jacquelinekay: I think the implementation is the best we can do for now, we just need to make a note of it in case it cases insane race conditions in the future.

rcl/src/rcl/wait.c
```diff
@@ -0,0 +1,448 @@
...
+rcl_wait_set_clear_timers(rcl_wait_set_t * wait_set)
+{
+  SET_CLEAR(timer)
+  return RCL_RET_OK;
+}
+
+rcl_ret_t
+rcl_wait_set_resize_timers(rcl_wait_set_t * wait_set, size_t size)
+{
+  SET_RESIZE(timer,;,;)  // NOLINT
```
jacquelinekay: what? why doesn't the size parameter get passed here?

rcl/CMakeLists.txt
```diff
@@ -3,14 +3,68 @@ cmake_minimum_required(VERSION 2.8.3)
...
+  ament_target_dependencies(${PROJECT_NAME}${target_suffix}
+    "rcl_interfaces"
+    "rmw"
+    "rosidl_generator_c"
+    "${rmw_implementation}"
+  )
+
+  # Causes the visibility macros to use dllexport rather than dllimport,
+  # which is appropriate when building the dll but not consuming it.
+  target_compile_definitions(${PROJECT_NAME}${target_suffix} PRIVATE "RCL_BUILDING_LIBRARY")
```
dirk-thomas: Why should we diverge from the name used in most other visibility files? Usually we call it `XXX_BUILDING_DLL`.

rcl/CMakeLists.txt
```diff
@@ -3,14 +3,68 @@ cmake_minimum_required(VERSION 2.8.3)
...
+find_package(rcl_interfaces REQUIRED)
+find_package(rmw REQUIRED)
+find_package(rmw_implementation REQUIRED)
+find_package(rmw_implementation_cmake REQUIRED)
+find_package(rosidl_generator_c REQUIRED)
 
+include_directories(include)
+
+if(NOT WIN32)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
```
dirk-thomas: Should we use C11 here?

rcl/CMakeLists.txt
```diff
@@ -3,14 +3,68 @@ cmake_minimum_required(VERSION 2.8.3)
...
 ament_export_dependencies(rmw)
+ament_export_dependencies(rmw_implementation)
+ament_export_dependencies(rosidl_generator_c)
 
 ament_export_include_directories(include)
 
+ament_export_libraries(${PROJECT_NAME})
+
 if(AMENT_ENABLE_TESTING)
+  find_package(ament_cmake_gtest REQUIRED)
```
dirk-thomas: Should we move this into the `test/CMakeLists.txt` file to get it closer to where its needed?

rcl/include/rcl/allocator.h
```diff
@@ -0,0 +1,62 @@
...
+  /* An error should be indicated by returning NULL. */
+  void * (*allocate)(size_t size, void * state);
+  /// Deallocate previously allocated memory, mimicking free().
+  void (* deallocate)(void * pointer, void * state);
+  /// Reallocate if possible, otherwise it deallocates and allocates.
+  /* If unsupported then do deallocate and then allocate.
+   * \TODO(wjwwood): should this behave as reallocf?
+   * This should behave as realloc is described, as opposed to reallocf, i.e.
+   * the memory given by pointer will not be free'd automatically if realloc
+   * fails.
```
dirk-thomas: I would agree to mimic `realloc` behavior. Deleting the passed memory when the allocation fails seems like a bad idea.

rcl/include/rcl/guard_condition.h
```diff
@@ -0,0 +1,161 @@
...
+typedef struct rcl_guard_condition_options_t
+{
+  /// Custom allocator for the guard condition, used for internal allocations.
+  rcl_allocator_t allocator;
+} rcl_guard_condition_options_t;
+
+/// Return a rcl_guard_condition_t struct with members set to NULL.
+RCL_PUBLIC
+rcl_guard_condition_t
+rcl_get_zero_initialized_guard_condition();
```
dirk-thomas: Does the function name need to include `get_zero_initialized`? In rosidl we used `create_` for this (vs. `alloc_` and `init_`).

rcl/include/rcl/guard_condition.h
```diff
@@ -0,0 +1,161 @@
...
+ *    ret = rcl_guard_condition_fini(&guard_condition, &node);
+ *    // ... error handling for rcl_guard_condition_fini()
+ *    ret = rcl_node_fini(&node);
+ *    // ... error handling for rcl_deinitialize_node()
+ *
+ * This function does allocate heap memory.
+ * This function is not thread-safe.
+ * This function is lock-free.
+ *
+ * \TODO(wjwwood): does this function need a node to be passed to it? (same for fini)
```
dirk-thomas: Currently `rmw_create_guard_condition` doesn't get any context passed. I think this should change in the future since we can't assume that every rmw impl. can perform (any) operation without context. Therefore I would argue to keep passing the node here (and sometime later do the same in the rmw interface).

rcl/test/memory_tools_common.cpp
```diff
@@ -0,0 +1,159 @@
...
+  enabled.store(false);
+  if (!malloc_expected) {
+    if (unexpected_malloc_callback) {
+      (*unexpected_malloc_callback)();
+    }
+  }
+  void * memory = malloc(size);
+  uint64_t fw_size = size;
+  MALLOC_PRINTF(
+    "malloc  expected(%s): %p %" PRIu64 "\n",
```
jacquelinekay: Is there a character missing here after the last %?

rcl/test/memory_tools_common.cpp
```diff
@@ -0,0 +1,159 @@
...
+{
+  if (!enabled.load()) {return free(memory);}
+  auto foo = SCOPE_EXIT(enabled.store(true););
+  enabled.store(false);
+  if (!free_expected) {
+    if (unexpected_free_callback) {
+      (*unexpected_free_callback)();
+    }
+  }
+  MALLOC_PRINTF("free    expected(%s): %p\n", malloc_expected ? "true " : "false", memory);
```
jacquelinekay: the spaces after `free` look a little weird. Maybe do something like `free():    expected...`

rcl/test/rcl/test_time.cpp
```diff
@@ -0,0 +1,109 @@
...
+  EXPECT_NE(now.nanoseconds, 0);
+  // Compare to std::chrono::system_clock time (within a second).
+  now = {0};
+  ret = rcl_system_time_point_now(&now);
+  {
+    std::chrono::system_clock::time_point now_sc = std::chrono::system_clock::now();
+    auto now_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(now_sc.time_since_epoch());
+    int64_t now_ns_int = now_ns.count();
+    int64_t now_diff = now.nanoseconds - now_ns_int;
+    EXPECT_LE(llabs(now_diff), RCL_MS_TO_NS(1000)) << "system_clock differs";
```
jacquelinekay: Perhaps the tolerance for this comparison (currently 1000) should be parameterized or a least stored in a const variable. Also, 1 second seems really large.

rcl/include/rcl/node.h
```diff
@@ -0,0 +1,231 @@
...
+
+/// Return the rmw node handle.
+/* The handle returned is a pointer to the internally held rmw handle.
+ * This function can fail, and therefore return NULL, if:
+ *   - node is NULL
+ *   - node has not been initialized (the implementation is invalid)
+ *
+ * The returned handle is only valid as long as the given node is valid.
+ *
+ * \TODO(wjwwood) should the return value of this be const?
```
dirk-thomas: If the user accesses this handle it is likely to call some rmw specific API. Therefore I would say not const. We could always add a second API to provide a const if needs comes up.

rcl/include/rcl/publisher.h
```diff
@@ -0,0 +1,277 @@
...
+ *                the interface, but I'll just mention it here.
+ *
+ * \TODO(wjwwood) should the return value of this be const?
+ *
+ * \param[in] publisher pointer to the rcl publisher
+ * \return rmw publisher handle if successful, otherwise NULL
+ */
+RCL_PUBLIC
+rmw_publisher_t *
+rcl_publisher_get_rmw_publisher_handle(const rcl_publisher_t * publisher);
```
dirk-thomas: Should we shorten the function name to not repeat `publisher`: `rcl_publisher_get_rmw_handle`?

rcl/include/rcl/publisher.h
```diff
@@ -0,0 +1,277 @@
...
+ *   - publisher is invalid (never called init, called fini, or invalid node)
+ *
+ * The returned handle is only valid as long as the given publisher is valid.
+ *
+ * \TODO(wjwwood) will the publisher ever need to remove and recreate the rmw
+ *                handle under the hood? Perhaps not now, but if we add the
+ *                ability to reconfigure, alias, or remap topics, and don't
+ *                implement that into rmw, then we should consider the
+ *                implications here. This argument applies to other parts of
+ *                the interface, but I'll just mention it here.
```
dirk-thomas: Good point. Probably a note in the doc would be enough for now that changes might result in the returned handle to not be valid forever and it should therefore be re-get manually?

rcl/include/rcl/subscription.h
```diff
@@ -0,0 +1,277 @@
...
+ * This function can fail, and therefore return NULL, if the:
+ *   - subscription is NULL
+ *   - subscription is invalid (never called init, called fini, or invalid)
+ *
+ * The returned string is only valid as long as the subscription is valid.
+ * The value of the string may change if the topic name changes, and therefore
+ * copying the string is recommended if this is a concern.
+ *
+ * This function is not thread-safe, and copying the result is not thread-safe.
+ * \TODO(wjwwood): should we provide a thread-safe copy_topic_name?
```
dirk-thomas: The doc is clear about how long the data is valid. I would offload that to the caller.

rcl/include/rcl/time.h
```diff
@@ -0,0 +1,127 @@
...
+
+#if __cplusplus
+extern "C"
+{
+#endif
+
+#include "rcl/types.h"
+#include "rcl/visibility_control.h"
+
+#define RCL_S_TO_NS(seconds) (seconds * 1000000000)
```
dirk-thomas: For better readability should we split long constant numbers into more readable factors:

    seconds * 1000 * 1000 * 1000

rcl/src/rcl/common.c
```diff
@@ -0,0 +1,55 @@
...
+rcl_impl_getenv(const char * env_name, const char ** env_value)
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(env_name, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(env_value, RCL_RET_INVALID_ARGUMENT);
+  *env_value = NULL;
+#if !defined(WIN32)
+  *env_value = getenv(env_name);
+  if (*env_value == NULL) {
+    *env_value = "";
+  }
```
dirk-thomas: With this the caller can not distinguish if the environment variable was not set or set to an empty string. Why not just return NULL for this?

rcl/src/rcl/common.h
```diff
@@ -0,0 +1,59 @@
...
+{
+#endif
+
+#include "rcl/error_handling.h"
+#include "rcl/types.h"
+
+#define RCL_CHECK_ARGUMENT_FOR_NULL(argument, error_return_type) \
+  RCL_CHECK_FOR_NULL_WITH_MSG(argument, #argument " argument is null", return error_return_type)
+
+#define RCL_CHECK_FOR_NULL_WITH_MSG(value, msg, error_statement) if (!value) { \
```
dirk-thomas: Suggesting to wrap before the `if` for better readability.

dirk-thomas: Also `value` should be wrapped in parenthesis since otherwise the `!` might modify the logic of an expression.

rcl/src/rcl/stdatomic_helper.h
dirk-thomas: For nested `#if` I would recommend to indent the preprocessor directives in the inner block with one space after the `#`. E.g. https://github.com/ros2/rmw_connext/blob/master/rmw_connext_cpp/src/functions.cpp#L26-L33 the same is being used in gtest. Btw. the style guide also recommends that (but doesn't require it): https://google.github.io/styleguide/cppguide.html#Preprocessor_Directives

rcl/test/CMakeLists.txt
```diff
@@ -0,0 +1,94 @@
+if(AMENT_ENABLE_TESTING)
```
dirk-thomas: This check shouldn't be necessary. It is checked before adding the subdirectory.

rcl/include/rcl/node.h
wjwwood: This is a good question. I don't know what the limits are for the DDS vendors or even if the specification deals with that detail. This remains an open question for me.

rcl/include/rcl/publisher.h
wjwwood: Ok, sounds good to me: cf450f5

rcl/include/rcl/subscription.h
wjwwood: I'll have to move this documentation somewhere first. I'll defer this until we have documentation setup: https://github.com/ros2/rcl/issues/6

rcl/src/rcl/timer.c
wjwwood: I think it was previously a more complicated operation, fixed with 8ea6b75

rcl/src/rcl/timer.c
wjwwood: You think there should be another note made still, or is the one in `rcl_timer_fini` that I linked to enough?

Can you suggest a change if you think another note needs to be in there? Should that note be on other functions?

rcl/src/rcl/wait.c
wjwwood: I don't remember, I think it used to be that the size of 0 was indicative of not setup, but that doesn't look to be the case anymore. I removed them in 68facdb, but we'll see if this causes an issue once I get to testing this (but it will be the last thing I test since it needs subscriptions which wake up and timers and guard conditions).

rcl/src/rcl/wait.c
wjwwood: Yeah, that's for concatenating things to the macro arguments. uncrustify insisted on putting spaces around them, which I didn't like, but couldn't find an easy way to change it so I just went with it.

rcl/src/rcl/wait.c
wjwwood: Yeah it's kind of nasty, basically it assumes size is defined in the local scope out side of the macro call.

rcl/CMakeLists.txt
wjwwood: Oh, I didn't notice, but I just copied it from `rclcpp`: https://github.com/ros2/rclcpp/blob/master/rclcpp/CMakeLists.txt#L56

I guess I could change them both.

wjwwood: It was actually broken before: 2398e98 I must have copied the CMake code from `rclcpp` when refactoring for the typesupport changes (refactored to build one library for each impl like `rclcpp`) after testing on Windows, because in the current state it would not work.

rcl/CMakeLists.txt
wjwwood: No, it causes issues on GCC 4.7 and it is not necessary to define that in order to use `stdatomic.h`, which is the only thing from "C11" that I use. Also Windows doesn't support it and it doesn't appear that it will.

rcl/CMakeLists.txt
wjwwood: I have a problem where no of the default linter tests would work if I put them in the subdirectory `CMakeLists.txt`, i.e. the `ament_lint_auto_find_test_dependencies()` call didn't work. So I moved them back into the top level `CMakeLists.txt`.

rcl/include/rcl/allocator.h
wjwwood: Well the `reallocf` behavior was introduced because the pattern for safely using `realloc` is a bit strange. It often leads to this subtle bug:

```c++
void * mem = malloc(...);
mem = realloc(mem, ...);
if (!mem) {
  // The memory allocated by malloc is now leaked,
  // since it was not free'd and the only pointer to it was overwritten with NULL.
}
```

Which leads to something correct being like this:

```c++
void * mem = malloc(...);
void * new_mem = realloc(mem, ...);
if (!new_mem) {
  free(mem);
}
mem = new_mem;
```

And `reallocf` avoids this issue completely with this pattern now being correct:

```c++
void * mem = malloc(...);
mem = reallocf(mem, ...);
if (!mem) {
  // No memory has been leaked.
}
```

The only downside to only having `reallocf` would be that you cannot keep the original memory if `realloc` fails, which might be a valid thing to do.

I think I'll instead make a wrapper function which does something similar to this:

```c++
void *
rcl_reallocf(
  void * (*realloc_func)(void *, size_t, void *),
  void (*free_func)(void *, void *),
  void * original_mem, size_t new_size, void * state)
{
  void * new_mem = realloc_func(original_mem, new_size, state);
  if (!new_mem) {
    free_func(original_mem, state);
  }
  return new_mem;
}
```

rcl/src/rcl/timer.c
jacquelinekay: I think it's okay not to make a note of that in the code, it's more of a general policy/coding practice that we should encapsulate in a tutorial or article.

rcl/include/rcl/allocator.h
wjwwood: I changed it in 898f389

rcl/include/rcl/guard_condition.h
wjwwood: I think `create_` is the wrong prefix, since that implies (at least to me) that you're doing allocation and initialization. This function is merely giving the appropriate NULL values to copy into your uninitialized, but allocated struct.

rcl/CMakeLists.txt
dirk-thomas: We already use `-std=c11` throughout rosidl. Which platform has only gcc 4.7 and what are the problems?

Since the line is in a condition it doesn't affect Windows.

rcl/CMakeLists.txt
dirk-thomas: I only referred to line 63. Is there a problem with moving that line to the subfolder?

rcl/include/rcl/guard_condition.h
dirk-thomas: I see, it currently doesn't perform any complex initialization. Do you expect it to do that in the future? If not I would rather not add a function like this to our core API - simply because the caller code can do exactly the same using a few characters (`= {0}`).

wjwwood: I generally agree, and I was about to remove these functions, but in testing I noticed that you can do this:

```c
rcl_node_t node = rcl_get_zero_initialized_node();
rcl_node_init(&node, ...);
// ... use the node
rcl_node_fini(&node);
node = rcl_get_zero_initialized_node();
```

But not this (at least not in C):

```c
rcl_node_t node = {0};
rcl_node_init(&node, ...);
// ... use the node
rcl_node_fini(&node);
node = {0};  // This is a compiler error.
```

The error looks like this:

```
error: expected expression
node = {0};
       ^
1 error generated.
```

So I left them in, but I can still remove them, even with this caveat.

rcl/include/rcl/guard_condition.h
wjwwood: I agree that some context may be needed, but I'm not sure that node is the right context. Should guard conditions become invalid when a node is destroyed? Should you be able to create guard conditions with out nodes (I think you should be abled to)?

I also have the `rcl_get_instance_id()` function which returns a unique number for each period between an `rcl_init` and the matching `rcl_shutdown`. Internally the implementation could use this in a map of resources by instance id, and the storage for that map could be created during `rcl_init`. This isn't fundamentally different from returning a "context handle" from `rcl_init` and passing it to pretty much every other init/fini/shutdown function.

rcl/test/memory_tools_common.cpp
wjwwood: Nope: http://en.cppreference.com/w/c/types/integer#Example

rcl/test/memory_tools_common.cpp
wjwwood: I implemented something like you suggested:

```
6: [==========] Running 1 test from 1 test case.
6: [----------] Global test environment set-up.
6: [----------] 1 test from TestMemoryTools
6: [ RUN      ] TestMemoryTools.test_allocation_checking_tools
6: test_memory_tools(26057,0x7fff7348f300) malloc: starting memory checking...
6: test_memory_tools(26057,0x7fff7348f300) malloc: malloc():  (    expected) 0x7fa791800000 1024
6: test_memory_tools(26057,0x7fff7348f300) malloc: realloc(): (    expected) 0x7fa791800000 0x7fa791800400 2048
6: test_memory_tools(26057,0x7fff7348f300) malloc: free():    (    expected) 0x7fa791800400
6: test_memory_tools(26057,0x7fff7348f300) malloc: malloc():  (not expected) 0x7fa791800000 1024
6: test_memory_tools(26057,0x7fff7348f300) malloc: realloc(): (    expected) 0x7fa791800000 0x7fa791800400 2048
6: test_memory_tools(26057,0x7fff7348f300) malloc: free():    (    expected) 0x7fa791800400
6: test_memory_tools(26057,0x7fff7348f300) malloc: malloc():  (not expected) 0x7fa792000000 1024
6: test_memory_tools(26057,0x7fff7348f300) malloc: realloc(): (    expected) 0x7fa792000000 0x7fa792000400 2048
6: test_memory_tools(26057,0x7fff7348f300) malloc: free():    (    expected) 0x7fa792000400
6: test_memory_tools(26057,0x7fff7348f300) malloc: malloc():  (    expected) 0x7fa792000000 1024
6: test_memory_tools(26057,0x7fff7348f300) malloc: realloc(): (    expected) 0x7fa792000000 0x7fa792000400 2048
6: test_memory_tools(26057,0x7fff7348f300) malloc: free():    (    expected) 0x7fa792000400
6: test_memory_tools(26057,0x7fff7348f300) malloc: malloc():  (    expected) 0x7fa792000000 1024
6: test_memory_tools(26057,0x7fff7348f300) malloc: realloc(): (    expected) 0x7fa792000000 0x7fa792000400 2048
6: test_memory_tools(26057,0x7fff7348f300) malloc: free():    (    expected) 0x7fa792000400
6: test_memory_tools(26057,0x7fff7348f300) malloc: malloc():  (    expected) 0x7fa792000000 1024
6: test_memory_tools(26057,0x7fff7348f300) malloc: realloc(): (    expected) 0x7fa792000000 0x7fa792000400 2048
6: test_memory_tools(26057,0x7fff7348f300) malloc: free():    (    expected) 0x7fa792000400
6: test_memory_tools(26057,0x7fff7348f300) malloc: stopping memory checking...
6: [       OK ] TestMemoryTools.test_allocation_checking_tools (1 ms)
6: [----------] 1 test from TestMemoryTools (1 ms total)
6:
6: [----------] Global test environment tear-down
6: [==========] 1 test from 1 test case ran. (2 ms total)
6: [  PASSED  ] 1 test.
```

And this is what it looked like before:

```
6: [==========] Running 1 test from 1 test case.
6: [----------] Global test environment set-up.
6: [----------] 1 test from TestMemoryTools
6: [ RUN      ] TestMemoryTools.test_allocation_checking_tools
6: test_memory_tools(26402,0x7fff7348f300) malloc: starting memory checking...
6: test_memory_tools(26402,0x7fff7348f300) malloc: malloc  expected(true ): 0x7ff163801800 1024
6: test_memory_tools(26402,0x7fff7348f300) malloc: realloc expected(true ): 0x7ff163801800 0x7ff163801c00 2048
6: test_memory_tools(26402,0x7fff7348f300) malloc: free    expected(true ): 0x7ff163801c00
6: test_memory_tools(26402,0x7fff7348f300) malloc: malloc  expected(false): 0x7ff163801800 1024
6: test_memory_tools(26402,0x7fff7348f300) malloc: realloc expected(true ): 0x7ff163801800 0x7ff163801c00 2048
6: test_memory_tools(26402,0x7fff7348f300) malloc: free    expected(true ): 0x7ff163801c00
6: test_memory_tools(26402,0x7fff7348f300) malloc: malloc  expected(false): 0x7ff164800000 1024
6: test_memory_tools(26402,0x7fff7348f300) malloc: realloc expected(true ): 0x7ff164800000 0x7ff164800400 2048
6: test_memory_tools(26402,0x7fff7348f300) malloc: free    expected(true ): 0x7ff164800400
6: test_memory_tools(26402,0x7fff7348f300) malloc: malloc  expected(true ): 0x7ff164800000 1024
6: test_memory_tools(26402,0x7fff7348f300) malloc: realloc expected(true ): 0x7ff164800000 0x7ff164800400 2048
6: test_memory_tools(26402,0x7fff7348f300) malloc: free    expected(true ): 0x7ff164800400
6: test_memory_tools(26402,0x7fff7348f300) malloc: malloc  expected(true ): 0x7ff164800000 1024
6: test_memory_tools(26402,0x7fff7348f300) malloc: realloc expected(true ): 0x7ff164800000 0x7ff164800400 2048
6: test_memory_tools(26402,0x7fff7348f300) malloc: free    expected(true ): 0x7ff164800400
6: test_memory_tools(26402,0x7fff7348f300) malloc: malloc  expected(true ): 0x7ff164800000 1024
6: test_memory_tools(26402,0x7fff7348f300) malloc: realloc expected(true ): 0x7ff164800000 0x7ff164800400 2048
6: test_memory_tools(26402,0x7fff7348f300) malloc: free    expected(true ): 0x7ff164800400
6: test_memory_tools(26402,0x7fff7348f300) malloc: stopping memory checking...
6: [       OK ] TestMemoryTools.test_allocation_checking_tools (1 ms)
6: [----------] 1 test from TestMemoryTools (1 ms total)
6:
6: [----------] Global test environment tear-down
6: [==========] 1 test from 1 test case ran. (1 ms total)
6: [  PASSED  ] 1 test.
```

Neither look very nice to me. Any other suggestions? Maybe `malloc`, `realloc`, and `free` should be right justified?

rcl/test/rcl/test_time.cpp
wjwwood: I can put it in a variable, and since you mentioned this I noticed the steady clock tests are wrong/bogus. I addressed both in 09a2aad.

For system clock, my expectation is that if they disagree then they will disagree by a lot. I'm not sure making this smaller will help since they will be different (since we take one time measurement and then the other), but not by a deterministic amount.

rcl/include/rcl/publisher.h
wjwwood: I removed the TODO and updated the docs to make a note about this in a44c14c.

I added similar notes for node, subscription, and guard condition.

I still have a lingering concern that in the future it may be possible for the handle to be invalidated asynchronously and therefore make using this function always unsafe. I imagine:

- Someone calls this function.
- Externally someone calls a remapping service call which causes the rmw handle to be destroyed and recreated.
- Someone trys to use the result of this function which is now invalid.

I guess it will be up to the user of rcl to make sure the callback for handling remapping or reconfiguring is locked against use of this function?

Having said all that I think it is best to leave it as is (with a note about the volatile nature of the handle) and address it later when we have the need.

rcl/include/rcl/subscription.h
wjwwood: Addressed in a44c14c.

rcl/include/rcl/time.h
wjwwood: 6ccbb35

rcl/src/rcl/common.c
wjwwood: On Windows unset and set to an empty string are the same thing. So even though I can tell on Unix, I cannot tell on Windows, so I made this function the lowest common denominator of functionality.

rcl/src/rcl/common.h
wjwwood: 7bceed4

rcl/test/CMakeLists.txt
wjwwood: I was concerned about how `add_subdirectory` behaves if I rerun cmake without `AMENT_ENABLE_TESTING` set, after having run it with it set. I'll remove it. b92148d

rcl/src/rcl/stdatomic_helper.h
wjwwood: f35c06e

rcl/CMakeLists.txt
wjwwood: I had trouble with the line `#include <time.h>` when I enabled `-std=c11`, but I never tracked it down to a bug report or tried with a newer version of gcc.

Windows doesn't support C11 and I can't find any indication that they plan to improve on that: http://herbsutter.com/2012/05/03/reader-qa-what-about-vc-and-c99/

I can tell you that they would need some extra features or compiler support to implement the C11 atomic api as described using their atomics primitives.

I haven't run into any C11 feature that I need yet and I had strange compiler behavior on Linux with gcc when using `-std=c11` (but it works fine on OS X), so I don't pass the flag.

rcl/CMakeLists.txt
wjwwood: 64dc7ac

rcl/include/rcl/publisher.h
wjwwood: In node I was thinking there might be different kinds of rmw handles you may need to ask for from the node (maybe a node and a guard condition) or something like that. So that spilled over into publisher and subscription because of that.

I'll shorten them for now and we can make them more specific later, if necessary.

wjwwood: c6a903f

wjwwood:
Thanks for the comments and style/typo/spelling fixes. I think I'm up-to-date on all comments. Please comment again if you think the ball is still in my court on any of the line items.

I'm going to add the warnings about using return codes and tests for node and then propose we merge it. I can continue to add tests and fix things in follow on pull requests.

rcl/test/memory_tools_common.cpp
wjwwood: I ended up with 5372c52, which looks like this:

```
6: [==========] Running 1 test from 1 test case.
6: [----------] Global test environment set-up.
6: [----------] 1 test from TestMemoryTools
6: [ RUN      ] TestMemoryTools.test_allocation_checking_tools
6: test_memory_tools(32191,0x7fff7348f300) malloc: starting memory checking...
6: test_memory_tools(32191,0x7fff7348f300) malloc:  malloc (    expected) 0x7f82cb800000 1024
6: test_memory_tools(32191,0x7fff7348f300) malloc: realloc (    expected) 0x7f82cb800000 0x7f82cb800400 2048
6: test_memory_tools(32191,0x7fff7348f300) malloc:    free (    expected) 0x7f82cb800400
6: test_memory_tools(32191,0x7fff7348f300) malloc:  malloc (not expected) 0x7f82cb800000 1024
6: test_memory_tools(32191,0x7fff7348f300) malloc: realloc (    expected) 0x7f82cb800000 0x7f82cb800400 2048
6: test_memory_tools(32191,0x7fff7348f300) malloc:    free (    expected) 0x7f82cb800400
6: test_memory_tools(32191,0x7fff7348f300) malloc:  malloc (not expected) 0x7f82cb800000 1024
6: test_memory_tools(32191,0x7fff7348f300) malloc: realloc (    expected) 0x7f82cb800000 0x7f82cc000000 2048
6: test_memory_tools(32191,0x7fff7348f300) malloc:    free (    expected) 0x7f82cc000000
6: test_memory_tools(32191,0x7fff7348f300) malloc:  malloc (    expected) 0x7f82cc000800 1024
6: test_memory_tools(32191,0x7fff7348f300) malloc: realloc (    expected) 0x7f82cc000800 0x7f82cc000000 2048
6: test_memory_tools(32191,0x7fff7348f300) malloc:    free (    expected) 0x7f82cc000000
6: test_memory_tools(32191,0x7fff7348f300) malloc:  malloc (    expected) 0x7f82cc000800 1024
6: test_memory_tools(32191,0x7fff7348f300) malloc: realloc (    expected) 0x7f82cc000800 0x7f82cc000000 2048
6: test_memory_tools(32191,0x7fff7348f300) malloc:    free (    expected) 0x7f82cc000000
6: test_memory_tools(32191,0x7fff7348f300) malloc:  malloc (    expected) 0x7f82cc000800 1024
6: test_memory_tools(32191,0x7fff7348f300) malloc: realloc (    expected) 0x7f82cc000800 0x7f82cc000000 2048
6: test_memory_tools(32191,0x7fff7348f300) malloc:    free (    expected) 0x7f82cc000000
6: test_memory_tools(32191,0x7fff7348f300) malloc: stopping memory checking...
6: [       OK ] TestMemoryTools.test_allocation_checking_tools (1 ms)
6: [----------] 1 test from TestMemoryTools (1 ms total)
6:
6: [----------] Global test environment tear-down
6: [==========] 1 test from 1 test case ran. (1 ms total)
6: [  PASSED  ] 1 test.
```

rcl/include/rcl/guard_condition.h
dirk-thomas: I think it comes down to what some rmw impl. might require. I can imagine that some implementation will need to create a context in `rmw_init` and needs that being passed into every other rmw function.

rcl/src/rcl/common.c
dirk-thomas: I see your point but I would draw a different conclusion: a user on non-Windows should be able to distinguish it. The limitation of Windows should not limit use cases on other platforms.

rcl/test/memory_tools_common.cpp
jacquelinekay: That looks better. You could also get rid of the whitespace in the
parentheses and transfer it to after the close paren so that the address is
still aligned (maybe?) but then the "expected" tokens wouldn't line up.
 *shrug*

rcl/src/rcl/common.c
dirk-thomas: Ok, if somebody needs this differentiation he just needs to use `getenv` directly.

rcl/include/rcl/guard_condition.h
wjwwood: I think we should wait until we refactor `rmw`, because this will be a recurring theme I think where `rmw` should actually provide this and `rcl` should just expose that data which is held and produced by `rmw`. Similar issues like the instance id should be managed by `rmw` and not `rcl`, and other related issues.

When we do that I think the right approach will be to have `rmw_init` return an "instance" or a "context" and pass that along to the other functions, but I think we should wait and see.

dirk-thomas: Sounds good to me.

wjwwood:
The Linux server is busy, so I'll start one of those later, but here are OS X and Windows for this branch:

- http://ci.ros2.org/job/ci_osx/636/
- http://ci.ros2.org/job/ci_windows/819/

wjwwood:
@tfoote @dirk-thomas I have the same problem on my OS X job, e.g. http://ci.ros2.org/job/ci_osx/636/testReport/junit/projectroot/test/test_memory_tools/.

wjwwood:
Starting a Linux CI job: http://ci.ros2.org/job/ci_linux/759/

wjwwood:
I think this is ready to merge after CI:

- http://ci.ros2.org/job/ci_linux/790/
- http://ci.ros2.org/job/ci_osx/659/
- http://ci.ros2.org/job/ci_windows/839/

dirk-thomas:
Hopefully fixed compiler warnings:
* http://ci.ros2.org/job/ci_linux/792/

dirk-thomas:
Hopefully fixed OS X missing library:

http://ci.ros2.org/job/ci_osx/661/

dirk-thomas:
+1

wjwwood:
Thanks for the fix ups @dirk-thomas, I'm going to merge and I'll address any new failures in a follow up pr if needed. :tada:

PR #7, opened 2015-12-18
fix for waitset api change

jacquelinekay:
Connects to ros2/rmw#49

### DIFF
--- a/rcl/src/rcl/wait.c
+++ b/rcl/src/rcl/wait.c
@@ -35,6 +35,7 @@ typedef struct rcl_wait_set_impl_t
   rmw_subscriptions_t rmw_subscriptions;
   size_t guard_condition_index;
   rmw_guard_conditions_t rmw_guard_conditions;
+  rmw_waitset_t * rmw_waitset;
   size_t timer_index;
   rcl_allocator_t allocator;
 } rcl_wait_set_impl_t;
@@ -113,6 +114,9 @@ rcl_wait_set_init(
   wait_set->impl->rmw_subscriptions.subscriber_count = 0;
   wait_set->impl->rmw_guard_conditions.guard_conditions = NULL;
   wait_set->impl->rmw_guard_conditions.guard_condition_count = 0;
+  wait_set->impl->rmw_waitset = rmw_create_waitset(
+    NULL, 2*number_of_subscriptions + number_of_guard_conditions);
+
   // Initialize subscription space.
   rcl_ret_t ret;
   if ((ret = rcl_wait_set_resize_subscriptions(wait_set, number_of_subscriptions)) != RCL_RET_OK) {
@@ -409,6 +413,7 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
     &wait_set->impl->rmw_guard_conditions,
     &dummy_services,
     &dummy_clients,
+    wait_set->impl->rmw_waitset,
     timeout_argument);
   // Check for error.
   if (ret != RMW_RET_OK) {
--- a/rcl/src/rcl/wait.c
+++ b/rcl/src/rcl/wait.c
@@ -115,7 +115,10 @@ rcl_wait_set_init(
   wait_set->impl->rmw_guard_conditions.guard_conditions = NULL;
   wait_set->impl->rmw_guard_conditions.guard_condition_count = 0;
   wait_set->impl->rmw_waitset = rmw_create_waitset(
-    NULL, 2*number_of_subscriptions + number_of_guard_conditions);
+    NULL, 2 * number_of_subscriptions + number_of_guard_conditions);
+  if (!wait_set->impl->rmw_waitset) {
+    goto fail;
+  }
 
   // Initialize subscription space.
   rcl_ret_t ret;
@@ -151,6 +154,12 @@ rcl_wait_set_init(
   wait_set->impl->allocator = allocator;
   return RCL_RET_OK;
 fail:
+  if (__wait_set_is_valid(wait_set)) {
+    rmw_ret_t ret = rmw_destroy_waitset(wait_set->impl->rmw_waitset);
+    if (ret != RMW_RET_OK) {
+      fail_ret = RCL_RET_WAIT_SET_INVALID;
+    }
+  }
   __wait_set_clean_up(wait_set, allocator);
   return fail_ret;
 }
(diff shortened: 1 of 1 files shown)


wjwwood:
It doesn't look like you ever destroy the `waitset`?

rcl/src/rcl/wait.c
```diff
@@ -111,6 +112,9 @@ rcl_wait_set_init(
   wait_set->impl->rmw_subscriptions.subscriber_count = 0;
   wait_set->impl->rmw_guard_conditions.guard_conditions = NULL;
   wait_set->impl->rmw_guard_conditions.guard_condition_count = 0;
+  wait_set->impl->rmw_waitset = rmw_create_waitset(
+    NULL, 2*number_of_subscriptions + number_of_guard_conditions);
```
wjwwood: You should check to make sure this function succeeded, and if it doesn't you should `goto fail` and make sure it is cleaned up in `__wait_set_clean_up`.

wjwwood:
C is like crawling on your hands and knees. I'd recommend putting the wait set clean up in `__wait_set_clean_up` that way it is cleanup on failure in the init function as well as in the fini function.

jacquelinekay:
Yeah man. Life is hard. c50ab45

rcl/src/rcl/wait.c
```diff
@@ -77,6 +78,7 @@ __wait_set_clean_up(rcl_wait_set_t * wait_set, rcl_allocator_t allocator)
     assert(ret == RCL_RET_OK);  // Defensive, shouldn't fail with size 0.
   }
   if (wait_set->impl) {
+    assert(rmw_destroy_waitset(wait_set->impl->rmw_waitset) == RMW_RET_OK);
```
wjwwood: This needs to be guarded by a `if (wait_set->impl->rmw_waitset) {...}` because it is possible for allocation of `wait_set->impl` to be successful, but `wait_set->impl->rmw_waitset` is still NULL or does not point to allocated memory.

wjwwood:
+1

rcl/src/rcl/wait.c
```diff
@@ -79,6 +80,9 @@ __wait_set_clean_up(rcl_wait_set_t * wait_set, rcl_allocator_t allocator)
     assert(ret == RCL_RET_OK);  // Defensive, shouldn't fail with size 0.
   }
   if (wait_set->impl) {
+    if (wait_set->impl->rmw_waitset) {
+      assert(rmw_destroy_waitset(wait_set->impl->rmw_waitset) == RMW_RET_OK);
```
dirk-thomas: Shouldn't this be checked in release builds too?

jacquelinekay: I'm following the pattern of the above code here. Since this function signature has no way of returning an error code, I'm not sure how to propagate the error to the highest level. @wjwwood any thoughts on how these functions should report errors in release builds?

wjwwood: Sorry, I had to do some research first. This is definitely a bug and needs to be fixed through out. I'm in the process of auditing `rcl` for this problem.

In the example above `rmw_destroy_waitset(...)` is never run if `-DNDEBUG` is passed to the compiler.

wjwwood: As to whether or not the check should be made, it is a situation where the it should never fail, but because it could in other scenarios there is a warning when you don't check the return value. This check is really defensive and should never be false.

wjwwood: I actually don't see any other problematic uses of asserts (where needed code is avoided with `NDEBUG`) right now.

Perhaps we need a custom assert (it's a fairly common thing in projects) which is always checked. I think there's a need for "I want to make a defensive check that should never be false, and if it is it is not recoverable" but that is not disabled when `-DNDEBUG` is used. The consensus on the internet is that you can easily make one with `printf` and `abort` which are portable.

jacquelinekay: ok, I've moved `rmw_destroy_waitset` out of the assert statement so that it better follows the pattern established in the above code. I'll leave changing the assert statement until after we've resolved the question about the custom assert.

Consider that in if the assert is truly defensive, it may be just fine to only check it in Debug mode (if the system is well tested) and ignore the assert in Release mode for performance purposes.

wjwwood: I agree, I think checking it in debug only is ok (that was the idea with the assert anyways), but then it results in a unused variable warnings which requires us to but in that obnoxious other "no lint" line. I'm not sure whether or not the performance difference is appreciable.

wjwwood: Another thing I'll mention here is that the other functions are implemented in this file and I can ensure they shouldn't fail in that scenario fairly easily. But the same might not be so easy to say of the `rmw_destroy_waitset` function since it is presumably implemented by each middleware implementation and might fail in some scenario. In which an assert might not be the right action to take here. But I think it's ok to handle that when it comes up and leave the assert for now.

dirk-thomas: Since the existing implementations of `rmw_destroy_waitset` do return an error code in certain situations I think the return value should definitely be checked here (meaning also in release). Otherwise the rmw impl. might detect a problem but it stays unnoticed here.

jacquelinekay: It's a good point that `rmw_destroy_waitset` isn't implemented in this file, unlike the other functions.

Should I change the signature of this function to return an error code to propagate it to the user?

jacquelinekay: I've decided the minimal viable solution here is to mirror what's in `publisher.c` and move the `rmw_destroy_waitset` call to `wait_set_fini`, which already has the right function signature to propagate an error code.

rcl/src/rcl/wait.c
```diff
@@ -156,7 +163,13 @@ rcl_wait_set_fini(rcl_wait_set_t * wait_set)
 {
   rcl_ret_t result = RCL_RET_OK;
   RCL_CHECK_ARGUMENT_FOR_NULL(wait_set, RCL_RET_INVALID_ARGUMENT);
+
   if (__wait_set_is_valid(wait_set)) {
+    rmw_ret_t ret = rmw_destroy_waitset(wait_set->impl->rmw_waitset);
```
wjwwood: Doesn't this need to be done in the fail label of the rcl_wait_set_init function?

jacquelinekay: It does; I will add it.

Is there an appropriate return code besides `RCL_RET_ERROR` for this case? I was considering `RCL_RET_WAIT_SET_INVALID`

wjwwood: Which ever.

Issue #8, opened 2015-12-19
rcl has failing tests on Windows


wjwwood:
E.g.: http://ci.ros2.org/job/ci_windows/867/

PR #9, opened 2015-12-19
fix tests on Windows

dirk-thomas:
Fixes all but one test. Connect to #8.

By using the same folder the test executables find the libraries of this package.

Before http://ci.ros2.org/job/ci_windows/868/testReport/
After http://ci.ros2.org/job/ci_windows/869/testReport/


gerkey:
+1

Does this mean that we must always declare tests in the same CMakeLists.txt where we declare libraries that are being used by those test?

wjwwood:
It's possible to make it work otherwise, but I'm not sure what the "right" approach is.

dirk-thomas:
That's exactly what I thought :wink:

spiked3:
dirk,
I'd like to help out - can we have a short discussion? spiked@gmail.com sorry, I do not know of a better way to contact you

wjwwood:
I think the right way to fix this would be to have an extra set of arguments for `ament_add_*test`:

- `APPEND_ENV`: pairs of `ENV_VAR`, `VALUE` where `VALUE` is appended to `ENV_VAR` (or set to if it if doesn't exist yet), this is the general form
- `APPEND_LIBRARY_DIRS`: list of paths which will be added to the system specific library path, i.e. `LD_LIBRARY_PATH`, `DYLD_LIBRARY_PATH`, or `PATH`

Then these values to be appended would be passed down to the test runner through a single `--append-env` argument in `run_test.py`.

We already have `ENV` which sets environment variables. I'm not sure if it is worth having `PREPEND` or `REMOVE` options for modifying environment variable as well.

wjwwood:
Closed in favor of the other solution in #10.

PR #11, opened 2015-12-31
allow fail of custom domain id on Win/OpenSplice

wjwwood:
It's a known issue that passing a custom domain id which doesn't match the `ROS_DOMAIN_ID` environment variable on Windows with OpenSplice doesn't work. I just adjusted the tests to allow for this. See: https://github.com/ros2/rmw_opensplice/blob/release-alpha3/rmw_opensplice_cpp/src/rmw_node.cpp#L76

Connects to ros2/rcl#8

### DIFF
--- a/rcl/test/rcl/test_node.cpp
+++ b/rcl/test/rcl/test_node.cpp
@@ -18,6 +18,7 @@
 
 #include "rcl/rcl.h"
 #include "rcl/node.h"
+#include "rmw/rmw.h"  // For rmw_get_implementation_identifier.
 
 #include "../memory_tools.hpp"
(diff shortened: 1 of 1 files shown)


wjwwood:
OpenSplice only Windows job (to see if the last failing tests are fixed): http://ci.ros2.org/job/ci_windows/885/

If that works I'll do a full set of CI jobs.

gerkey:
+1

wjwwood:
Full CI:
- http://ci.ros2.org/job/ci_linux/835/
- http://ci.ros2.org/job/ci_osx/697/
- http://ci.ros2.org/job/ci_windows/888/

Issue #15, opened 2016-01-05
Unused variable warning when compiling in `RelWithDebInfo`


gerkey:
The problem is that the variable is only referenced inside an `assert()`, which is presumably being compiled out under `RelWithDebInfo`.

~~~
Building C object CMakeFiles/rcl__rmw_opensplice_cpp.dir/src/rcl/timer.c.o
/home/gerkey/ros2_ws/src/ros2/rcl/rcl/src/rcl/wait.c: In function ‘__wait_set_clean_up’:
/home/gerkey/ros2_ws/src/ros2/rcl/rcl/src/rcl/wait.c:66:13: warning: variable ‘ret’ set but not used [-Wunused-but-set-variable]
   rcl_ret_t ret;
             ^
/home/gerkey/ros2_ws/src/ros2/rcl/rcl/src/rcl/wait.c: In function ‘rcl_wait’:
/home/gerkey/ros2_ws/src/ros2/rcl/rcl/src/rcl/wait.c:431:15: warning: variable ‘rcl_ret’ set but not used [-Wunused-but-set-variable]
     rcl_ret_t rcl_ret;
               ^
~~~

wjwwood:
I'd just add a `(void)ret;` after each of those lines.

wjwwood:
I'm looking into a pr for this.

dirk-thomas:
Can this be closed?

gerkey:
Confirmed fixed when building `RelWithDebInfo` on Linux.

PR #16, opened 2016-01-05
Fix warnings

wjwwood:
Should address https://github.com/ros2/rcl/issues/15 as well as a few other warnings I noticed.

### DIFF
--- a/rcl/src/rcl/wait.c
+++ b/rcl/src/rcl/wait.c
@@ -63,18 +63,19 @@ __wait_set_is_valid(const rcl_wait_set_t * wait_set)
 static void
 __wait_set_clean_up(rcl_wait_set_t * wait_set, rcl_allocator_t allocator)
 {
-  rcl_ret_t ret;
-  (void)ret;  // NO LINT
   if (wait_set->subscriptions) {
-    ret = rcl_wait_set_resize_subscriptions(wait_set, 0);
+    rcl_ret_t ret = rcl_wait_set_resize_subscriptions(wait_set, 0);
+    (void)ret;  // NO LINT
     assert(ret == RCL_RET_OK);  // Defensive, shouldn't fail with size 0.
   }
   if (wait_set->guard_conditions) {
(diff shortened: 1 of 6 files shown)


rcl/src/rcl/wait.c
```diff
@@ -64,6 +64,7 @@ static void
 __wait_set_clean_up(rcl_wait_set_t * wait_set, rcl_allocator_t allocator)
 {
   rcl_ret_t ret;
+  (void)ret;  // NO LINT
```
dirk-thomas: Why not declare the `ret` variable in each local scope since the variable seems to never be used in this bigger scope?

Also shouldn't all the checks with `assert` also be done when building release?

wjwwood: No particular reason, but I don't see an advantage to changing that.

As each of the asserts are commented those are defensive checks, so I'd say they don't need to be checked at runtime. I'm sort of agnostic one way or the other.

dirk-thomas: I am argueing for the smaller scope based on out programming conventions (https://github.com/ros2/ros2/wiki/Developer-Guide#programming-conventions):

> Declare variables in the narrowest scope possible.

If you think the checks are fine for debug builds only that is ok for me.

wjwwood: Well, I interpret that guideline as particularly for a scenario like this:

```c++
// This way:
int foo()
{
  {
    int i;
    i  = 3;
  }
}
// Rather than this way:
int foo()
{
  int i;
  {
    i = 3;
  }
}
```

But in this scenario we're actually reusing the variable across multiple scopes. But since we're not getting any advantage with this (other than less code) I changed it in 3310359.

dirk-thomas: Well, the variables in the different scopes only have the same name but are never used to pass information. They could easily have unique names. Therefore I think the smallest scope rule applies.

dirk-thomas:
Lgtm

PR #19, opened 2016-02-05
some changes to C typesupport tests for Connext

jacquelinekay:
I tried to generalize this to work with Connext and it works for me locally.


wjwwood:
Should be fine, I'll test it locally too.

Issue #22, opened 2016-02-12
rcl test failures when run with only opensplice on windows


tfoote:
We're seeing some test failures on the windows build when connext and fastrtps are disabled. 

They were first observed in #12 and #21 merge reviews and have been confirmed to be a bug in master: http://ci.ros2.org/job/ci_windows/1054/  

The failures are in these tests. 
TestNodeFixture.test_rcl_node_accessors
TestNodeFixture.test_rcl_node_life_cycle

In #21 @wjwwood has also observed it with connext too, still windows only.

jacquelinekay:
error output from one of the failing tests:

```
17: [ RUN      ] TestNodeFixture.test_rcl_node_accessors
17: ========================================================================================
17: Report      : ERROR
17: Date        : Fri Feb 12 10:56:10 Pacific Standard Time 2016
17: Description : Failed to connect to/create the domain with ID 42
17: Node        : leon
17: Process     : test_node__rmw_opensplice_cpp.exe <2724>
17: Thread      : 3004
17: Internals   : V6.4.140407OSS/3b83e82/3b83e82/gapi::gapi_domainParticipantFactory::gapi_domainParticipantFactory_create_participant/gapi_domainParticipantFactory.c/285/0/1455303
370.637426897
17: starting memory checking... not available
17: stopping memory checking... not available
17: C:\cygwin64\home\Jackie\dev\ros2\src\ros2\rcl\rcl\test\rcl\test_node.cpp(78): error: Value of: ret
17:   Actual: 100
17: Expected: 0
17: stopping memory checking... not available
17: [  FAILED  ] TestNodeFixture.test_rcl_node_accessors (5 ms)
17: [ RUN      ] TestNodeFixture.test_rcl_node_life_cycle
17: starting memory checking... not available
17: -- run_test.py: invocation failed: 'utf-8' codec can't decode byte 0xd2 in position 60: invalid continuation byte
17: stopping memory checking... not available
17: C:\cygwin64\home\Jackie\dev\ros2\src\ros2\rcl\rcl\test\rcl\test_node.cpp(237): error: Value of: ret
17:   Actual: 0
17: Expected: 101
17: Expected RCL_RET_NOT_INIT
17: stopping memory checking... not available
17: [  FAILED  ] TestNodeFixture.test_rcl_node_life_cycle (1314 ms)
```

tfoote:
@wjwwood and I looked into this some last night. 

[...]

PR #23, opened 2016-02-12
extend the library path with the location of the memory tools also

wjwwood:
This commit got excluded from #21 somehow. It should address the failing `rcl` tests on Windows.

### DIFF
--- a/rcl/test/CMakeLists.txt
+++ b/rcl/test/CMakeLists.txt
@@ -5,8 +5,9 @@ include(rcl_add_custom_gtest.cmake)
 
 set(extra_test_libraries)
 set(extra_test_env)
+set(extra_lib_dirs "${rcl_lib_dir}")
 
-# This subdirectory extends both extra_test_libraries and extra_test_env.
+# This subdirectory extends both extra_test_libraries, extra_test_env, and extra_lib_dirs.
 add_subdirectory(memory_tools)
 
 macro(test_target)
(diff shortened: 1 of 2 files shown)


jacquelinekay:
Aha!

wjwwood:
Is that a +1?

Here is CI (only Windows, only OpenSplice): http://ci.ros2.org/job/ci_windows/1058/

But I'm pretty confident that this will work since it was included in this CI from #21: https://github.com/ros2/rcl/pull/21#issuecomment-183173668

jacquelinekay:
I'm testing it locally right now.

jacquelinekay:
Locally all the tests are fixed except for test_node__rmw_opensplice_cpp, which fails on test_rcl_node_accessors due to a failure in gapi_domainParticipantFactory. (How do I copy text out of a Windows developer prompt? It's like I can't even highlight things.)

wjwwood:
Right-click->mark, select things, press enter, and then paste where ever you want.

wjwwood:
I'm pretty sure @tfoote already ticketed that: https://github.com/ros2/rcl/issues/22

wjwwood:
Unless it's a new error that I'm not familiar with, but it sounds connected.

jacquelinekay:
It sounds like the same one. I'll take my discussion to that issue.

wjwwood:
CI job passed and has the failures reported by #24

PR #24, opened 2016-02-12
Fix node test for opensplice on Windows

jacquelinekay:
so, this "fixes" the failing node test case on Windows, though I'm not sure if the test should be changed or the behavior of `rcl_node_init`.

Connects to #22

### DIFF
--- a/rcl/src/rcl/node.c
+++ b/rcl/src/rcl/node.c
@@ -107,6 +107,7 @@ fail:
   if (node->impl) {
     allocator->deallocate(node->impl, allocator->state);
   }
+  *node = rcl_get_zero_initialized_node();
   return fail_ret;
 }
 
--- a/rcl/test/rcl/test_node.cpp
+++ b/rcl/test/rcl/test_node.cpp
@@ -261,7 +261,6 @@ TEST_F(TestNodeFixture, test_rcl_node_life_cycle) {
   ret = rcl_node_init(&node, name, &options_with_invalid_allocator);
   EXPECT_EQ(RCL_RET_INVALID_ARGUMENT, ret) << "Expected RCL_RET_INVALID_ARGUMENT";
   rcl_reset_error();
-  node = rcl_get_zero_initialized_node();
   // Try with failing allocator.
   rcl_node_options_t options_with_failing_allocator = rcl_node_get_default_options();
   options_with_failing_allocator.allocator.allocate = failing_malloc;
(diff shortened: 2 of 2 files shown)


tfoote:
Yeah, this should get rid of the failure, however I think it's mostly avoiding the issue rather than fixing it.

tfoote:
I've cleaned up the patch and kicked off CI jobs, but they're all backlogged. 
http://ci.ros2.org/job/ci_linux/980
http://ci.ros2.org/job/ci_osx/814
http://ci.ros2.org/job/ci_windows/1065

@wjwwood @jacquelinekay please review

wjwwood:
lgtm, don't forget to squash.

tfoote:
Squashed. 
CI is green on linux and osx. 
Windows reported one test failure, [details](http://ci.ros2.org/job/ci_windows/1065/testReport/projectrootC_.J.workspace.ci_windows.ws.src.ros2.rcl/rcl/test_node__rmw_opensplice_cpp/) It looks to me like the tests pass, Each test reports PASS as I would expect. However there's a utf-8 warning. My guess is that one of the error messages contains a print of uninitialized data.

Unfortunately the next job has blown away the workspace and so I can't verify what's in the xml.

dirk-thomas:
I retriggered the same build again: http://ci.ros2.org/job/ci_windows/1073/ We will see if the test fails the same way and if it happens you maybe get a chance to look into the workspace for more information.

tfoote:
Looking deeper I found an error not cleaning up the rmw error. I added commit 1eb1745 to fix that. And now the tests pass. 

New ci: 
http://ci.ros2.org/job/ci_linux/998/
http://ci.ros2.org/job/ci_osx/829
http://ci.ros2.org/job/ci_windows/1078/

Does anyone know how this error would cause the test to fail that reported pass?

jacquelinekay:
Ideally `rmw_reset_error` would not be needed in that location. That implies to me that `rmw_reset_error` must be added to the fail block of any function that uses `rmw/rcl_set_error_state`. We should try to find a fix that is more local to the function to make the fix.

tfoote:
I think the `rmw_reset_error` is required to be there in this case. This function calls an rmw method which fails. It catches that error and then triggers an rcl error. At that point the rmw error should be cleared as it has been caught and passed to a higher level logic. Anyone that processes the rcl error later will not have any knowledge of the original source of the error, except through introspecting the string, and consequently cannot cleanup the error. 

It's not true that any fail block needs this, only ones which can enter the fail block due to an rmw error and forward the error to an rcl error will need to reset the rmw error state. 

We could change the `RCL_CHECK_FOR_NULL_WITH_MSG` to also support an additional cleanup function. However that is the purpose of the fail block, to clean up any state, which includes the `rmw_error_state`. We could also extend the `rmw/rcl` error API to have a more atomic, "get and clear" method for use in this case. But however we do this it needs to be cleared.

tfoote:
We need to investigate why it's failing on jenkins.  There must be something else happening to fail the test. 

We'll merge the change to the node, w/o the rmw_reset if it's aproblem at the jenkins level. 

Longer term we should look at expanding the macro to be able to do an update to the error message, or consider different error storage for rcl and rmw error strings.

tfoote:
rebuilding with ros2/rmw#54 and the last commit stripped

http://ci.ros2.org/job/ci_windows/1080/

tfoote:
Talked with @wjwwood merging in conjunction with ros2/rmw#54

PR #26, opened 2016-02-19
fix style of header include guard

dirk-thomas:
Connect ament/ament_lint#46

I looked into this after the additional prefix was necessary in https://github.com/ros2/rcl/pull/21#discussion_r52572364


wjwwood:
+1

esteve:
+1

Issue #27, opened 2016-02-23
Services


jacquelinekay:
Implement service/client API for rcl.

- C typesupport for services
  - [x] rosidl_generator_c and rosidl_typesupport_introspection_c ros2/rosidl#97
  - [x] rosidl_typesupport_opensplice_c ros2/rmw_opensplice#117
  - [x] rosidl_typesupport_connext_c ros2/rmw_connext#141
- [x] rmw: add service name to client and service structs ros2/rmw#56
- [x] rcl implementation and tests ros2/rcl#30

PR #30, opened 2016-02-26
C clients/services

jacquelinekay:
Connects to ros2/rcl#27

Implement C clients and services.

Not ready for review quite yet, but I would welcome a preliminary review of the .h files and the tests.

TODO:

- [ ] Discuss design choices
  - [x] Storing and incrementing the sequence_number happens at the rcl layer. Should this variable be atomic?
  - [x] Do the fini methods need the node argument? Should I be doing something with them? (Resolution: keep the unused node argument for now and consider changing signature in rmw)
  - [ ] the `taken` boolean is not exposed as an argument to the `take` functions, like it is in rmw. Instead, if take failed, a new error code (CLIENT_TAKE_FAILED or SERVICE_TAKE_FAILED) is returned. Is this OK? I see that `rcl_take` in `subscription.h` has a pointer to a `taken` boolean in the function signature. I suppose the `take` functions should at least be consistent with each other. @wjwwood any thoughts?
  - [x] request_header. Currently it's type-erased in rcl and rmw. Does it need to be? See ros2/rmw#57
- [x] Comments. I copy pasted the headers and the comments currently reflect that. I'll take the time to correct them and write comments that are at least as extensive as the publisher/subscription comments.
  - [ ] In particular, comments about multithreading and allocation behavior
- [ ] Code cleanup/linting
- [ ] Multithreaded/multi-process test (optional?)

### DIFF
--- a/rcl/include/rcl/client.h
+++ b/rcl/include/rcl/client.h
@@ -1,0 +1,301 @@
+// Copyright 2016 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCL__CLIENT_H_
+#define RCL__CLIENT_H_
+
+#if __cplusplus
+extern "C"
+{
+#endif
+
+#include "rosidl_generator_c/service_type_support.h"
+
+#include "rcl/macros.h"
+#include "rcl/node.h"
+#include "rcl/visibility_control.h"
+
+/// Internal rcl client implementation struct.
+struct rcl_client_impl_t;
+
+/// Handle for a rcl client.
+typedef struct rcl_client_t
+{
+  struct rcl_client_impl_t * impl;
+} rcl_client_t;
+
+/// Options available for a rcl client.
+typedef struct rcl_client_options_t
+{
+  /// Middleware quality of service settings for the client.
+  rmw_qos_profile_t qos;
+  /// Custom allocator for the client, used for incidental allocations.
+  /* For default behavior (malloc/free), use: rcl_get_default_allocator() */
+  rcl_allocator_t allocator;
+} rcl_client_options_t;
+
+/// Return a rcl_client_t struct with members set to NULL.
+/* Should be called to get a null rcl_client_t before passing to
+ * rcl_initalize_client().
+ * It's also possible to use calloc() instead of this if the rcl_client is
+ * being allocated on the heap.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_client_t
+rcl_get_zero_initialized_client(void);
+
+
+/// Initialize a rcl client.
+/* After calling this function on a rcl_client_t, it can be used to send requests of the given
+ * type by calling rcl_send_request().
+ * If the request is received by a (possibly remote) service and if the service sends a response,
+ * the client can access the response through rcl_take_response once the response is available to
+ * the client.
+ *
+ * The given rcl_node_t must be valid and the resulting rcl_client_t is only
+ * valid as long as the given rcl_node_t remains valid.
+ *
+ * The rosidl_service_type_support_t is obtained on a per .srv type basis.
+ * When the user defines a ROS service, code is generated which provides the
+ * required rosidl_service_type_support_t object.
+ * This object can be obtained using a language appropriate mechanism.
+ * \TODO(wjwwood) probably should talk about this once and link to it instead
+ * For C this macro can be used (using example_interfaces/AddTwoInts as an example):
+ *
+ *    #include <rosidl_generator_c/service_type_support.h>
+ *    #include <example_interfaces/srv/add_two_ints.h>
+ *    rosidl_service_type_support_t * ts =
+ *      ROSIDL_GET_SERVICE_TYPE_SUPPORT(example_interfaces, AddTwoInts);
+ *
+ * For C++ a template function is used:
+ *
+ *    #include <rosidl_generator_cpp/service_type_support.hpp>
+ *    #include <example_interfaces/srv/add_two_ints.h>
+ *    rosidl_service_type_support_t * ts = rosidl_generator_cpp::get_service_type_support_handle<
+ *      example_interfaces::srv::AddTwoInts>();
+ *
+ * The rosidl_service_type_support_t object contains service type specific
+ * information used to send or take requests and responses.
+ *
+ * \TODO(wjwwood) update this once we've come up with an official scheme.
+ * The service name must be a non-empty string which follows the topic/service naming
+ * format.
+ *
+ * The options struct allows the user to set the quality of service settings as
+ * well as a custom allocator which is used when initializing/finalizing the
+ * client to allocate space for incidentals, e.g. the service name string.
+ *
+ * Expected usage (for C services):
+ *
+ *    #include <rcl/rcl.h>
+ *    #include <rosidl_generator_c/service_type_support.h>
+ *    #include <example_interfaces/srv/add_two_ints.h>
+ *
+ *    rcl_node_t node = rcl_get_zero_initialized_node();
+ *    rcl_node_options_t node_ops = rcl_node_get_default_options();
+ *    rcl_ret_t ret = rcl_node_init(&node, "node_name", &node_ops);
+ *    // ... error handling
+ *    rosidl_service_type_support_t * ts = ROSIDL_GET_SERVICE_TYPE_SUPPORT(
+ *      example_interfaces, AddTwoInts);
+ *    rcl_client_t client = rcl_get_zero_initialized_client();
+ *    rcl_client_options_t client_ops = rcl_client_get_default_options();
+ *    ret = rcl_client_init(&client, &node, ts, "add_two_ints", &client_ops);
+ *    // ... error handling, and on shutdown do finalization:
+ *    ret = rcl_client_fini(&client, &node);
+ *    // ... error handling for rcl_client_fini()
+ *    ret = rcl_node_fini(&node);
+ *    // ... error handling for rcl_node_fini()
+ *
+ * This function is not thread-safe.
+ *
+ * \param[inout] client preallocated client structure
+ * \param[in] node valid rcl node handle
+ * \param[in] type_support type support object for the service's type
+ * \param[in] service_name the name of the service to request
+ * \param[in] options client options, including quality of service settings
+ * \return RCL_RET_OK if the client was initialized successfully, or
+ *         RCL_RET_NODE_INVALID if the node is invalid, or
+ *         RCL_RET_ALREADY_INIT if the client is already initialized, or
+ *         RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
+ *         RCL_RET_BAD_ALLOC if allocating memory fails, or
+ *         RCL_RET_ERROR if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_client_init(
+  rcl_client_t * client,
+  const rcl_node_t * node,
+  const rosidl_service_type_support_t * type_support,
+  const char * service_name,
+  const rcl_client_options_t * options);
+
+/// Finalize a rcl_client_t.
+/*
+ * After calling, calls to rcl_send_request and rcl_take_response will fail when using this client.
+ * However, the given node handle is still valid.
+ *
+ * This function is not thread-safe.
+ *
+ * \param[inout] client handle to the client to be finalized
+ * \param[in] node handle to the node used to create the client
+ * \return RCL_RET_OK if client was finalized successfully, or
+ *         RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
+ *         RCL_RET_ERROR if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_client_fini(rcl_client_t * client, rcl_node_t * node);
+
+/// Return the default client options in a rcl_client_options_t.
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_client_options_t
+rcl_client_get_default_options(void);
+
+/// Send a ROS request using a client.
+/* It is the job of the caller to ensure that the type of the ros_request
+ * parameter and the type associate with the client (via the type support)
+ * match.
+ * Passing a different type to send_request produces undefined behavior and cannot
+ * be checked by this function and therefore no deliberate error will occur.
+ *
+ * send_request is an non-blocking call.
+ *
+ * The ROS request message given by the ros_request void pointer is always owned by the
+ * calling code, but should remain constant during send_request.
+ *
+ * This function is thread safe so long as access to both the client and the
+ * ros_request is synchronized.
+ * That means that calling rcl_send_request from multiple threads is allowed, but
+ * calling rcl_send_request at the same time as non-thread safe client functions
+ * is not, e.g. calling rcl_send_request and rcl_client_fini concurrently
+ * is not allowed.
+ * Before calling rcl_send_request the message can change and after calling
+ * rcl_send_request the message can change, but it cannot be changed during the
+ * send_request call.
+ * The same ros_request, however, can be passed to multiple calls of
+ * rcl_send_request simultaneously, even if the clients differ.
+ * The ros_request is unmodified by rcl_send_request.
+ *
+ * \param[in] client handle to the client which will make the response
+ * \param[in] ros_request type-erased pointer to the ROS request message
+ * \param[out] sequence_number the sequence number
+ * \return RCL_RET_OK if the request was sent successfully, or
+ *         RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
+ *         RCL_RET_CLIENT_INVALID if the client is invalid, or
+ *         RCL_RET_ERROR if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_send_request(const rcl_client_t * client, const void * ros_request, int64_t * sequence_number);
+
+
+// Take a ROS response using a client
+/* It is the job of the caller to ensure that the type of the ros_response
+ * parameter and the type associate with the client (via the type support)
+ * match.
+ * Passing a different type to take_response produces undefined behavior and cannot
+ * be checked by this function and therefore no deliberate error will occur.
+ * The request_header is an rmw struct for meta-information about the request sent
+ * (e.g. the sequence number).
+ * ros_response should point to an already allocated ROS response message struct of the
+ * correct type, into which the response from the service will be copied.
+ *
+ * \param[in] client handle to the client which will take the response
+ * \param[inout] request_header pointer to the request header
+ * \param[inout] ros_response type-erased pointer to the ROS response message
+ * \return RCL_RET_OK if the response was taken successfully, or
+ *         RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
+ *         RCL_RET_CLIENT_INVALID if the client is invalid, or
+ *         RCL_RET_ERROR if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_take_response(
+  const rcl_client_t * client,
+  rmw_request_id_t * request_header,
+  void * ros_response);
+
+/// Get the name of the service that this client will request a response from.
+/* This function returns the client's internal service name string.
+ * This function can fail, and therefore return NULL, if the:
+ *   - client is NULL
+ *   - client is invalid (never called init, called fini, or invalid node)
+ *
+ * The returned string is only valid as long as the rcl_client_t is valid.
+ * The value of the string may change if the service name changes, and therefore
+ * copying the string is recommended if this is a concern.
+ *
+ * This function is not thread-safe, and copying the result is not thread-safe.
+ *
+ * \param[in] client pointer to the client
+ * \return name string if successful, otherwise NULL
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+const char *
+rcl_client_get_service_name(const rcl_client_t * client);
+
+/// Return the rcl client options.
+/* This function returns the client's internal options struct.
+ * This function can fail, and therefore return NULL, if the:
+ *   - client is NULL
+ *   - client is invalid (never called init, called fini, or invalid node)
+ *
+ * The returned struct is only valid as long as the rcl_client_t is valid.
+ * The values in the struct may change if the options of the client change,
+ * and therefore copying the struct is recommended if this is a concern.
+ *
+ * This function is not thread-safe, and copying the result is not thread-safe.
+ *
+ * \param[in] client pointer to the client
+ * \return options struct if successful, otherwise NULL
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+const rcl_client_options_t *
+rcl_client_get_options(const rcl_client_t * client);
+
+/// Return the rmw client handle.
+/* The handle returned is a pointer to the internally held rmw handle.
+ * This function can fail, and therefore return NULL, if the:
+ *   - client is NULL
+ *   - client is invalid (never called init, called fini, or invalid node)
+ *
+ * The returned handle is made invalid if the client is finalized or if
+ * rcl_shutdown() is called.
+ * The returned handle is not guaranteed to be valid for the life time of the
+ * client as it may be finalized and recreated itself.
+ * Therefore it is recommended to get the handle from the client using
+ * this function each time it is needed and avoid use of the handle
+ * concurrently with functions that might change it.
+ *
+ * \param[in] client pointer to the rcl client
+ * \return rmw client handle if successful, otherwise NULL
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rmw_client_t *
+rcl_client_get_rmw_handle(const rcl_client_t * client);
+
+#if __cplusplus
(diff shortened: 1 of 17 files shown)


jacquelinekay:
The documentation should be fairly coherent now (?).

rcl/include/rcl/client.h
```diff
@@ -0,0 +1,300 @@
...
+ *
+ *    #include <rcl/rcl.h>
+ *    #include <rosidl_generator_c/service_type_support.h>
+ *    #include <example_interfaces/srv/add_two_ints.h>
+ *
+ *    rcl_node_t node = rcl_get_zero_initialized_node();
+ *    rcl_node_options_t node_ops = rcl_node_get_default_options();
+ *    rcl_ret_t ret = rcl_node_init(&node, "node_name", &node_ops);
+ *    // ... error handling
+ *    rosidl_service_type_support_t * ts = ROSIDL_GET_MESSAGE_TYPE_SUPPORT(
```
wjwwood: Should be `ROSIDL_GET_SERVICE_TYPE_SUPPORT`?

rcl/include/rcl/client.h
```diff
@@ -0,0 +1,300 @@
...
+ *      example_interfaces, AddTwoInts);
+ *    rcl_client_t client = rcl_get_zero_initialized_client();
+ *    rcl_client_options_t client_ops = rcl_client_get_default_options();
+ *    ret = rcl_client_init(&client, &node, ts, "chatter", &client_ops);
```
wjwwood: Probably should use "add_two_ints" as the service name.

rcl/include/rcl/client.h
```diff
@@ -0,0 +1,300 @@
...
+ *    #include <example_interfaces/srv/add_two_ints.h>
+ *    rosidl_service_type_support_t * ts = rosidl_generator_cpp::get_service_type_support_handle<
+ *      example_interfaces::srv::AddTwoInts>();
+ *
+ * The rosidl_service_type_support_t object contains service type specific
+ * information used to send or take requests and responses.
+ *
+ * \TODO(wjwwood) update this once we've come up with an official scheme.
+ * The topic name must be a non-empty string which follows the topic naming
+ * format.
```
wjwwood: I think the term "topic" is specific to pub/sub, and we should use "service" name instead. Services may also have a different naming format since topics are content oriented and services could also be location oriented, i.e. `service://node/service_name` versus `service://*/service_name` (or something like that).

rcl/include/rcl/client.h
```diff
@@ -0,0 +1,300 @@
...
+ *    ret = rcl_client_fini(&client, &node);
+ *    // ... error handling for rcl_client_fini()
+ *    ret = rcl_node_fini(&node);
+ *    // ... error handling for rcl_deinitialize_node()
+ *
+ * This function is not thread-safe.
+ *
+ * \param[inout] client preallocated client structure
+ * \param[in] node valid rcl node handle
+ * \param[in] type_support type support object for the topic's type
```
wjwwood: "for the service's type"

rcl/include/rcl/client.h
```diff
@@ -0,0 +1,300 @@
...
+rcl_client_init(
+  rcl_client_t * client,
+  const rcl_node_t * node,
+  const rosidl_service_type_support_t * type_support,
+  const char * service_name,
+  const rcl_client_options_t * options);
+
+/// Finalize a rcl_client_t.
+/* After calling, the node will no longer be advertising that it is publishing
+ * on this topic (assuming this is the only client on this topic).
```
wjwwood: This sentence doesn't look right. I'm not 100% sure what it should say, I guess that the node is no longer calling a particular service.

rcl/include/rcl/client.h
```diff
@@ -0,0 +1,300 @@
...
+ *
+ * After calling, calls to rcl_publish will fail when using this client.
+ * However, the given node handle is still valid.
```
wjwwood: Not `rcl_publish`, but the client specific functions.

rcl/include/rcl/client.h
```diff
@@ -0,0 +1,300 @@
...
+/* It is the job of the caller to ensure that the type of the ros_request
+ * parameter and the type associate with the client (via the type support)
+ * match.
+ * Passing a different type to send_request produces undefined behavior and cannot
+ * be checked by this function and therefore no deliberate error will occur.
+ *
+ * send_request is an non-blocking call. 
+ *
+ * The ROS request message given by the ros_request void pointer is always owned by the
+ * calling code, but should remain constant during publish.
```
wjwwood: "during send"?

rcl/include/rcl/client.h
```diff
@@ -0,0 +1,300 @@
...
+ * TODO: Is this true for send_request?
+ * This function is thread safe so long as access to both the client and the
+ * ros_request is synchronized.
+ * That means that calling rcl_send_request from multiple threads is allowed, but
+ * calling rcl_send_request at the same time as non-thread safe client functions
+ * is not, e.g. calling rcl_send_request and rcl_client_fini concurrently
+ * is not allowed.
+ * Before calling rcl_send_request the message can change and after calling
+ * rcl_send_request the message can change, but it cannot be changed during the
+ * publish call.
```
wjwwood: "send call"

rcl/include/rcl/client.h
```diff
@@ -0,0 +1,300 @@
...
+ * match.
+ * Passing a different type to send_request produces undefined behavior and cannot
+ * be checked by this function and therefore no deliberate error will occur.
+ *
+ * send_request is an non-blocking call. 
+ *
+ * The ROS request message given by the ros_request void pointer is always owned by the
+ * calling code, but should remain constant during publish.
+ *
+ * TODO: Is this true for send_request?
```
wjwwood: I think it should be, but of course that depends on the implementation. If there's a compelling reason for it not to be true then we can change this.

rcl/include/rcl/service.h
```diff
@@ -0,0 +1,319 @@
...
+ * The given rcl_node_t must be valid and the resulting rcl_service_t is
+ * only valid as long as the given rcl_node_t remains valid.
+ *
+ * The rosidl_service_type_support_t is obtained on a per .srv type basis.
+ * When the user defines a ROS service, code is generated which provides the
+ * required rosidl_service_type_support_t object.
+ * This object can be obtained using a language appropriate mechanism.
+ * \TODO(wjwwood) probably should talk about this once and link to it instead
+ * \TODO(jacquelinekay) reworded this for services with substitutions, should it refer to messages?
+ * For C this macro can be used (using std_msgs/String as an example):
```
wjwwood: `std_msgs/String` should be `example_interfaces/AddTwoInts`?

rcl/include/rcl/service.h
```diff
@@ -0,0 +1,319 @@
...
+ *    #include <example_interfaces/srv/add_two_ints.h>
+ *    rosidl_service_type_support_t * ts = rosidl_generator_cpp::get_service_type_support_handle<
+ *      example_interfaces::srv::AddTwoInts>();
+ *
+ * The rosidl_service_type_support_t object contains service type specific
+ * information used to send or take requests and responses.
+ *
+ * \TODO(wjwwood) update this once we've come up with an official scheme.
+ * The topic name must be a non-empty string which follows the topic naming
+ * format.
```
wjwwood: Same comment about topic vs service name.

rcl/include/rcl/service.h
```diff
@@ -0,0 +1,319 @@
...
+ *
+ *    #include <rcl/rcl.h>
+ *    #include <rosidl_generator_c/service_type_support.h>
+ *    #include <example_interfaces/srv/add_two_ints.h>
+ *
+ *    rcl_node_t node = rcl_get_zero_initialized_node();
+ *    rcl_node_options_t node_ops = rcl_node_get_default_options();
+ *    rcl_ret_t ret = rcl_node_init(&node, "node_name", &node_ops);
+ *    // ... error handling
+ *    rosidl_service_type_support_t * ts = ROSIDL_GET_MESSAGE_TYPE_SUPPORT(
```
wjwwood: `ROSIDL_GET_SERVICE_TYPE_SUPPORT`

rcl/include/rcl/service.h
```diff
@@ -0,0 +1,319 @@
...
+ *      example_interfaces, AddTwoInts);
+ *    rcl_service_t service = rcl_get_zero_initialized_service();
+ *    rcl_service_options_t service_ops = rcl_service_get_default_options();
+ *    ret = rcl_service_init(&service, &node, ts, "chatter", &service_ops);
```
wjwwood: "chatter" should probably be "add_two_ints"

rcl/include/rcl/service.h
```diff
@@ -0,0 +1,319 @@
...
+ *    // ... error handling, and on shutdown do finalization:
+ *    ret = rcl_service_fini(&service, &node);
+ *    // ... error handling for rcl_service_fini()
+ *    ret = rcl_node_fini(&node);
+ *    // ... error handling for rcl_deinitialize_node()
```
wjwwood: `rcl_node_fini` (maybe wrong where you copied from too)

rcl/include/rcl/service.h
```diff
@@ -0,0 +1,319 @@
...
+ *         RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
+ *         RCL_RET_CLIENT_INVALID if the service is invalid, or
+ *         RCL_RET_ERROR if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_send_response(
+  const rcl_service_t * service,
+  void * request_header,
```
wjwwood: Should this be response header? Also there are a different number of parameters here and what is in the doc block.

wjwwood:
I got through the headers with only minor comments and questions (I'll try to make a pr tonight to address the obvious changes).

As for your OP questions:

> Storing and incrementing the sequence_number happens at the rcl layer. Should this variable be atomic?

Probably, since calling send_request and send_response should probably be thread safe, like publish is.

> Do the fini methods need the node argument? Should I be doing something with them?

They are passed through to the rmw layer and at least one of the functions does need the original node handle to unregister types or something like that. We could argue that the client/service/publisher/etc should hold on to the dds participant or what ever it needs and use it during fini. The problem then becomes ownership of the pointer to that object. Since node has to be passed to the client's fini function, it's obvious that the node needs to be valid, otherwise it might make sense to fini the node and then the client (which would obviously be an issue if the client's fini function used a thing that is owned by the node that just got fini).

> the taken boolean is not exposed as an argument to the take functions, like it is in rmw. Instead, if take failed, a new error code (CLIENT_TAKE_FAILED or SERVICE_TAKE_FAILED) is returned. Is this OK? I see that rcl_take in subscription.h has a pointer to a taken boolean in the function signature. I suppose the take functions should at least be consistent with each other. @wjwwood any thoughts?

This is probably historical. I think we either were copying the style of the matching DDS function or we have some part of the take function chain that doesn't have many error codes (just ok or error), so taken was passed separately back and we copied that behavior. Either way I think using an error code would be ok, since we also do this for timeout. The only semantic squabble is that take can technically succeed even if no data is taken so it's not quite correct to say that taking no data is a failure.

> request_header. Currently it's type-erased in rcl and rmw. Does it need to be?

I don't know. If the type of `request_header` is something we can share (in an rcl header) and its type doesn't change based on the message type, then no it does not need to be type erased.

jacquelinekay:
"They are passed through to the rmw layer and at least one of the functions does need the original node handle to unregister types or something like that. We could argue that the client/service/publisher/etc should hold on to the dds participant or what ever it needs and use it during fini. The problem then becomes ownership of the pointer to that object. Since node has to be passed to the client's fini function, it's obvious that the node needs to be valid, otherwise it might make sense to fini the node and then the client (which would obviously be an issue if the client's fini function used a thing that is owned by the node that just got fini)."

I guess I was confused because the function signatures rmw_destroy_client and rmw_destroy_service don't include a rmw_node pointer:

https://github.com/ros2/rmw/blob/master/rmw/include/rmw/rmw.h#L120
https://github.com/ros2/rmw/blob/master/rmw/include/rmw/rmw.h#L151

jacquelinekay:
whoops, didn't see your thing about making a PR for the obvious stuff so I took care of it in 174feb6. More small adjustments are always welcome...

wjwwood:
Oh, I was just going to make a pr to help you out, but if you already found time to fix it up then that's fine. 

Yeah I'm not sure about node being passed to the client service finis. Maybe they shouldn't.

jacquelinekay:
Regarding the node pointer in client_fini and service_fini:
Part of your original point was that the node pointer should be passed to fini to indicate ownership.
I agree with this in principal, but it bothers me that the rmw functions do not do this.
rmw_opensplice has a roundabout implementation of destroy_service (and destroy_client), but it does everything it eventually needs to do:
```
callbacks->destroy_responder(service_info->responder_, &rmw_free);
```
https://github.com/ros2/rmw_opensplice/blob/master/rmw_opensplice_cpp/src/rmw_service.cpp#L206
Which, through the typesupport callbacks, calls Responder::teardown. Responder stores a pointer to the DDS Participant and calls the `Participant::delete_...` functions you would expect to clean up the service:
https://github.com/ros2/rmw_opensplice/blob/master/rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/responder.hpp#L168
So `destroy_service` eventually accesses the DDS Participant. The rmw_node that the service was added to points to the same participant, but the participant owned by Responder has no idea about the rmw_node pointer.
However, in Connext the Participant is never accessed directly in destroy due to the usage of Requester/Replier API.
I think I'm going to leave the node pointer in rcl_client_fini and rcl_service_fini and ticket adding the rmw_node pointer to rmw_destroy_client and rmw_destroy_service if needed.

Regarding `request_header` and `response_header`:
The type of this object seems poorly defined right now. Here's a rough sketch:
In rclcpp:
In Client and Service, `create_request_header` returns a shared_ptr<void>, but the underlying type is rmw_request_id_t.
https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/client.hpp#L111
https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/service.hpp#L109

In rcl (current state of this PR):
The pointer is passed straight to rmw as type void *.

In rmw:
`request_header` in take_request, send_response, and take_response is of type void.
https://github.com/ros2/rmw/blob/master/rmw/include/rmw/rmw.h#L133-L137
https://github.com/ros2/rmw/blob/master/rmw/include/rmw/rmw.h#L156-L160
https://github.com/ros2/rmw/blob/master/rmw/include/rmw/rmw.h#L165-L168

In typesupport:
`request_header` is always cast to an `rmw_request_id_t*`.
https://github.com/ros2/rmw_opensplice/blob/master/rosidl_typesupport_opensplice_cpp/resource/srv__type_support.cpp.template#L434
https://github.com/ros2/rmw_connext/blob/master/rosidl_typesupport_connext_cpp/resource/srv__type_support.cpp.template#L132-L133

I don't see why the `request_header` argument is sometimes type-erased when the type is always the same. So I think we can make motions to make the type always `rmw_request_id_t`. The other option would be to create an `rmw_request_header_t` struct in case we anticipate adding more metadata at the rcl layer.

rcl/include/rcl/service.h
jacquelinekay: After some poking around I confirmed that this should indeed be called response_header and not request_header. As a side note, this is kind of annoying because the struct definition is called rmw_request_id_t, but the type is used for both response and request metadata.

wjwwood: Well the `rmw_request_id_t` is just that a request id, and both the request header and the response header may contain the original request id, but for example the response header may have more information. I'm not sure that in practice the request and the response headers will be different or even contain anything other than the request id.

wjwwood:
> I think I'm going to leave the node pointer in rcl_client_fini and rcl_service_fini and ticket adding the rmw_node pointer to rmw_destroy_client and rmw_destroy_service if needed.

Sounds good to me.

> I don't see why the request_header argument is sometimes type-erased when the type is always the same. So I think we can make motions to make the type always rmw_request_id_t. The other option would be to create an rmw_request_header_t struct in case we anticipate adding more metadata at the rcl layer.

That sounds appropriate, especially if there is both a request and a response header.

jacquelinekay:
I think the C Services PRs are ready for review. I can try separating out the changes a bit more if it makes things easier.

http://ci.ros2.org/job/ci_linux/1071/

wjwwood:
@jacquelinekay I'm starting a review now.

jacquelinekay:
I fixed the warnings and the test failure in rclpy: http://ci.ros2.org/job/ci_linux/1073/

jacquelinekay:
OSX passes after some fiddling with linking in the generators:
http://ci.ros2.org/job/ci_osx/875/

Onto Windows!

jacquelinekay:
OK, this Windows CI job appears to have the same number of warnings and test failures as the nightly Release job:

http://ci.ros2.org/job/ci_windows/1140/

rcl/include/rcl/client.h
```diff
@@ -0,0 +1,301 @@
...
+RCL_WARN_UNUSED
+rcl_client_t
+rcl_get_zero_initialized_client(void);
+
+
+/// Initialize a rcl client.
+/* After calling this function on a rcl_client_t, it can be used to send requests of the given
+ * type by calling rcl_send_request().
+ * Once a response has been sent by the service, the response is available to the client when
+ * rcl_take_response() is called.
```
wjwwood: Is it more correct to say "Once a response has been sent by the service, a response may become available in the future and if so it can be retrieved using rcl_take_response()."? As it is currently worded it sounds like calling `rcl_take_response` is what makes it available, which I think is misleading.

rcl/include/rcl/client.h
```diff
@@ -0,0 +1,301 @@
...
+ *
+ * The given rcl_node_t must be valid and the resulting rcl_client_t is only
+ * valid as long as the given rcl_node_t remains valid.
+ *
+ * The rosidl_service_type_support_t is obtained on a per .srv type basis.
+ * When the user defines a ROS service, code is generated which provides the
+ * required rosidl_service_type_support_t object.
+ * This object can be obtained using a language appropriate mechanism.
+ * \TODO(wjwwood) probably should talk about this once and link to it instead
+ * \TODO(jacquelinekay) reworded this for services with substitutions, should it refer to messages?
```
wjwwood: I do not believe so.

rcl/include/rcl/client.h
```diff
@@ -0,0 +1,301 @@
...
+ * match.
+ * Passing a different type to send_request produces undefined behavior and cannot
+ * be checked by this function and therefore no deliberate error will occur.
+ *
+ * send_request is an non-blocking call.
+ *
+ * The ROS request message given by the ros_request void pointer is always owned by the
+ * calling code, but should remain constant during send_request.
+ *
+ * TODO: Is this true for send_request?
```
wjwwood: I think I previously commented on this one. I believe it should be true for `send_request`. So, if you agree, then I'd say this TODO either needs to be removed or updated to say it is not currently but should be.

rcl/include/rcl/client.h
```diff
@@ -0,0 +1,301 @@
...
+ * match.
+ * Passing a different type to take_response produces undefined behavior and cannot
+ * be checked by this function and therefore no deliberate error will occur.
+ * The request_header is an rmw struct for meta-information about the request sent
+ * (e.g. the sequence number).
+ * ros_response should point to an already allocated ROS response message struct of the
+ * correct type, into which the response from the service will be copied.
+ *
+ * \param[in] client handle to the client which will take the response
+ * \param[inout] request_header type-erased pointer to the request header
```
wjwwood: No longer type-erased.

rcl/include/rcl/service.h
```diff
@@ -0,0 +1,318 @@
...
+ * rcl_initalize_service().
+ * It's also possible to use calloc() instead of this if the rcl_service_t
+ * is being allocated on the heap.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_service_t
+rcl_get_zero_initialized_service(void);
+
+/// Initialize a ROS service.
```
wjwwood: In `client.h` the equivalent line is something like "Initialize a rcl client". I don't have a preference for one over the other, but I think they should be consistent.

rcl/include/rcl/client.h
jacquelinekay: I agree with you, I think I just forgot to remove this TODO.

rcl/include/rcl/service.h
```diff
@@ -0,0 +1,318 @@
...
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_service_init(
+  rcl_service_t * service,
+  const rcl_node_t * node,
+  const rosidl_service_type_support_t * type_support,
+  const char * service_name,
+  const rcl_service_options_t * options);
+
+/// Deinitialize a rcl_service_t.
```
wjwwood: Finalize rather than Deinitialize? I'm not sure what we're using throughout.

jacquelinekay: I personally prefer "finalize" to "deinitialize" since it is more specific (and deinitialize technically isn't a word...?)

wjwwood: I agree. I was actually just wondering why we use `fini`, which isn't really a shortened version of `finalize`, but rather French/German for "finished" or "to terminate". Which is appropriate but not strictly English. I guess it's like saying "fin" at the end of a film. We've co-opted it :smile:

jacquelinekay: Yeah, and I guess "init" technically isn't English but an abbreviation for "initialize"...

rcl/include/rcl/service.h
```diff
@@ -0,0 +1,318 @@
...
+ * The ros_request pointer should point to an already allocated ROS request message
+ * struct of the correct type, into which the taken ROS request will be copied
+ * if one is available.
+ * If taken is false after calling, then the ROS request will be unmodified.
+ *
+ * If allocation is required when taking the request, e.g. if space needs to
+ * be allocated for a dynamically sized array in the target message, then the
+ * allocator given in the service options is used.
+ *
+ * The request_header is a type-erased pointer to pre-allocated a rmw struct containing
```
wjwwood: No longer type-erased I think.

rcl/include/rcl/service.h
```diff
@@ -0,0 +1,318 @@
...
+ * match.
+ * Passing a different type to send_response produces undefined behavior and cannot
+ * be checked by this function and therefore no deliberate error will occur.
+ *
+ * send_response is an non-blocking call.
+ *
+ * The ROS response message given by the ros_response void pointer is always owned by the
+ * calling code, but should remain constant during send_response.
+ *
+ * TODO: Is this still true?
```
wjwwood: Another instance that should probably be removed.

rcl/include/rcl/service.h
```diff
@@ -0,0 +1,318 @@
...
+ e This function is thread safe so long as access to both the service and the
```
wjwwood:
```
 * ...
 *
 * ...
 e (day 26, they still haven't noticed me)
 * ...
 * ...
```

rcl/src/rcl/stdatomic_helper.h
```diff
@@ -62,6 +62,14 @@ rcl_atomic_load_bool(atomic_bool * a_bool)
 }
 
 static inline uint64_t
+rcl_atomic_load_int64_t(atomic_int_least64_t * a_int64_t)
```
wjwwood: Shouldn't this return a `int64_t`?

rcl/src/rcl/wait.c
```diff
@@ -236,7 +283,7 @@ rcl_wait_set_get_allocator(const rcl_wait_set_t * wait_set, rcl_allocator_t * al
   memset( \
     wait_set->impl->RMWStorage, \
     0, \
-    sizeof(rmw_subscription_t *) * wait_set->impl->RMWCount);
+    sizeof(rmw_ ## Type ## _t *) * wait_set->impl->RMWCount);
```
wjwwood: Hmm, good catch. I guess it didn't show up in testing since it's always the sizeof a pointer, but still a good thing to fix.

rcl/test/rcl/test_client.cpp
```diff
@@ -0,0 +1,186 @@
...
+  const char * topic_name = "add_two_ints";
+  rcl_client_options_t client_options = rcl_client_get_default_options();
+
+  const rosidl_service_type_support_t * ts = ROSIDL_GET_TYPE_SUPPORT(
+    example_interfaces, srv, AddTwoInts);
+  ret = rcl_client_init(&client, this->node_ptr, ts, topic_name, &client_options);
+
+  // Check the return code of initialization and that the service name matches what's expected
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
+  EXPECT_EQ(strcmp(rcl_client_get_service_name(&client), topic_name), 0);
```
wjwwood: I have been (trying) to do `EXPECTED_EQ(expected, actual)`, but it seems that will eventually not matter:

From: https://github.com/google/googletest/blob/master/googletest/docs/Primer.md#binary-comparison
> Historical note: Before February 2016 *_EQ had a convention of calling it as ASSERT_EQ(expected, actual), so lots of existing code uses this order. Now *_EQ treats both parameters in the same way.

wjwwood:
Other than my comments, lgtm.

Btw, won't this require a related change in `rclpy` since the signature of `rcl_wait_set_init` has changed?

jacquelinekay:
Yep, that's one of the many PRs linked to #30 on Waffle: ros2/rclpy#7

The values for number of clients and number of services are currently both hardcoded as zero until Python services are implemented.

rcl/include/rcl/client.h
```diff
@@ -0,0 +1,300 @@
...
+ * ros_response should point to an already allocated ROS response message struct of the
+ * correct type, into which the response from the service will be copied.
+ *
+ * \param[in] client handle to the client which will take the response
+ * \param[inout] request_header pointer to the request header
+ * \param[inout] ros_response type-erased pointer to the ROS response message
+ * \return RCL_RET_OK if the response was taken successfully, or
+ *         RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
+ *         RCL_RET_CLIENT_INVALID if the client is invalid, or
+ *         RCL_RET_ERROR if an unspecified error occurs.
```
wjwwood: `RCL_RET_CLIENT_TAKE_FAILED` doesn't seem to be documented here. Also did we ever decide if that was better (preferable) to passing a `bool * taken` parameter?

wjwwood:
> Yep, that's one of the many PRs linked to #27 on Waffle: ros2/rclpy#7

Awesome, sorry I missed that connection.

PR #35, opened 2016-03-30
Remove check argument for null from int64_t

jacquelinekay:
I believe a timeout value of 0 is a valid input to `rcl_wait` (no blocking). However, there is check for a null argument that interprets the argument as a pointer, and if timeout is equal to zero then it is interpreted as a null pointer and the function returns an error.


wjwwood:
Good catch, I think it might have been a pointer to a struct at some point. +1

PR #36, opened 2016-03-31
Add fix and regression test for resize to zero bug

jacquelinekay:
Found a bug where `rcl_wait_set_t::size_of_subscriptions` does not get set to zero if `rcl_wait_set_resize_subscriptions` receives an input size of zero. The first commit adds a regression test to show that this is a bug, the second commit makes the fix.


dirk-thomas:
LGTM

Did you trigger a CI before / after?

jacquelinekay:
I tested it manually on Linux only using rcl (before and after). I can launch jobs on all 3 platforms though.

jacquelinekay:
http://ci.ros2.org/job/ci_linux/1132
http://ci.ros2.org/job/ci_osx/918
http://ci.ros2.org/job/ci_windows/1196

wjwwood:
+1

PR #37, opened 2016-03-31
take out test case

jacquelinekay:
I made a mistake when I merged #36 and forgot to take out some a commit that added a test case with faulty assumptions..


wjwwood:
+1 to merge if the test isn't ready, but can you expand on what's wrong with it?

jacquelinekay:
This test case assumes that after `clear_subscriptions` is called on the wait set, then `size_of_subscriptions` is set to zero. After our discussion offline I decided that this was wrong, since you said that `size_of_subscriptions` represents the capacity of the array, not the number of valid (non-null) entries in the array.

jacquelinekay:
Note that the other test case, which tests `resize` instead of `clear`, still remains.

wjwwood:
Ah I see, yeah that distinction isn't clear. I would be ok if we changed the variables names to a more consistent terminology, if there is one. I think I was following the nomenclature of C array, where there is no distinction between size and capacity but elements are allowed to be null (empty).

PR #38, opened 2016-03-31
fix unknownEvaluationOrder warning

dirk-thomas:
Before: http://ci.ros2.org/job/ci_osx/919/
After: http://ci.ros2.org/job/ci_osx/922/


wjwwood:
+1

dirk-thomas:
Of course testing only OS X was not enough :worried:  This is currently breaking the Linux and Windows builds. Running CI builds for a potential fix now - otherwise will revert to unblock master.

PR #39, opened 2016-04-01
fix syntax on Linux and Windows

dirk-thomas:
Fix regression of #38.

Of course testing only OS X was not enough:
* http://ci.ros2.org/job/ci_linux/1137/
* http://ci.ros2.org/job/ci_windows/1200/


wjwwood:
lgtm

PR #40, opened 2016-04-01
Fix rmw array semantics

jacquelinekay:
Tweak behavior to agree with the comments in ros2/rmw#61 and add a few coments to describe this behavior.

### DIFF
--- a/rcl/include/rcl/wait.h
+++ b/rcl/include/rcl/wait.h
@@ -164,9 +164,6 @@ rcl_wait_set_get_allocator(const rcl_wait_set_t * wait_set, rcl_allocator_t * al
 /* This function does not guarantee that the subscription is not already in the
  * wait set.
  *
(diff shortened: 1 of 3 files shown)


wjwwood:
It's a little weird that we're not keeping track of how big our rmw storage is anymore (at least directly), but since free and realloc don't care, neither do I.

I have slight reservations about the new documentation, which is documenting implementation details which shouldn't affect the user (as opposed to an implementation detail that they might care about like does it use locks or is it thread-safe). So, it's something to consider if that documentation should be added.

Overall, +1

jacquelinekay:
I could move it to the code rather than the doxygen-style header comments.

jacquelinekay:
I did it! I used the squash and merge button! I was kinda scared that it was an April Fool's joke that they finally implemented that feature.

PR #41, opened 2016-04-02
Make timer callback optional and fix timer logic

jacquelinekay:
two tweaks to timers that were giving me trouble in the rcl/rclcpp refactor:

1. Make the timer callback optional. If the callback is null, it is simply ignored. In rclcpp, timers take care of their own callbacks instead of storing them in rcl, since lambdas, bound functions, etc. have unexpected function signatures.

2. Fix the logic determining when to wake up `rmw_wait`.

### DIFF
--- a/rcl/include/rcl/timer.h
+++ b/rcl/include/rcl/timer.h
@@ -153,7 +153,8 @@ rcl_timer_fini(rcl_timer_t * timer);
  * the timer's period has not yet elapsed.
  * It is up to the calling code to make sure the period has elapsed by first
  * calling rcl_timer_is_ready().
- * If the callback option was set to null in init, no callback is fired.
+ * If the callback pointer is null (either set in init or exchanged after
+ * initialized), no callback is fired.
  * However, this function should still be called by the client library to
  * update the state of the timer.
  * The order of operations in this command are as follows:
--- a/rcl/include/rcl/timer.h
+++ b/rcl/include/rcl/timer.h
@@ -77,7 +77,12 @@ rcl_get_zero_initialized_timer(void);
  * The period is a duration (rather an absolute time in the future).
  * If the period is 0 then it will always be ready.
  *
- * The callback must be a function which returns void and takes two arguments,
+ * The callback is an optional argument. Valid inputs are either a pointer
+ * to the function callback, or NULL to indicate that no callback will be
+ * stored in rcl.
+ * If the callback is null, the caller client library is responsible for
+ * firing the timer callback.
+ * Else, it must be a function which returns void and takes two arguments,
  * the first being a pointer to the associated timer, and the second a uint64_t
  * which is the time since the previous call, or since the timer was created
  * if it is the first call to the callback.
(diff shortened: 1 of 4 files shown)


jacquelinekay:
I can write a test for this, but it will also be indirectly tested in rclcpp.

rcl/include/rcl/timer.h
```diff
@@ -77,7 +77,12 @@ rcl_get_zero_initialized_timer(void);
  * The period is a duration (rather an absolute time in the future).
  * If the period is 0 then it will always be ready.
  *
- * The callback must be a function which returns void and takes two arguments,
+ * The callback is an optional argument. Valid inputs are either a pointer
```
wjwwood: Please place new sentences on new lines.

wjwwood: Beat me to it. :smile:

jacquelinekay: <(￣︶￣)>

rcl/include/rcl/timer.h
```diff
@@ -148,6 +153,9 @@ rcl_timer_fini(rcl_timer_t * timer);
  * the timer's period has not yet elapsed.
  * It is up to the calling code to make sure the period has elapsed by first
  * calling rcl_timer_is_ready().
+ * If the callback option was set to null in init, no callback is fired.
```
wjwwood: It might be better to phrase this as "If the callback is NULL (either set in init or exchanged for NULL later) ..." or something to that effect.

rcl/src/rcl/wait.c
```diff
@@ -512,44 +512,45 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
...
-  }
+
   if (timeout == 0) {
     // Then it is non-blocking, so set the temporary storage to 0, 0 and pass it.
     temporary_timeout_storage.sec = 0;
     temporary_timeout_storage.nsec = 0;
     timeout_argument = &temporary_timeout_storage;
+  } else {
+    int64_t min_timeout = INT64_MAX;
+    // If min_timeout is > 0, then compare it to the time until each timer.
```
wjwwood: if timeout is > 0?

jacquelinekay:
CI (to be fancy/save time I passed `--end-with rclpy` to the job):
http://ci.ros2.org/job/ci_linux/1151/
http://ci.ros2.org/job/ci_osx/932/
http://ci.ros2.org/job/ci_windows/1210/

jacquelinekay:
Do you think 1969044 is enough to fix the issue we discussed offline?

rcl/src/rcl/wait.c
```diff
@@ -512,44 +512,45 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
...
+        return ret;  // The rcl error state should already be set.
+      }
+      if (timer_timeout < min_timeout) {
+        min_timeout = timer_timeout;
+      }
+    }
+    if (min_timeout == INT64_MAX) {
+      timeout_argument = NULL;
+    } else {
+      temporary_timeout_storage.sec = RCL_NS_TO_S(min_timeout);
```
wjwwood: As discussed off-line I think we need to check if `min_timeout` is negative here. I think it will currently "do the right thing" if it ends up being negative, but I think it would be better if we didn't have multiple states which mean the same thing. I.e., currently time.sec = 0 and time.nsec = 0 mean non-blocking, but negative values for that also mean non-blocking.

wjwwood:
> Do you think 1969044 is enough to fix the issue we discussed offline?

Yeah, lgtm

PR #43, opened 2016-04-05
handle negative timeout

jacquelinekay:
oops. When I changed the logic for wait in #41, I forgot a key case.


wjwwood:
Did this break a test or an example?

jacquelinekay:
No, it doesn't break any existing merged tests or examples because we don't have any code that uses rcl_wait merged yet. This would probably be a good time to introduce those tests.

I discovered this as I was debugging my problem with hanging services/clients in Connext while using rcl in rclcpp. However this didn't appear to solve the problem.

wjwwood:
@jacquelinekay yeah, as we discussed before, tests would be good. But this is incrementally better, imo, so I'd say add tests if you have time to, otherwise merge now.

jacquelinekay:
More tests will come soon.

PR #45, opened 2016-04-18
tests for rcl_wait

jacquelinekay:
I wrote a few tests for simple cases in `rcl_wait` and made a fix to the way timeouts and timers are handled in `rcl_wait`.

The cases are:

- positive timeout argument (1 ms)
- negative timeout argument (block forever) and a timer with 1 ms callback
- zero timeout (nonblocking)
- negative timeout argument with guard condition in the waitset that is triggered from another thread.

### DIFF
--- a/rcl/test/rcl/test_wait.cpp
+++ b/rcl/test/rcl/test_wait.cpp
@@ -31,7 +31,7 @@
 #endif
 
 
-#define TOLERANCE RCL_US_TO_NS(200)
+#define TOLERANCE RCL_US_TO_NS(600)
 
 TEST(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), test_resize_to_zero) {
   // Initialize a waitset with a subscription and then resize it to zero.
@@ -54,7 +54,7 @@ TEST(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), finite_timeout) {
   rcl_ret_t ret = rcl_wait_set_init(&wait_set, 0, 0, 1, 0, 0, rcl_get_default_allocator());
   EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
 
-  int64_t timeout = RCL_MS_TO_NS(1);  // nanoseconds
+  int64_t timeout = RCL_MS_TO_NS(10);  // nanoseconds
   std::chrono::steady_clock::time_point before_sc = std::chrono::steady_clock::now();
   ret = rcl_wait(&wait_set, timeout);
   std::chrono::steady_clock::time_point after_sc = std::chrono::steady_clock::now();
@@ -80,7 +80,7 @@ TEST(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), negative_timeout) {
   EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
 
   rcl_timer_t timer = rcl_get_zero_initialized_timer();
-  ret = rcl_timer_init(&timer, RCL_MS_TO_NS(1), nullptr, rcl_get_default_allocator());
+  ret = rcl_timer_init(&timer, RCL_MS_TO_NS(10), nullptr, rcl_get_default_allocator());
   EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
   ret = rcl_wait_set_add_timer(&wait_set, &timer);
   EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
@@ -103,7 +103,7 @@ TEST(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), negative_timeout) {
   ASSERT_EQ(RCL_RET_TIMEOUT, ret) << rcl_get_error_string_safe();
   // Check time
   int64_t diff = std::chrono::duration_cast<std::chrono::nanoseconds>(after_sc - before_sc).count();
-  EXPECT_LE(diff, RCL_MS_TO_NS(1) + TOLERANCE);
+  EXPECT_LE(diff, RCL_MS_TO_NS(10) + TOLERANCE);
 }
 
 // Test rcl_wait with a timeout value of 0 (non-blocking)
@@ -166,7 +166,7 @@ TEST(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), guard_condition) {
   std::chrono::nanoseconds trigger_diff;
   std::thread trigger_thread([&p, &guard_cond, &trigger_diff]() {
     std::chrono::steady_clock::time_point before_trigger = std::chrono::steady_clock::now();
-    std::this_thread::sleep_for(std::chrono::milliseconds(1));
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
     rcl_ret_t ret = rcl_trigger_guard_condition(&guard_cond);
     std::chrono::steady_clock::time_point after_trigger = std::chrono::steady_clock::now();
     trigger_diff = std::chrono::duration_cast<std::chrono::nanoseconds>(
--- a/rcl/test/rcl/test_wait.cpp
+++ b/rcl/test/rcl/test_wait.cpp
@@ -16,7 +16,7 @@
 #include <future>
 #include <thread>
 
-#include <gtest/gtest.h>
+#include "gtest/gtest.h"
 
 #include "rcl/rcl.h"
 #include "rcl/error_handling.h"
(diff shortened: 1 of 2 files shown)


rcl/test/rcl/test_wait.cpp
```diff
@@ -27,6 +30,8 @@
 #endif
 
 
+#define TOLERANCE 100000  // clock error tolerance in nanoseconds
```
wjwwood: You could use `RCL_S_TO_NS` here if you feel you need to describe it anyways.

rcl/test/rcl/test_wait.cpp
```diff
@@ -41,3 +46,132 @@ TEST(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), test_resize_to_zero) {
...
   ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
 }
+
+// Some test cases for the waitset
+TEST(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), finite_timeout) {
+  rcl_wait_set_t wait_set = rcl_get_zero_initialized_wait_set();
+  rcl_ret_t ret = rcl_wait_set_init(&wait_set, 0, 0, 1, 0, 0, rcl_get_default_allocator());
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
+
+  int64_t timeout = 1000000;  // nanoseconds
```
wjwwood: Again I'd recommend `RCL_MS_TO_NS` or something like that.

rcl/test/rcl/test_wait.cpp
```diff
@@ -41,3 +46,132 @@ TEST(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), test_resize_to_zero) {
   ret = rcl_wait_set_fini(&wait_set);
   ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
 }
+
+// Some test cases for the waitset
```
wjwwood: Is there a more appropriate comment you can put here, one that perhaps summarizes what particular API or behaviors of the waitset are being tested?

rcl/test/rcl/test_wait.cpp
wjwwood: I guess that there isn't a precise enough conversion for this in the `rcl` macros, i.e. there is no `RCL_US_TO_NS`.

jacquelinekay: https://github.com/ros2/rcl/blob/master/rcl/include/rcl/time.h#L27-L33 ?

wjwwood: Ah, just kidding there is... Past self is smarter than current self apparently. Yes those macros are the one's I was talking about.

rcl/test/rcl/test_wait.cpp
```diff
@@ -41,3 +46,132 @@ TEST(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), test_resize_to_zero) {
...
+  rcl_wait_set_t wait_set = rcl_get_zero_initialized_wait_set();
+  rcl_ret_t ret = rcl_wait_set_init(&wait_set, 0, 0, 1, 0, 0, rcl_get_default_allocator());
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
+
+  int64_t timeout = 1000000;  // nanoseconds
+  std::chrono::steady_clock::time_point before_sc = std::chrono::steady_clock::now();
+  ret = rcl_wait(&wait_set, timeout);
+  std::chrono::steady_clock::time_point after_sc = std::chrono::steady_clock::now();
+  // Check time
+  int64_t diff = (after_sc - before_sc).count();
```
wjwwood: Are you sure `.count` will always return nanoseconds? I think it might depend on the default storage of the clock being used (which might be implementation specific). I'd recommend doing an explicit `duration_cast` to nanoseconds first.

rcl/test/rcl/test_wait.cpp
```diff
@@ -41,3 +46,132 @@ TEST(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), test_resize_to_zero) {
...
+
+  // Add a dummy guard condition to avoid an error
+  rcl_guard_condition_t guard_cond = rcl_get_zero_initialized_guard_condition();
+  ret = rcl_guard_condition_init(&guard_cond, rcl_guard_condition_get_default_options());
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
+  ret = rcl_wait_set_add_guard_condition(&wait_set, &guard_cond);
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
+
+
+  // TODO(jacquelinekay) rcl timer tests?
```
wjwwood: More tests or is this a stale TODO?

rcl/test/rcl/test_wait.cpp
```diff
@@ -41,3 +46,132 @@ TEST(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), test_resize_to_zero) {
...
+  rcl_timer_t timer = rcl_get_zero_initialized_timer();
+  rcl_timer_callback_t callback;
+  ret = rcl_timer_init(&timer, 1000000, callback, rcl_get_default_allocator());
```
wjwwood: Should you pass `NULL` or `nullptr` (since this is a cpp file) instead of an uninitialized `rcl_timer_callback_t` (essentially a function pointer)?

jacquelinekay: hmm, yeah, I forgot that `rcl_timer_callback_t` is a typedef to a pointer type--I glanced at the header file and thought that I had to pass an object by const reference, which come to think of it doesn't make much sense...

rcl/test/rcl/test_wait.cpp
```diff
@@ -41,3 +46,132 @@ TEST(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), test_resize_to_zero) {
...
+      after_trigger - before_trigger);
+    p.set_value(ret);
+  }
+  );
+  auto f = p.get_future();
+
+  std::chrono::steady_clock::time_point before_sc = std::chrono::steady_clock::now();
+  ret = rcl_wait(&wait_set, timeout);
+  std::chrono::steady_clock::time_point after_sc = std::chrono::steady_clock::now();
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
```
wjwwood: Looks like this would cause the guard condition and the wait set to memory leak, but that's probably not a big issue. Nodes and `rcl_init`/`rcl_shutdown` are more important to clean up in exception cases.

wjwwood:
lgtm other than some comments.

rcl/test/rcl/test_wait.cpp
jacquelinekay: yeah I mean, I could change ASSERT to EXPECT, or I could add a `fail:` block (here and in several other places). If I added a fail block I would probably generalize some of these cases better by adding custom startup/shutdown behaviors in the `WaitSetTestFixture` class. But also, I'm lazy.

rcl/test/rcl/test_wait.cpp
jacquelinekay: I was considering adding a new `test_timer` file that tests timers without a wait_set involved, in anticipation of seeing more regressions with rclcpp using rcl. But, because I didn't see those regressions in my other branch, I think I'm just going to take out this TODO and then add timer unit tests in the future as they become an issue. Does that sound OK?

wjwwood: Fine by me.

rcl/test/rcl/test_wait.cpp
wjwwood: I think the common pattern is to use a "scope exit" to handle the clean up, e.g.:

https://github.com/ros2/rcl/blob/master/rcl/test/rcl/test_node.cpp#L90-L94

jacquelinekay:
I think I've addressed all comments (I decided to go ahead and implement scope exits here, it wasn't so bad).

I launched the CI with `--end-with rcl`. There are some failures in the services nosetest on default which I will address separately.

http://ci.ros2.org/job/ci_linux/1202/
http://ci.ros2.org/job/ci_osx/972/
http://ci.ros2.org/job/ci_windows/1251/

jacquelinekay:
I think maybe the tolerances are too strict for our Jenkins machines.

jacquelinekay:
relaunched
http://ci.ros2.org/job/ci_linux/1205/
http://ci.ros2.org/job/ci_osx/973/
http://ci.ros2.org/job/ci_windows/1252/
If these are OK I'll merge and then relaunch CI for use_rcl.

PR #46, opened 2016-04-18
include launch in the package.xml

jacquelinekay:
This should fix the failing launchtests in rcl.

http://ci.ros2.org/job/ci_linux/1203/ :boom:


wjwwood:
+1

PR #48, opened 2016-04-19
Keep service around until client exits

jacquelinekay:
Connects to ros2/launch#27
The remote services test is flaky for Connext because the service process needs to stick around until the client exits.

### DIFF
--- a/rcl/test/rcl/service_fixture.cpp
+++ b/rcl/test/rcl/service_fixture.cpp
@@ -151,7 +151,7 @@ int main(int argc, char ** argv)
     // Our scope exits should take care of fini for everything
     // stick around until we are killed by the client
     // To avoid an infinite loop and signal handling, just sleep for a "long" time
-    std::this_thread::sleep_for(std::chrono::seconds(10));
+    std::this_thread::sleep_for(std::chrono::seconds(3));
   }
   return main_ret;
 }
--- a/rcl/test/rcl/service_fixture.cpp
+++ b/rcl/test/rcl/service_fixture.cpp
@@ -151,7 +151,9 @@ int main(int argc, char ** argv)
     // Our scope exits should take care of fini for everything
     // stick around until we are killed by the client
     // To avoid an infinite loop and signal handling, just sleep for a "long" time
-    std::this_thread::sleep_for(std::chrono::seconds(3));
+    while (true) {
+      std::this_thread::sleep_for(std::chrono::milliseconds(1));
+    }
   }
   return main_ret;
 }
--- a/rcl/test/rcl/service_fixture.cpp
+++ b/rcl/test/rcl/service_fixture.cpp
@@ -147,8 +147,7 @@ int main(int argc, char ** argv)
       return -1;
     }
     // Our scope exits should take care of fini for everything
-    // stick around until we are killed by the client
-    // To avoid an infinite loop and signal handling, just sleep for a "long" time
+    // stick around until launch gives us a signal to exit
     while (true) {
       std::this_thread::sleep_for(std::chrono::milliseconds(1));
     }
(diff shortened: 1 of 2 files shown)


jacquelinekay:
the "while true" loop in the service was leading to some problems with the return code, so I've changed that to a sleep for 5 seconds, which shouldn't give an error if interrupted.

http://ci.ros2.org/job/ci_osx/979/
http://ci.ros2.org/job/ci_windows/1257

I believe the alternative is replicating the signal handling code in rclcpp in the rcl test suite.

rcl/test/rcl/service_fixture.cpp
```diff
@@ -149,6 +149,9 @@ int main(int argc, char ** argv)
       return -1;
     }
     // Our scope exits should take care of fini for everything
+    // stick around until we are killed by the client
+    // To avoid an infinite loop and signal handling, just sleep for a "long" time
+    std::this_thread::sleep_for(std::chrono::seconds(3));
```
wjwwood: Why not `spin` here instead? Or sleep in a loop?

wjwwood:
Why is this sleep needed: https://github.com/ros2/rcl/pull/48/files#diff-dfc3a1157a17788738b1748a8bce57d4R116 ?

rcl/test/rcl/service_fixture.cpp
wjwwood: I guess `spin` is not there.

wjwwood: After discussing with @jacquelinekay offline, we agreed that adding a new exit handler to launch which allows for the "sigterm" return code after the server was interrupted by launch is the right way to go. Then the server can just busy-loop wait forever rather than do a single sleep here.

jacquelinekay:
This change now needs ros2/launch#27

dirk-thomas:
I think the way this uses the exit handlers doesn't work as you expect it to. If the second executable (which is the client) has a a primary exit handler its return code will determine the rc of the launch file. When the client exits it will sigint / sigkill all other processes. But the new `ignore_signal_exit_handler` will override the launch file rc with `0` when it is being signaled - independent of what the intended rc from the primary exit handler is.

jacquelinekay:
I see. If I only want the launch file to return 0 if the client returned zero and the service returned zero or sigint/sigkill, how would I implement this in launch? Can I access the original return code of the other process containing the primary exit handler from the process that is getting shut down? I assume `context.task_state` refers to the task state of the current process.

dirk-thomas:
The *ignore signal* handler should simply not be a primary exit handler, since you want to use it on the non-primary executable. It never needs to change the launch rc directly.

If `context.task_state.returncode` is one of the expected signals then just send that one to `0`. Then do the default exit handling.

The primary exit handler can then still decide to use `ignore_returncode` or not.

jacquelinekay:
Continuing this discussion in ros2/launch#27

jacquelinekay:
I removed the extraneous sleep that @wjwwood pointed out. I don't think it's need since `wait_for_service` is used. From some local testing, the inter-process services test still passes. Testing on OSX since it failed in the earlier link that I gave (before I made changes to the exit handler).

http://ci.ros2.org/job/ci_osx/1004

tfoote:
@jacquelinekay to run CI for other platforms.

jacquelinekay:
http://ci.ros2.org/job/ci_linux/1251
http://ci.ros2.org/job/ci_osx/1006/
http://ci.ros2.org/job/ci_windows/1288/

jacquelinekay:
The changed test passes for all platforms. Therefore I think this is good to merge.

wjwwood:
lgtm, but since @dirk-thomas was previously commenting, I'd wait for feedback from him too.

dirk-thomas:
My comment about the exit handler was addressed within the launch PR. +1

PR #49, opened 2016-04-23
Fix atomics on Windows

jacquelinekay:
in the macros for atomic instructions on Windows, the `object` variable is a pointer to an int, so it will always be 64-bit for 64-bit systems or 32-bit for 32-bit systems. The macros were not defined for 8-bit integer sizes so I added them in here. Also added error-checking for unsupported types.


wjwwood:
lgtm

Issue #28, opened 2016-02-23
Parameters


jacquelinekay:
Implement parameter API for C

jacquelinekay:
This ticket will probably take some more forethought and design than its prerequisite, services (#27).

- [ ] How much parameter support should live in rcl and how much should be implemented in `rclc`?
  - My suspicion is that parameter storage should primarily happen at the client library layer.
- [ ] How to implement parameter nodes vs non-parameter nodes?
  - In rclcpp, all nodes are parameter nodes. A more modular design is desirable.

clalancette:
It looks like #54 is connected into this, and that this is a larger story.  Going to label this as enhancement.

PR #54, opened 2016-04-28
C parameters

jacquelinekay:
**This is currently an API stub, NOT a fully functional implementation.**

I'm submitting this PR to get initial feedback on the approach before I fully implement the API. I'm making an effort to make sure this branch builds and links so that I can also prototype how it will be used in `rclcpp`.

I suggest starting with `test_parameters.cpp`, which lays out how the API calls will be used.

Here are the remaining items to address before I will consider this ready for merging:
- [ ] Prototype usage in `rclcpp` and refine API as needed
- [ ] Add array assign, array copy, and array resize functions in `rosidl_generator_c`
- [ ] Change `rcl_interfaces::msg::ParameterValue` to allow arrays of primitives (bool, int, float). (optional, but I think this would be useful!)
- [x] Finish implementation in this PR
  - Error checking!
  - Sanitize ParameterValue messages by setting unset fields to "NOT_SET"
  - Finish `convert_changes_to_event` convenience function
  - Add better test coverage (e.g. unsetting parameters)

Some questions that came up while prototyping this:
- In the current `rclcpp` implementation, the parameter events topic is global. That is, all nodes publish parameter event changes to the topic `"parameter_events"`. Should the topic instead be node local (e.g.`"node_name__parameter_events"`)? Should this be a configuration option?
- Should ParameterEvent messages be timestamped?
- Should the floating point type, which has an underlying storage type of "double", be called "float" or "double"?
- Should `set_parameters_atomically` be a special case of `set_parameters` or should they use separate services and clients as they do in this current prototype?
- Should there be utilities in `rcl` for managing parameter namespacing? In this implementation, `rcl` currently has no conception of parameter namespaces, it simply passes a "prefixes" option to `list_parameters`.
- Am I abusing macros too much in the .c files? Or am I not using them enough (i.e. should I use them in the header files too)?

Connects to #28

### DIFF
--- a/rcl/include/rcl/parameter.h
+++ b/rcl/include/rcl/parameter.h
@@ -1,0 +1,161 @@
+// Copyright 2016 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCL__PARAMETER_H_
+#define RCL__PARAMETER_H_
+
+#if __cplusplus
+extern "C"
+{
+#endif
+
+#include "rcl_interfaces/msg/parameter.h"
+#include "rcl_interfaces/msg/parameter_event.h"
+
+typedef int rcl_param_action_t;
+#define RCL_GET_PARAMETERS 0
+#define RCL_GET_PARAMETER_TYPES 1
+#define RCL_SET_PARAMETERS 2
+#define RCL_SET_PARAMETERS_ATOMICALLY 3
+#define RCL_LIST_PARAMETERS 4
+#define RCL_NUMBER_OF_PARAMETER_ACTIONS 5
+#define RCL_PARAMETER_ACTION_UNKNOWN 6
+
+/// rcl_parameter_client_set_<TYPE> adds the parameter key, value pair to the
+/// pending set_parameters_request
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_parameter_set_bool(
+  rcl_interfaces__msg__Parameter * parameter, const char * parameter_name, bool value);
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_parameter_set_integer(
+  rcl_interfaces__msg__Parameter * parameter, const char * parameter_name, int value);
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_parameter_set_double(
+  rcl_interfaces__msg__Parameter * parameter, const char * parameter_name, double value);
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_parameter_set_string(
+  rcl_interfaces__msg__Parameter * parameter,
+  const char * parameter_name,
+  const char * value);
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_parameter_set_bytes(
+  rcl_interfaces__msg__Parameter * parameter,
+  const char * parameter_name,
+  const uint8_t * value);
+
+// etc. for all types
+
+/// Get the boolean value of the parameter, or return error if the parameter is not a boolean.
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_parameter_get_bool(const rcl_interfaces__msg__Parameter * parameter, bool * output);
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_parameter_get_int(const rcl_interfaces__msg__Parameter * parameter, int * output);
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_parameter_get_double(const rcl_interfaces__msg__Parameter * parameter, double * output);
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_parameter_get_string(
+  const rcl_interfaces__msg__Parameter * parameter, rosidl_generator_c__String * output);
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_parameter_get_bytes(
+  const rcl_interfaces__msg__Parameter * parameter, rosidl_generator_c__byte__Array * output);
+
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_parameter_set_value_bool(
+  rcl_interfaces__msg__ParameterValue * parameter_value, bool value);
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_parameter_set_value_integer(
+  rcl_interfaces__msg__ParameterValue * parameter_value, int value);
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_parameter_set_value_double(
+  rcl_interfaces__msg__ParameterValue * parameter_value, double value);
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_parameter_set_value_string(
+  rcl_interfaces__msg__ParameterValue * parameter_value,
+  const char * value);
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_parameter_set_value_bytes(
+  rcl_interfaces__msg__ParameterValue * parameter_value,
+  const uint8_t * value);
+
+// etc. for all types
+
+// Some other convenience function ideas:
+// Get type enum from a ParameterValue
+// rcl_parameter_get_type(rcl_interfaces__msg__ParameterValue * parameter_value, uint8 type)
+
+// Get type enum from a Parameter
+// rcl_parameter_get_type(rcl_interfaces__msg__Parameter * parameter, uint8 type)
+
+bool rcl_parameter_value_compare(
+  const rcl_interfaces__msg__ParameterValue * parameter1,
+  const rcl_interfaces__msg__ParameterValue * parameter2);
+
+// diff the new state and the old state of parameters to populate a ParameterEvent message struct
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_parameter_convert_changes_to_event(
+  const rcl_interfaces__msg__Parameter__Array * prior_state,
+  const rcl_interfaces__msg__Parameter__Array * new_state,
+  rcl_interfaces__msg__ParameterEvent * parameter_event);
+
+#if __cplusplus
+}
+#endif
+
+#endif  // RCL__PARAMETER_H_
--- a/rcl/include/rcl/parameter_client.h
+++ b/rcl/include/rcl/parameter_client.h
@@ -1,0 +1,182 @@
+// Copyright 2016 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCL__PARAMETER_CLIENT_H_
+#define RCL__PARAMETER_CLIENT_H_
+
+#if __cplusplus
+extern "C"
+{
+#endif
+
+#include <rcl_interfaces/msg/list_parameters_result__struct.h>
+#include <rcl_interfaces/msg/parameter__struct.h>
+#include <rcl_interfaces/msg/parameter_event__struct.h>
+#include <rcl_interfaces/msg/set_parameters_result__struct.h>
+
+#include "rcl/client.h"
+#include "rcl/node.h"
+#include "rcl/parameter.h"
+#include "rcl/wait.h"
+
+/// TODO: documentation!!!
+
+/// Internal rcl implementation struct
+struct rcl_parameter_client_impl_t;
+
+/// There is no sync/async parameter client distinction in rcl.
+typedef struct rcl_parameter_client_t
+{
+  struct rcl_parameter_client_impl_t * impl;
+} rcl_parameter_client_t;
+
+typedef struct rcl_parameter_client_options_t
+{
+  // quality of service settings for all parameter-related services
+  rmw_qos_profile_t qos;
+  rcl_allocator_t allocator;
+  char * remote_node_name;
+  rmw_qos_profile_t parameter_event_qos;
+} rcl_parameter_client_options_t;
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_parameter_client_options_t
+rcl_parameter_client_get_default_options(void);
+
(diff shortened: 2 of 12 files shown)


rcl/test/rcl/test_parameters.cpp
```diff
@@ -0,0 +1,383 @@
...
+    set_on_unexpected_free_callback(nullptr);
+    rcl_ret_t ret = rcl_node_fini(this->node_ptr);
+    delete this->node_ptr;
+
+
+    ret = rcl_wait_set_fini(this->wait_set);
+    delete this->wait_set;
+
+    ret = rcl_parameter_service_fini(this->parameter_service);
+    delete this->parameter_service;
```
wjwwood: You should probably delete these in reverse order, i.e. you should probably fini and delete the parameter service before the node.

rcl/include/rcl/parameter.h
```diff
@@ -0,0 +1,177 @@
...
+rcl_parameter_set_ ## TYPE ## ( \
+  rcl_interfaces__msg__Parameter * parameter, const char * name, CTYPE(TYPE) value); \
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_parameter_set_bool(
+  rcl_interfaces__msg__Parameter * parameter, const char * parameter_name, bool value);
+
+// Or, with the above macro, it would simply be RCL_DECLARE_SET_PARAMETER_REQUEST(bool)
```
wjwwood: If there are only five variants (that's how many I count anyways) then maybe the macro isn't necessary for the function prototypes. In the header I think it would be best to avoid MACRO's for documentation purposes. In the C files where common lines of code should automatically distribute fixes (fix one line, fix all the set or get functions), I think the macro's make more sense.

jacquelinekay: I agree, that's why I opted for writing out all the headers. I'll delete the macro then.

rcl/include/rcl/parameter.h
```diff
@@ -0,0 +1,177 @@
...
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_parameter_set_integer(
+  rcl_interfaces__msg__Parameter * parameter, const char * parameter_name, int value);
+
+// can't distinguish between floats and doubles. should use doubles underneath for more precision.
+// But should we still call them floats?
```
wjwwood: If the underlying storage is always a double, then I think we should only provide a double interface. That forces the user to convert their floats to doubles themselves and deal with any conversion issues that might come with that. It's more important for `get_float` which might require us to truncate the stored double. Instead of internalizing that issue, we should only return a double and the user can decide how to cast it into a float if required.

wjwwood:
For the functions in `parameter.h`, I think you should consider type erasing the set and get functions. For example:

```c++
rcl_ret_t
rcl_parameter_set_integer(
  void * parameter,
  rcl_parameter_type_support_t * ts,
  const char * parameter_name,
  int value);
```

Where the `rcl_parameter_type_support_t` struct would look like this:

```c++
typedef struct rcl_parameter_type_support_t
{
  rcl_ret_t (*set_integer) (void * parameter, const char * name, int value);
  rcl_ret_t (*set_double) (void * parameter, const char * name, double value);
  // And so on...
} rcl_parameter_type_support_t;
```

There would be a version of this structure for each language supported and could live in a static (not generated) header in the `rcl_interfaces` package.

What this lets you do is use either C or C++ versions of the Parameter messages, for example C++:

```c++
#include <rcl_interfaces/msg/Parameter.hpp>
#include <rcl_interfaces/cpp/type_support.hpp>

rcl_interfaces::msg::Parameter param;
rcl_parameter_set_integer(
  &param,
  rcl_interfaces::get_cpp_parameter_type_support(),
  "foo", 42);
```

And the same for C using the same `rcl_parameter*` functions:

```c++
#include <rcl_interfaces/msg/Parameter.hpp>
#include <rcl_interfaces/c/type_support.h>

rcl_interfaces__msg__Parameter param;
rcl_parameter_set_integer(
  &param,
  rcl_interfaces_get_c_parameter_type_support(),
  "foo", 42);
```

And in the `rcl_parameter_set_integer` function you'd do something like this:

```c++
rcl_ret_t
rcl_parameter_set_integer(
  void * param,
  rcl_parameter_type_support_t * ts,
  const char * name,
  int value)
{
  return ts->set_integer(param, name, value);
}
```

This is all sort of roundabout, but allows you to add new language types (maybe `PyObj *` for Python or `Local<Object>` for NodeJS) without modifying `rcl`'s API. The alternative is to unroll this and have `get_int`/`set_int` for functions in `rcl` for each supported language. In this pr only C is support, but you'd at least need to expand that to include C++ I think.

I suppose it's also worth taking a step back to consider if it is necessary to support more than just C messages in this parameter API. I don't have a high enough level view yet to say one way or the other.

wjwwood:
For the functions in `parameter.h` and my last comment those could almost just go into `rcl_interfaces` completely, since they are just operations on the data structures themselves.

However, I think my above argument about using type erasure and type support structures to delegate work for different types applies more to functions like this:

```c++
RCL_PUBLIC
RCL_WARN_UNUSED
rcl_ret_t
rcl_parameter_client_take_set_response(
  const rcl_parameter_client_t * parameter_client,
  rmw_request_id_t * request_header,
  rcl_interfaces__msg__SetParametersResult__Array * set_parameters_result);
```

This function only supports the C type, and as you mentioned offline, it would be nice to use this interface in `rclcpp` with the C++ version of the parameter messages. You could imagine something more like this:

```c++
RCL_PUBLIC
RCL_WARN_UNUSED
rcl_ret_t
rcl_parameter_client_take_set_response(
  const rcl_parameter_client_t * parameter_client,
  rmw_request_id_t * request_header,
  void * set_parameters_result,
  rcl_message_type_support_t * set_parameters_result_ts);
```

Not sure about the types and stuff, but you get the idea. In this way these functions would emulate the underlying functions like `publish`, `take`, `send_request`, `take_response`, and so on.

rcl/include/rcl/parameter_client.h
```diff
@@ -0,0 +1,185 @@
...
+
+/// Internal rcl implementation struct
+struct rcl_parameter_client_impl_t;
+
+/// There is no sync/async parameter client distinction in rcl.
+typedef struct rcl_parameter_client_t
+{
+  // the parameter client provides storage and utility functions for parameter event subscription
+  // Should instead have a create_parameter_event_subscription function for this parameter client?
+  // rcl_subscription_t * parameter_event_subscription;
```
wjwwood: Naively, I'd say that the fact that a subscription is involved is an implementation detail of parameters. A related question is whether or not the user can get a handle on the subscription object that the client uses or change the QoS options of that subscription.

If we don't expose these "details" of how parameters work, then I think it's clear that the subscription should be created and destroyed within `rcl` without extra action from the client library.

I think you've already done the other necessary bits to "abstract" the implementation of the parameter client and server from the user of `rcl` with `rcl_wait_set_add_parameter_client` and the matching `take` functions.

More to your question, can you describe in more detail what `create_parameter_event_subscription` would do and what the alternative default behavior is?

rcl/include/rcl/parameter_client.h
```diff
@@ -0,0 +1,185 @@
...
+rcl_parameter_client_options_t
+rcl_parameter_client_get_default_options(void);
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_parameter_client_t
+rcl_get_zero_initialized_parameter_client(void);
+
+
+// Should pointers to the internal entities be allocated outside or...
```
wjwwood: Can you expand on what "internal entities" are?

rcl/src/rcl/parameter.c
```diff
@@ -0,0 +1,96 @@
...
+#endif
+
+#include <rosidl_generator_c/string_functions.h>
+#include <rcl_interfaces/msg/parameter__struct.h>
+#include <rcl_interfaces/msg/parameter_event__struct.h>
+#include <rcl_interfaces/msg/parameter_type__functions.h>
+
+#include "rcl/types.h"
+
+// TODO: Can we assume that other ParameterValue fields are set to NOT_ASSIGNED or should we reassign them?
```
wjwwood: I guess this is a question about the "default" constructor of the ParameterValue's generated structure and what it sets its values to initially? I suppose you should be able to specify the default values in the message definition if the implicit behavior is not correct.

rcl/src/rcl/parameter_client.c
```diff
@@ -0,0 +1,301 @@
...
+  rcl_node_t * node;
+  rcl_client_t get_client;
+  rcl_client_t get_types_client;  // Also referred to as describe parameters in docs
+  rcl_client_t set_client;
+  rcl_client_t set_atomically_client;
+  rcl_client_t list_client;
+
+  rcl_subscription_t event_subscription;
+
+  // Storage for requests/responses for each client
```
wjwwood: Is this storage necessary? Whenever these events are taken the "taker" should pass in storage shouldn't they? Maybe I'm missing part of the puzzle here...

rcl/src/rcl/parameter_client.c
```diff
@@ -0,0 +1,301 @@
...
+  // Initialize all clients in impl storage
+  RCL_PARAMETER_INITIALIZE_CLIENT(get, GetParameters, "__get_parameters");
+  RCL_PARAMETER_INITIALIZE_CLIENT(get_types, GetParameterTypes, "__get_parameter_types");
+  RCL_PARAMETER_INITIALIZE_CLIENT(set, SetParameters, "__set_parameters");
+  RCL_PARAMETER_INITIALIZE_CLIENT(set_atomically, SetParametersAtomically, "__set_parameters_atomically");
+  RCL_PARAMETER_INITIALIZE_CLIENT(list, ListParameters, "__list_parameters");
+
+  const rosidl_message_type_support_t * event_ts = ROSIDL_GET_TYPE_SUPPORT(
+    rcl_interfaces, msg, ParameterEvent);
+  // TODO Is the parameter_event topic name global or node local?
```
wjwwood: Global I think, what does "node local" mean?

wjwwood:
Ok, I've finished my first review. I tried to give some useful feedback on any questions I saw and raise questions where I didn't understand, but now I'm just hoping it was more help than hindrance 😄.

rcl/src/rcl/parameter.c
jacquelinekay: It would be a defensive measure to set all the unused fields to NOT_ASSIGNED whenever setting a type field, to enforce that only one type is set in a ParameterValue. Someone could arbitrarily set the fields of struct at any point, which is why I think this is a good idea.

rcl/src/rcl/parameter_client.c
jacquelinekay: Summarizing from our conversation offline: this comment is trying to ask whether or not the `parameter_events` topic should be shared among all parameter nodes, or if each node should have its own associated "`<node_name>__parameter_events`" topic. In the current implementation, there is one `parameter_events` topic for all nodes. I think it makes more sense to have a namespaced parameter_events topic to partition the parameter events to a particular node.

In the current implementation, all nodes will receive callbacks on all parameter events, and there is no way of distinguishing which node the parameter events is for, because parameter keys are not namespaced, and the ParameterEvent message definition contains no field for the node name.

I haven't addressed the logic of namespacing parameter keys yet. I will probably have more thoughts about that as the prototyping continues.

rcl/src/rcl/parameter_client.c
jacquelinekay: Basically I did this so that the parameter client could internally handle filling and accessing the request/response structures so that the API directly accesses the fields that are interesting, and the client library calling rcl doesn't need to instantiate storage for this. But, if I change the take/send calls of the rcl parameter API to be type-erased, then this probably won't be relevant anymore.

rcl/src/rcl/parameter_client.c
tfoote: Summary from meeting, we'll table this discussion as to not hold up this pr. I'd suggest opening a new issue to discuss it.

rcl/src/rcl/parameter.c
```diff
@@ -0,0 +1,263 @@
...
+
+  size_t new_array_idx = 0;
+  size_t deleted_idx = 0;
+  size_t changed_idx = 0;
+
+  for (prior_idx = 0; prior_idx < prior_state->size; ++prior_idx) {
+    rcl_interfaces__msg__Parameter * prior_entry = &prior_state->data[prior_idx];
+    for (new_idx = 0; new_idx < new_state->size; ++new_idx) {
+      rcl_interfaces__msg__Parameter * new_entry = &new_state->data[new_idx];
+      // If the parameter has the same name but a different type or value, count it as changed.
```
mikaelarguedas: Here we assume uniqueness of parameter `name` in parameter arrays. Is there a place where we prevent user to have duplicated parameter names in a given `Parameter__Array` ?
For example if one append a new parameter with the same name at the end of the array but different type or value, this parameter won't appear in the parameter event message (given that it wont be catched neither as a changed one nor a new one)
I agree that parameter names should be unique for a given node, I just can't find a place where it's enforced.

jacquelinekay: Where would you suggest adding the check for uniqueness of parameter keys?

rcl/src/rcl/parameter.c
```diff
@@ -0,0 +1,263 @@
...
+
+  rcl_interfaces__msg__Parameter__Array__fini(&parameter_event->deleted_parameters);
+  rcl_interfaces__msg__Parameter__Array__fini(&parameter_event->changed_parameters);
+  rcl_interfaces__msg__Parameter__Array__fini(&parameter_event->new_parameters);
+
+  rcl_interfaces__msg__Parameter__Array__init(&parameter_event->deleted_parameters,
+    num_deleted_params);
+  rcl_interfaces__msg__Parameter__Array__init(&parameter_event->changed_parameters,
+    num_changed_params);
+  rcl_interfaces__msg__Parameter__Array__init(&parameter_event->new_parameters, num_new_params);
```
mikaelarguedas: Should we check for return values of these `rcl_interfaces__msg__Parameter__Array__init` calls in case memory cannot be allocated ?

rcl/src/rcl/parameter.c
mikaelarguedas: I'm not sure what the right place would be. I have trouble to figure the scope of this feature. If it's going to be only used for parameters and nowhere else I would put it in the [send request client macro](https://github.com/ros2/rcl/pull/54/files#diff-e5b489889a483f301eea46906d80e9e6R258) so that the client ensure the integrity of the data before interacting with the server and the server never has to deal with duplicates

rcl/src/rcl/parameter_client.c
```diff
@@ -0,0 +1,437 @@
...
+          client_ptr = &parameter_client->impl->list_client;
+          break;
+        default:
+          return RCL_RET_ERROR;
+      }
+      if (client_ptr == wait_set->clients[i]) {
+        return RCL_RET_OK;
+      }
+    }
+    *action = RCL_PARAMETER_ACTION_UNKNOWN;
```
mikaelarguedas: I don't know if this is relevant ; using `actions` here can get confusing if we implement ROS1-like actions in the future in rcl.

jacquelinekay:
bf055492685576777fc29c95b50cc50484bd107c will ignore the parameters test for Connext Dynamic. With the `fix_cpp_convert` branches the test should pass for other middlewares.

rcl/test/rcl/test_service.cpp
```diff
@@ -22,6 +22,9 @@
 
 #include "rcl/rcl.h"
 
+#include "rcl_interfaces/srv/set_parameters.h"
+#include "rcl_interfaces/msg/set_parameters_result.h"
+#include "rcl_interfaces/msg/parameter.h"
```
wjwwood: I don't think these are used.

rcl/test/rcl/test_subscription.cpp
```diff
@@ -213,7 +213,7 @@ TEST_F(CLASSNAME(TestSubscriptionFixture, RMW_IMPLEMENTATION), test_subscription
     ret = rcl_publish(&publisher, &msg);
     // TODO(wjwwood): re-enable this fini when ownership of the string is resolved.
     //                currently with Connext we will spuriously get a double free here.
-    // std_msgs__msg__String__fini(&msg);
+    std_msgs__msg__String__fini(&msg);
```
wjwwood: I had forgotten about this. Did you have to do anything else to make this work (some other patch somewhere)?

jacquelinekay: Yeah, I believe I do this in this pull request: https://github.com/ros2/rmw_connext/pull/164/files#diff-d3deaebdd6625290a633bebc6ead9d88R232

ivanpauno:
Closing, as it's a bit outdated (same as https://github.com/ros2/rclcpp/pull/276#issuecomment-621480064).

Issue #52, opened 2016-04-26
Linking errors for client_fixture and service_fixture tests


dhood:
The `service_fixture` and `client_fixture` tests are not linking properly on my machine. The rest work fine.

```
Linking CXX executable client_fixture__rmw_connext_cpp
cd /home/dhood/src/rcl/build/rcl/test && /usr/bin/cmake -E cmake_link_script CMakeFiles/client_fixture__rmw_connext_cpp.dir/link.txt --verbose=1
[...]

PR #55, opened 2016-04-28
Workaround for the change in libnddscore.so in connext 5.2.0 between gcc4.6.3 and gcc4.8.2

dhood:
Fixes https://github.com/ros2/rcl/issues/52, which I currently understand to be caused by the connext 5.2.0 package labelled as built with gcc4.8.2 not exporting a dependency on `dl` (in contrast to the one labelled with gcc4.6.3).

The executables do have `-ldl` in the linking stage, but `--no-as-needed` is needed to ensure that the linking actually happens.

Adding a `dlopen(NULL, RTLD_NOW);` call to the `service_fixture` and `client_fixture` files is an alternative workaround because it forces `dl` to not be considered unneeded. 

No regressions that I can see
http://ci.ros2.org/job/ci_linux/1243/
http://ci.ros2.org/job/ci_windows/1281/
http://ci.ros2.org/job/ci_osx/998/


rcl/test/CMakeLists.txt
```diff
@@ -139,4 +146,19 @@ function(test_target_function)
   )
 endfunction()
 
+macro(connext_workaround target)
+  if(NOT "${ARGN} " STREQUAL " ")
+    message(FATAL_ERROR "append_link_flags() called with "
+      "unused arguments: ${ARGN}")
+  endif()
```
dirk-thomas: Since this is a local macro which is not used externally I would not bother with this check and save the four lines.

rcl/test/CMakeLists.txt
```diff
@@ -139,4 +146,19 @@ function(test_target_function)
...
+  if("${CMAKE_CXX_COMPILER_ID} " STREQUAL "GNU ")
+    get_target_property(current_link_flags target LINK_FLAGS)
+    set(link_flags "-Wl,--no-as-needed")
+    if(current_link_flags)
+      set(link_flags ${current_link_flags} ${link_flags})
+    endif()
+    set_target_properties(${target} PROPERTIES LINK_FLAGS "${link_flags}")
+  endif()
```
dirk-thomas: Macros should prefix all local variable with an underscore to avoid breaking variables in the parent scope.

dhood: thanks - addressed

dirk-thomas:
lgtm

wjwwood:
lgtm too

dirk-thomas:
With Connext 5.2.3 on Xenial I needed the same patch for these targets (note all failing targets are ending with `rmw_fastrtps_cpp`):

* intra_process_demo
  * cyclic_pipeline__rmw_fastrtps_cpp
  * two_node_pipeline__rmw_fastrtps_cpp
* rclcpp_examples
  * list_parameters_async__rmw_fastrtps_cpp
  * parameter_events_async__rmw_fastrtps_cpp
* test_communication
  * test_replier_cpp__rmw_fastrtps_cpp
  * test_subscriber_cpp__rmw_fastrtps_cpp
  * test_subscription_valid_data_cpp__rmw_fastrtps_cpp
* test_rclcpp
  * gtest_multithreaded__rmw_fastrtps_cpp
  * gtest_subscription__rmw_fastrtps_cpp
  * test_services_client_cpp__rmw_fastrtps_cpp
* tlsf_cpp
  * test_tlsf__rmw_fastrtps_cpp
* pendulum_contol
  * pendulum_demo__rmw_fastrtps_cpp
  * pendulum_logger__rmw_fastrtps_cpp
  * pendulum_teleop__rmw_fastrtps_cpp

I don't know why only for those targets are affected and not for the others. And therefore I hesitate to apply the same patch in these cases.

Since the RTI Connext release notes mentioned that `--no-as-needed` is necessary I am leaning to expose this in `connext_cmake_module` and make sure it is being applied everywhere.

PR #57, opened 2016-05-03
add graph functions to support wait_for_service

wjwwood:
Connects to ros2/ros2#215

### DIFF
--- a/rcl/include/rcl/node.h
+++ b/rcl/include/rcl/node.h
@@ -143,6 +143,37 @@ RCL_PUBLIC
 rcl_node_options_t
 rcl_node_get_default_options(void);
 
+/// Return true if the node is valid, else false.
+/* Also return false if the node pointer is nullptr.
+ *
+ * A node is invalid if:
+ *   - the implementation is null (rcl_node_init not called or failed)
+ *   - rcl_shutdown has been called since the node has been initialized
+ *   - the node has been finalized with rcl_node_fini
+ *
+ * There is a possible validity race condition.
+ * Consider:
+ *
+ *   assert(rcl_node_is_valid(node));  <-- thread 1
+ *   rcl_shutdown();                   <-- thread 2
+ *   // use node as if valid           <-- thread 1
+ *
+ * In the third line the node is now invalid, even though on the previous line
+ * of thread 1 it was checked to be valid.
+ * This is why this function is considered not thread-safe.
+ *
+ * This function does not manipulate heap memory.
+ * This function is not thread-safe.
+ * This function is lock-free so long as the C11's stdatomic.h function
+ * atomic_is_lock_free() returns true for atomic_uint_least64_t.
+ *
+ * \param[in] node handle to the node to validated
+ * \return true if the node is valid, otherwise false.
+ */
+RCL_PUBLIC
+bool
+rcl_node_is_valid(const rcl_node_t * node);
+
 /// Get the name of the node.
 /* This function returns the node's internal name string.
  * This function can fail, and therefore return NULL, if:
--- a/rcl/include/rcl/node.h
+++ b/rcl/include/rcl/node.h
@@ -29,6 +29,7 @@ extern "C"
 
 #define RCL_NODE_OPTIONS_DEFAULT_DOMAIN_ID SIZE_MAX
 
+struct rcl_guard_condition_t;
 struct rcl_node_impl_t;
 
 /// Handle for a ROS node.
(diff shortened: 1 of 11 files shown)


jacquelinekay:
Do you think it would be possible to push more logic from rmw into rcl?

wjwwood:
In which functions? I don't have any ideas.

jacquelinekay:
Anything in the API that is left up to the rmw implementations means that the code has to be written and maintained `n` times for each rmw implementation, which is why I'm generally interested in looking into making rcl do work that rmw would do otherwise. So, the fact that `rcl_get_topic_names_and_types`, `rcl_destroy_topic_names_and_types`, `rcl_count_publishers`, `rcl_count_subscribers`, and `rcl_service_server_is_available` bothers me a little, but I also don't have any good suggestions for changing it. Maybe once the full implementation is available I'll have more feedback.

dirk-thomas:
Each rmw impl. needs to be able to provide its own impl. of these functions since some vendors have very efficient (non standardized) functions for that.

If multiple rmw impl. end up using the same algo to collect the information based on other api that can easily be refactored into a single place.

wjwwood:
Yeah, unfortunately I think @dirk-thomas is right about the topic listing and pub/sub counting. It's also true for the service is ready call, which might be implemented different ways in each rmw impl.

wjwwood:
I pushed some new commits which add a function to create a `rcl_guard_condition_t` from an existing `rmw_guard_condition_t`, use that function in cread node to automatically create a graph guard condition which is used to get notified about graph changes when using `rcl_wait`, and some basic tests.

rcl/test/rcl/test_graph.cpp
```diff
@@ -0,0 +1,153 @@
...
+      time_to_sleep = min_sleep;
+    }
+    ret = rcl_wait_set_clear_guard_conditions(&wait_set);
+    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
+    ret = rcl_wait_set_add_guard_condition(&wait_set, graph_guard_condition);
+    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
+    printf("waiting up to %lld nanoseconds for graph changes\n", time_to_sleep.count());
+    ret = rcl_wait(&wait_set, time_to_sleep.count());
+    if (ret == RCL_RET_TIMEOUT) {
+      printf("timeout!\n");
```
jacquelinekay: Is there something more useful we can do here than printing? It's ambiguous to me if this expected behavior or not. Should we expect `rcl_wait` to spuriously time out in this case?

wjwwood: Sorry, that was left over from debug. I removed it. There's an identical check one below.

Incidentally, I have a few more changes for this pr.

wjwwood:
Ok, this pr is also ready for review.

wjwwood:
In the most recent commits I added a function called `rcl_node_is_valid` and lots of tests for the graph API.

rcl/test/rcl/test_graph.cpp
```diff
@@ -0,0 +1,536 @@
...
+      continue;
+    }
+    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
+    ret = rcl_service_server_is_available(this->node_ptr, &client, &is_available);
+    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
+    if (is_available) {
+      break;
+    }
+  }
+  ASSERT_TRUE(is_available);
```
tfoote: One simple extension to this test would be to tear down the service and assert is_available goes back to false to make sure all state is cleared.

tfoote:
+1 the extra test case would be great if you have time, but is not necessary

rcl/include/rcl/node.h
```diff
@@ -263,6 +295,33 @@ RCL_WARN_UNUSED
...
+/// Return a guard condition which is triggered when the ROS graph changes.
+/* The handle returned is a pointer to an internally held rcl guard condition.
+ * This function can fail, and therefore return NULL, if:
+ *   - node is NULL
+ *   - node is invalid
+ *
+ * The returned handle is made invalid if the node is finialized or if
+ * rcl_shutdown() is called.
+ *
+ * The guard condition will be triggered anytime change to the ROS graph occurs.
```
dhood: anytime change -> anytime a change

rcl/test/rcl/test_graph.cpp
wjwwood: Good idea, I updated this test to do that: 631898e

rcl/include/rcl/node.h
wjwwood: Thanks: 77d9666

PR #59, opened 2016-06-02
Clarify wait concurrency

wjwwood:
Previously the thread-safety rules around `rcl_wait` were unclear.

I categorically eliminated "thread-safe, even for the same rcl_wait_set_t instance", since it would mean concurrent read and write to the contents. 

I tested "thread-safe for different wait sets, but with shared content". I created two wait sets and one guard condition, added the guard condition to both, and waited on the wait sets each in their own thread while I triggered the guard condition in the main thread. I observed (with opensplice) that only one of the wait's would wake up per trigger, i.e. equivalent to `notify_one()` versus `notify_all()`. So that's a problem we can't work around. Another issue is that `rcl_wait` will reset any triggered guard conditions after waking up from the underlying DDS implementation. So the only way to know if a particular guard condition was triggered is to check to see if the pointer to it is null or not after `rcl_wait` wakes up. The problem is that when the guard condition is shared, there is a race condition between the DDS wait waking up and the implementation of `rmw_wait` checking to see if the condition is set. Based on those issues, and in favor of simplicity I excluded this as an option for the rule too.

That left me with "thread-safe for unique wait set instances and unique contents". `rcl_wait` does not guarantee that the wait sets are unique (that's left up to the caller to ensure), but waiting on more than one wait set concurrently which share state is declared as undefined behavior.

This pr, states this in the API documentation for `rcl_wait` and adds a test to see if that actually works as well (waiting concurrently on multiple wait sets, each with their own guard conditions).

The ramification of this is that the `rcl_node_get_graph_guard_condition` will probably need to return more than one graph guard condition object so that multiple locations can wait on their own copy. When the graph changes, each guard condition out on loan will need to be triggered. The other option is to just return the one, but make the client libraries only use one at a time. My impression is that will be overly constraining for at least rclcpp. In particular if you want to be able to wait on graph state changes without spinning and at the same time as spinning (in that case both the executor and the wait function would need their own graph guard condition).

I'll look into that tomorrow, but either way, in my opinion, this pr is still valid and ready for review.

### DIFF
--- a/rcl/include/rcl/wait.h
+++ b/rcl/include/rcl/wait.h
@@ -513,11 +513,11 @@ rcl_wait_set_resize_services(rcl_wait_set_t * wait_set, size_t size);
  * comes first.
  * Passing a timeout struct with uninitialized memory is undefined behavior.
  *
- * \TODO(wjwwood) this function should probably be thread-safe with itself but
- *                it's not clear to me what happens if the wait sets being
- *                waited on can be overlapping or not or if we can even check.
- * This function is not thread-safe and cannot be called concurrently, even if
- * the given wait sets are not the same and non-overlapping in contents.
+ * This function is thread-safe for unique wait sets with unique contents.
+ * This function cannot operate on the same wait set in multiple threads, and
+ * the wait sets may not share content.
+ * For example, calling rcl_wait in two threads on two different wait sets that
+ * both contain a single, shared guard condition is undefined behavior.
  *
  * \param[inout] wait_set the set of things to be waited on and to be pruned if not ready
  * \param[in] timeout the duration to wait for the wait set to be ready, in nanoseconds
--- a/rcl/src/rcl/wait.c
+++ b/rcl/src/rcl/wait.c
@@ -505,8 +505,12 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
     RCL_SET_ERROR_MSG("wait set is invalid");
     return RCL_RET_WAIT_SET_INVALID;
   }
-  if (wait_set->size_of_subscriptions == 0 && wait_set->size_of_guard_conditions == 0 &&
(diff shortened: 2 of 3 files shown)


wjwwood:
For reference, the DDS specification is a little vague about this:

> set_trigger_value

> This operation sets the trigger_value of the GuardCondition. WaitSet objects behavior depends on the changes of the trigger_value of their attached conditions. Therefore, any WaitSet to which is attached the GuardCondition is potentially affected by this operation.

So from that it sounds like you could have "notify all" or "notify one" or none based on the trigger value.

dirk-thomas:
LGTM

wjwwood:
Thanks for the review. I'll start CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1381/)
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1078/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/1392/)

wjwwood:
New Linux CI with the style issues addressed:
- [Build Status](http://ci.ros2.org/job/ci_linux/1382/)

wjwwood:
New CI with (9f57dba) explicit capture of const local variables in the lambda's (Windows doesn't allow implicit):
- [Build Status](http://ci.ros2.org/job/ci_windows/1394/)

wjwwood:
New OS X CI (with style fixes) since it'll still finish before Windows:

- [Build Status](http://ci.ros2.org/job/ci_osx/1079/)

wjwwood:
CI lgtm, merging.

wjwwood:
After more consideration, I think we can live with a single graph guard condition per node. It may require the use of condition variables and/or additional threads in the client library, but I think it's worth some extra complexity there to keep rmw and rcl simple and lock free.

PR #60, opened 2016-06-06
Add sleep_for utility and use it to correct flaky Windows test

jacquelinekay:
We previously discussed problems with `std::this_thread::sleep` waking up spuriously on Windows. In this PR I add a spinning wait as a utility in the rcl tests, use it in the trigger thread, use `std::this_thread::yield` before going to sleep to provide hints to the Windows thread scheduler for when to context switch, and increase the trigger period. 

CI jobs with arguments `--end-with rcl --ctest-args --repeat-until-fail 20 -R test_wait__rmw_opensplice_cpp --`

Windows, Build type=None
[Build Status](http://ci.ros2.org/job/ci_windows/1412/)

Windows, Build type=Release
[Build Status](http://ci.ros2.org/job/ci_windows/1413/)

(problem is not present on Debug job)


rcl/test/rcl/test_wait.cpp
```diff
@@ -178,7 +179,9 @@ TEST(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), multi_wait_set_threaded)
             ss << "[thread " << test_set.thread_id << "] Waiting (try #" << wake_try_count << ")";
             // printf("%s\n", ss.str().c_str());
           }
+          std::this_thread::yield();
           ret = rcl_wait(&test_set.wait_set, wait_period);
+          ASSERT_NE(ret, RCL_RET_ERROR);
```
wjwwood: I'd move this into the else section of the `if (ret != RCL_RET_TIMEOUT)` and change it to `ASSERT_EQ(ret, RCL_RET_OK)`.

rcl/test/rcl/test_wait.cpp
```diff
@@ -178,7 +179,9 @@ TEST(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), multi_wait_set_threaded)
             ss << "[thread " << test_set.thread_id << "] Waiting (try #" << wake_try_count << ")";
             // printf("%s\n", ss.str().c_str());
           }
+          std::this_thread::yield();
```
wjwwood: Why is this needed? Shouldn't the wait function do this at the bottom for us?

rcl/test/rcl/test_wait.cpp
```diff
@@ -269,15 +273,16 @@ TEST(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), multi_wait_set_threaded)
...
   while (loop_test()) {
     loop_count++;
-    // print_state("triggering, current states: ");
     for (const auto & test_set : test_sets) {
       ret = rcl_trigger_guard_condition(&test_set.guard_condition);
       EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
     }
-    std::this_thread::sleep_for(trigger_period);
+    std::this_thread::yield();
+    rcl_test::sleep_for(std::chrono::duration_cast<std::chrono::nanoseconds>(trigger_period));
```
wjwwood: `sleep_for` should not wake up spuriously: http://stackoverflow.com/questions/30538508/can-stdthis-threadsleep-for-have-spurious-wakeups

We had a problem with `std::condition_variable::wait_for` waking up spuriously, but that's documented behavior (that's why you have a predicate in condition variable).

rcl/test/rcl/test_wait.cpp
```diff
@@ -22,6 +22,7 @@
 #include "gtest/gtest.h"
 
 #include "../scope_exit.hpp"
+#include "../sleep_for.hpp"
```
wjwwood: I don't think you included this file.

rcl/test/rcl/test_wait.cpp
```diff
@@ -269,15 +273,16 @@ TEST(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), multi_wait_set_threaded)
...
   size_t loop_count = 0;
   while (loop_test()) {
     loop_count++;
-    // print_state("triggering, current states: ");
     for (const auto & test_set : test_sets) {
       ret = rcl_trigger_guard_condition(&test_set.guard_condition);
       EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
     }
-    std::this_thread::sleep_for(trigger_period);
+    std::this_thread::yield();
```
wjwwood: Again, not sure why we need this when we're sleeping. Usually you use `yield` when doing a busy wait, like this:

```c++
while (!condition) {
  std::this_thread::yield();
}
```

So that your busy wait doesn't starve the other threads. I think `yield` is basically implied in other blocking system calls like sleep, select, and in our case probably the DDS wait function too.

jacquelinekay: http://stackoverflow.com/questions/35741656/stdthis-threadsleep-for-sleeps-shorter-than-expected-in-vs2015/35743145

wjwwood: That answer is talking about the fact that the OP is using a different method to compare to sleep_for and the accuracies might be different. However, the reference website (and likely the standard) says that `sleep_for` uses a steady clock to gage the duration and is very specific that it should sleep for *at least* the requested period.

From http://en.cppreference.com/w/cpp/thread/sleep_for:
> Blocks the execution of the current thread for at least the specified sleep_duration.
A steady clock is used to measure the duration. This function may block for longer than sleep_duration due to scheduling or resource contention delays.

The SO link you sent also points out that the `sleep_for` function on Windows is already a loop around `Sleep` function to handle spurious wake ups.

jacquelinekay:
ok, through instrumenting the call to sleep_for I convinced myself that a custom sleep_for is not needed, so I will remove the sleep_for.hpp header.

In order to convince myself one way or the other about `yield`, I collected a small amount of data (16 runs). I timed how long the entire test took when "wait_period" was equal to 5ms and "trigger_period" was equal to 10ms. @wjwwood's hypothesis was that calls to `yield` were introducing a delay that happened to make the test pass and therefore aren't needed.

With `std::this_thread::yield` called before `sleep_for` and `rcl_wait`, the tests took an average of 466 ms, and the test succeeded 100% of the time.
Without `std::this_thread::yield`, the tests took an average of 906 ms. I had to run the test 30 times to get 16 successes and 14 failures (I didn't count timing information for the failures).

16 samples each might not be statistically significant but to me this suggests that the Windows thread scheduler doesn't necessarily know to context switch without `yield`. It also isn't specified in the C++ standard that `sleep_for` has to force a context switch so I see no specification for why that would have to be (other than common sense).

There certainly are more precise ways to measure this, but the system calls for counting context switches in Windows look kind of awful and I didn't delve too deeply into them. I can make a standalone test with only the standard library calls if others think I need more evidence.

jacquelinekay:
I've decided on a fix that William and I can probably agree on:

I reverted my other changes and changed `wait_period` to a very large number (100ms, compared to the trigger_period of 2ms). I think this test was flaky before because after rcl_wait timed out the test_set threads would continuously try waiting again but the trigger thread would not get rescheduled, leading to the trigger condition never happening. By significantly increasing the wait time, the trigger thread will continuously trigger and the test_set threads will wait and wake up when they received a guard condition. I think this is also fairly consistent with a common use case for `rcl_wait`, which is to call it from `spin`, which will pass rcl_wait a "block indefinitely until woken up" argument (-1). Does that make sense?

jacquelinekay:
[Build Status](http://ci.ros2.org/job/ci_windows/1418/) (with arguments `--end-with rcl --ctest-args --repeat-until-fail 20 -R test_wait__rmw_opensplice_cpp --` to check if the test is still flaky)

rcl/test/rcl/test_wait.cpp
```diff
@@ -179,11 +181,11 @@ TEST(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), multi_wait_set_threaded)
             // printf("%s\n", ss.str().c_str());
           }
           ret = rcl_wait(&test_set.wait_set, wait_period);
+          ASSERT_NE(ret, RCL_RET_ERROR);
           if (ret != RCL_RET_TIMEOUT) {
             {
               std::stringstream ss;
-              ss << "[thread " << test_set.thread_id << "] Wakeup  (try #" << wake_try_count << ")";
-              // printf("%s\n", ss.str().c_str());
+              ss << "[thread " << test_set.thread_id << "] Wakeup (try #" << wake_try_count << ")";
```
wjwwood: I left the print statements commented out in case they're helpful for debugging later. Can you either put the printf back as a comment or remove all of them? Either is fine by me. (also the extra whitespace was so the output of `(try ..` lined up)

rcl/test/rcl/test_wait.cpp
```diff
@@ -179,11 +181,11 @@ TEST(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), multi_wait_set_threaded)
             // printf("%s\n", ss.str().c_str());
           }
           ret = rcl_wait(&test_set.wait_set, wait_period);
+          ASSERT_NE(ret, RCL_RET_ERROR);
```
wjwwood: My comment about this still stands. As-is it would miss a problem if, for example, `RCL_RET_INVALID_ARGUMENT` was returned, which is an error but not `RCL_RET_ERROR`. The safe thing to check is "if not RCL_RET_TIMEOUT or RCL_RET_OK then error occurred". Thanks for catching this gap in the logic though.

wjwwood:
+1 It makes sense to me! Thanks for iterating on this with me @jacquelinekay.

rcl/test/rcl/test_wait.cpp
jacquelinekay: I removed most of the printfs except for the one that occurs during timeout (since it's useful as a warning in case the test fails). I noticed that adding or removing printfs "artificially" improved the performance of this test when it was flaky due to the timing overhead of printf.

wjwwood: Well, what I meant is that if you're not going to leave the commented printf in, for example here, then the stringstream should probably also go.

jacquelinekay: I think if you take a look at the current state of the diff I've removed all commented printfs and unused stringstreams except for the one used to report a timeout.

rcl/test/rcl/test_wait.cpp
```diff
@@ -254,32 +246,15 @@ TEST(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), multi_wait_set_threaded)
     }
     return false;
   };
-  // *INDENT-ON*
```
wjwwood: This needs to stay I think.

PR #62, opened 2016-06-10
fix compiler warning

dirk-thomas:
Before: http://ci.ros2.org/job/ci_linux/1430/
After: http://ci.ros2.org/job/ci_linux/1431/


mikaelarguedas:
+1

PR #63, opened 2016-06-14
remove incorrect asserts

wjwwood:
Connects to ros2/ros2#215


jacquelinekay:
Is there a better assert we can make, like using the `current_index` of the rcl wait set and comparing that to the count in the rmw array?

wjwwood:
I think the assertion isn't necessary anymore. I did think about modifying the loop logic to stop checking the value of `rmw_guard_conditions.guard_conditions[i]` after `i` exceeded `rmw_guard_conditions.guard_condition_count`, but I think the resize functions guarantee that the rmw and rcl sets are identical in size anyways. So I decided to just remove the assertions.

jacquelinekay:
OK. I agree the asserts aren't necessary as long, as the user of the rcl_wait_set_* functions accesses the wait_set in a single-threaded fashion.

+1

PR #67, opened 2016-07-06
pass node to client and service destroy funcs and enable tests for connext

wjwwood:
Connects to ros2/rmw_connext#168.

### DIFF
--- a/rcl/test/CMakeLists.txt
+++ b/rcl/test/CMakeLists.txt
@@ -61,7 +61,10 @@ function(test_target_function)
   )
 
   # TODO(wjwwood): remove this when the graph API works properly for more than OpenSplice.
-  if(rmw_implementation STREQUAL "rmw_opensplice_cpp")
+  if(
+    rmw_implementation STREQUAL "rmw_opensplice_cpp"
+    OR rmw_implementation STREQUAL "rmw_connext_cpp"
+  )
     rcl_add_custom_gtest(test_graph${target_suffix}
       SRCS rcl/test_graph.cpp
       ENV ${extra_test_env}
--- a/rcl/test/CMakeLists.txt
+++ b/rcl/test/CMakeLists.txt
@@ -60,11 +60,13 @@ function(test_target_function)
     AMENT_DEPENDENCIES ${rmw_implementation}
   )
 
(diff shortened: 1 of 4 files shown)


rcl/test/CMakeLists.txt
```diff
@@ -61,7 +61,10 @@ function(test_target_function)
   )
 
   # TODO(wjwwood): remove this when the graph API works properly for more than OpenSplice.
-  if(rmw_implementation STREQUAL "rmw_opensplice_cpp")
+  if(
+    rmw_implementation STREQUAL "rmw_opensplice_cpp"
+    OR rmw_implementation STREQUAL "rmw_connext_cpp"
```
dirk-thomas: The `OR` should be trailing the previous line.

rcl/test/rcl/test_graph.cpp
```diff
@@ -44,6 +44,9 @@
 # define CLASSNAME(NAME, SUFFIX) NAME
 #endif
 
+bool is_connext =
+  std::string(rmw_get_implementation_identifier()).find("connext") != std::string::npos;
```
dirk-thomas: Is this `find` specific enough or should it better check for `find("rmw_connext") == 0`?

rcl/test/CMakeLists.txt
wjwwood: Hmm, maybe this whole check should be removed now. I started these changes before FastRTPS got some upgrades. I'm testing that out right now.

rcl/test/rcl/test_graph.cpp
wjwwood: I think it is good enough, but I'll change it anyways: 0d9ecc1

rcl/test/CMakeLists.txt
wjwwood: I changed the logic (596c0ec9faaa81cb4ba9dc064c27ba340412c023) and opened issues to track why:

- Fast RTPS: https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/issues/64
- Connext Dynamic: https://github.com/ros2/rmw_connext/issues/203

PR #75, opened 2016-08-24
Avoid overflow on 32-bit machines when working with large ns values

gerkey:
During the investigation of ros2/rclcpp#248, I found that the underlying clock functions weren't working properly. It turns out that we're experiencing overflow in the step where we combine the `timespec` fields into a single large ns value. The fields in `timespec` are signed longs and it seems that, at least on 32-bit ARM, that type allowed for integer overflow to occur, which results in crazy clock values.

Without this change, lots of things fail, at least intermittently, including comparisons between our time functions and `std::chrono`, like this:

https://github.com/ros2/rcl/blob/master/rcl/test/rcl/test_time.cpp#L76-L87

In other words, we already have the tests for this problem, but we haven't been running them on the architecture where it presents itself.

Life is hard when you can't figure out what time it is.

### DIFF
--- a/rcl/src/rcl/time_unix.c
+++ b/rcl/src/rcl/time_unix.c
@@ -66,7 +66,7 @@ rcl_system_time_now(rcl_time_point_value_t * now)
     RCL_SET_ERROR_MSG("unexpected negative time");
     return RCL_RET_ERROR;
   }
-  *now = RCL_S_TO_NS(timespec_now.tv_sec) + timespec_now.tv_nsec;
+  *now = RCL_S_TO_NS((uint64_t)timespec_now.tv_sec) + timespec_now.tv_nsec;
   return RCL_RET_OK;
 }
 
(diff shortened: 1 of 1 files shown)


wjwwood:
Seems reasonable to me. I think it's probably the implicit conversion of `signed long` to an `uint64_t`. On 64-bit machines `time_t` might be 64-bit, whereas it might be 32-bit on other machines. Either way, if this patch improves test stability I think it makes sense to merge it.

gerkey:
FYI, here's a simple program to demonstrate the problem in isolation:
~~~
#include <inttypes.h>
#include <stdio.h>
#include <time.h>

#define RCL_S_TO_NS(seconds) (seconds * (1000 * 1000 * 1000))

int
main(void)
{
  struct timespec timespec_now;
  uint64_t t0, t1;
  while(1) {
    clock_gettime(CLOCK_REALTIME, &timespec_now);
    t0 = RCL_S_TO_NS(timespec_now.tv_sec) + timespec_now.tv_nsec;
    t1 = RCL_S_TO_NS((uint64_t)timespec_now.tv_sec) + timespec_now.tv_nsec;
    if (t0 != t1) {
      printf("t0: %llu\tt1: %llu\n", t0, t1);
    }
  }
  return 0;
}
~~~
On my x86_64 box, that program outputs nothing. On the rpi3 (arm32), it gives two kinds of wrong answers in batches that alternate at about 1Hz, like this:
~~~
t0: 502410287	t1: 1472081164825144367
~~~
and like this:
~~~
t0: 18446744072213988357	t1: 1472081313151026181
~~~

PR #76, opened 2016-08-24
add timer overflow test

dirk-thomas:
I just created this based on the discussion in ros2/rclcpp#248. I am fine just closing this PR without merging.

This test requires access to the timer implementation in order to set the last call time. But adding `_rcl_timer_set_last_call_time` to the public interface seems pretty ugly.


gerkey:
Fair point about the interface change being ugly. And while it helped me in debugging, the new test didn't actually catch any problems. Thanks for doing it, though!

Issue #78, opened 2016-09-02
rcl_send_response  can return a RCL_RET_CLIENT_INVALID?


firesurfer:
According to the documentation the `rcl_send_response` function can return a RCL_RET_CLIENT_INVALID.

``` c++
 * \return RCL_RET_OK if the response was sent successfully, or
 *         RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
 *         RCL_RET_CLIENT_INVALID if the service is invalid, or
 *         RCL_RET_ERROR if an unspecified error occurs.
 */
RCL_PUBLIC
RCL_WARN_UNUSED
rcl_ret_t
rcl_send_response(
  const rcl_service_t * service,
  rmw_request_id_t * response_header,
  void * ros_response);
```

Shouldn't this be a RCL_RET_SERVICE_INVALID ?

dirk-thomas:
Yes, it looks like a copy-n-paste error. Can you please create a PR with the proposed change.

But the current implementations never return this code anyway. @wjwwood ?

wjwwood:
Definitely a copy-paste bug. That return code is reserved for checking if the service is valid or not. In the publisher code, an equivalent return code is returned if the argument is null, but perhaps that's not the best idea, and should instead return `RCL_RET_INVALID_ARGUMENT` as the implementation of this function does. When I think of an "invalid service" I mean:

- rcl_shutdown has been called
- the associated node is invalid

Currently the node can only be invalid if rcl_shutdown has been called, which is a bit redundant. However, I wanted to build in the notion that a node could have a valid/invalid state separate from the init/shutdown state.

A more general issue is that we need to check these states in all functions, and provide `rcl_X_is_valid()` functions for each type of thing.

firesurfer:
@dirk-thomas  Then I'm going to create a PR that corrects the documentation from `RCL_RET_CLIENT_INVALID`
to `RCL_RET_SERVICE_INVALID` ?

@wjwwood 
I would expect this error in case the passed service is invalid (is this even possible to determine?). In case the node is invalid or rcl_shutdown has been called I would return the corresponding value. E.g in case the node is invalid I would expect the function to return `RCL_RET_NODE_INVALID` 

By a `rcl_X_is_valid() ` function you mean for example a function that checks if a given node is valid: `bool rcl_node_is_valid(rcl_node_t * node)` ?

wjwwood:
[...]

PR #80, opened 2016-09-16
Removed RCL_RET_CLIENT_INVALID from documentation for rcl_send_response

firesurfer:
See: https://github.com/ros2/rcl/issues/78
I simply removed the RCL_RET_CLIENT_INVALID documentation entry from the rcl_send_response  function because the function doesn't return RCL_RET_CLIENT_INVALID.


wjwwood:
Please instead replace this with `RCL_RET_SERVICE_INVALID`, which was the original intention. Just because it doesn't return it now, doesn't mean we don't want to potentially send it in the future.

firesurfer:
I added the RCL_RET_SERVICE_INVALID return value to the documentation

wjwwood:
Awesome, thanks!

PR #84, opened 2016-10-19
Add rosnode and rostopic functionality (alpha8)

vmayoral:
Follows from https://github.com/ros2/rcl/pull/34

### DIFF
--- a/rcl/CMakeLists.txt
+++ b/rcl/CMakeLists.txt
@@ -31,6 +31,7 @@ set(${PROJECT_NAME}_sources
   src/rcl/time.c
   src/rcl/timer.c
   src/rcl/wait.c
+  src/rcl/topic.c
 )
 
 macro(target)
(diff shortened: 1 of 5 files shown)


firesurfer:
Perhaps it might be better to implement these functions in the graph.c because the graph.h already contains functions with the same name (and I guess the same function).

mikaelarguedas:
thanks @firesurfer for the comment. This will not be merged as is indeed, a discussion has been started on discourse to come up with a design for the implementation of these tools: http://discourse.ros.org/t/ros-graph-information-tools-implementation-discussion/674

mikaelarguedas:
We now have a first implementation of the ros commandline tools available on https://github.com/ros2/ros2cli. Feel free to comment on https://github.com/ros2/ros2/issues/197#issuecomment-308577572 for any feedback.

Thanks a lot @vmayoral for all this preliminary work !

PR #85, opened 2016-10-27
Zero init topic names and types

mikaelarguedas:
Adding a zero_initialized function for rcl_topic_names_and_types_t structure + sanity checks in rcl_get_topic_names_and_types

### DIFF
--- a/rcl/include/rcl/graph.h
+++ b/rcl/include/rcl/graph.h
@@ -44,8 +44,7 @@ rcl_get_zero_initialized_topic_names_and_types(void);
  *
  * The node parameter must not be NULL, and must point to a valid node.
  *
- * The topic_names_and_types parameter must not be NULL, and must point to an
- * already allocated rcl_topic_names_and_types_t struct.
+ * The topic_names_and_types parameter must be allocated and zero initalized.
  * The topic_names_and_types is the output for this function, and contains
  * allocated memory.
  * Therefore, it should be passed to rcl_destroy_topic_names_and_types() when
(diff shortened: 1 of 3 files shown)


mikaelarguedas:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/1830)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1365/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/1737/)

mikaelarguedas:
please also review https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/63 I cannot put label on this one

dirk-thomas:
While all our usage of the rmw interface goes through rcl anyone could use the rmw interface directly. In that case I think it would be helpful if each implementation would perform the sanity checking itself. Maybe the check can be implemented in a single generic function somewhere which can then be reused by each implementation?

wjwwood:
In my opinion, I don't think it is necessary for each rmw implementation to check it, so long as it is documented. I'd also be fine with a common check function that each calls, but I am not in favor of duplicating the check in each rmw implementation and rcl.

mikaelarguedas:
So I'll go for a common check function in the rmw package that every rmw_implementation can call, sounds good to everyone ?

wjwwood:
That's fine by me.

As I said before, I think it is unnecessary, but it also doesn't hurt to check it explicitly.

mikaelarguedas:
done https://github.com/ros2/rmw/issues/76

mikaelarguedas:
new set of CI with rmw changes:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/1837)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1372/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/1744/)

mikaelarguedas:
damn copy paste XD https://github.com/ros2/rcl/blob/master/rcl/include/rcl/node.h#L83

wjwwood:
Oops lol 😄

PR #86, opened 2016-11-04
cast timing at macro expansion

mikaelarguedas:
Without a cast at macro expansion we get the following warning as soon as we want to wait for more that 2 seconds
```c
warning: integer overflow in expression [-Woverflow]
 #define RCL_S_TO_NS(seconds) (seconds * (1000 * 1000 * 1000))
```

### DIFF
--- a/rcl/include/rcl/time.h
+++ b/rcl/include/rcl/time.h
@@ -24,9 +24,9 @@ extern "C"
 #include "rcl/types.h"
 #include "rcl/visibility_control.h"
 
-#define RCL_S_TO_NS(seconds) (seconds * (1000 * 1000 * 1000))
-#define RCL_MS_TO_NS(milliseconds) (milliseconds * (1000 * 1000))
-#define RCL_US_TO_NS(microseconds) (microseconds * 1000)
+#define RCL_S_TO_NS(seconds) ((int64_t)seconds * (1000 * 1000 * 1000))
+#define RCL_MS_TO_NS(milliseconds) ((int64_t)milliseconds * (1000 * 1000))
+#define RCL_US_TO_NS(microseconds) ((int64_t)microseconds * 1000)
 
 #define RCL_NS_TO_S(nanoseconds) (nanoseconds / (1000 * 1000 * 1000))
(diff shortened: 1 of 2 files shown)


rcl/include/rcl/time.h
```diff
@@ -24,9 +24,9 @@ extern "C"
 #include "rcl/types.h"
 #include "rcl/visibility_control.h"
 
-#define RCL_S_TO_NS(seconds) (seconds * (1000 * 1000 * 1000))
-#define RCL_MS_TO_NS(milliseconds) (milliseconds * (1000 * 1000))
-#define RCL_US_TO_NS(microseconds) (microseconds * 1000)
+#define RCL_S_TO_NS(seconds) ((int64_t)seconds * (1000 * 1000 * 1000))
+#define RCL_MS_TO_NS(milliseconds) ((int64_t)milliseconds * (1000 * 1000))
+#define RCL_US_TO_NS(microseconds) ((int64_t)microseconds * 1000)
```
mikaelarguedas: good question, I didn't see any part of the codebase using it for anything else than for `rcl_wait` that uses `int64_t` that's why I went for it.
I would naively assume that an implicit cast from `int64_t` to `uint64_t` would just work without problem  but maybe I'm wrong

dhood: the[ `other lines of this PR`](https://github.com/ros2/rcl/pull/86/files#diff-eebe14c889bc8ee26a92052b7ae5ae08L69) were using uint64_t and are now being cast to int64_t which could cause overflow problems

wjwwood: This cannot be cast to `uint64_t` though, because if the given value is negative it shouldn't be cast to an unsigned type.

I would revert this and ensure the temporary storage is large enough at the point of use, as in the lines referred to @dhood: https://github.com/ros2/rcl/pull/86/files#diff-eebe14c889bc8ee26a92052b7ae5ae08L69

mikaelarguedas: Sound good, thanks for the feedback @dhood @wjwwood.
Reverting

PR #89, opened 2016-11-16
triggering a guard condition is not const

wjwwood:
Also fixed a line of the documentation.


wjwwood:
I tested this, on a separate branch to disconnect it from https://github.com/ros2/rclcpp/pull/277, but it failed because I'll need changes from the other pr before merging this. I may break out the necessary commits to get this merged sooner.

PR #90, opened 2016-11-24
Mark blacklisted tests as skipped

mikaelarguedas:
connects to connects to https://github.com/ros2/ros2/issues/290

### DIFF
--- a/rcl/test/CMakeLists.txt
+++ b/rcl/test/CMakeLists.txt
@@ -60,29 +60,23 @@ function(test_target_function)
     AMENT_DEPENDENCIES ${rmw_implementation}
   )
 
+  set(_SKIP_TEST "")
   # TODO(wjwwood): remove this when the graph API works properly for FastRTPS
   if(
     rmw_implementation STREQUAL "rmw_connext_dynamic_cpp" OR
     rmw_implementation STREQUAL "rmw_fastrtps_cpp"
   )
     message(STATUS "Skipping test_graph${target_suffix} test.")
-    rcl_add_custom_gtest(test_graph${target_suffix}
-      SRCS rcl/test_graph.cpp
-      ENV ${extra_test_env}
-      APPEND_LIBRARY_DIRS ${extra_lib_dirs}
-      LIBRARIES ${PROJECT_NAME}${target_suffix} ${extra_test_libraries}
-      AMENT_DEPENDENCIES ${rmw_implementation} "example_interfaces" "std_msgs"
-      SKIP_TEST
-    )
-  else()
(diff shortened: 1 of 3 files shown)


rcl/test/CMakeLists.txt
```diff
@@ -66,6 +66,14 @@ function(test_target_function)
...
   )
     message(STATUS "Skipping test_graph${target_suffix} test.")
+    rcl_add_custom_gtest(test_graph${target_suffix}
+      SRCS rcl/test_graph.cpp
+      ENV ${extra_test_env}
+      APPEND_LIBRARY_DIRS ${extra_lib_dirs}
+      LIBRARIES ${PROJECT_NAME}${target_suffix} ${extra_test_libraries}
+      AMENT_DEPENDENCIES ${rmw_implementation} "example_interfaces" "std_msgs"
+      SKIP_TEST
+    )
```
dirk-thomas: Instead of duplicating the function call with all its arguments I would suggest to use a variable to either pass or not pass `SKIP_TEST`.

mikaelarguedas: fixed in https://github.com/ros2/rcl/pull/90/commits/593906752f461a1d299c1f0e2c9c193f47a2296e

rcl/test/rcl_add_custom_gtest.cmake
```diff
@@ -67,9 +67,14 @@ macro(rcl_add_custom_gtest target)
   if(_ARG_APPEND_LIBRARY_DIRS)
     set(_ARG_APPEND_LIBRARY_DIRS "APPEND_LIBRARY_DIRS" ${_ARG_APPEND_LIBRARY_DIRS})
   endif()
+  if(_ARG_SKIP_TEST)
+    set(_ARG_SKIP_TEST "SKIP_TEST")
+  else()
+    set(_ARG_SKIP_TEST "")
```
dirk-thomas: I don't think this is necessary?

mikaelarguedas: for some reason before adding this, `"FALSE"` was passed to the script instead of an empty string, resulting in the following error: `run_test.py: error: --env argument 'FALSE' contains no equal sign` 
haven't found why FALSE is passed yet

dirk-thomas: Hm, I see. Looks like `cmake_parse_arguments` is actually setting it to either `TRUE` or `FALSE`. I thought that was different some (probably long) time ago. Sounds good as it is then. Sorry for the noise.

PR #92, opened 2016-12-07
update to use refactored typesupport

dirk-thomas:
Connect to ros2/rosidl#182.


mikaelarguedas:
@dirk-thomas can this branch be deleted?

dirk-thomas:
Oh yeah, missed that. Thanks - gone.

PR #97, opened 2016-12-13
move "getenv" function to rmw for more reuse

wjwwood:
follow up of ros2/rmw_fastrtps#79

Connects to ros2/rcutils#30


mikaelarguedas:
@wjwwood

wjwwood:
I will refactor this pr to use `c_utilities` once we finish the rename of that package.

clalancette:
This one is blocked essentially on ros2/rcutils#30 , so we'll wait on that outcome of that before moving on this.  I'll add the "question" tag for now, though it might be more appropriate to create a new "blocked" tag instead.

mikaelarguedas:
I don't think that waffle handles multiple "Connects to" keywords properly because I dont think there is a good way to know to which one it should be connected. You can refer to the first issue but connect to the open one it depends on. If possible, update the label given that this issue is not a question

wjwwood:
Closing in favor of https://github.com/ros2/rcl/issues/119

PR #98, opened 2016-12-15
Doxygen setup

wjwwood:
This pull request sets up Doxygen and fixes the documentation to look correct when rendered and to reference external data, like cppreference.com (where appropriate).

I've pushed the current state up to http://docs.ros2.org/beta1/api/rcl/. It doesn't have cross referencing to rmw or rosidl since I haven't set them up yet, but it would once I have their tag files.

### DIFF
--- a/rcl/Doxyfile
+++ b/rcl/Doxyfile
@@ -1,7 +1,7 @@
 # All settings not listed here will use the Doxygen default values.
 
 PROJECT_NAME           = "rcl"
-PROJECT_NUMBER         = beta1
+PROJECT_NUMBER         = master
 PROJECT_BRIEF          = "C API providing common ROS client library functionality."
 
 INPUT                  = ./include
@@ -13,6 +13,9 @@ SORT_MEMBER_DOCS       = NO
 
 GENERATE_LATEX         = NO
 
+# Tag files that do not exist will produce a warning and cross-project linking will not work.
 TAGFILES += "../../../../doxygen_tag_files/cppreference-doxygen-web.tag.xml=http://en.cppreference.com/w/"
-TAGFILES += "../../../../doxygen_tag_files/rmw.tag=http://docs.ros2.org/beta1/api/rmw/"
-GENERATE_TAGFILE = "../../../../doxygen_tag_files/rcl.tag"
+# Consider changing "latest" to the version you want to reference (e.g. beta1 or 1.0.0)
+TAGFILES += "../../../../doxygen_tag_files/rmw.tag=http://docs.ros2.org/latest/api/rmw/"
+# Uncomment to generate tag files for cross-project linking.
+#GENERATE_TAGFILE = "../../../../doxygen_tag_files/rcl.tag"
--- a/rcl/Doxyfile
+++ b/rcl/Doxyfile
@@ -1,0 +1,16 @@
+# All settings not listed here will use the Doxygen default values.
+
+PROJECT_NAME           = "rcl"
+PROJECT_NUMBER         = beta1
+PROJECT_BRIEF          = "C API providing common ROS client library functionality."
+
+INPUT                  = ./include
+RECURSIVE              = YES
+OUTPUT_DIRECTORY       = doc_output
+
+EXTRACT_ALL            = YES
+SORT_MEMBER_DOCS       = NO
+
+GENERATE_LATEX         = NO
+
+TAGFILES += "../../../../doxygen_tag_files/cppreference-doxygen-web.tag.xml=http://en.cppreference.com/w/"
--- a/rcl/Doxyfile
+++ b/rcl/Doxyfile
@@ -13,6 +13,12 @@ SORT_MEMBER_DOCS       = NO
 
 GENERATE_LATEX         = NO
 
+ENABLE_PREPROCESSING   = YES
+MACRO_EXPANSION        = YES
+EXPAND_ONLY_PREDEF     = YES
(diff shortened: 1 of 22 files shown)


rcl/Doxyfile
```diff
@@ -0,0 +1,17 @@
+# All settings not listed here will use the Doxygen default values.
+
+PROJECT_NAME           = "rcl"
+PROJECT_NUMBER         = beta1
```
dirk-thomas: I am not sure if embedding "beta1" here is s good idea. I looks like that would be yet another location in many repos which needs to be updated.

rcl/Doxyfile
```diff
@@ -0,0 +1,17 @@
...
+INPUT                  = ./include
+RECURSIVE              = YES
+OUTPUT_DIRECTORY       = doc_output
+
+EXTRACT_ALL            = YES
+SORT_MEMBER_DOCS       = NO
+
+GENERATE_LATEX         = NO
+
+TAGFILES += "../../../../doxygen_tag_files/cppreference-doxygen-web.tag.xml=http://en.cppreference.com/w/"
```
dirk-thomas: Why the need for these relative paths? They will likely be a problem on other machines.

rcl/src/rcl/client.c
```diff
@@ -117,6 +117,7 @@ rcl_client_fini(rcl_client_t * client, rcl_node_t * node)
 rcl_client_options_t
 rcl_client_get_default_options()
 {
+  // !!! MAKE SURE THAT CHANGES TO THESE DEFAULTS ARE REFLECTED IN THE HEADER DOC STRING
```
dirk-thomas: I don't these why this comment is necessary. The same applies basically to every line of code. If the developer changes anything he needs to consider upgrading related documentation.

rcl/include/rcl/rcl.h
```diff
@@ -12,6 +12,52 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/** \mainpage rcl: Common functionality for other ROS Client Libraries
```
Karsten1987: Shouldn't that be a brief description?

rcl/include/rcl/subscription.h
```diff
@@ -69,25 +69,29 @@ rcl_get_zero_initialized_subscription(void);
  * When the user defines a ROS message, code is generated which provides the
  * required rosidl_message_type_support_t object.
  * This object can be obtained using a language appropriate mechanism.
- * \TODO(wjwwood) probably should talk about this once and link to it instead
- * For C this macro can be used (using std_msgs/String as an example):
+ * \todo TODO(wjwwood) write these instructions once and link to it instead
```
Karsten1987: double todo necessary here?

rcl/Doxyfile
wjwwood: It's the version number, it has to be updated every time we change the version. What should I put there instead?

rcl/src/rcl/client.c
wjwwood: Because in this case (as opposed to "every line of code") this is very easy to forget or not even know that the documentation exists elsewhere and the information about what the defaults are used is basically duplicated information, which we can avoid in most code only cases. I'm just trying to help future developers out, you may think it unnecessary, but I think it helps, so unless there is a good reason to exclude it, I will leave it.

rcl/include/rcl/rcl.h
wjwwood: I don't know what you mean? This is the title of the main page.

rcl/include/rcl/subscription.h
wjwwood: Yes, the `\todo` causes doxygen to render it, but the `TODO` is something that can be found by our linters and editors automatically. Doxygen will not render `\TODO` and my editor will not highlight `\todo` only.

rcl/Doxyfile
dirk-thomas: I would say "master". And when doing the documentation building run temporary replace the string locally before invoking Doxygen. But no need to check the specific build number in.

rcl/Doxyfile
wjwwood: Yeah, it's definitely fragile, but without templating the Doxygen file each time or doing some weird `cat` + `echo` stuff (recommendation of Doxygen http://stackoverflow.com/a/11038467/671658) I cannot create/use these files dynamically.

rcl/Doxyfile
wjwwood: Ok, I'll put in `master` in the pr. If I had more time I would make it into a CMake custom command and template the file with the commit hash or branch or tag automatically and handle the tag files through the build/install spaces.

wjwwood:
@Karsten1987 I looked into it a bit and even with different themes and stuff, the bottom issue is that the return type and the function name/parameters are in side by side table cells and so the longest return type drives the width. Unfortunately that is caused by the actual type on most pages, e.g. `const std::vector< rclcpp::callback_group::CallbackGroup::WeakPtr > &` on http://docs.ros2.org/beta1/api/rclcpp/classrclcpp_1_1node_1_1_node.html. The macros are distracting, but not the main issue in my opinion. The templates and parameters rendering are also problematic too.

I'd look into it more, but I don't think I have time to right now, so I'd like to leave it for now and address some of the look and feel stuff later.

wjwwood:
Now that the beta is out, I'd like to get these basic configurations and documentation updates merged. Looking for a review.

PR #99, opened 2016-12-15
add missing wait-for-server-being-available

dirk-thomas:
Fixes the recently failing test in the nightly builds across all platforms (e.g. http://ci.ros2.org/view/nightly/job/nightly_linux_debug/lastCompletedBuild/testReport/(root)/projectroot/test_services__rmw_fastrtps_cpp/).

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2077/)
* Mac OS [Build Status](http://ci.ros2.org/job/ci_osx/1606/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2097/)

### DIFF
--- a/rcl/test/rcl/client_fixture.cpp
+++ b/rcl/test/rcl/client_fixture.cpp
@@ -24,6 +24,33 @@
 #include "../memory_tools/memory_tools.hpp"
 #include "../scope_exit.hpp"
 #include "rcl/error_handling.h"
+#include "rcl/graph.h"
+
+bool
+wait_for_server_to_be_available(
+  rcl_node_t * node,
+  rcl_client_t * client,
+  size_t max_tries,
+  int64_t period_ms)
+{
+  size_t iteration = 0;
+  do {
+    ++iteration;
+    bool is_ready;
+    rcl_ret_t ret = rcl_service_server_is_available(node, client, &is_ready);
+    if (ret != RCL_RET_OK) {
+      fprintf(
+        stderr, "Error in rcl_service_server_is_available: %s\n",
+        rcl_get_error_string_safe());
+      return false;
+    }
+    if (is_ready) {
+      return true;
+    }
+    std::this_thread::sleep_for(std::chrono::milliseconds(period_ms));
+  } while (iteration < max_tries);
+  return false;
+}
 
 bool
 wait_for_client_to_be_ready(
@@ -111,6 +138,12 @@ int main(int argc, char ** argv)
         main_ret = -1;
       }
     });
+
+    // Wait until server is available
+    if (!wait_for_server_to_be_available(&node, &client, 1000, 100)) {
+      fprintf(stderr, "Server never became available\n");
+      return -1;
+    }
 
     // Initialize a request.
     example_interfaces__srv__AddTwoInts_Request client_request;


rcl/test/rcl/client_fixture.cpp
```diff
@@ -112,6 +139,12 @@ int main(int argc, char ** argv)
       }
     });
 
+    // Wait unti server is available
```
mikaelarguedas: Typo: until (on my phone so can't push directly to fix it, hence the comment)

rcl/test/rcl/client_fixture.cpp
```diff
@@ -24,6 +24,33 @@
...
+  rcl_client_t * client,
+  size_t max_tries,
+  int64_t period_ms)
+{
+  size_t iteration = 0;
+  do {
+    ++iteration;
+    bool is_ready;
+    rcl_ret_t ret = rcl_service_server_is_available(node, client, &is_ready);
+    if (ret != RCL_RET_OK) {
```
Karsten1987: why not check for RCL_RET_OK and then return?

rcl/test/rcl/client_fixture.cpp
```diff
@@ -24,6 +24,33 @@
 #include "../memory_tools/memory_tools.hpp"
 #include "../scope_exit.hpp"
 #include "rcl/error_handling.h"
+#include "rcl/graph.h"
+
+bool
+wait_for_server_to_be_available(
```
wjwwood: I'd recommend the name `wait_for_service_to_be_available`, as `wait_for_server_to_be_available` is a bit ambiguous and `wait_for_service_server_to_be_available`, while explicit, is quite tedious.

rcl/test/rcl/client_fixture.cpp
wjwwood: Isn't that what he's doing?

rcl/test/rcl/client_fixture.cpp
```diff
@@ -24,6 +24,33 @@
 #include "../memory_tools/memory_tools.hpp"
 #include "../scope_exit.hpp"
 #include "rcl/error_handling.h"
+#include "rcl/graph.h"
+
+bool
```
wjwwood: Personally, I think `bool` is ok here, since it is a utility function just for the tests.

rcl/test/rcl/client_fixture.cpp
Karsten1987: I thought more like:
```
if (ret == RCL_RET_OK) {
  return ret;
}

File: rcl/test/rcl/client_fixture.cpp (hunk 4 above)
Comment (Karsten1987) [inline rcl/test/rcl/client_fixture.cpp]: agreed.

File: rcl/test/rcl/client_fixture.cpp (hunk 1 above)
Comment (dirk-thomas) [inline rcl/test/rcl/client_fixture.cpp]: Fixed: 62c73de2bb880f59a7bcd97c953fb5ae31cc173c

File: rcl/test/rcl/client_fixture.cpp (hunk 4 above)
Comment (dirk-thomas) [inline rcl/test/rcl/client_fixture.cpp]: Since the function only return either "true" or "false" I don't think the type should be `rcl_ret_t` which implies multiple other values are possible.

File: rcl/test/rcl/client_fixture.cpp [hunk 5]
```diff
@@ -24,6 +24,33 @@
...
+wait_for_server_to_be_available(
+  rcl_node_t * node,
+  rcl_client_t * client,
+  size_t max_tries,
+  int64_t period_ms)
```
dirk-thomas: The argument uses the same type as the already existing function. Therefore I would prefer to keep it similar.

PR #101, opened 2016-12-23
Require C11

allenh1:
This is the first Gentoo patch for ROS2. This really just allows compilation with the C-89 standard (I believe the C++-style for-loop declarations was added to C in the C-99 standard, and the default for GCC-5 is C-11).

I didn't actually have to change all that much! There's another patch that will be coming for one of the demos (there's an include that needs GCC-5 or later -- `experimental/filesystem`). This just effects the console demo, though.


dirk-thomas:
ROS 2 already requires a C++14 compiler. While we haven't made explicit requirements for C I would expect that we will also a require a similarly new C compiler.

Therefore I am not sure if this patch should be expected. Maybe we need to specify the required C compiler in CMake explicitly as we do for C++.

allenh1:
That's a sensible solution. I removed the code changes and set up the CMake to force the C11 standard.

dirk-thomas:
Thanks for the update. I used the option to swap both lines.

dirk-thomas:
I bet the same will be necessary for other packages. Please feel free to open PRs for those as well since you seem to have an easy way of detecting those on your platform.

allenh1:
This is actually the only one that was problematic. The only fix left for Gentoo is the missing `experimental/filesystem` include.

Issue #81, opened 2016-09-23
Implement rcl_x_is_valid methods


firesurfer:
In order to track what was discussed in #78 I open this issue tracker. 

@wjwwood  suggested that in the future it might become necessary/useful to implement functions in the style `rcl_x_is_valid` .

An example would be:  `bool rcl_service_is_valid(rcl_service_t * service)` which would check if this service is valid. (Perhaps check if it's not null, check if it was initialized, and so on)

wjwwood:
Thanks!

allenh1:
@tfoote Is this the kind of thing I could work on?

tfoote:
Yeah, We've tagged several issues ["help wanted"](https://waffle.io/ros2/ros2?search=help%20wanted) or also "unclaimed"](https://waffle.io/ros2/ros2?search=unclaimed) that you could look into.

There are many more, but we tried to pick out some good places for people to get involved.

allenh1:
We'll I'll start implementing this, then!

allenh1:
For now, I'm putting these in the `rcl.h` header -- is this appropriate?

allenh1:
Actually, after further investigation, I'm now thinking they should be implemented in several different header files (`bool rcl_service_is_valid(rcl_service_t * service)` should be in `service.h`, and similar).

Alternatively, I suppose I could put them in a `check.h` header, but I'll implement them in the above fashion in the meantime.

wjwwood:
Putting them in the same header with the type you're checking seems appropriate to me.

Issue #100, opened 2016-12-19
calling shutdown fails in examples_rclpy


dirk-thomas:
When running `examples_rclpy_minimal_service` and `examples_rclpy_minimal_client_async` with Connext the second one fails with the following error message after handling the response properly:

```
[rmw|error_handling.c:93] error string being overwritten: rcl node is invalid, rcl instance id does not match, at .../src/ros2/rcl/rcl/src/rcl/node.c:191
```

In that case the code tries to destroy the rcl client but the rcl node handle already has a mismatch with the underlying rmw node id since the Python code has called `shutdown` (https://github.com/ros2/examples/blob/b57f9591d96432671978f0a4be8a1f928bde5016/rclpy/services/minimal_client/client_async.py#L48).

clalancette:
Looking at the recent nightly builds, I believe this particular error has been fixed.  There are some other instances of "error string is being overwritten", but I am going to close this out and open new issues about those.

dirk-thomas:
> I am going to close this out and open new issues about those.

Please reference these new tickets here otherwise it is unclear where they are. Also consider spending time on resolving them during this sprint.

clalancette:
I'm not sure I agree; the new issue doesn't really have anything to do with this one.  But it's #155 , for reference.

PR #103, opened 2016-12-29
Implement is valid methods

allenh1:
Working on #81 .

This PR adds the `rcl_x_is_valid` calls for:
 * publishers
 * subscriptions
 * services
 * clients

Connect to https://github.com/ros2/rcl/issues/81

### DIFF
--- a/rcl/include/rcl/client.h
+++ b/rcl/include/rcl/client.h
@@ -391,6 +391,7 @@ rcl_client_get_rmw_handle(const rcl_client_t * client);
  */
 bool
 rcl_client_is_valid(const rcl_client_t * client);
+
 #if __cplusplus
 }
 #endif
--- a/rcl/include/rcl/client.h
+++ b/rcl/include/rcl/client.h
@@ -374,6 +374,8 @@ rcl_client_get_rmw_handle(const rcl_client_t * client);
 /**
  * The bool returned is `false` if client is invalid
  * The bool returned is `true` otherwise.
+ * In the case where `false` is to be returned, an
+ * error message is set.
  * This function cannot fail.
  *
  * <hr>
--- a/rcl/include/rcl/client.h
+++ b/rcl/include/rcl/client.h
@@ -338,7 +338,8 @@ RCL_PUBLIC
 RCL_WARN_UNUSED
 const rcl_client_options_t *
 rcl_client_get_options(const rcl_client_t * client);
-
+#define _client_get_options(client) \
+  &client->impl->options;
 /// Return the rmw client handle.
 /**
  * The handle returned is a pointer to the internally held rmw handle.
@@ -370,6 +371,8 @@ RCL_WARN_UNUSED
 rmw_client_t *
 rcl_client_get_rmw_handle(const rcl_client_t * client);
 
+bool
+rcl_client_is_valid(const rcl_client_t * client);
 #if __cplusplus
 }
 #endif
--- a/rcl/include/rcl/client.h
+++ b/rcl/include/rcl/client.h
@@ -375,6 +375,27 @@ RCL_WARN_UNUSED
 rmw_client_t *
 rcl_client_get_rmw_handle(const rcl_client_t * client);
 
+/// Check that the client is valid.
+/**
+ * The bool returned is `false` if:
+ *   - the argument `client` is `NULL`
+ *   - the argument's option pointer is `NULL`
+ *   - the argument's mw implementation is `NULL`
+ *   - the argument's rmw handle is `NULL`
+ * The bool returned is `true` otherwise.
+ * This function cannot fail.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] client pointer to the rcl client
+ * \return `true` if `client` is valid, otherwise `false`
+ */
 bool
 rcl_client_is_valid(const rcl_client_t * client);
 #if __cplusplus
--- a/rcl/include/rcl/client.h
+++ b/rcl/include/rcl/client.h
@@ -339,11 +339,6 @@ RCL_WARN_UNUSED
 const rcl_client_options_t *
 rcl_client_get_options(const rcl_client_t * client);
 
-///  Non-safe get_options function.
-/* *INDENT-OFF* */
-#define _client_get_options(client) &client->impl->options;
-/* *INDENT-ON* */
-
 /// Return the rmw client handle.
 /**
  * The handle returned is a pointer to the internally held rmw handle.
@@ -377,11 +372,7 @@ rcl_client_get_rmw_handle(const rcl_client_t * client);
 
 /// Check that the client is valid.
 /**
- * The bool returned is `false` if:
- *   - the argument `client` is `NULL`
- *   - the argument's option pointer is `NULL`
- *   - the argument's mw implementation is `NULL`
- *   - the argument's rmw handle is `NULL`
+ * The bool returned is `false` if client is invalid
  * The bool returned is `true` otherwise.
  * This function cannot fail.
  *
--- a/rcl/include/rcl/client.h
+++ b/rcl/include/rcl/client.h
@@ -338,8 +338,12 @@ RCL_PUBLIC
 RCL_WARN_UNUSED
 const rcl_client_options_t *
 rcl_client_get_options(const rcl_client_t * client);
-#define _client_get_options(client) \
-  &client->impl->options;
+
+///  Non-safe get_options function.
+/* *INDENT-OFF* */
+#define _client_get_options(client) &client->impl->options;
+/* *INDENT-ON* */
+
 /// Return the rmw client handle.
 /**
  * The handle returned is a pointer to the internally held rmw handle.
--- a/rcl/include/rcl/publisher.h
+++ b/rcl/include/rcl/publisher.h
@@ -305,10 +305,6 @@ RCL_WARN_UNUSED
 const rcl_publisher_options_t *
 rcl_publisher_get_options(const rcl_publisher_t * publisher);
 
-/* *INDENT-OFF* */
-#define _publisher_get_options(pub) &pub->impl->options
-/* *INDENT-ON* */
-
 /// Return the rmw publisher handle.
 /**
  * The handle returned is a pointer to the internally held rmw handle.
@@ -342,9 +338,7 @@ rcl_publisher_get_rmw_handle(const rcl_publisher_t * publisher);
 
 /// Check that the publisher is valid
 /**
- * The bool returned is `false` if:
- *   - the argument `publisher` is `NULL`
- *   - the argument's `publisher->impl` is `NULL`
+ * The bool returned is `false` if `publisher` is invalid.
  * The bool returned is `true` otherwise.
  * This function cannot fail.
  *
--- a/rcl/include/rcl/publisher.h
+++ b/rcl/include/rcl/publisher.h
@@ -305,6 +305,8 @@ RCL_WARN_UNUSED
 const rcl_publisher_options_t *
 rcl_publisher_get_options(const rcl_publisher_t * publisher);
 
+#define _publisher_get_options(pub) &pub->impl->options
+
 /// Return the rmw publisher handle.
 /**
  * The handle returned is a pointer to the internally held rmw handle.
--- a/rcl/include/rcl/publisher.h
+++ b/rcl/include/rcl/publisher.h
@@ -339,6 +339,25 @@ rmw_publisher_t *
 rcl_publisher_get_rmw_handle(const rcl_publisher_t * publisher);
 
 /// Check that the publisher is valid
+/**
+ * The bool returned is `false` if:
+ *   - the argument `publisher` is `NULL`
+ *   - the argument's `publisher->impl` is `NULL`
+ * The bool returned is `true` otherwise.
+ * This function cannot fail.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] publisher pointer to the rcl publisher
+ * \return `true` if the publisher is valid, otherwise `false`
+ */
+
 bool
 rcl_publisher_is_valid(const rcl_publisher_t * publisher);
 
--- a/rcl/include/rcl/publisher.h
+++ b/rcl/include/rcl/publisher.h
@@ -336,6 +336,10 @@ RCL_WARN_UNUSED
 rmw_publisher_t *
 rcl_publisher_get_rmw_handle(const rcl_publisher_t * publisher);
 
+/// Check that the publisher is valid
+bool
+rcl_publisher_is_valid(const rcl_publisher_t * publisher);
+
 #if __cplusplus
 }
 #endif
--- a/rcl/include/rcl/publisher.h
+++ b/rcl/include/rcl/publisher.h
@@ -340,6 +340,8 @@ rcl_publisher_get_rmw_handle(const rcl_publisher_t * publisher);
 /**
  * The bool returned is `false` if `publisher` is invalid.
  * The bool returned is `true` otherwise.
+ * In the case where `false` is to be returned, an
+ * error message is set.
  * This function cannot fail.
  *
  * <hr>
--- a/rcl/include/rcl/publisher.h
+++ b/rcl/include/rcl/publisher.h
@@ -305,7 +305,9 @@ RCL_WARN_UNUSED
 const rcl_publisher_options_t *
 rcl_publisher_get_options(const rcl_publisher_t * publisher);
 
+/* *INDENT-OFF* */
 #define _publisher_get_options(pub) &pub->impl->options
+/* *INDENT-ON* */
 
 /// Return the rmw publisher handle.
 /**
--- a/rcl/include/rcl/subscription.h
+++ b/rcl/include/rcl/subscription.h
@@ -339,6 +339,10 @@ RCL_WARN_UNUSED
 rmw_subscription_t *
 rcl_subscription_get_rmw_handle(const rcl_subscription_t * subscription);
 
+/// Check if a subscription is valid
+bool
+rcl_subscription_is_valid(const rcl_subscription_t * subscription);
+
 #if __cplusplus
 }
 #endif
--- a/rcl/include/rcl/subscription.h
+++ b/rcl/include/rcl/subscription.h
@@ -339,7 +339,26 @@ RCL_WARN_UNUSED
 rmw_subscription_t *
 rcl_subscription_get_rmw_handle(const rcl_subscription_t * subscription);
 
-/// Check if a subscription is valid
+/// Check that the subscription is valid.
+/**
+ * The bool returned is `false` if:
+ *   - the argument `subscription` is `NULL`
+ *   - the argument's `subscription->impl` is `NULL`
+ *   - the argument's `subscription->impl->rmw_handle` is `NULL`
+ * The bool returned is `true` otherwise.
+ * This function cannot fail.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] subscription pointer to the rcl subscription
+ * \return `true` if `subscription` is valid, otherwise `false`
+ */
 bool
 rcl_subscription_is_valid(const rcl_subscription_t * subscription);
 
--- a/rcl/include/rcl/subscription.h
+++ b/rcl/include/rcl/subscription.h
@@ -343,6 +343,8 @@ rcl_subscription_get_rmw_handle(const rcl_subscription_t * subscription);
 /**
  * The bool returned is `false` if `subscription` is invalid.
  * The bool returned is `true` otherwise.
+ * In the case where `false` is to be returned, an
+ * error message is set.
  * This function cannot fail.
  *
  * <hr>
--- a/rcl/include/rcl/subscription.h
+++ b/rcl/include/rcl/subscription.h
@@ -308,8 +308,9 @@ RCL_WARN_UNUSED
 const rcl_subscription_options_t *
 rcl_subscription_get_options(const rcl_subscription_t * subscription);
 /// unsafe _subscription_get_options
-#define _subscription_get_options(subscription) \
-  &subscription->impl->options
+/* *INDENT-OFF* */
+#define _subscription_get_options(subscription) &subscription->impl->options
+/* *INDENT-ON* */
 
 /// Return the rmw subscription handle.
 /**
--- a/rcl/include/rcl/subscription.h
+++ b/rcl/include/rcl/subscription.h
@@ -307,6 +307,9 @@ RCL_PUBLIC
 RCL_WARN_UNUSED
 const rcl_subscription_options_t *
 rcl_subscription_get_options(const rcl_subscription_t * subscription);
+/// unsafe _subscription_get_options
+#define _subscription_get_options(subscription) \
+  &subscription->impl->options
 
 /// Return the rmw subscription handle.
 /**
@@ -343,6 +346,7 @@ rcl_subscription_get_rmw_handle(const rcl_subscription_t * subscription);
 /**
  * The bool returned is `false` if:
  *   - the argument `subscription` is `NULL`
+ *   - the argument's option pointer is `NULL`
  *   - the argument's `subscription->impl` is `NULL`
  *   - the argument's `subscription->impl->rmw_handle` is `NULL`
  * The bool returned is `true` otherwise.
--- a/rcl/include/rcl/subscription.h
+++ b/rcl/include/rcl/subscription.h
@@ -307,10 +307,6 @@ RCL_PUBLIC
 RCL_WARN_UNUSED
 const rcl_subscription_options_t *
 rcl_subscription_get_options(const rcl_subscription_t * subscription);
-/// unsafe _subscription_get_options
-/* *INDENT-OFF* */
-#define _subscription_get_options(subscription) &subscription->impl->options
-/* *INDENT-ON* */
 
 /// Return the rmw subscription handle.
 /**
@@ -345,11 +341,7 @@ rcl_subscription_get_rmw_handle(const rcl_subscription_t * subscription);
 
 /// Check that the subscription is valid.
 /**
- * The bool returned is `false` if:
- *   - the argument `subscription` is `NULL`
- *   - the argument's option pointer is `NULL`
- *   - the argument's `subscription->impl` is `NULL`
- *   - the argument's `subscription->impl->rmw_handle` is `NULL`
+ * The bool returned is `false` if `subscription` is invalid.
  * The bool returned is `true` otherwise.
  * This function cannot fail.
  *
--- a/rcl/src/rcl/client.c
+++ b/rcl/src/rcl/client.c
@@ -201,29 +201,25 @@ rcl_client_get_default_options()
 const char *
 rcl_client_get_service_name(const rcl_client_t * client)
 {
-  const rcl_client_options_t * options = rcl_client_get_options(client);
-  if (!options) {
-    return NULL;  // error already set
+  if (!rcl_client_is_valid(client)) {
+    return NULL; // error already set
   }
-  RCL_CHECK_FOR_NULL_WITH_MSG(
-    client->impl->rmw_handle, "client is invalid", return NULL, options->allocator);
   return client->impl->rmw_handle->service_name;
 }
 
 const rcl_client_options_t *
 rcl_client_get_options(const rcl_client_t * client)
 {
-  RCL_CHECK_ARGUMENT_FOR_NULL(client, NULL, rcl_get_default_allocator());
-  RCL_CHECK_FOR_NULL_WITH_MSG(
-    client->impl, "client is invalid", return NULL, rcl_get_default_allocator());
-  return &client->impl->options;
+  if (!rcl_client_is_valid(client)) {
+    return NULL; // error already set
+  }
+  return _client_get_options(client);
 }
 
 rmw_client_t *
 rcl_client_get_rmw_handle(const rcl_client_t * client)
 {
-  const rcl_client_options_t * options = rcl_client_get_options(client);
-  if (!options) {
+  if (!rcl_client_is_valid(client)) {
     return NULL;  // error already set
   }
   return client->impl->rmw_handle;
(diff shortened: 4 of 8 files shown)


wjwwood:
I think this might be related to this: https://github.com/ros2/rcl/issues/100

Basically one of the things we check is whether or not `rcl_shutdown()` has been called, and if so return an error code saying it is invalid. We may need to relax that check for "getters".

wjwwood:
`SIGINT` is already being caught, it's one of the responsibilities of the language specific client libraries, e.g. `rclcpp` or `rclpy`. Here is `rclcpp`'s version of this that is done in `rclcpp::init`:

https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/utilities.cpp#L111-L123

allenh1:
That looks really good to me.

But, other than the regression, do these changes make sense to you? Did I miss any checks?

I tried to conform with the macro calls used in each function's checking.

rcl/src/rcl/publisher.c
```diff
@@ -94,10 +94,10 @@ rcl_publisher_init(
 rcl_ret_t
 rcl_publisher_fini(rcl_publisher_t * publisher, rcl_node_t * node)
 {
+  if (!rcl_publisher_is_valid(publisher)) return RCL_RET_PUBLISHER_INVALID;
+  else if (!rcl_node_is_valid(node)) return RCL_RET_NODE_INVALID;
```
wjwwood: We always use `{}` to enclose if like statements. You'll want to run our linters over this code to catch any other issues like this. You can either run the tests or you can just run the linters directly:

```
$ ament_uncrustify /path/to/ws/src/ros2/rcl
$ ament_cpplint /path/to/ws/src/ros2/rcl
```

Uncrustify can also automatically fix the style with the `--reformat` option.

rcl/src/rcl/publisher.c
```diff
@@ -94,10 +94,10 @@ rcl_publisher_init(
...
   rcl_ret_t result = RCL_RET_OK;
-  RCL_CHECK_ARGUMENT_FOR_NULL(publisher, RCL_RET_INVALID_ARGUMENT);
-  RCL_CHECK_ARGUMENT_FOR_NULL(node, RCL_RET_INVALID_ARGUMENT);
-  if (publisher->impl) {
+   if (publisher->impl) {
```
wjwwood: Indentation is off here.

rcl/src/rcl/publisher.c
```diff
@@ -124,46 +124,47 @@ rcl_publisher_get_default_options()
...
-  RCL_CHECK_FOR_NULL_WITH_MSG(
-    publisher->impl, "publisher is invalid", return RCL_RET_PUBLISHER_INVALID);
-  if (rmw_publish(publisher->impl->rmw_handle, ros_message) != RMW_RET_OK) {
+  if (!rcl_publisher_is_valid(publisher)) return RCL_RET_PUBLISHER_INVALID;
+  else if (rmw_publish(publisher->impl->rmw_handle, ros_message) != RMW_RET_OK) {
     RCL_SET_ERROR_MSG(rmw_get_error_string_safe());
     return RCL_RET_ERROR;
-  }
-  return RCL_RET_OK;
+  } return RCL_RET_OK;
```
wjwwood: `return` should be on its own line.

rcl/src/rcl/publisher.c
```diff
@@ -124,46 +124,47 @@ rcl_publisher_get_default_options()
...
   return publisher->impl->rmw_handle;
 }
 
+bool
+rcl_publisher_is_valid(const rcl_publisher_t * publisher)
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(publisher, false);
+  RCL_CHECK_FOR_NULL_WITH_MSG(publisher->impl,
+                              "publisher implementation is invalid",
+                              return NULL);
```
wjwwood: Please wrap all or none of the arguments. I'm not sure if the linter will automatically correct this. This is what I would do:

```c++
RCL_CHECK_FOR_NULL_WITH_MSG(
  publisher->impl, "publisher implementation is invalid", return NULL);
```

Or this:

```c++
RCL_CHECK_FOR_NULL_WITH_MSG(
  publisher->impl,
  "publisher implementation is invalid",
  return NULL);
```

rcl/src/rcl/publisher.c
```diff
@@ -124,46 +124,47 @@ rcl_publisher_get_default_options()
...
 
 const char *
 rcl_publisher_get_topic_name(const rcl_publisher_t * publisher)
 {
-  RCL_CHECK_ARGUMENT_FOR_NULL(publisher, NULL);
-  RCL_CHECK_FOR_NULL_WITH_MSG(
-    publisher->impl, "publisher is invalid", return NULL);
-  RCL_CHECK_FOR_NULL_WITH_MSG(
-    publisher->impl->rmw_handle, "publisher is invalid", return NULL);
+  if (!rcl_publisher_is_valid(publisher)) return NULL;
```
wjwwood: Unfortunately, I don't know if we can use the rcl_publisher_is_valid function you created here. The check has become stricter, before this it checked publisher and publisher->impl for NULL, but with this change it also checks that publisher->impl->rmw_handle is not NULL. That might be ok, but we recently ran into an issue which might indicate that this is too strict to simply access the data in the structure. See: #100

The new function makes sense in `rcl_publish` and `rcl_publisher_fini`, but for the "getters" (like this function and others) it might be too strict (again see #100).

rcl/src/rcl/publisher.c
wjwwood: This applies here and everywhere else that you made a change.

rcl/src/rcl/service.c
```diff
@@ -203,6 +203,25 @@ rcl_send_response(
...
+                              return false);
+  RCL_CHECK_FOR_NULL_WITH_MSG(service->impl->rmw_handle,
+                              "rmw handle is invalid",
+                              return false);
+  RCL_CHECK_FOR_NULL_WITH_MSG(service->impl->rmw_handle->service_name,
+                              "service name is invalid",
+                              return false);
+  RCL_CHECK_FOR_NULL_WITH_MSG(*service->impl->rmw_handle->service_name,
+                              "service name is invalid",
+                              return false);
```
wjwwood: What is this one checking? That the string isn't empty, i.e. the first character is not `\0`?

wjwwood: Also, why check the service name here, but not check the topic name in publisher and subscription? I'm not sure you need to check it anywhere. I don't see any way that it could end up invalid.

rcl/src/rcl/publisher.c
allenh1: Got it -- I'll use the linter before I push in the future.

rcl/src/rcl/publisher.c
allenh1: Fixed

rcl/src/rcl/publisher.c
allenh1: Fixed

allenh1:
Sorry for disappearing for a little there (first week back at school). When I ran uncrustify, I got the following:

```
allenh1@host-147-10 ~/ros2_ws $ ament_uncrustify src/rcl/
Traceback (most recent call last):
  File "/home/allenh1/ros2_ws/install/bin/ament_uncrustify", line 6, in <module>
    from pkg_resources import load_entry_point
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 958, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 664, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 634, in _load_backward_compatible
  File "/usr/lib64/python3.5/site-packages/setuptools-32.2.0-py3.5.egg/pkg_resources/__init__.py", line 3019, in <module>
  File "/usr/lib64/python3.5/site-packages/setuptools-32.2.0-py3.5.egg/pkg_resources/__init__.py", line 3003, in _call_aside
  File "/usr/lib64/python3.5/site-packages/setuptools-32.2.0-py3.5.egg/pkg_resources/__init__.py", line 3032, in _initialize_master_working_set
  File "/usr/lib64/python3.5/site-packages/setuptools-32.2.0-py3.5.egg/pkg_resources/__init__.py", line 655, in _build_master
  File "/usr/lib64/python3.5/site-packages/setuptools-32.2.0-py3.5.egg/pkg_resources/__init__.py", line 963, in require
  File "/usr/lib64/python3.5/site-packages/setuptools-32.2.0-py3.5.egg/pkg_resources/__init__.py", line 849, in resolve
pkg_resources.DistributionNotFound: The 'ament-uncrustify' distribution was not found and is required by the application
```

I assume this is going to be Gentoo specific. If you'd like, I'll see if I can't find a patch for this.

rcl/src/rcl/publisher.c
allenh1: Ok -- I just removed the extra check. Now it just checks that `publisher->impl` is not `NULL`. I also changed it to return `false` instead of `NULL`, because I'm not sure why I had it as `NULL` instead of `false`.

rcl/src/rcl/service.c
allenh1: Ok - I'll remove those checks.

allenh1:
@wjwwood Resolved conflicts. Am I missing anything else with this?

wjwwood:
@allenh1 not sure, I'll catch up on it as soon as I can now that you've had a fresh look at it.

One thing I can see is that there is not documentation on some of the new functions and the ones that have them just have a brief. If you don't have time to improve the docs I'll do it, but if you're looking for something to do I recommend taking some inspiration from the docs of the other functions and consider cases like what if `NULL` passed and in what cases is each of those considered invalid. I believe it would be best if at least those two categories of things were covered if not more.

allenh1:
ok -- will do.

allenh1:
@wjwwood Just got around to doing the documentation here. I also moved a check I saw in ` rcl_subscription_get_topic_name` to the `rcl_subscription_is_valid` function (checks that the `option` pointer is not `NULL`).

rcl/src/rcl/subscription.c
```diff
@@ -186,13 +186,19 @@ rcl_subscription_get_rmw_handle(const rcl_subscription_t * subscription)
 bool
 rcl_subscription_is_valid(const rcl_subscription_t * subscription)
 {
+  const rcl_subscription_options_t * options = rcl_subscription_get_options(subscription);
```
wjwwood: Isn't this an infinite loop? `rcl_subscription_get_options` calls this function.

wjwwood:
Looks like client is still missing, from my review above:

> Also, missing from this list is client (the counterpart to service).

Thanks for working on it 😄!

rcl/src/rcl/subscription.c
allenh1: Uh -- wups. Let me just undo that then.

wjwwood: You could split `rcl_subscription_get_options` up like this:

```c
return_type
_subscription_get_options(params...) {
  // just return the options, no checks
  return options
}

return_type
rcl_subscription_get_options(params....) {
   // do checks, like call is_valid
   return _subscription_get_options(params...);
}
```

Then you could use the internal `_subscription_get_options` function in `is_valid`.

Or since it is so simple you could just duplicate the logic to extract the options in both places.

allenh1:
@wjwwood I pushed the fix for the `rcl_subscription_get_options` function if you wanna look at it. Tested it and it works (as far as I can tell).

Next on my list is the client checks. After that, we should be good right?

wjwwood:
> Tested it and it works (as far as I can tell).

Have you run all of the tests (even for downstream packages)? If not I can show you how or we can start a CI job to test it for you.

> Next on my list is the client checks. After that, we should be good right?

When you've got that, I'll do another code review, and we'll run CI. If both of those don't turn anything up, then yeah I think this should be done.

wjwwood:
I started some CI for you:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2708/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/249/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2165/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2832/)

mikaelarguedas:
@allenh1 this should be rebased on top of master (and the conflicts resolved) before running ci.

allenh1:
@mikaelarguedas I already did so, and it was definitely not a good time.

rcl/include/rcl/client.h
```diff
@@ -370,6 +375,8 @@ RCL_WARN_UNUSED
 rmw_client_t *
 rcl_client_get_rmw_handle(const rcl_client_t * client);
 
+bool
```
wjwwood: Needs docs still.

rcl/src/rcl/subscription.c
```diff
@@ -227,34 +227,50 @@ rcl_take(
...
+  if (!rcl_subscription_is_valid(subscription)) {
+    return NULL;
+  }
   return subscription->impl->rmw_handle;
 }
 
+bool
+rcl_subscription_is_valid(const rcl_subscription_t * subscription)
+{
+  const rcl_subscription_options_t * options = _subscription_get_options(subscription);
```
wjwwood: This will fail if subscription is `NULL`.

allenh1:
@wjwwood Got it -- will do.

rcl/src/rcl/subscription.c
allenh1: This is very true. Thank you.

rcl/include/rcl/client.h
```diff
@@ -339,6 +339,11 @@ RCL_WARN_UNUSED
 const rcl_client_options_t *
 rcl_client_get_options(const rcl_client_t * client);
 
+///  Non-safe get_options function.
+/* *INDENT-OFF* */
+#define _client_get_options(client) &client->impl->options;
```
wjwwood: Does this need to be in the header? It would be better to put it in the `.c` file if possible. It shouldn't be part of the public API if possible (even if it has the leading `_`).

wjwwood: Same applies to the similar functions in all the other headers of this pr.

wjwwood: Also, there is some inconsistency in that this kind of function is in client, pub, and sub, but not service. Which is strange.

rcl/include/rcl/client.h
```diff
@@ -370,6 +375,29 @@ RCL_WARN_UNUSED
 rmw_client_t *
 rcl_client_get_rmw_handle(const rcl_client_t * client);
 
+/// Check that the client is valid.
+/**
+ * The bool returned is `false` if:
+ *   - the argument `client` is `NULL`
+ *   - the argument's option pointer is `NULL`
+ *   - the argument's mw implementation is `NULL`
```
wjwwood: typo `mw` -> `rmw`

rcl/include/rcl/publisher.h
```diff
@@ -336,6 +340,29 @@ RCL_WARN_UNUSED
 rmw_publisher_t *
 rcl_publisher_get_rmw_handle(const rcl_publisher_t * publisher);
 
+/// Check that the publisher is valid
+/**
+ * The bool returned is `false` if:
+ *   - the argument `publisher` is `NULL`
+ *   - the argument's `publisher->impl` is `NULL`
```
wjwwood: The list of reasons for client is different (more exhaustive). I don't know if we need the list to be that exhaustive, but it ought to be consistent amongst similar elements.

rcl/include/rcl/subscription.h
```diff
@@ -307,6 +307,10 @@ RCL_PUBLIC
 RCL_WARN_UNUSED
 const rcl_subscription_options_t *
 rcl_subscription_get_options(const rcl_subscription_t * subscription);
+/// unsafe _subscription_get_options
```
wjwwood: Again, here but not every where. Should be removed here or added everywhere else.

rcl/src/rcl/client.c
```diff
@@ -280,7 +277,16 @@ rcl_take_response(
   return RCL_RET_OK;
 }
 
-
+bool rcl_client_is_valid(const rcl_client_t * client)
+{
+  const rcl_client_options_t * options;
+  RCL_CHECK_ARGUMENT_FOR_NULL(client, RCL_RET_INVALID_ARGUMENT, rcl_get_default_allocator());        
+  options = _client_get_options(client);
+  RCL_CHECK_FOR_NULL_WITH_MSG(options, "client is invalid", return false, options->allocator);
```
wjwwood: If `options` is null, then you shouldn't access it to get the allocator. You should use `rcl_get_default_allocator()` here and optionally you could use `options->allocator` in the next null check.

rcl/src/rcl/publisher.c
```diff
@@ -199,16 +204,10 @@ rcl_publisher_get_default_options()
...
-  if (!options) {
+  if (!rcl_publisher_is_valid(publisher)) {
     return RCL_RET_PUBLISHER_INVALID;
-  }
-  RCL_CHECK_ARGUMENT_FOR_NULL(ros_message, RCL_RET_INVALID_ARGUMENT, options->allocator);
-  RCL_CHECK_FOR_NULL_WITH_MSG(
-    publisher->impl, "publisher is invalid", return RCL_RET_PUBLISHER_INVALID, options->allocator);
-  if (rmw_publish(publisher->impl->rmw_handle, ros_message) != RMW_RET_OK) {
-    RCL_SET_ERROR_MSG(rmw_get_error_string_safe(), options->allocator);
+  } else if (rmw_publish(publisher->impl->rmw_handle, ros_message) != RMW_RET_OK) {
```
wjwwood: This doesn't need to be an `else if` since the previous `if` returns.

rcl/src/rcl/service.c
```diff
@@ -280,6 +280,21 @@ rcl_send_response(
...
+  RCL_CHECK_FOR_NULL_WITH_MSG(service->impl,
+    "rcl service implementation is invalid",
+    return false,
+    rcl_get_default_allocator());
+  RCL_CHECK_FOR_NULL_WITH_MSG(service->impl->rmw_handle,
+    "rmw handle is invalid",
+    return false,
+    rcl_get_default_allocator());
+  return true;
+}
```
wjwwood: Why does this one not check the options of the service like it does for client and pub/sub?

rcl/src/rcl/subscription.c
```diff
@@ -231,34 +231,51 @@ rcl_take(
...
+bool
+rcl_subscription_is_valid(const rcl_subscription_t * subscription)
+{
+  const rcl_subscription_options_t * options;
+  RCL_CHECK_ARGUMENT_FOR_NULL(subscription, false, rcl_get_default_allocator());
+  options = _subscription_get_options(subscription);
+  RCL_CHECK_FOR_NULL_WITH_MSG(options,
+    "subscription implementation is invalid",
+    return false,
+    options->allocator);
```
wjwwood: Again, you shouldn't use `options->allocator` if `options` is null...

allenh1:
@wjwwood Thanks! I'll get to these shortly.

rcl/include/rcl/publisher.h
allenh1: Is this sufficient?

```c
/**
 * The bool returned is `false` if the argument `publisher` is invalid.
 * The bool returned is `true` otherwise.
 */
```

(not leaving out the details beneath, but I didn't retype them here for clarity)

rcl/src/rcl/client.c
allenh1: That's a very good point. Not sure why I thought that would be a good idea.

wjwwood: No worries, probably a copy-paste error or something.

rcl/include/rcl/publisher.h
wjwwood: That's fine. As long as it is consistently done on each of the entities.

allenh1:
After linting and whatnot, I ran the tests and hit the following:

```
2: Test command: /usr/bin/python3 "-u" "/home/allenh1/ros2_ws/install/share/ament_cmake_test/cmake/run_test.py" "/home/allenh1/ros2_ws/build/rcl_lifecycle/test_results/rcl_lifecycle/test_multiple_instances.gtest.xml" "--output-file" "/home/allenh1/ros2_ws/build/rcl_lifecycle/ament_cmake_gtest/test_multiple_instances.txt" "--command" "/home/allenh1/ros2_ws/build/rcl_lifecycle/test_multiple_instances" "--gtest_output=xml:/home/allenh1/ros2_ws/build/rcl_lifecycle/test_results/rcl_lifecycle/test_multiple_instances.gtest.xml"
2: Test timeout computed to be: 60
2: -- run_test.py: invoking following command in '/home/allenh1/ros2_ws/src/ros2/rcl/rcl_lifecycle':
2:  - /home/allenh1/ros2_ws/build/rcl_lifecycle/test_multiple_instances --gtest_output=xml:/home/allenh1/ros2_ws/build/rcl_lifecycle/test_results/rcl_lifecycle/test_multiple_instances.gtest.xml
2: Running main() from gtest_main.cc
2: [==========] Running 1 test from 1 test case.
2: [----------] Global test environment set-up.
2: [----------] 1 test from TestMultipleInstances
2: [ RUN      ] TestMultipleInstances.default_sequence_error_unresolved
2: 
2: >>> [rcutils|error_handling.c:148] rcutils_set_error_state()
2: This error state is being overwritten:
2: 
2:   'options argument is null, at /home/allenh1/ros2_ws/src/ros2/rcl/rcl/src/rcl/publisher.c:255'
2: 
2: with this new error message:
2: 
2:   'could not free lifecycle com interface. Leaking memory!
2: , at /home/allenh1/ros2_ws/src/ros2/rcl/rcl_lifecycle/src/rcl_lifecycle.c:104'
2: 
2: rcutils_reset_error() should be called after error handling to avoid this.
2: <<<
2: /home/allenh1/ros2_ws/src/ros2/rcl/rcl_lifecycle/test/test_multiple_instances.cpp:111: Failure
2: Value of: ret
2:   Actual: 1
2: Expected: 0
2: could not free lifecycle com interface. Leaking memory!
2: , at /home/allenh1/ros2_ws/src/ros2/rcl/rcl_lifecycle/src/rcl_lifecycle.c:104
2: 
2: >>> [rcutils|error_handling.c:148] rcutils_set_error_state()
2: This error state is being overwritten:
2: 
2:   'could not free lifecycle com interface. Leaking memory!
2: , at /home/allenh1/ros2_ws/src/ros2/rcl/rcl_lifecycle/src/rcl_lifecycle.c:104'
2: 
2: with this new error message:
2: 
2:   'options argument is null, at /home/allenh1/ros2_ws/src/ros2/rcl/rcl/src/rcl/publisher.c:255'
2: 
2: rcutils_reset_error() should be called after error handling to avoid this.
2: <<<
2: 
2: >>> [rcutils|error_handling.c:148] rcutils_set_error_state()
2: This error state is being overwritten:
2: 
2:   'options argument is null, at /home/allenh1/ros2_ws/src/ros2/rcl/rcl/src/rcl/publisher.c:255'
2: 
2: with this new error message:
2: 
2:   'could not free lifecycle com interface. Leaking memory!
2: , at /home/allenh1/ros2_ws/src/ros2/rcl/rcl_lifecycle/src/rcl_lifecycle.c:104'
2: 
2: rcutils_reset_error() should be called after error handling to avoid this.
2: <<<
2: /home/allenh1/ros2_ws/src/ros2/rcl/rcl_lifecycle/test/test_multiple_instances.cpp:113: Failure
2: Value of: ret
2:   Actual: 1
2: Expected: 0
2: could not free lifecycle com interface. Leaking memory!
2: , at /home/allenh1/ros2_ws/src/ros2/rcl/rcl_lifecycle/src/rcl_lifecycle.c:104
2: 
2: >>> [rcutils|error_handling.c:148] rcutils_set_error_state()
2: This error state is being overwritten:
2: 
2:   'could not free lifecycle com interface. Leaking memory!
2: , at /home/allenh1/ros2_ws/src/ros2/rcl/rcl_lifecycle/src/rcl_lifecycle.c:104'
2: 
2: with this new error message:
2: 
2:   'options argument is null, at /home/allenh1/ros2_ws/src/ros2/rcl/rcl/src/rcl/publisher.c:255'
2: 
2: rcutils_reset_error() should be called after error handling to avoid this.
2: <<<
2: 
2: >>> [rcutils|error_handling.c:148] rcutils_set_error_state()
2: This error state is being overwritten:
2: 
2:   'options argument is null, at /home/allenh1/ros2_ws/src/ros2/rcl/rcl/src/rcl/publisher.c:255'
2: 
2: with this new error message:
2: 
2:   'could not free lifecycle com interface. Leaking memory!
2: , at /home/allenh1/ros2_ws/src/ros2/rcl/rcl_lifecycle/src/rcl_lifecycle.c:104'
2: 
2: rcutils_reset_error() should be called after error handling to avoid this.
2: <<<
2: /home/allenh1/ros2_ws/src/ros2/rcl/rcl_lifecycle/test/test_multiple_instances.cpp:115: Failure
2: Value of: ret
2:   Actual: 1
2: Expected: 0
2: could not free lifecycle com interface. Leaking memory!
2: , at /home/allenh1/ros2_ws/src/ros2/rcl/rcl_lifecycle/src/rcl_lifecycle.c:104
2: [  FAILED  ] TestMultipleInstances.default_sequence_error_unresolved (3 ms)
2: [----------] 1 test from TestMultipleInstances (3 ms total)
2: 
2: [----------] Global test environment tear-down
2: [==========] 1 test from 1 test case ran. (3 ms total)
2: [  PASSED  ] 0 tests.
2: [  FAILED  ] 1 test, listed below:
2: [  FAILED  ] TestMultipleInstances.default_sequence_error_unresolved
2: 
2:  1 FAILED TEST
2: -- run_test.py: return code 1
2: -- run_test.py: verify result file '/home/allenh1/ros2_ws/build/rcl_lifecycle/test_results/rcl_lifecycle/test_multiple_instances.gtest.xml'
2/8 Test #2: test_multiple_instances ..........***Failed    0.08 sec
```

I take this to mean that the checks I implemented are too aggressive. Should I be checking the options pointer?

wjwwood:
> I take this to mean that the checks I implemented are too aggressive.

Maybe, I'm not sure. It could also be that these tests or the implementation of the lifecycle stuff is making an incorrect assumption. You'll just have to dig into the test failures and see what needs to change.

> Should I be checking the options pointer?

I think so, but that's just my impression without looking at it in detail.

allenh1:
Found it. It was an unnecessary check in the publisher's fini function.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2897/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/382/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2334/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3019/)

allenh1:
(I just changed one of the strings that said "subscription" in the service.c file to say "service" -- don't think this is worth re-triggering CI over)

---

@wjwwood can I merge this after the CI is green?

rcl/src/rcl/client.c
```diff
@@ -280,7 +281,18 @@ rcl_take_response(
   return RCL_RET_OK;
 }
 
-
+bool rcl_client_is_valid(const rcl_client_t * client)
+{
+  const rcl_client_options_t * options;
+  RCL_CHECK_ARGUMENT_FOR_NULL(
+    client, RCL_RET_INVALID_ARGUMENT, rcl_get_default_allocator());
```
dirk-thomas: `RCL_RET_INVALID_ARGUMENT`? This function can only return `bool`.

rcl/src/rcl/publisher.c
```diff
@@ -217,35 +212,49 @@ rcl_publish(const rcl_publisher_t * publisher, const void * ros_message)
...
+{
+  const rcl_publisher_options_t * options;
+  RCL_CHECK_ARGUMENT_FOR_NULL(publisher, false, rcl_get_default_allocator());
+  options = _publisher_get_options(publisher);
+  RCL_CHECK_FOR_NULL_WITH_MSG(
+    options, "publisher is invalid", return false, rcl_get_default_allocator());
+  RCL_CHECK_FOR_NULL_WITH_MSG(
+    publisher->impl, "publisher is invalid", return false, options->allocator);
+  RCL_CHECK_FOR_NULL_WITH_MSG(
+    publisher->impl->rmw_handle, "publisher is invalid", return false, options->allocator);
```
dirk-thomas: It would be better if the different checks use a different error message. Otherwise the user won't be able to distinguish the error case without additional debugging.

rcl/include/rcl/client.h
```diff
@@ -370,6 +370,25 @@ RCL_WARN_UNUSED
...
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] client pointer to the rcl client
+ * \return `true` if `client` is valid, otherwise `false`
+ */
```
dirk-thomas: Should the docblocks mention that in case these functions return `false` an error message is being set?

rcl/src/rcl/client.c
allenh1: Good catch! I believe that would have been the same as `return 1`, which is very not correct.

rcl/src/rcl/publisher.c
allenh1: That's a good point.  I'll update other files as well.

rcl/include/rcl/client.h
allenh1: I have no idea -- @wjwwood what do you think?

wjwwood: We normally don't mention that an error message is set, but in this case I think it makes sense because we're not returning an error code. I think we mention it explicitly in a few other functions where a pointer is returned (an error is set if NULL is returned instead). Can't hurt.

dirk-thomas: Will the user need to reset the error message before being able to call the next function "safely"?

wjwwood: Well it's always safe to leave the error message, but you do need to do it before calling something else if you don't want the overwriting an error message print. Elsewhere it is implied that a non `RCL_RET_OK` return code means an error message is set. Here it's not so clear that one would be set by a `false` return value. So I think it's appropriate.

allenh1: Ok. I just added a simple

```c
/**
 * In the case where `false` is to be returned, an
 * error message is set.
 */
```

Should I re-run the CI?

rcl/include/rcl/client.h
```diff
@@ -370,6 +370,27 @@ RCL_WARN_UNUSED
...
+bool
+rcl_client_is_valid(const rcl_client_t * client);
```
mikaelarguedas: Nit: missing empty line here

allenh1: fixed

rcl/include/rcl/client.h
dirk-thomas: Are we expecting users to call these `rcl_x_is_valid` functions? If no, I think the current state documenting that an error message is being set in the case the functions return `false` is good.

If yes, I am not sure if that is a convenient API. I can imagine many cases where the caller doesn't care about the specific error message and will either have to write boiler plate code around every call to reset the error message or even forget about it (which will result in the error message being printed when attempted to be overwritten). Maybe in that case the functions should have a flag to optionally not set (or internall reset) the error message.

allenh1: My understanding is that these functions are just to clean up the code so that, if a check is added in the future, it only needs to go in one place.

But that may be completely off-base, so I'll defer to @wjwwood.

wjwwood: I think the interface is fine as-is. I don't expect people to use these functions often (only client library implementers) and I expect even fewer cases where a return of false doesn't trigger an error case or at least a logging event. Also resetting the error is a single line, so it's not much boiler plate.

Having another option to not set the error seems unnecessarily complicated for the number of times it would be used.

dirk-thomas: Sounds good to me than. :+1:

allenh1:
Ok, here's the new CI.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2898/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/383/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2335/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3020/)

wjwwood:
I added an issue to add tests for these functions in the future: https://github.com/ros2/rcl/issues/150

PR #106, opened 2017-01-05
use rmw implementation

dirk-thomas:
See ros2/ros2#302


dirk-thomas:
The CI builds for this as well as all referenced PRs:

Before:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2204/)
* Mac OS [Build Status](http://ci.ros2.org/job/ci_osx/1670/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2197/)

After:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2206/)
* Mac OS [Build Status](http://ci.ros2.org/job/ci_osx/1672/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2199/)

Both sets of builds without `--retest-until-pass` to not alter the test count. So all builds can contain flaky test failures.

This PR as well as the referenced ones are ready for review.

dirk-thomas:
The latest builds with the rmw selection logic removed from `rclpy`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2208/)
* Mac OS [Build Status](http://ci.ros2.org/job/ci_osx/1674/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2201/)

Waiting for review / comment on ros2/rclpy#66.

wjwwood:
Shouldn't you also now remove this file:

https://github.com/ros2/rcl/blob/72e7d0319841a1db91b2f96f9fe4854bf917ba5b/rcl/cmake/get_rcl_information.cmake

Now that `rcl` only ever passes on exactly one library? This also means that you can get rid of the extras file:

https://github.com/ros2/rcl/blob/72e7d0319841a1db91b2f96f9fe4854bf917ba5b/rcl/rcl-extras.cmake

And remove that from the `CMakeLists.txt`:

https://github.com/ros2/rcl/blob/72e7d0319841a1db91b2f96f9fe4854bf917ba5b/rcl/CMakeLists.txt#L77

I think this would also apply to `rcl_lifecycle`:

https://github.com/ros2/rcl/blob/72e7d0319841a1db91b2f96f9fe4854bf917ba5b/rcl_lifecycle/cmake/get_rcl_lifecycle_information.cmake
https://github.com/ros2/rcl/blob/72e7d0319841a1db91b2f96f9fe4854bf917ba5b/rcl_lifecycle/rcl_lifecycle-extras.cmake
https://github.com/ros2/rcl/blob/72e7d0319841a1db91b2f96f9fe4854bf917ba5b/rcl_lifecycle/CMakeLists.txt#L80

As well as `rclcpp` (though you cannot get rid of the extras file):

https://github.com/ros2/rclcpp/blob/fee3118bdd79467b6d9306e0b9d3d520549a32d0/rclcpp/cmake/get_rclcpp_information.cmake
https://github.com/ros2/rclcpp/blob/fee3118bdd79467b6d9306e0b9d3d520549a32d0/rclcpp/rclcpp-extras.cmake#L28

And would apply to `rclcpp_lifecycle` but it never seemed to have one in the first place.

dirk-thomas:
Please see ros2/ros2#302 where those are already being tracked.

dirk-thomas:
And yet another round of builds (done in approximately one hour):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2210/)
* Mac OS [Build Status](http://ci.ros2.org/job/ci_osx/1676/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2203/)

esteve:
@dirk-thomas couldn't the `get_rcl_*` CMake functions be preserved or at least moved elsewhere? `rcljava` relies on them to build the `rmw` glue for each `rmw` implementation. Android can't use environment variables, so the `rmw` implementation is loaded explicitly via `RCLJava.setRMWImplementation`. The same would happen if a Java application is deployed as a `WAR` into a Servlet container, as it would require changing the environment variables of the container (one normally just passes a properties file to set that sort of configuration).

dirk-thomas:
The `get_rcl*` functions are generating custom variable to reference the rmw specific libraries of `rcl` / `rclcpp`. But since we don't build multiple variations of these libraries anymore (only one `librcl.so` and `librclcpp.so`) those functions can't provide that functionality anymore. So I don't see how they could be preserved (here or anywhere else) if we don't want to keep N variations of these libraries.

I don't know how the Java implementation works but it can just use `rmw_implementation` / `rcl` / `rclcpp` directly. Currently `rmw_implemenation` chooses only based on the environment variable (https://github.com/ros2/rmw_implementation/blob/fb8072150b555bcc0a716cf9a73b5a013c886ee1/rmw_implementation/src/functions.cpp#L114). We could add a select function to choose a specific implementation to override the environment variable?

dirk-thomas:
I don't know enough about Java / Android development but didn't you mention in a different thread that you can't do dynamic loading of shared libraries? In that case you can only statically link against one rmw implementation anyway or am I missing something?

dirk-thomas:
And another round of builds after addressing the comment on https://github.com/ros2/system_tests/pull/187#pullrequestreview-15593247:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2212/)
* Mac OS [Build Status](http://ci.ros2.org/job/ci_osx/1677/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2205/)

esteve:
> I don't know enough about Java / Android development but didn't you mention in a different thread that you can't do dynamic loading of shared libraries?

Both Java and Android allow loading libraries dynamically from within the Java code, but libraries that call `dlopen` to load additional libraries might be blocked by the JVM security manager, I have to check that.

> In that case you can only statically link against one rmw implementation anyway or am I missing something?

That's applicable to iOS, not Java/Android.

> We could add a select function to choose a specific implementation to override the environment variable?

That would be fantastic. Would that function go into `rcl` or into `rmw`? Just asking, I may have time to implement such function and submit it for review.

dirk-thomas:
> > We could add a select function to choose a specific implementation to override the environment variable?

> That would be fantastic. Would that function go into `rcl` or into `rmw`? Just asking, I may have time to implement such function and submit it for review.

It would need to go into `rmw_implementation`.

PR #108, opened 2017-01-26
Rosnode list

Karsten1987:
Connect to ros2/cli_tools#1

### DIFF
--- a/rcl/test/CMakeLists.txt
+++ b/rcl/test/CMakeLists.txt
@@ -63,6 +63,17 @@ function(test_target_function)
     AMENT_DEPENDENCIES ${rmw_implementation}
   )
 
+  rcl_add_custom_gtest(test_get_node_names${target_suffix}
+    SRCS rcl/test_get_node_names.cpp
+    ENV
+      ${extra_test_env}
+      EMPTY_TEST=
+      NORMAL_TEST=foo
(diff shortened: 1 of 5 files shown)


rcl/test/CMakeLists.txt
```diff
@@ -63,6 +63,17 @@ function(test_target_function)
     AMENT_DEPENDENCIES ${rmw_implementation}
   )
 
+  rcl_add_custom_gtest(test_get_node_names${target_suffix}
+    SRCS rcl/test_get_node_names.cpp
+    ENV
+      ${extra_test_env}
+      EMPTY_TEST=
+      NORMAL_TEST=foo
```
dirk-thomas: What are these two environment variables for? They don't seem to be used in this patch.

mikaelarguedas: looks like a copy-paste artifact and should be removed. these are used to [test getenv in common .c](https://github.com/ros2/rcl/blob/d3a27df6e2dda9638acf2d0e57eefeb22197742f/rcl/test/rcl/test_common.cpp#L29-L37) and are not used for the test_get_node_names test

dirk-thomas: Another unrelated question: why are these tests run for each rmw implementation?

Karsten1987: clearly copy&paste error. I'll open another PR to correct this.

PR #112, opened 2017-03-29
add concept of node namespace to rcl API

wjwwood:
This not only exposes the node namespace in the API (via the `rcl_node_init()` and `rcl_node_get_namespace()` functions), but it also validates the node name and the new namespace arguments.

This also relies on https://github.com/ros2/rmw/pull/93 to help validate the node namespace.

Connects to ros2/rmw#95

### DIFF
--- a/rcl/include/rcl/node.h
+++ b/rcl/include/rcl/node.h
@@ -89,19 +89,18 @@ rcl_get_zero_initialized_node(void);
  * The name of the node cannot coincide with another node of the same name.
  * If a node of the same name is already in the domain, it will be shutdown.
  *
- * The namespace of the node should not be NULL and should also follow the
- * rules for topics, see:
+ * The namespace of the node should not be NULL and should also pass the
+ * rmw_validate_namespace() function's rules.
  *
- *   http://design.ros2.org/articles/topic_and_service_names.html
- *
- * And therefore the namespace should pass validation of the
- * rmw_validate_topic_name() function.
- * The namespace should be absolute, i.e. it should begin with a forward slash.
- * Even though this is a namespace, with additional relative namespace and/or
- * a basename to follow, the namespace string passed here should not include
- * a trailing forward slash.
- * For an empty namespace, an empty string, i.e. "", should be used.
- * The namespace "/" is considered invalid.
+ * Additionally this function allows namespaces which lack a leading forward
+ * slash.
+ * Because there is no notion of a relative namespace, there is no difference
+ * between a namespace which lacks a forward and the same namespace with a
+ * leasing forward slash.
+ * Therefore, a namespace like ``"foo/bar"`` is automatically changed to
+ * ``"/foo/bar"`` by this function.
+ * Similarly, the namespace ``""`` will implicitly become ``"/"`` which is a
+ * valid namespace.
  *
  * \todo TODO(wjwwood):
  *   Parameter infrastructure is currently initialized in the language specific
--- a/rcl/include/rcl/node.h
+++ b/rcl/include/rcl/node.h
@@ -81,15 +81,16 @@ rcl_get_zero_initialized_node(void);
  * After calling, the ROS node object can be used to create other middleware
  * primitives like publishers, services, parameters, etc.
  *
- * The name of the node must adhere to naming restrictions, see the
- * rmw_validate_node_name() function for rules.
+ * The name of the node must not be NULL and adhere to naming restrictions,
+ * see the rmw_validate_node_name() function for rules.
  *
  * \todo TODO(wjwwood): node name uniqueness is no yet enforced
  *
  * The name of the node cannot coincide with another node of the same name.
  * If a node of the same name is already in the domain, it will be shutdown.
  *
- * The namespace of the node should follow the rules for topics, see:
+ * The namespace of the node should not be NULL and should also follow the
+ * rules for topics, see:
  *
  *   http://design.ros2.org/articles/topic_and_service_names.html
  *
@@ -143,8 +144,8 @@ rcl_get_zero_initialized_node(void);
  * \post the node handle is valid and can be used in other `rcl_*` functions
  *
  * \param[inout] node a preallocated rcl_node_t
- * \param[in] name the name of the node
- * \param[in] namespace_ the namespace of the node
+ * \param[in] name the name of the node, must be a valid c-string
+ * \param[in] namespace_ the namespace of the node, must be a valid c-string
  * \param[in] options the node options
  * \return `RCL_RET_OK` if the node was initialized successfully, or
  * \return `RCL_RET_ALREADY_INIT` if the node has already be initialized, or
--- a/rcl/include/rcl/node.h
+++ b/rcl/include/rcl/node.h
@@ -44,9 +44,12 @@ typedef struct rcl_node_t
 typedef struct rcl_node_options_t
 {
   // bool anonymous_name;
+
   // rmw_qos_profile_t parameter_qos;
+
   /// If true, no parameter infrastructure will be setup.
   // bool no_parameters;
+
   /// If set, then this value overrides the ROS_DOMAIN_ID environment variable.
   /**
    * It defaults to RCL_NODE_OPTIONS_DEFAULT_DOMAIN_ID, which will cause the
@@ -59,6 +62,7 @@ typedef struct rcl_node_options_t
    *   (currently max size_t)
    */
   size_t domain_id;
+
   /// Custom allocator used for internal allocations.
   rcl_allocator_t allocator;
 } rcl_node_options_t;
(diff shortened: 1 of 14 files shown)


rcl/include/rcl/node.h
```diff
@@ -124,17 +144,24 @@ rcl_get_zero_initialized_node(void);
...
  * \return `RCL_RET_ERROR` if an unspecified error occurs.
  */
 RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_ret_t
-rcl_node_init(rcl_node_t * node, const char * name, const rcl_node_options_t * options);
+rcl_node_init(
+  rcl_node_t * node,
+  const char * name,
+  const char * name_space,
```
Karsten1987: why not putting `name_space` into `rcl_node_options_t` ? We can expose this struct on a high level language.

rcl/test/rcl/test_node.cpp
```diff
@@ -351,3 +380,145 @@ TEST_F(CLASSNAME(TestNodeFixture, RMW_IMPLEMENTATION), test_rcl_node_life_cycle)
...
+  {
+    rcl_node_t node = rcl_get_zero_initialized_node();
+    ret = rcl_node_init(&node, name, "/~/", &default_options);
+    ASSERT_EQ(RCL_RET_NODE_INVALID_NAMESPACE, ret);
+    rcl_ret_t ret = rcl_node_fini(&node);
+    EXPECT_EQ(RCL_RET_OK, ret);
+  }
+
+  // Node namespace with a trailing / which is not allowed.
+  {
```
Karsten1987: could you add a test with two consecutive slashes? Because on ROS1 history, this case appears quite often when concatenating NS and Topic. Just to make this case explicit.

rcl/include/rcl/node.h
wjwwood: I had two reasons for placing it like this.

The first was because I believe the namespace should be of elevated importance (similar to node name). Your same argument could be made for `name`, specifically that it should also be in the options struct, but to emulate the interface of ROS 1, node name and node namespace are high level concepts.

On the technical side, it is nice that the options structure remains a POD. As soon as I add a pointer (like a `const char *`) it is no longer "trivially copyable" (borrowing a term from C++). 

That all being said, I don't feel strongly about it. I definitely agree that most options should end up in the structure, but in this instance I made a decision to leave it out.

rcl/src/rcl/node.c
```diff
@@ -72,18 +79,46 @@ rcl_node_init(rcl_node_t * node, const char * name, const rcl_node_options_t * o
...
+  // Make sure the node name is valid before allocating memory.
+  int validation_result = 0;
+  size_t invalid_index = 0;
+  ret = rmw_validate_node_name(name, &validation_result, &invalid_index);
+  if (ret != RMW_RET_OK) {
+    RCL_SET_ERROR_MSG(rmw_get_error_string_safe());
+    return ret;
+  }
+  if (validation_result != RMW_NODE_NAME_VALID) {
+    const char * msg = rmw_node_name_validation_result_string(validation_result);
```
Karsten1987: i think this was discussed in another PR already. But why not apply the same logic as above here?
```
// sets the error message inside
ret = rmw_node_name_validation_result_string(validation_result);
if (ret != RMW_RET_OK)  // or similar for this use-case
{
  RCL_GET_ERROR_MSG_SAFE...
}
```

rcl/test/rcl/test_node.cpp
```diff
@@ -351,3 +380,145 @@ TEST_F(CLASSNAME(TestNodeFixture, RMW_IMPLEMENTATION), test_rcl_node_life_cycle)
...
+  }
+  {
+    rcl_node_t node = rcl_get_zero_initialized_node();
+    ret = rcl_node_init(&node, name, "/~/", &default_options);
+    ASSERT_EQ(RCL_RET_NODE_INVALID_NAMESPACE, ret);
+    rcl_ret_t ret = rcl_node_fini(&node);
+    EXPECT_EQ(RCL_RET_OK, ret);
+  }
+
+  // Node namespace with a trailing / which is not allowed.
```
Karsten1987: I know this is not allowed by design, but I wonder if we should manually filter this constraint to be allowed. It looks to me like a rather easy check for the last character. Just an idea, nothing critical to this PR.

rcl/test/rcl/test_node.cpp
```diff
@@ -351,3 +380,145 @@ TEST_F(CLASSNAME(TestNodeFixture, RMW_IMPLEMENTATION), test_rcl_node_life_cycle)
...
+  {
+    rcl_node_t node = rcl_get_zero_initialized_node();
+    ret = rcl_node_init(&node, name, "/ns/foo/", &default_options);
+    ASSERT_EQ(RCL_RET_NODE_INVALID_NAMESPACE, ret);
+    rcl_ret_t ret = rcl_node_fini(&node);
+    EXPECT_EQ(RCL_RET_OK, ret);
+  }
+
+  // Node namespace which is not absolute.
```
Karsten1987: same thinking as above.

rcl/src/rcl/node.c
wjwwood: `rmw_node_name_validation_result_string()` returns the validation issue description as a string, because I didn't want to force it into being an error. Also, this way you can formulate the error however you want (possibly using the input topic name and the violation index to make a better error message).

rcl/test/rcl/test_node.cpp
wjwwood: I'd rather leave this kind of flexibility to the top level API's (like rclcpp and rclpy). Each corner case makes it harder to describe and harder to test, so I'd like to keep `rcl` simple by avoiding them. That's just my personal preference though. I have no problem with allowing this exception to the rule here or in the higher level API's.

rcl/test/rcl/test_node.cpp
wjwwood: What if I add some "smoothing" here when I do the code which expands topics and stuff automatically (upcoming prs)? That way I can reuse those functions to automatically remove trailing `/` and add leading `/` if necessary.

Karsten1987: sounds fine to me.

rcl/include/rcl/node.h
```diff
@@ -77,11 +81,27 @@ rcl_get_zero_initialized_node(void);
...
+ *
+ *   http://design.ros2.org/articles/topic_and_service_names.html
+ *
+ * And therefore the namespace should pass validation of the
+ * rmw_validate_topic_name() function.
+ * The namespace should be absolute, i.e. it should begin with a forward slash.
+ * Even though this is a namespace, with additional relative namespace and/or
+ * a basename to follow, the namespace string passed here should not include
+ * a trailing forward slash.
+ * For an empty namespace, an empty string, i.e. "", should be used.
```
wjwwood: Good idea: dce05d6

rcl/src/rcl/node.c
```diff
@@ -72,18 +79,46 @@ rcl_node_init(rcl_node_t * node, const char * name, const rcl_node_options_t * o
     allocator->allocate, "allocate not set", return RCL_RET_INVALID_ARGUMENT);
   RCL_CHECK_FOR_NULL_WITH_MSG(
     allocator->deallocate, "deallocate not set", return RCL_RET_INVALID_ARGUMENT);
+  // Make sure the node name is valid before allocating memory.
+  int validation_result = 0;
+  size_t invalid_index = 0;
+  ret = rmw_validate_node_name(name, &validation_result, &invalid_index);
```
dirk-thomas: Since `invalid_index` isn't used it would be good to make the argument optional in the `rmw_validate_node_name` function.

rcl/src/rcl/node.c
```diff
@@ -72,18 +79,46 @@ rcl_node_init(rcl_node_t * node, const char * name, const rcl_node_options_t * o
...
+      msg = "unknown validation_result, this should not happen";
+    }
+    RCL_SET_ERROR_MSG(msg);
+    return RCL_RET_NODE_INVALID_NAME;
+  }
+  // Make sure the node namespace is valid too, but only if it is not an empty string.
+  if (strlen(namespace_) > 0) {
+    validation_result = 0;
+    invalid_index = 0;
+    ret = rmw_validate_topic_name(namespace_, &validation_result, &invalid_index);
```
dirk-thomas: Since the `namespace_` alone is not a FQN yet (it doesn't contain the topic name at the end, right?) doesn't this need a separate validation function (`rmw_validate_namespace`)?

wjwwood: It is either `/` or it is a valid FQN, so the topic name validation function is sufficient once I account for the `/` case. The code as written (and documented and tested) covers all the cases for namespace that I'm aware of. I don't think we need a separate namespace validation function myself.

dirk-thomas: A namespace can be empty. A topic name can't be. So both types have separate validation criteria, right? Therefore I would suggest to create separate validation functions for them. These implementation could share common functionality though if that makes sense.

wjwwood: See: https://github.com/ros2/rmw/pull/95#issuecomment-292370855

rcl/src/rcl/node.c
wjwwood: See: https://github.com/ros2/rmw/pull/95#issuecomment-292370855

wjwwood:
In 33e7a35 I started using `rmw_validate_namespace()`.

@Karsten1987 I also made `rcl_node_init()` more flexible in 33e7a35 to allow: `""`, `"/"`, and `"foo/bar"`.

Also @Karsten1987 , another possible use case for `utilities_strcat` which takes an allocator:

https://github.com/ros2/rcl/pull/112/commits/394048d32a89db65b73567069b50d29edeb4834a#diff-b1ed1b132e1e72fee905fe7bc3b4c62bR111

rcl/src/rcl/node.c
wjwwood: 33f084c

rcl/src/rcl/node.c
```diff
@@ -72,18 +80,64 @@ rcl_node_init(rcl_node_t * node, const char * name, const rcl_node_options_t * o
...
+  validation_result = 0;
+  ret = rmw_validate_namespace(local_namespace_, &validation_result, NULL);
+  if (ret != RMW_RET_OK) {
+    RCL_SET_ERROR_MSG(rmw_get_error_string_safe());
+    return ret;
+  }
+  if (validation_result != RMW_NAMESPACE_VALID) {
+    const char * msg = rmw_namespace_validation_result_string(validation_result);
+    if (!msg) {
+      msg = "unknown validation_result, this should not happen";
```
dirk-thomas: It would be good if the message included the value of `validation_result`.

wjwwood: Similar to the change in `rmw`: f0bec07

PR #116, opened 2017-04-11
fix cmake for test - remove unused params

Karsten1987:
addresses https://github.com/ros2/rcl/pull/108#discussion_r110696489

### DIFF
--- a/rcl/test/CMakeLists.txt
+++ b/rcl/test/CMakeLists.txt
@@ -65,10 +65,6 @@ function(test_target_function)
 
   rcl_add_custom_gtest(test_get_node_names${target_suffix}
     SRCS rcl/test_get_node_names.cpp
-    ENV
-      ${extra_test_env}
-      EMPTY_TEST=
-      NORMAL_TEST=foo
     APPEND_LIBRARY_DIRS ${extra_lib_dirs}
     LIBRARIES ${PROJECT_NAME} ${extra_test_libraries}
     AMENT_DEPENDENCIES ${rmw_implementation}
--- a/rcl/test/CMakeLists.txt
+++ b/rcl/test/CMakeLists.txt
@@ -65,6 +65,7 @@ function(test_target_function)
 
   rcl_add_custom_gtest(test_get_node_names${target_suffix}
     SRCS rcl/test_get_node_names.cpp
+    ENV ${extra_test_env}
     APPEND_LIBRARY_DIRS ${extra_lib_dirs}
     LIBRARIES ${PROJECT_NAME} ${extra_test_libraries}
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
@Karsten1987 can you also address https://github.com/ros2/rcl/pull/108#discussion_r110705337

Karsten1987:
Are you talking about running the tests for each rmw? If so, I need some feedback. I don't know why these tests are running for every rmw and why they shouldn't run for every rmw.

mikaelarguedas:
ok, I think it extends further than this single test anyway so let's do that in a follow-up

rcl/test/CMakeLists.txt
```diff
@@ -65,10 +65,6 @@ function(test_target_function)
 
   rcl_add_custom_gtest(test_get_node_names${target_suffix}
     SRCS rcl/test_get_node_names.cpp
-    ENV
-      ${extra_test_env}
```
mikaelarguedas: If you keep several rmw for now, please also pass the RMW_IMPLEMENTATION you want to test (by keeping line 68-69)

Karsten1987: done. I am not sure tho if this is actually needed.

mikaelarguedas: It is not strictly needed but if we register one test target per RMW_IMPLEMENTATION I think it makes sense to not run all of them with the default one

dirk-thomas:
> Are you talking about running the tests for each rmw? If so, I need some feedback. I don't know why these tests are running for every rmw and why they shouldn't run for every rmw.

If a test doesn't check anything which is rmw implementation specific it shouldn't run N times. Running it once would be sufficient (since the result of all N runs would be exactly the same) and therefore save resources.

Karsten1987:
In this particular case, I believe it makes sense since we want to verify that `nodes_names` are correctly set for all RMW.

dirk-thomas:
> In this particular case, I believe it makes sense since we want to verify that nodes_names are correctly set for all RMW.

My concern was not specific to the `test_get_node_names.cpp`. It was more general about all the tests in the `rcl` package. E.g. `test_common.cpp`, `test_time.cpp`.

mikaelarguedas:
In that case, it makes a lot of sense to pass RMW_IMPLEMENTATION to the `test_get_node_names` test. Also should the tests assert that the right rmw implementation is use? (using RCL_ASSERT_RMW_ID_MATCHES)

wjwwood:
Some of the tests probably test on multiple just because when we refactored the type support we didn't think about which tests should and should not test for multiple rmw. I'd be in favor of reducing tests to only use the default rmw in any cases it makes sense.

PR #118, opened 2017-04-11
use string_array_t from c_utilities package

Karsten1987:
connects to ros2/rmw#100

### DIFF
--- a/rcl/include/rcl/graph.h
+++ b/rcl/include/rcl/graph.h
@@ -101,6 +101,35 @@ rcl_ret_t
 rcl_destroy_topic_names_and_types(
   rcl_topic_names_and_types_t * topic_names_and_types);
 
+/// Return a list of available nodes in the ROS graph.
+/**
+ * This function returns a list of nodes in the ROS graph.
+ *
+ * The node parameter must not be `NULL`, and must point to a valid node.
+ *
+ * The node_names  parameter must be allocated and zero initialized.
+ * The node_names is the output for this function, and contains
+ * allocated memory.
+ * Use utilities_get_zero_initialized_string_array() for initializing an empty
+ * string_array_t struct.
+ * This node_names struct therefore  should be passed to utilities_string_array_fini()
+ * when it is no longer needed.
+ * Failing to do so will result in leaked memory.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Maybe [1]
+ * <i>[1] implementation may need to protect the data structure with a lock</i>
+ *
+ * \param[in] node the handle to the node being used to query the ROS graph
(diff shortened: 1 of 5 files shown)


rcl/include/rcl/graph.h
```diff
@@ -112,13 +106,7 @@ RCL_WARN_UNUSED
 rcl_ret_t
 rcl_get_node_names(
```
wjwwood: While you're here, please add some documentation for this function.

rcl/test/rcl/test_graph.cpp
```diff
@@ -150,42 +150,6 @@ TEST_F(
...
-  ret = rcl_destroy_node_names(nullptr);
-  EXPECT_EQ(RCL_RET_INVALID_ARGUMENT, ret) << rcl_get_error_string_safe();
-  rcl_reset_error();
-  // valid calls
-  ret = rcl_get_node_names(this->node_ptr, &node_names);
-  EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
-  ret = rcl_destroy_node_names(&node_names);
-  EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
-}
-
```
wjwwood: Do these tests not need to stay? And maybe be extended to actually check the contents of the node names listed? Or is that tested elsewhere (maybe system tests)?

Karsten1987: I'll removed them because of duplication.

The create and fini functions are covered in this one: https://github.com/ros2/c_utilities/blob/master/test/test_string_array.cpp
The semantics of `get_node_names` is covered here:
https://github.com/ros2/rcl/blob/master/rcl/test/rcl/test_get_node_names.cpp

Are we missing any tests on that?

wjwwood: Oh ok, I didn't notice the other tests in the file. Looks good, sorry for the noise.

rcl/test/rcl/test_get_node_names.cpp
```diff
@@ -75,14 +77,16 @@ TEST_F(CLASSNAME(TestGetNodeNames, RMW_IMPLEMENTATION), test_rcl_get_node_names)
...
-  rcl_string_array_t node_names = rcl_get_zero_initialized_string_array();
+  utilities_string_array_t node_names = utilities_get_zero_initialized_string_array();
   ret = rcl_get_node_names(node1_ptr, &node_names);
   ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
 
   EXPECT_EQ(size_t(2), node_names.size);
   EXPECT_EQ(0, strcmp(node1_name, node_names.data[0]));
   EXPECT_EQ(0, strcmp(node2_name, node_names.data[1]));
 
+  utilities_string_array_fini(&node_names);
```
wjwwood: This has a return code, I think you ought to check it. It's related to my comment about this function needing the "warn unused" attribute applied to it.

PR #122, opened 2017-04-13
move allocator and error handling to c utilities

wjwwood:
Connects to ros2/c_utilities#16


rcl/test/rcl/test_guard_condition.cpp
```diff
@@ -0,0 +1,201 @@
...
+    rcl_guard_condition_get_default_options();
+  options_with_failing_allocator.allocator.allocate = failing_malloc;
+  options_with_failing_allocator.allocator.deallocate = failing_free;
+  options_with_failing_allocator.allocator.reallocate = failing_realloc;
+  ret = rcl_guard_condition_init(&guard_condition, options_with_failing_allocator);
+  ASSERT_EQ(RCL_RET_BAD_ALLOC, ret) << "Expected RCL_RET_BAD_ALLOC";
+  // The error will not be set because the allocator will not work.
+  // It should, however, print a message to the screen and get the bad alloc ret code.
+  // ASSERT_TRUE(rcl_error_is_set());
+  // rcl_reset_error();
```
mikaelarguedas: did you keep these 2 commented lines on purpose? same below

rcl_lifecycle/src/rcl_lifecycle.c
```diff
@@ -67,6 +71,13 @@ rcl_lifecycle_state_machine_init(
   if (default_states) {
     rcl_lifecycle_init_default_state_machine(state_machine);
```
mikaelarguedas: this return code should be checked, doesn't it ?

rcl/test/rcl/test_guard_condition.cpp
wjwwood: Yeah, I explicitly kept those as an example of what I would normally do, but I am not due to the comment above.

rcl_lifecycle/src/rcl_lifecycle.c
wjwwood: It should, but the current implementation only ever returns `RCL_RET_OK`. Also it wasn't added in this pr 😛.

Anyways: 73c5f2c

PR #125, opened 2017-04-18
add support for DDS Security

mikaelarguedas:
connects to ros2/ros2#339

### DIFF
--- a/rcl/src/rcl/node.c
+++ b/rcl/src/rcl/node.c
@@ -23,7 +23,6 @@ extern "C"
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <sys/stat.h>
 
 #include "c_utilities/file_permissions.h"
 #include "c_utilities/get_env.h"
--- a/rcl/src/rcl/node.c
+++ b/rcl/src/rcl/node.c
@@ -247,8 +247,8 @@ rcl_node_init(
     } else {
       if (RMW_SECURITY_ENFORCEMENT_ENFORCE == node_security_options.enforce_security) {
         RCL_SET_ERROR_MSG(
-          "SECURITY ERROR: unable to find ROS_SECURE_ROOT directory while "
-          "security strategy requires it", *allocator);
+          "SECURITY ERROR: unable to find " RCUTILS_STRINGIFY(ROS_SECURITY_ROOT_DIRECTORY_VAR_NAME)
+          " directory while the requested security strategy requires it", *allocator);
         return RCL_RET_ERROR;
       }
     }
--- a/rcl/src/rcl/node.c
+++ b/rcl/src/rcl/node.c
@@ -25,6 +25,8 @@ extern "C"
 #include <string.h>
 #include <sys/stat.h>
 
+#include "c_utilities/file_permissions.h"
+#include "c_utilities/get_env.h"
 #include "rcl/rcl.h"
 #include "rmw/error_handling.h"
 #include "rmw/rmw.h"
@@ -53,8 +55,10 @@ typedef struct rcl_node_impl_t
 static const char * rcl_get_secure_root(const char * node_name)
 {
   const char * env_var_name = "ROS_SECURE_ROOT";
-  const char * ros_secure_root_env = NULL;  // todo: branch on OS
-  ros_secure_root_env = getenv(env_var_name);  // only on POSIX at the moment
+  const char * ros_secure_root_env = NULL;
+  if (!get_env(env_var_name, &ros_secure_root_env)) {
+    return NULL;
+  }
   if (!ros_secure_root_env) {
     return NULL;  // environment variable not defined
   }
@@ -62,26 +66,16 @@ static const char * rcl_get_secure_root(const char * node_name)
   if (!ros_secure_root_size) {
     return NULL;  // environment variable was empty
   }
-  const char node_secure_root_maxlen =
+  size_t node_secure_root_maxlen =
     strlen(ros_secure_root_env) + 1 + strlen(node_name) + 1;
-  char * node_secure_root = malloc(node_secure_root_maxlen);  // leak leak leak
-  // todo: use windows separator if needed
+  char * node_secure_root = malloc(node_secure_root_maxlen * sizeof(char));
   const char * separator =
     ros_secure_root_env[ros_secure_root_size - 1] == '/' ? "" : "/";
   snprintf(node_secure_root, node_secure_root_maxlen, "%s%s%s",
     ros_secure_root_env, separator, node_name);
-  struct stat buf;
-  if (stat(node_secure_root, &buf) < 0) {
+  if (!is_directory(node_secure_root)) {
     free(node_secure_root);
-    return NULL;  // path doesn't exist
-  }
-#ifndef WIN32
-  if (!(buf.st_mode & S_IRUSR)) {
-// #else
-//   if (!(buf.st_mode & _S_IREAD)) {
-#endif
-    free(node_secure_root);
-    return NULL;  // path is not readable by our process
+    return NULL;
   }
   return node_secure_root;
 }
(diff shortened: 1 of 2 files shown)


rcl/src/rcl/node.c
```diff
@@ -39,6 +42,10 @@ extern "C"
   _snprintf_s(buffer, buffer_size, _TRUNCATE, format, __VA_ARGS__)
 #endif
 
+static const char * sros_root_var_name = "ROS_SECURE_ROOT";
+static const char * sros_security_strategy_var_name = "ROS_SECURITY_STRATEGY";
+static const char * sros_enable_security_var_name = "ROS_ENABLE_SECURITY";
```
mikaelarguedas: Looking for feedback on variables name

rcl/src/rcl/node.c
```diff
@@ -179,7 +211,47 @@ rcl_node_init(
   }
   // actual domain id
   node->impl->actual_domain_id = domain_id;
-  node->impl->rmw_node_handle = rmw_create_node(name, local_namespace_, domain_id);
+
+  const char * ros_enable_security = NULL;
+  const char * ros_enforce_security = NULL;
+
+  if (rcutils_get_env(sros_enable_security_var_name, &ros_enable_security)) {
```
dirk-thomas: Since `rcutils_get_env` is not thread safe it make all code using that function not thread safe which I think is a problem here. I think we need a thread safe version of this if we want to use it throughout the code like here.

rcl/src/rcl/node.c
```diff
@@ -179,7 +211,47 @@ rcl_node_init(
...
+    RCL_SET_ERROR_MSG("fail fetching environment variable", rcl_get_default_allocator());
+    return RCL_RET_ERROR;
+  }
+
+  bool use_security = (0 == strcmp(ros_enable_security, "true")) ? true : false;
```
dirk-thomas: The `? true : false` is unnecessary. That is exactly what your expression already is.

mikaelarguedas: oh right I forgot I simplified the condition here, so you're right it's not necessary anymore

rcl/src/rcl/node.c
wjwwood: Not that it matters, but it is only not thread-safe on Windows. You could make a thread safe version by making this global thread-local:

https://github.com/ros2/rcutils/blob/master/src/get_env.c#L27

wjwwood: Actually looks like the posix `getenv()` need not be thread-safe either: http://pubs.opengroup.org/onlinepubs/9699919799/functions/getenv.html

So we'd have to use the thread-local storage on all platforms and copy the requested value into the thread local storage. That would at least make it thread-safe with concurrent calls to `rcutils_get_env` though it would still not be thread-safe with setenv, but that's less important I think.

mikaelarguedas: thanks @wjwwood for looking into it. I'll add a "pseudo thread-safe" version of it in rcutils then

mikaelarguedas: Given that getenv is used in various places in the codebase already. I think that implementing a pseudo thread-safe version of it goes along refactoring all calls to get env throughout the codebase. So I'm going to submit this and the associated PRs for review right now and iterate on the getenv improvement/refactor in another set of PRs.

dirk-thomas: Sounds good to me. Can you please create a ticket to track making `getenv` thread safe.

rcl/src/rcl/node.c
dirk-thomas: I think it would be better if all SROS specific environment variable share a common prefix, e.g. `ROS_SECURITY_*`.

rcl/src/rcl/node.c
```diff
@@ -179,7 +211,47 @@ rcl_node_init(
...
+  }
+
+  rmw_node_security_options_t node_security_options =
+    rmw_get_zero_initialized_node_security_options();
+  node_security_options.enforce_security = (0 == strcmp(ros_enforce_security, "Enforce")) ?
+    RMW_SECURITY_ENFORCEMENT_ENFORCE : RMW_SECURITY_ENFORCEMENT_PERMISSIVE;
+
+  if (!use_security) {
+    node_security_options.enforce_security = RMW_SECURITY_ENFORCEMENT_PERMISSIVE;
+    node_security_options.security_root_path = NULL;
```
dirk-thomas: Is this not covered by `rmw_get_zero_initialized_node_security_options`?

rcl/src/rcl/node.c
```diff
@@ -179,7 +211,47 @@ rcl_node_init(
   }
   // actual domain id
   node->impl->actual_domain_id = domain_id;
-  node->impl->rmw_node_handle = rmw_create_node(name, local_namespace_, domain_id);
+
+  const char * ros_enable_security = NULL;
+  const char * ros_enforce_security = NULL;
+
+  if (rcutils_get_env(sros_enable_security_var_name, &ros_enable_security)) {
+    RCL_SET_ERROR_MSG("fail fetching environment variable", rcl_get_default_allocator());
```
dirk-thomas: The error message should mention which env var wasn't found. Also I think "fetch" is not a common verb for env vars. Maybe better: "Env var 'xxx' not set".

rcl/src/rcl/node.c
mikaelarguedas: sounds good https://github.com/ros2/rcutils/issues/30

rcl/src/rcl/node.c
mikaelarguedas: That would be great indeed, I don't remember RCL_SET_ERROR_MSG taking formatting arguments so I didn't do it. I could write directly the variable name in the string for each condition but that seems wrong to me so I stuck with just saying that it failed. I can replace 'fetching' with 'to read' though

wjwwood: If you made the constants `#define`'s then you could do `RCUTILS_STRINGIFY(...)` in both the function call and in this error message.

rcl/src/rcl/node.c
mikaelarguedas: This is my (arguably dirty) way of making sure that the underlying rmw will not create a secure node if use_security is false. I invalidate the path in rcl rather than adding a field to the security_options structure (to say enable or disable) allowing to avoid duplicating the logic in each rmw_implementation

rcl/src/rcl/node.c
mikaelarguedas: oh cool I didn't know about that. thanks @wjwwood for the pointer! https://github.com/ros2/rcl/pull/125/commits/1407fe7ab09c123ad045cbf6604fd5b2192ac0eb

rcl/src/rcl/node.c
```diff
@@ -39,6 +43,10 @@ extern "C"
   _snprintf_s(buffer, buffer_size, _TRUNCATE, format, __VA_ARGS__)
 #endif
 
+#define SROS_ROOT_VAR_NAME "ROS_SECURITY_ROOT_DIRECTORY"
+#define SROS_SECURITY_STRATEGY_VAR_NAME "ROS_SECURITY_STRATEGY"
+#define SROS_SECURITY_ENABLE_VAR_NAME "ROS_SECURITY_ENABLE"
```
dirk-thomas: Why not use a consistent define name, like `ROS_SECURITY_xxx_VAR_NAME`?

rcl/src/rcl/node.c
dirk-thomas: Ok, but isn't `node_security_options.security_root_path` already `NULL` before this line (due to `node_security_options = rmw_get_zero_initialized_node_security_options()`)?

mikaelarguedas: correct :confounded: , fixed in https://github.com/ros2/rcl/pull/125/commits/e653ae685e98855c3fb51cf3ebbbc83d4c7bfe2e

PR #132, opened 2017-04-27
Namespace expansion

wjwwood:
This pull request adds two main functions:

- `rcl_validate_topic_name()`
  - In contrast to `rmw_validate_full_topic_name()` this function validates relative topic names and topic names with things like `~` and `{substitutions}` in them
- `rcl_expand_topic_name()`
  - This function takes a topic and expands it to a fully qualified name, using a node name, namespace, and a map of other possible substitutions

I would like to merge these functions (and the related prs) before taking the final step and calling these functions from within the `rcl_create_*` functions which take a topic. It would also be good to do that last step after, or at the same time as, the namespaces work @Karsten1987 has been doing under the rmw layer.


wjwwood:
CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2563/)
- Linux-aarch64:
  - [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/160/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/2032/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2670/)

Still looking for reviews, thanks!

wjwwood:
Can someone please review these changes? They've been ready for 10 days now.

rcl/include/rcl/expand_topic_name.h
```diff
@@ -0,0 +1,132 @@
...
+  const rcutils_string_map_t * substitutions,
+  rcl_allocator_t allocator,
+  char ** output_topic_name);
+
+/// Fill a given string map with the default substitution pairs.
+/**
+ * If the string map is not initialized RCL_RET_INVALID_ARGUMENT is returned.
+ *
+ * \param[inout] string_map rcutils_string_map_t map to be filled with pairs
+ * \return const pointer to default substitutions as a string map
```
Karsten1987: the string map is not a return value?

rcl/src/rcl/expand_topic_name.c
```diff
@@ -0,0 +1,252 @@
...
+    // These assumptions are taken because this is checked in the validation function.
+    const char * current_output = (local_output) ? local_output : input_topic_name;
+    char * next_opening_brace = NULL;
+    while ((next_opening_brace = strchr(current_output, '{')) != NULL) {
+      char * next_closing_brace = strchr(current_output, '}');
+      // conclusion based on above assumptions: next_closing_brace - next_opening_brace > 1
+      size_t substitution_substr_len = next_closing_brace - next_opening_brace + 1;
+      // figure out what the replacement is for this substitution
+      const char * replacement = NULL;
+      if (strncmp("{node}", next_opening_brace, substitution_substr_len) == 0) {
```
Karsten1987: I'd propose exposing the substitution literals with constants in a common place. This gets easily lost like this.

rcl/src/rcl/expand_topic_name.c
```diff
@@ -0,0 +1,252 @@
...
+        replacement = node_name;
+      } else if (  // NOLINT
+        strncmp("{ns}", next_opening_brace, substitution_substr_len) == 0 ||
+        strncmp("{namespace}", next_opening_brace, substitution_substr_len) == 0)
```
Karsten1987: why not putting `node`, `ns` and `namespace` as default values when initializing the substitution map? With this, you simply have to iterate over the map. This also comes along with the previous comment.

rcl/src/rcl/expand_topic_name.c
```diff
@@ -0,0 +1,252 @@
...
+    // Assumptions entering this scope about the topic string:
+    //
+    // - All {} are matched and balanced
+    // - There is no nesting, i.e. {{}}
+    // - There are no empty substitution substr, i.e. '{}' versus '{something}'
+    //
+    // These assumptions are taken because this is checked in the validation function.
+    const char * current_output = (local_output) ? local_output : input_topic_name;
+    char * next_opening_brace = NULL;
+    while ((next_opening_brace = strchr(current_output, '{')) != NULL) {
```
Karsten1987: I think this while loop deserves a quick comment on what's happening with `current_output` and `local_output`. It makes it easier to follow throughout the rest of this code.

rcl/test/rcl/test_expand_topic_name.cpp
```diff
@@ -0,0 +1,221 @@
...
+  rcutils_string_map_t subs = rcutils_get_zero_initialized_string_map();
+  rcutils_ret_t rcu_ret = rcutils_string_map_init(&subs, 0, allocator);
+  ASSERT_EQ(RCUTILS_RET_OK, rcu_ret);
+  ret = rcl_get_default_topic_name_substitutions(&subs);
+  ASSERT_EQ(RCL_RET_OK, ret);
+
+  // {node}/chatter example
+  {
+    const char * topic = "{node}/chatter";
+    const char * ns = "/my_ns";
```
Karsten1987: I think on `rcl` level we should allow non-absolute namespaces and attach the root slash ourselves.

rcl/test/rcl/test_expand_topic_name.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+  std::vector<std::vector<std::string>> topics_that_should_expand_to = {
+    // {"input_topic", "node_name", "/namespace", "expected result"},
+    {"/chatter", "my_node", "/my_ns", "/chatter"},
+    {"chatter", "my_node", "/my_ns", "/my_ns/chatter"},
+    {"{node}/chatter", "my_node", "/my_ns", "/my_ns/my_node/chatter"},
+    {"/{node}", "my_node", "/my_ns", "/my_node"},
+    {"{node}", "my_node", "/my_ns", "/my_ns/my_node"},
+    {"{ns}", "my_node", "/my_ns", "/my_ns"},
+    {"{namespace}", "my_node", "/my_ns", "/my_ns"},
```
Karsten1987: can you add a test with double substitutions?
`"{namespace}/{node}/chatter", "my_node", "/my_ns", "/my_ns/my_node/chatter"}`

rcl/src/rcl/expand_topic_name.c
wjwwood: What comment do you want? I don't know what I would put there to make it clearer.

rcl/src/rcl/expand_topic_name.c
wjwwood: > This gets easily lost like this.

What do you mean? It is documented in the docstring for this function. How does it get lost? Get lost for whom?

rcl/src/rcl/expand_topic_name.c
wjwwood: I didn't put it in the default substitutions for two reasons:

- so the default substitutions were not tied to a single node and could be reused
- because you have to pass the node name and namespace to this function separately anyways because they are used to expand `~` and make relative topics absolute too

I considered putting them in the substitutions, but because of the reasons above I decided not to do that. It also avoids needing to copy the name and namespace into the map.

rcl/test/rcl/test_expand_topic_name.cpp
wjwwood: It is allowed when creating a node, but here (in `rcl_expand_topic_name`) I'm assuming that this is not the case so the logic does not have to be repeated:

https://github.com/ros2/rcl/blob/90176d9f1a3551deb861e698f36f81534e7763ac/rcl/src/rcl/node.c#L121

wjwwood:
New CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2582/)
- Linux-aarch64:
  - [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/174/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/2050/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2704/)

rcl/src/rcl/expand_topic_name.c
Karsten1987: i was thinking about putting these literals somewhere in constant variables, to have them exposed in a single place and can be changed solely there in case they have to be reused at some point. But just my opinion,

rcl/src/rcl/expand_topic_name.c
Karsten1987: I was just thinking about mentioning that this local_output will be piecewise replaced within this while loop. But nvmd, maybe it was just me.

rcl/src/rcl/expand_topic_name.c
wjwwood: I can put them in a constant, but they were only used in the code in one place. Anyways, it's fine by me now that I understand what you mean: ec85415

rcl/src/rcl/expand_topic_name.c
wjwwood: I don't understand the use of the term `piecewise` here. I happy to add a comment, but I still don't know what would make it clearer for you.

wjwwood: I took a stab at it in 655b871, is that clearer?

rcl/test/rcl/test_expand_topic_name.cpp
wjwwood: I added a test for this in 8ff0e82

wjwwood:
The previous windows failure was due to a time delta being too large:

http://ci.ros2.org/job/ci_windows/2704/testReport/(root)/projectroot/test_wait__rmw_connext_cpp/

It looks unrelated to this change.

Otherwise CI was ok. I recompiled and tested locally for ec85415, so I think it's safe.

rcl/src/rcl/expand_topic_name.c
Karsten1987: All I had in mind was to briefly explain what is going to happen in this while loop as it's big long and does quite some operations. 
But that comment helps already.

Karsten1987:
lgtm

wjwwood:
@Karsten1987 is this good to go or are there still some things to address?

PR #135, opened 2017-05-08
Namespace impl

Karsten1987:
adding a test for namespaces. 
It's one simple test for services, but as they are implemented with pub/sub internally, these are also covered by that test.

Connects to ros2/ros2#327


wjwwood:
The test doesn't use the `RMW_IMPLEMENTATION` in the test names, like these test:

https://github.com/ros2/rcl/blob/master/rcl/test/rcl/test_publisher.cpp#L28-L35

rcl/test/rcl/test_namespace.cpp
```diff
@@ -0,0 +1,137 @@
+// Copyright 2016 Open Source Robotics Foundation, Inc.
```
wjwwood: The copyright year should be 2017.

wjwwood: Also, please consider putting this file in `rcl/test` rather than `rcl/test/rcl`. I've tried to keep tests in the `rcl/test/rcl` folder a one-to-one match with source files in `rcl/src/rcl`. Since this doesn't correspond to a `rcl/src/rcl/namespace.c` file (AFAICT), I think it can be raised up a folder level.

rcl/test/rcl/test_namespace.cpp
```diff
@@ -0,0 +1,137 @@
...
+  rcl_node_t * node_ptr;
+  void SetUp()
+  {
+    stop_memory_checking();
+    rcl_ret_t ret;
+    ret = rcl_init(0, nullptr, rcl_get_default_allocator());
+    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
+    this->node_ptr = new rcl_node_t;
+    *this->node_ptr = rcl_get_zero_initialized_node();
+    const char * name = "node_name";
```
wjwwood: Since you changed the name in `test_graph.cpp`, maybe this should be something like `test_namespace_node`?

rcl/test/rcl/test_namespace.cpp
```diff
@@ -0,0 +1,137 @@
...
+  bool is_available = false;
+  for (auto i = 0; i < timeout; ++i) {
+    ret = rcl_service_server_is_available(this->node_ptr, &unmatched_client, &is_available);
+    if (is_available) {
+      // this should not happen
+      break;
+    }
+    std::this_thread::sleep_for(1s);
+  }
+  ASSERT_FALSE(is_available);
```
wjwwood: I don't see a different way to assert it, but this test will always take 10s when it succeeds which kind of sucks in terms of keeping the tests running quickly.

wjwwood:
There are still comments on this one that need to be addressed I think.

PR #137, opened 2017-05-19
Use namespaces

wjwwood:
This pr is part of the final set of changes needed to support namespaces. It actually change the `rcl_create_*` functions which take a topic or service name to expand those names and validate them before passing the fully qualified version down to `rmw` functions.

I still need to expose this stuff into `rclcpp` and `rclpy`, as well as implement the "avoid ros prefix option".

### DIFF
--- a/rcl/src/rcl/client.c
+++ b/rcl/src/rcl/client.c
@@ -77,7 +77,7 @@ rcl_client_init(
     return RCL_RET_ALREADY_INIT;
   }
   // Expand the given service name.
-  rcutils_allocator_t rcutils_allocator = rcutils_get_default_allocator();
+  rcutils_allocator_t rcutils_allocator = *allocator;  // implicit conversion to rcutils version
   rcutils_string_map_t substitutions_map = rcutils_get_zero_initialized_string_map();
   rcutils_ret_t rcutils_ret = rcutils_string_map_init(&substitutions_map, 0, rcutils_allocator);
   if (rcutils_ret != RCUTILS_RET_OK) {
--- a/rcl/src/rcl/client.c
+++ b/rcl/src/rcl/client.c
@@ -19,10 +19,13 @@ extern "C"
 
 #include "rcl/client.h"
 
+#include <stdio.h>
 #include <string.h>
 
+#include "rcl/expand_topic_name.h"
 #include "rmw/error_handling.h"
 #include "rmw/rmw.h"
+#include "rmw/validate_full_topic_name.h"
 
 #include "./common.h"
 #include "./stdatomic_helper.h"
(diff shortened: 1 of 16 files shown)


wjwwood:
I was going to add the option to disable the ROS specific prefix in this pull request, but it makes it harder to review in my opinion (I've been applying those changes locally) because it touches a lot of otherwise unrelated stuff like how options are propagated for the publishers and subscribers. So I'd rather merge this set of pr's and open new ones for the new option.

I'm putting these in review, here's a first set of CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2642/)
- Linux-aarch64:
  - [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/199/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/2100/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2763/)

wjwwood:
Another set of CI with some small test related fixes:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2644/)
- Linux-aarch64:
  - [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/201/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/2106/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2765/)

The remaining macOS failure is a test timeout. I think that timeout is just too close to a normal runtime. Each Connext test takes like 3-6 seconds to shutdown before then next can run. On my machine it is flaky I am going to up the timeout from 30 to 60 seconds and rerun the macOS CI, but I don't consider that a regression atm.

Please review these if you have time. Thanks.

wjwwood:
See ros2/system_tests@ba60562 for the updated timeout.

rcl/src/rcl/client.c
```diff
@@ -73,6 +76,66 @@ rcl_client_init(
     RCL_SET_ERROR_MSG("client already initialized, or memory was unintialized", *allocator);
     return RCL_RET_ALREADY_INIT;
   }
+  // Expand the given service name.
+  rcutils_allocator_t rcutils_allocator = rcutils_get_default_allocator();
+  rcutils_string_map_t substitutions_map = rcutils_get_zero_initialized_string_map();
+  rcutils_ret_t rcutils_ret = rcutils_string_map_init(&substitutions_map, 0, rcutils_allocator);
```
Karsten1987: why not using the allocator from above?

wjwwood:
Updated macOS CI is green (with updated timeout):

[Build Status](http://ci.ros2.org/job/ci_osx/2107/)

rcl/src/rcl/client.c
wjwwood: Well, my original reasoning was that the rcutils allocator and rcl allocator might be different in the future, but I think that's not as important as using the given allocator as much as possible. I'll change this (and look for other places where I made this choice).

wjwwood:
> I believe this expansion could be excluded in a separate function called get_fully_expanded_topic_name or similar as the changes in publisher, subscriber, .. are all pretty much the same given a topic/service name and a node namespace

That's true, but they are repeated because they return different errors based on some things (especially between pub/sub and client/service).

In the interest of moving this forward I'll ticket this and do it after we merge this.

rcl/src/rcl/client.c
wjwwood: Have a look at 3284737.

wjwwood:
CI after addressing comments and getting approval from @Karsten1987:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2650/)
- Linux-aarch64:
  - [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/205/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/2108/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2769/)

PR #139, opened 2017-05-23
Warn unused

Karsten1987:
adding unused macros for `rcl_lifecycle`


rcl_lifecycle/src/com_interface.h
```diff
@@ -20,12 +20,15 @@ extern "C"
 {
 #endif
 
+#include "rcl/macros.h"
```
mikaelarguedas: Nit: `#include <rcl/macros.h>` like above

Same below

PR #140, opened 2017-06-01
move time files to rcutils

dirk-thomas:
Connect to ros2/rcutils#25.

Please review.


rcl/include/rcl/time.h
```diff
@@ -487,68 +488,6 @@ rcl_ret_t
...
- *
- * \param[out] now a struct in which the current time is stored
- * \return `RCL_RET_OK` if the current time was successfully obtained, or
- * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
- * \return `RCL_RET_ERROR` an unspecified error occur.
- */
-RCL_PUBLIC
-RCL_WARN_UNUSED
-rcl_ret_t
-rcl_steady_time_now(rcl_time_point_value_t * now);
```
wjwwood: If you alias these two functions you can get rid of the pull request in `rclcpp`: https://github.com/ros2/rclcpp/pull/332/files

And the API of `rcl`'s time functions will remain unchanged due to these pr's.

wjwwood: It will also remove the need for several other changes in this pr.

dirk-thomas: I don't think the functions should be aliases. First, that makes it harder to see where the functionality is coming from. Second, in this case the function is not even supposed to be used by users of `rcl` anyway. Users will use the time source / time source type instead. Therefore I would prefer to leave this as-is.

PR #141, opened 2017-06-06
Explicitly set C11 and C++14 via the CMake variables

esteve:
This PR uses `CMAKE_C_STANDARD` and `CMAKE_CXX_STANDARD` to set C11 and C++14 respectively.

### DIFF
--- a/rcl/CMakeLists.txt
+++ b/rcl/CMakeLists.txt
@@ -23,8 +23,7 @@ if(NOT CMAKE_CXX_STANDARD)
 endif()
 
 if(NOT WIN32)
-  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
+  add_compile_options("-Wall -Wextra")
 endif()
 
 set(${PROJECT_NAME}_sources
--- a/rcl/CMakeLists.txt
+++ b/rcl/CMakeLists.txt
@@ -23,7 +23,7 @@ if(NOT CMAKE_CXX_STANDARD)
 endif()
 
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-  add_compile_options("-Wall -Wextra")
+  add_compile_options(-Wall -Wextra)
 endif()
 
 set(${PROJECT_NAME}_sources
--- a/rcl/CMakeLists.txt
+++ b/rcl/CMakeLists.txt
@@ -12,8 +12,15 @@ find_package(rosidl_generator_c REQUIRED)
 
 include_directories(include)
 
-set(CMAKE_C_STANDARD 99)
-set(CMAKE_CXX_STANDARD 14)
+# Default to C99
+if(NOT CMAKE_C_STANDARD)
+  set(CMAKE_C_STANDARD 99)
+endif()
+
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
 
 if(NOT WIN32)
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra")
--- a/rcl/CMakeLists.txt
+++ b/rcl/CMakeLists.txt
@@ -12,9 +12,9 @@ find_package(rosidl_generator_c REQUIRED)
 
 include_directories(include)
 
-# Default to C99
+# Default to C11
 if(NOT CMAKE_C_STANDARD)
-  set(CMAKE_C_STANDARD 99)
+  set(CMAKE_C_STANDARD 11)
 endif()
 
 # Default to C++14
--- a/rcl/CMakeLists.txt
+++ b/rcl/CMakeLists.txt
@@ -22,7 +22,7 @@ if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 14)
 endif()
 
-if(NOT WIN32)
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   add_compile_options("-Wall -Wextra")
 endif()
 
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
:+1: If you want I think I have a "compile_options" branch on all repos to do this and pass the additional flags using `add_compile_options` rather than appending the flags to both the C and C++ flags

rcl/CMakeLists.txt
```diff
@@ -12,9 +12,12 @@ find_package(rosidl_generator_c REQUIRED)
 
 include_directories(include)
 
+set(CMAKE_C_STANDARD 99)
+set(CMAKE_CXX_STANDARD 14)
```
dirk-thomas: These variables should only be set if they are not yet defined in order to not prevent overriding the value.

Same below.

esteve: Fixed in https://github.com/ros2/rcl/pull/141/commits/4e092bc560ff7eeaa242a6da05ae13ae6cfcb3e8

rcl/CMakeLists.txt
```diff
@@ -12,9 +12,19 @@ find_package(rosidl_generator_c REQUIRED)
...
+
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+
 if(NOT WIN32)
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14 -Wall -Wextra")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
```
mikaelarguedas: These two lines could become a one-liner by using add_compile_options (I don't mind making the change if it's ok with you @esteve )

rcl_lifecycle/CMakeLists.txt
```diff
@@ -11,6 +11,21 @@ find_package(rmw REQUIRED)
...
+endif()
+
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+
+if(NOT WIN32)
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -Wpedantic")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wpedantic")
```
mikaelarguedas: These two lines could become a one-liner by using `add_compile_options`.
Also I think the previous condition was more accurate: `CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang"` because these flags are compiler dependant rather than Platform dependant

rcl/CMakeLists.txt
esteve: That'd be great, thanks!

rcl_lifecycle/CMakeLists.txt
esteve: That's true, though wouldn't that be only for the case where `rcl` is being compiled on Windows with `Clang` or `GCC`? The rest of ROS2 uses the same style `if(NOT WIN32) ...`

mikaelarguedas: Yeah my goal is to start converting the other ones to use the same logic. I agree that there are for now few configurations where this doesn't give the same result but that will be more future proof if we add more platforms in the future

rcl/CMakeLists.txt
esteve: Done in https://github.com/ros2/rcl/pull/141/commits/f4cb53273e1e8869675311823d9de22887ad36f6

rcl_lifecycle/CMakeLists.txt
esteve: I went ahead and revert to check the compiler instead of the operating system in https://github.com/ros2/rcl/pull/141/commits/2ecb36d556db75c1c99d07182e9ad5cc21e20939

rcl/CMakeLists.txt
esteve: Also done for `rcl_lifecycle` in https://github.com/ros2/rcl/pull/141/commits/cc2c9a5f8ead5708b26fae0f3fa9f56238594339

rcl_lifecycle/CMakeLists.txt
esteve: Also done for `rcl_lifecycle` in https://github.com/ros2/rcl/pull/141/commits/cc2c9a5f8ead5708b26fae0f3fa9f56238594339

rcl/CMakeLists.txt
```diff
@@ -12,9 +12,18 @@ find_package(rosidl_generator_c REQUIRED)
...
+  set(CMAKE_C_STANDARD 99)
+endif()
+
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options("-Wall -Wextra")
```
mikaelarguedas: I think that we didn't because of the memory tools. see https://github.com/ros2/ros2/issues/305

esteve:
I just rebased this to resolve any potential conflicts, let me know if there's any further work that needs to be done.

esteve:
Also, pushed a last minute change to pass the compiler flags as a list, not as a string, to `add_compile_options()`

mikaelarguedas:
Thanks @esteve, I'm putting CI results on this ticket https://github.com/ros2/ros2/issues/362.

I think that we're good to go. Though this package actually needs c11 otherwise there are warnings on macos. Can you cherry-pick this commit https://github.com/ros2/rcl/commit/71bb5bb52041acaf190692df69a6138aa626242f? thanks!

esteve:
@mikaelarguedas done. I've updated the PR title, but unfortunately the branch name is c99, shall I close this and open a new one with a better name?

mikaelarguedas:
It's ok, the title of the PR is what will appear in the history of this repo and you already updated it accrdingly :+1:

PR #148, opened 2017-06-27
ensure node is valid before using it

dirk-thomas:
Otherwise `rcl_node_get_rmw_handle` is being called, sets an error, returns null, and the next function like `rmw_destroy_*` is being called with `NULL` as the rmw node.

I didn't trigger CI since there is currently no test which has this case.

Ready for review.

### DIFF
--- a/rcl/src/rcl/client.c
+++ b/rcl/src/rcl/client.c
@@ -173,8 +173,11 @@ rcl_client_fini(rcl_client_t * client, rcl_node_t * node)
   RCL_CHECK_ARGUMENT_FOR_NULL(node, RCL_RET_INVALID_ARGUMENT, rcl_get_default_allocator());
   if (client->impl) {
     rcl_allocator_t allocator = client->impl->options.allocator;
-    rmw_ret_t ret =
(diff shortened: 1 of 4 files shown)


allenh1:
This looks connected to #81, so I'm tagging it here.

dirk-thomas:
> I think this pr should use the return value of `rcl_node_get_rmw_handle()` rather than using `rcl_node_is_valid()` directly.

Makes sense. Updated.

allenh1:
> This is an improvement because it is handling the error case better. However, I would say you should check the return value of rcl_node_get_rmw_handle for NULL rather than using rcl_node_is_valid. The reason being, that the same check is done within rcl_node_get_rmw_handle and because I think the conditions under which that function will work might be less strict that "is valid".

@wjwwood In #103 I check that the `rmw_handle` is not `NULL` and, if it is, `is_valid` should return `false`. Should I change this behavior?

wjwwood:
@allenh1 I'm not sure yet, but I've pointed this out before: https://github.com/ros2/rcl/pull/103#issuecomment-270222010

dirk-thomas:
@allenh1 Can you please move the discussion related to #103 to that ticket.

PR #152, opened 2017-07-25
remove fprintf, use logging macros

Karsten1987:
connects to ros2/rclcpp#344

title is self-explanatory

### DIFF
--- a/rcl_lifecycle/src/rcl_lifecycle.c
+++ b/rcl_lifecycle/src/rcl_lifecycle.c
@@ -27,6 +27,8 @@ extern "C"
 #include "rcl_lifecycle/rcl_lifecycle.h"
 #include "rcl_lifecycle/transition_map.h"
 
+#include "rcutils/logging_macros.h"
+
 #include "com_interface.h"  // NOLINT
 #include "default_state_machine.h"  // NOLINT
 #include "states.h"  // NOLINT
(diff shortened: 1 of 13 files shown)


rcl_lifecycle/src/rcl_lifecycle.c
```diff
@@ -163,14 +165,14 @@ rcl_lifecycle_trigger_transition(
 
   // If we have a faulty transition pointer
   if (!transition) {
-    fprintf(stderr, "No transition found for node %s with key %d\n",
-      state_machine->current_state->label, key);
+    RCUTILS_LOG_ERROR("No transition found for node %s with key %d",
+      state_machine->current_state->label, key)
```
mikaelarguedas: Nitpicks: some macros have `;` after the closing parenthesis and some don't. I would expect (according to how the macros are implemented) to either fail to compile without the semi-colon or to produce a pedantic warning if they present and not needed

rcl_lifecycle/src/rcl_lifecycle.c
```diff
@@ -27,6 +27,8 @@ extern "C"
 #include "rcl_lifecycle/rcl_lifecycle.h"
 #include "rcl_lifecycle/transition_map.h"
 
+#include "rcutils/logging_macros.h"
```
mikaelarguedas: Nit: I think hat this should be imported with `<` like in the file above

rcl_lifecycle/src/rcl_lifecycle.c
dirk-thomas: The macro invocation should **not** be followed by a semicolon. The rational is that the macros might evaluate to *nothing* in case the macro (or at least the specific log level) is being compiled out for efficiency reasons. At that point a trailing semicolon would be left.

PR #156, opened 2017-08-08
Timer waitset modifications

dhood:
Opening for visibility.

This is an abandoned branch that from the comments in https://github.com/ros2/rclpy/pull/54#issuecomment-261298344 aims to improve timer wakeup

Does anyone have any context for why there wasn't a PR opened originally for this?

### DIFF
--- a/rcl/src/rcl/wait.c
+++ b/rcl/src/rcl/wait.c
@@ -556,6 +556,21 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
     &wait_set->impl->rmw_clients,
     wait_set->impl->rmw_waitset,
     timeout_argument);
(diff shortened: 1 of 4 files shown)


mikaelarguedas:
I don't have context but remember that I thought some of it could be reused to improve performace when I was investigating timers  and waitset issues when implementing them in python. If this can improve the performace we may be able to [reenable some tests](https://github.com/ros2/rclpy/blob/c49f318efb80eb1f2f32fd65c3ca11b18ddbbf29/rclpy/test/test_timer.py#L139-L140)

wjwwood:
@mikaelarguedas should we close this or do you still want to consider the changes to improve test reliability?

clalancette:
There's been no activity on this for a long time, so I'm just going to close it out.  We can resurrect things in the future if we think it is useful.

PR #157, opened 2017-08-10
Rcl use rcutils

mikaelarguedas:
first commit switch prints to logging macro
second commit uses rcutils_snprintf rather than snprintf
the other ones are style

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2975/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/414/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2404/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3084/)

### DIFF
--- a/rcl/include/rcl/expand_topic_name.h
+++ b/rcl/include/rcl/expand_topic_name.h
@@ -64,7 +64,7 @@ extern "C"
  * if (ret != RCL_RET_OK) {
  *   // ... error handling
  * } else {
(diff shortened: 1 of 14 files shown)


mikaelarguedas:
Done, do you want me to create a merge commit for this to keep both commits separated in history ? or squash and merge just so that this PR is cleaner but the history sees only one commit ?

dirk-thomas:
I would suggest a merge commit.

PR #158, opened 2017-08-10
Fix memory problems in rcl.

clalancette:
There are actually two problems here.  In one of the problems,
we were sometimes taking an rcl_guard_condition that we did *not*
allocate, and trying to deallocate it.  This was leading to
double frees.

The second problem was forgetting to call guard_condition_fini
during node_fini.

connects to ros2/ros2#387

CI jobs are in the above issue

### DIFF
--- a/rcl/src/rcl/guard_condition.c
+++ b/rcl/src/rcl/guard_condition.c
@@ -27,6 +27,7 @@ extern "C"
 typedef struct rcl_guard_condition_impl_t
 {
   rmw_guard_condition_t * rmw_handle;
+  bool allocated_rmw_guard_condition;
   rcl_guard_condition_options_t options;
 } rcl_guard_condition_impl_t;
 
@@ -76,6 +77,7 @@ __rcl_guard_condition_init_from_rmw_impl(
   if (rmw_guard_condition) {
     // If given, just assign (cast away const).
     guard_condition->impl->rmw_handle = (rmw_guard_condition_t *)rmw_guard_condition;
+    guard_condition->impl->allocated_rmw_guard_condition = false;
   } else {
     // Otherwise create one.
     guard_condition->impl->rmw_handle = rmw_create_guard_condition();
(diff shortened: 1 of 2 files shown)


rcl/src/rcl/node.c
```diff
@@ -325,11 +325,21 @@ rcl_node_fini(rcl_node_t * node)
...
+  rmw_ret_t rmw_ret = rmw_destroy_node(node->impl->rmw_node_handle);
+  if (rmw_ret != RMW_RET_OK) {
+    RCL_SET_ERROR_MSG(rmw_get_error_string_safe(), allocator);
+    result = RCL_RET_ERROR;
+  }
+  rcl_ret_t rcl_ret = rcl_guard_condition_fini(node->impl->graph_guard_condition);
+  if (rcl_ret != RCL_RET_OK) {
+    if (result != RCL_RET_OK) {
+      // Here, we are overwriting a previous error; log it
+      fprintf(stderr, "overwriting error in rcl_node_fini: previous error was '%s'\n", rcl_get_error_string_safe());
```
tfoote: Would it be better append this string to the other error message?

dirk-thomas: Isn't the macro already printing a warning about this?

clalancette: Oh, that's a good point; I didn't realize that rcutils_set_error_state() already did that.  OK, I can drop this piece of code, thanks.

rcl/src/rcl/guard_condition.c
```diff
@@ -119,7 +122,7 @@ rcl_guard_condition_fini(rcl_guard_condition_t * guard_condition)
   if (guard_condition->impl) {
     // assuming the allocator is valid because it is checked in rcl_guard_condition_init()
     rcl_allocator_t allocator = guard_condition->impl->options.allocator;
-    if (guard_condition->impl->rmw_handle) {
+    if (guard_condition->impl->rmw_handle && guard_condition->impl->allocated_rmw_guard_condition) {
```
mikaelarguedas: Are the linters passing on this (line looks long)

clalancette: Ah, possibly not.  I forgot to run them locally (I'll do that now).

clalancette: Huh, interesting, linters seem OK with the change.  It is exactly 100 characters long, which I think is the absolute allowed length, so I got lucky :).

clalancette:
Approved, CI looks green.  Merging.

PR #160, opened 2017-08-22
nullify canceled timers

mikaelarguedas:
Opening this for bookkeeping, I'll provide a more detailed description once I have time to investigate.

Right now `rcl_wait` returns a non-null pointer for timers that passed the deadline but are canceled. Forcing functions processing the waitset values to check if they are actually ready (e.g. [here](https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/__init__.py#L94)). The current fix sets the cancelled timers to null before passing them to the rmw waitset

Needs some investigation to find out if this is the right fix and if it's the good thing to do to not update cancel timer and skip them in the waitset. Feedback in that matter is welcome!

### DIFF
--- a/rcl/src/rcl/wait.c
+++ b/rcl/src/rcl/wait.c
@@ -548,6 +548,7 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
         return ret;  // The rcl error state should already be set.
       }
       if (is_canceled) {
+        wait_set->timers[i] = NULL;
         continue;  // Skip canceled timers.
       }
 


wjwwood:
I guess this is fine. I suppose it might prevent the user of the waitset to check each timer again. Does this address a particular bug?

mikaelarguedas:
This address the bug of `rcl_wait` returning when nothing is ready execpt a canceled timer. For example the following code should send one message and then hang forever:
```python
import rclpy

from std_msgs.msg import String


def main(args=None):
    rclpy.init(args=args)

    node = rclpy.create_node('minimal_publisher')
    publisher = node.create_publisher(String, 'topic')

    msg = String()
    i = 0

    def timer_callback():
        nonlocal i
        nonlocal timer
        msg.data = 'Hello World: %d' % i
        i += 1
        print('Publishing: "%s"' % msg.data)
        publisher.publish(msg)
        timer.cancel()

    timer_period = 0.5  # seconds
    timer = node.create_timer(timer_period, timer_callback)

    loop_count = 0
    while rclpy.ok() and loop_count < 10:
        print('spin_once %d' % loop_count)
        # this is supposed to wait forever or wakeup if something needs to be processed
        rclpy.spin_once(node)
        loop_count += 1

    # Destroy the timer attached to the node explicitly
    # (optional - otherwise it will be done automatically
    # when the garbage collector destroys the node object)
    node.destroy_timer(timer)
    node.destroy_node()
    rclpy.shutdown()


if __name__ == '__main__':
    main()
```

Running it with the current master gives the following output:
```
before spin_once 0
Publishing: "Hello World: 0"
before spin_once 1
before spin_once 2
before spin_once 3
before spin_once 4
before spin_once 5
before spin_once 6
before spin_once 7
before spin_once 8
before spin_once 9

real	0m1.484s
user	0m0.868s
sys	0m0.112s
```

mikaelarguedas:
This is more a ticket to get feedback on "how do we expect rcl_wait to process timers" than something I'm expecting to merge as is.
Should a timer that exceeded its deadline but has never been called be waking up the waitset even if it has been canceled?

I know that's not very specific, I will spend more time looking at different use cases and report here with an updated description. Once I know where to go I'll update the PR and add regression tests accordingly

wjwwood:
But isn't that a problem that the canceled timer keeps getting put into the wait set? Otherwise it should only be an issue one time (or never?), right?

wjwwood:
(sorry didn't see your new post when I answered before)

> Should a timer that exceeded its deadline but has never been called be waking up the waitset even if it has been canceled?

I would say that canceled timers should not be in the wait set.

mikaelarguedas:
Just to be sure I got your point:
"Canceled timers should not be in the waitset. If a canceled timer happens to be in the waitset, it should not be 'incremented' and should not make rcl_wait exit". Is that correct?

wjwwood:
What do you mean by incremented? I think it is ok to wake up wait, but it should not be placed into the waitset after being canceled, I think.

mikaelarguedas:
> What do you mean by incremented? 

Sorry I meant processed (taken into account to compute the timeout passed to the `rmw_wait()`)

After looking into this a little bit more, I think the issue is in the way we process timeouts for waitsets.
Timers are handled by rcl and not passed to rmw. So what we currently do is that we [compute](https://github.com/ros2/rcl/blob/3a026e45f4ce64edb239cfb7c76e9fb5685be4cd/rcl/src/rcl/wait.c#L536-L571) the timeout of `rmw_wait()` based on the minimum time between all timer expiration dates and the timeout passed by the user to rcl.

Then we call `rmw_wait()` with that timeout and if the timeout is reached we return `RCL_RET_TIMEOUT`. There are 2 confusting things about this:
- the user receives a timeout error code while the timeout he specified has never been reached (if I ask to wait forever I dont expect to ever receive a timeout return code)
- when rmw_wait returns a timeout we null every waitset entity but none of the timers so all the timers are "not null".

See #163 for a potential fix

wjwwood:
Closing in favor of https://github.com/ros2/rcl/pull/163

PR #163, opened 2017-08-31
Timer tests

mikaelarguedas:
wip: opening this for visibility, replaces #160. Still need to write some tests for this and make sure it doesnt break anything

Note: following recommendation in https://github.com/ros2/rcl/pull/160#issuecomment-324172203 wait will wake up if a timer reached it's deadline even if it has been canceled. It's up to the client library to make sure to not add canceled timers to the waitset


rcl/test/rcl/test_timer.cpp
```diff
@@ -0,0 +1,110 @@
+// Copyright 2015 Open Source Robotics Foundation, Inc.
```
wjwwood: Is this a new file and not a moved file? If so should this be 2017?

mikaelarguedas: it's a new file and indeed should be updated :+1:

wjwwood:
Offline @mikaelarguedas and I looked at this and came up with 370b7fe57873cc2b66e3f159a0ee823514db34cb.

This should fix the issues in `rclpy` as well as in `rclcpp` (https://github.com/ros2/rclcpp/issues/319).

It should also replace https://github.com/ros2/rcl/pull/160 and makes https://github.com/ros2/rclpy/pull/108 not necessary anymore.

mikaelarguedas:
Added a few more tests, let me know if there is any use case that you think should be covered and are not currently covered by these few tests

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3144/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/492/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2501/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3197/)

mikaelarguedas:
hmmm looks like we add more test exhibiting the timer inaccuracy on windows. I'll wait a bit to merge this to see if we can improve the windows situation

dirk-thomas:
> I'll wait a bit to merge this to see if we can improve the windows situation

:+1: to only merge this once it doesn't introduce new failing / flaky tests.

mikaelarguedas:
> :+1: to only merge this once it doesn't introduce new failing / flaky tests.

Not sure if I'll have time to investigate the windows timing issue that has been biting us for some time. The new tests are testing features that were not tested before. We can remove them from this PR and add them later when we figure out the windows timing issue. I do think that the code change should get in before beta3 though because it fixes existing bugs and provide a better behavior overall.

dirk-thomas:
> Not sure if I'll have time to investigate the windows timing issue that has been biting us for some time. The new tests are testing features that were not tested before. We can remove them from this PR and add them later when we figure out the windows timing issue. I do think that the code change should get in before beta3 though because it fixes existing bugs and provide a better behavior overall.

I would suggest to split the PR into a part which can be merged now (with the same or less failing tests) and a second part which needs further work in order to not introduce new failing / flaky tests.

mikaelarguedas:
new (flaky) tests have been removed in https://github.com/ros2/rcl/pull/163/commits/30a53cbedd0b525404345f0e0d1d8cd7c7b7feaa and pushed to the [timer_tests2](https://github.com/ros2/rcl/tree/timer_tests2) branch

rcl/src/rcl/wait.c
```diff
@@ -528,12 +528,34 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
...
+        return ret;  // The rcl error state should already be set.
+      }
+      if (is_canceled) {
+        number_of_valid_timers--;
+        wait_set->timers[i] = NULL;
+      }
+    }
+  }
+
+  bool using_provided_timeout_not_timer_timeout = true;
```
dirk-thomas: With an inverted logic this variable could be named `is_timer_timeout` which sounds a bit more readable / intuitive.

rcl/src/rcl/wait.c
```diff
@@ -591,25 +623,13 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
     assert(rcl_ret == RCL_RET_OK);  // Defensive, shouldn't fail with valid wait_set.
     rcl_ret = rcl_wait_set_clear_clients(wait_set);
     assert(rcl_ret == RCL_RET_OK);  // Defensive, shouldn't fail with valid wait_set.
-    return RCL_RET_TIMEOUT;
-  }
-  // Check for error.
-  if (ret != RMW_RET_OK) {
+    if (using_provided_timeout_not_timer_timeout) {
+      return RCL_RET_TIMEOUT;
```
dirk-thomas: I am not sure I fully understand the current semantic correctly. In which cases does `rmw_wait` return `RMW_RET_TIMEOUT`? Only when the passed timeout has happened? Does it matter if any of the conditions has been triggered?

With this change the semantic of `RMW_RET_TIMEOUT` and `RCL_RET_TIMEOUT` seem to be slightly different which seems confusing to me. Could instead the behavior of the rmw interface be changed (if necessary)?

mikaelarguedas: > I am not sure I fully understand the current semantic correctly. In which cases does rmw_wait return RMW_RET_TIMEOUT? Only when the passed timeout has happened?

correct, if a condition has been triggered `rmw_wait` would return `RMW_RET_OK` and the condition being non-null.

> With this change the semantic of `RMW_RET_TIMEOUT` and `RCL_RET_TIMEOUT` seem to be slightly different which seems confusing to me.

Before this PR:
`rmw_wait` returns `RMW_RET_TIMEOUT` only when the provided timeout has been reached without any condition being triggered.
`rcl_wait` returns `RCL_RET_TIMEOUT` when the provided timeout has been reached **or any timer (even canceled ones)** has reached its deadline.

After this PR:
`rmw_wait` returns `RMW_RET_TIMEOUT` only when the provided timeout has been reached without any condition being triggered.
`rcl_wait` returns `RCL_RET_TIMEOUT` only when the provided timeout has been reached. If any condition has been triggered including timers it will return `RCL_RET_OK` with the corresponding timer non-nullified

The new new implementation matches my understanding of how wait should behave. Could you give more details (or an example) to support the inconsistency of the new behavior, and what type of change in the rmw interface would make more sense here ?

rcl/src/rcl/wait.c
mikaelarguedas: yeah we could come up with a good name at the time of the writing. I'll update this to be clearer, thanks for the variable name suggestion

rcl/src/rcl/wait.c
dirk-thomas: I guess we are clear that both functions should:

* (1) return `RET_TIMEOUT` when the timeout expires and no other condition being triggered and
* (2) return `RET_OK` when at least one condition has been triggered and the function returned before the given timeout.

The question is what should be returned when (3) the function only returns after a given timeout has expired **and** at the same time at least one condition has been triggered.

Without this patch there was a clear correlation: when `rmw` returned `RMW_RET_TIMEOUT` than `rcl` returned `RCL_RET_TIMEOUT` (https://github.com/ros2/rcl/pull/163/files#diff-8fda46206fc38bfdbcac0e5213bffddaL594).

With this patch the `rmw` API seems to return `RMW_RET_TIMEOUT` in the third case but the` rcl` API will return `RCL_RET_OK`. Am I ready that correctly? If yes, the question is if that different semantic is intentional.

mikaelarguedas: I'm not sure I have the same reading here.
(3) the function only returns after a given timeout has expired and at the same time at least one condition has been triggered.

In this case:
the "function" is `rcl_wait`
the "conditions" are "all rmw_waitset entities" + "rcl timers".
"a given timeout" is the user specified timeout

>  the function only returns after a given timeout has expired

so we reach the user specified timeout and not a timer, correct? In that case we do return `RCL_RET_TIMEOUT` regardless of if a condition has been met or not (same behavior as before)

So I don't think we changed anything in the existing semantic here.

Then if we use a timer timeout and not the one specified by the user, that means that the function will return before the user specified timeout, thus should return RCL_RET_OK regardless of the return code of rmw_wait.
Is that correct ?

mikaelarguedas: @dirk-thomas @wjwwood can you confirm if my reading is correct here or if I need to change something in this PR to keep the correct semantic?

wjwwood: > Without this patch there was a clear correlation: when rmw returned RMW_RET_TIMEOUT than rcl returned RCL_RET_TIMEOUT (https://github.com/ros2/rcl/pull/163/files#diff-8fda46206fc38bfdbcac0e5213bffddaL594).

I think that was what was inconsistent with it. A condition may have been met (i.e. a timer was ready), but timeout was returned by rcl too despite the fact rcl had more conditions to consider. To me this makes rcl and rmw more consistent with each other, i.e. ret timeout only returned when the given timeout was exceeded and not when a condition was met instead.

@mikaelarguedas I think it looks right too.

dirk-thomas: After some offline discussion we were able to resolve the misunderstanding. `rmw` returns `TIMEOUT` when it wakes up by the timeout specified rather than any other condition. `rcl` might either pass the user provided timeout to `rmw` or the shortest timer - whichever is lower.

So before `rcl` was returning a `TIMEOUT` even though the timeout signaled by `rmw` was due to a timer and not due to the user provided timeout. With this patch `rcl` will not return `TIMEOUT` anymore when the timeout coming from `rmw` was due to a timer.

With this being said the change in the `test_wait.cpp` to expect `OK` rather than `TIMEOUT` in this case makes a lot more sense :wink:

:+1: :heavy_plus_sign: :ship:

Issue #150, opened 2017-07-18
add tests for rcl_*_is_valid() functions


wjwwood:
Follow up to https://github.com/ros2/rcl/pull/103.

mikaelarguedas:
@allenh1 do you think you will work on this in the near future ?

allenh1:
I certainly could, if needed.

mikaelarguedas:
It looks like you committed to it in https://github.com/ros2/rcl/pull/103#issuecomment-310299057 so it would be great if you can follow up on it :+1: . Otherwise it should be assigned to someone else. The overall goal being to decrease the amount of untested code rather than increasing it

allenh1:
Ok, then. I'll go ahead and assign myself to this one.

mikaelarguedas:
@allenh1 friendly ping

allenh1:
@mikaelarguedas ping received

PR #166, opened 2017-09-11
Implementing `rcl_*_is_valid` tests.

allenh1:
Opening for visibility and feedback as this is under way.

I'd like to ask whether or not these tests should be in one, self-contained, `rcl_x_is_valid.c` file, or if they should be in the file to which the `x` is valued (`rcl_publisher_is_valid` is tested in the `test_publisher.cpp` file). @wjwwood thoughts?

connects to ros2/rcl#150

### DIFF
--- a/rcl/test/rcl/test_client.cpp
+++ b/rcl/test/rcl/test_client.cpp
@@ -130,18 +130,18 @@ TEST_F(TestClientFixture, test_client_init_fini) {
   rcl_reset_error();
 
   // Check if null publisher is valid
-  EXPECT_EQ(rcl_client_is_valid(nullptr), false);
+  EXPECT_FALSE(rcl_client_is_valid(nullptr));
   rcl_reset_error();
 
   // Check if zero initialized client is valid
   client = rcl_get_zero_initialized_client();
-  EXPECT_EQ(rcl_client_is_valid(&client), false);
+  EXPECT_FALSE(rcl_client_is_valid(&client));
   rcl_reset_error();
 
-  // Chek that a valid client is valid
+  // Check that a valid client is valid
   client = rcl_get_zero_initialized_client();
   ret = rcl_client_init(&client, this->node_ptr, ts, topic_name, &default_client_options);
-  EXPECT_EQ(rcl_client_is_valid(&client), true);
+  EXPECT_TRUE(rcl_client_is_valid(&client));
   rcl_reset_error();
 
   // Try passing an invalid (uninitialized) node in init.
--- a/rcl/test/rcl/test_client.cpp
+++ b/rcl/test/rcl/test_client.cpp
@@ -141,6 +141,7 @@ TEST_F(TestClientFixture, test_client_init_fini) {
   // Check that a valid client is valid
   client = rcl_get_zero_initialized_client();
   ret = rcl_client_init(&client, this->node_ptr, ts, topic_name, &default_client_options);
+  EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
   EXPECT_TRUE(rcl_client_is_valid(&client));
   rcl_reset_error();
 
--- a/rcl/test/rcl/test_client.cpp
+++ b/rcl/test/rcl/test_client.cpp
@@ -133,16 +133,22 @@ TEST_F(TestClientFixture, test_client_init_fini) {
   EXPECT_EQ(rcl_client_is_valid(nullptr), false);
   rcl_reset_error();
 
+  // Check if zero initialized client is valid
+  client = rcl_get_zero_initialized_client();
+  EXPECT_EQ(rcl_client_is_valid(&client), false);
+  rcl_reset_error();
+
+  // Chek that a valid client is valid
+  client = rcl_get_zero_initialized_client();
+  ret = rcl_client_init(&client, this->node_ptr, ts, topic_name, &default_client_options);
+  EXPECT_EQ(rcl_client_is_valid(&client), true);
+  rcl_reset_error();
+
   // Try passing an invalid (uninitialized) node in init.
   client = rcl_get_zero_initialized_client();
   rcl_node_t invalid_node = rcl_get_zero_initialized_node();
   ret = rcl_client_init(&client, &invalid_node, ts, topic_name, &default_client_options);
   EXPECT_EQ(RCL_RET_NODE_INVALID, ret) << rcl_get_error_string_safe();
-  rcl_reset_error();
-
-  // Check if zero initialized client is valid
-  rcl_client_t zero_client = rcl_get_zero_initialized_client();
-  EXPECT_EQ(rcl_client_is_valid(&zero_client), false);
   rcl_reset_error();
 
   // Try passing null for the type support in init.
--- a/rcl/test/rcl/test_client.cpp
+++ b/rcl/test/rcl/test_client.cpp
@@ -129,11 +129,20 @@ TEST_F(TestClientFixture, test_client_init_fini) {
   EXPECT_EQ(RCL_RET_INVALID_ARGUMENT, ret) << rcl_get_error_string_safe();
   rcl_reset_error();
 
+  // Check if null publisher is valid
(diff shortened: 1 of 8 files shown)


wjwwood:
I would do the latter, test each in the corresponding file, but either is fine.

allenh1:
Sorry everyone, this one kind of slipped my mind. I've rebased and added some preliminary tests for client, subscription, publisher, and service (really, really trivial ones).

@ros2/team does anyone have any specific things I should test for?

rcl/test/rcl/test_publisher.cpp
```diff
@@ -155,23 +155,25 @@ TEST_F(CLASSNAME(TestPublisherFixture, RMW_IMPLEMENTATION), test_publisher_init_
...
-  // Try passing null for the type support in init.
+  // Check if zero initialized node is valid
   publisher = rcl_get_zero_initialized_publisher();
-  ret = rcl_publisher_init(
-    &publisher, this->node_ptr, nullptr, topic_name, &default_publisher_options);
-  EXPECT_EQ(RCL_RET_INVALID_ARGUMENT, ret) << rcl_get_error_string_safe();
+  EXPECT_EQ(rcl_publisher_is_valid(&publisher), false);
   rcl_reset_error();
 
-  // Try passing null for the topic name in init.
```
allenh1: Not sure why this got removed -- putting this back now...

allenh1:
CI:
------
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3313/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/578/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2643/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3352/)

mikaelarguedas:
I don't know where the right place to add this is but it would be great to also have tests to confirm it does return `true` when the entity is valid

allenh1:
@mikaelarguedas that's a good point... Will do.

allenh1:
Alright, I feel much better about these tests now. Ready for round two of reviews (I'm also assuming something's not quite right with the build farm atm?).

rcl/test/rcl/test_client.cpp
```diff
@@ -129,6 +129,21 @@ TEST_F(TestClientFixture, test_client_init_fini) {
...
+  // Check if null publisher is valid
+  EXPECT_EQ(rcl_client_is_valid(nullptr), false);
+  rcl_reset_error();
+
+  // Check if zero initialized client is valid
+  client = rcl_get_zero_initialized_client();
+  EXPECT_EQ(rcl_client_is_valid(&client), false);
+  rcl_reset_error();
+
+  // Chek that a valid client is valid
```
mikaelarguedas: typo: Chek -> Check (same below)

rcl/test/rcl/test_client.cpp
```diff
@@ -129,6 +129,21 @@ TEST_F(TestClientFixture, test_client_init_fini) {
...
+  client = rcl_get_zero_initialized_client();
+  ret = rcl_client_init(&client, this->node_ptr, ts, topic_name, &default_client_options);
```
mikaelarguedas: Would it make sense to check the value of ret before testing if it's valid? same below

rcl/test/rcl/test_client.cpp
```diff
@@ -129,6 +129,21 @@ TEST_F(TestClientFixture, test_client_init_fini) {
   EXPECT_EQ(RCL_RET_INVALID_ARGUMENT, ret) << rcl_get_error_string_safe();
   rcl_reset_error();
 
+  // Check if null publisher is valid
+  EXPECT_EQ(rcl_client_is_valid(nullptr), false);
```
mikaelarguedas: Tip: you can use `EXPECT_FALSE` / `EXPECT_TRUE`. If you use `EXPECT_EQ` please place the expected value first and the value to check second (same everywhere in this PR)

mikaelarguedas:
> I'm also assuming something's not quite right with the build farm atm?

Yes the master shut down earlier resulting in several builds being cancelled or to fail. It seems to be back online for now

rcl/test/rcl/test_client.cpp
allenh1: I was wondering about this myself, actually. I didn't see it as completely necessary, but I don't mind adding an `EXPECT_EQ(RCL_RET_OK, ret)` if requested. I'll defer to you.

mikaelarguedas: Given that every other test does it I think it makes sense to do it in the newly added tests as well.

Please also print the error in case it's not returning the right error code:
`EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();`

allenh1:
Ok, spinning up the ci:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3324/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/586/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2652/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3364/)

mikaelarguedas:
lgtm once the linking issue on windows is resolved

mikaelarguedas:
@allenh1 Any progress on this ?

allenh1:
> @allenh1 Any progress on this ?

@mikaelarguedas Please excuse the delay (I had to patch the virtualbox driver to build with my kernel). I got the virtualbox up and, consistent with the CI, the build is not working. I've been really frustrated with this one, and have no idea how to approach this kind of linker problem, as I can't see how the `rcl_*_is_valid` functions are any different from the other functions (and, moreover, why the previously existing `rcl_node_is_valid` function works just fine).

Do you have any idea what might be going on? Do I need to do something in the CMake file? Also, I should point out that `rcl` builds just fine, and tests for functions calling `rcl_*_is_valid` pass.

rcl/test/rcl/test_client.cpp
```diff
@@ -129,6 +129,22 @@ TEST_F(TestClientFixture, test_client_init_fini) {
   EXPECT_EQ(RCL_RET_INVALID_ARGUMENT, ret) << rcl_get_error_string_safe();
   rcl_reset_error();
 
+  // Check if null publisher is valid
+  EXPECT_FALSE(rcl_client_is_valid(nullptr));
+  rcl_reset_error();
+
+  // Check if zero initialized client is valid
+  client = rcl_get_zero_initialized_client();
+  EXPECT_FALSE(rcl_client_is_valid(&client));
```
sloretz: Looks like the symbol `rcl_client_is_valid` is not being made visible on the dll on windows.
https://github.com/ros2/rcl/blob/38d3725262e25d5beb4c2b5c1155b74402d371d1/rcl/include/rcl/client.h#L391-L393

On windows symbols that need to be visible must have `__declspec(dllexport)` when the dll is being built, and `__declspec(dllimport)` when the dll is being used.

https://msdn.microsoft.com/en-us/library/a90k134d.aspx

`RCL_PUBLIC` macro takes care of this. Add it before the `rcl_*_is_valid()` function declarations and the linking error you're seen should go away.

allenh1: @sloretz thank you so much!

allenh1:
Ok, that should do it.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3393/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/638/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2722/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3440/)

allenh1:
@mikaelarguedas could I get one last look before I merge this?

allenh1:
> Nitpick: Documentation format for this return value seems inconsistent with the other rcl_x_is_valid functions

@mikaelarguedas I'm failing to see the inconsistency -- can you be a little more specific?

PR #170, opened 2017-10-23
*_raw function

Karsten1987:
Connects to ros2/demos#185

### DIFF
--- a/rcl/src/rcl/publisher.c
+++ b/rcl/src/rcl/publisher.c
@@ -247,7 +247,7 @@ rcl_publish(const rcl_publisher_t * publisher, const void * ros_message)
 rcl_ret_t
 rcl_publish_raw(const rcl_publisher_t * publisher, const rcl_message_raw_t * raw_message)
 {
-  if (!rcl_publisher_is_valid(publisher)) {
+  if (!rcl_publisher_is_valid(publisher, NULL)) {
     return RCL_RET_PUBLISHER_INVALID;
   }
   if (rmw_publish_raw(publisher->impl->rmw_handle, raw_message) != RMW_RET_OK) {
--- a/rcl/src/rcl/publisher.c
+++ b/rcl/src/rcl/publisher.c
@@ -244,6 +244,19 @@ rcl_publish(const rcl_publisher_t * publisher, const void * ros_message)
   return RCL_RET_OK;
 }
 
+rcl_ret_t
+rcl_publish_raw(const rcl_publisher_t * publisher, const rcl_message_raw_t * raw_message)
+{
+  if (!rcl_publisher_is_valid(publisher)) {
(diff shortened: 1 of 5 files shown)


rcl/src/rcl/subscription.c
```diff
@@ -272,6 +272,41 @@ rcl_take(
   return RCL_RET_OK;
 }
 
+rcl_ret_t
+rcl_take_raw(
+  const rcl_subscription_t * subscription,
+  rcl_message_raw_t * raw_message,
+  rmw_message_info_t * message_info)
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(subscription, RCL_RET_INVALID_ARGUMENT, rcl_get_default_allocator());
```
sloretz: It looks like a few of these checks (`subscription`, `subscription->impl`, subscription options, `subscription->impl->rmw_handle`) can be replaced with with a call `rcl_subscription_is_valid()`.

rcl/src/rcl/publisher.c
```diff
@@ -244,6 +244,19 @@ rcl_publish(const rcl_publisher_t * publisher, const void * ros_message)
...
 
+rcl_ret_t
+rcl_publish_raw(const rcl_publisher_t * publisher, const rcl_message_raw_t * raw_message)
+{
+  if (!rcl_publisher_is_valid(publisher, NULL)) {
+    return RCL_RET_PUBLISHER_INVALID;
+  }
+  if (rmw_publish_raw(publisher->impl->rmw_handle, raw_message) != RMW_RET_OK) {
+    RCL_SET_ERROR_MSG(rmw_get_error_string_safe(), rcl_get_default_allocator());
+    return RCL_RET_ERROR;
```
wjwwood: This hides more specific errors from `rmw`, like `RMW_RET_BAD_ALLOC`. I'd advise catch those specific cases, and returning the equivalent `rcl` error type.

rcl/src/rcl/subscription.c
```diff
@@ -272,6 +266,38 @@ rcl_take(
...
+  // If message_info is NULL, use a place holder which can be discarded.
+  rmw_message_info_t dummy_message_info;
+  rmw_message_info_t * message_info_local = message_info ? message_info : &dummy_message_info;
+  // Call rmw_take_with_info.
+  bool taken = false;
+  rmw_ret_t ret =
+    rmw_take_raw_with_info(subscription->impl->rmw_handle, raw_message, &taken, message_info_local);
+  if (ret != RMW_RET_OK) {
+    RCL_SET_ERROR_MSG(rmw_get_error_string_safe(), error_allocator);
+    return RCL_RET_ERROR;
```
wjwwood: Same thing here, loosing error specifics.

PR #172, opened 2017-11-01
Out-of-scope memory use and memory leak issue

gaoethan:
`temp` going out of scope leaks the storage it points to
resource leak from `local_namespace_`

Signed-off-by: Ethan Gao <ethan.gao@linux.intel.com>

### DIFF
--- a/rcl/src/rcl/node.c
+++ b/rcl/src/rcl/node.c
@@ -165,7 +165,9 @@ rcl_node_init(
   ret = rmw_validate_namespace(local_namespace_, &validation_result, NULL);
   if (ret != RMW_RET_OK) {
     RCL_SET_ERROR_MSG(rmw_get_error_string_safe(), *allocator);
-    allocator->deallocate((char *)local_namespace_, allocator->state);
+    if (should_free_local_namespace_) {
+      allocator->deallocate((char *)local_namespace_, allocator->state);
+    }
     return ret;
   }
   if (validation_result != RMW_NAMESPACE_VALID) {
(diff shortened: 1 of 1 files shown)


rcl/src/rcl/node.c
```diff
@@ -163,6 +165,7 @@ rcl_node_init(
   ret = rmw_validate_namespace(local_namespace_, &validation_result, NULL);
   if (ret != RMW_RET_OK) {
     RCL_SET_ERROR_MSG(rmw_get_error_string_safe(), *allocator);
+    allocator->deallocate((char *)local_namespace_, allocator->state);
```
wjwwood: This cannot be done if line 147 is executed (if `local_namespace_` is a literal string). I think it needs to check `should_free_local_namespace_`.

Here and below.

gaoethan: exactly, thanks !

wjwwood:
Thanks @gaoethan!

dirk-thomas:
I don't think the scope of this variable needs to be increased. `temp` is being used to avoid changing `local_namespace_` in case of errors. As long as `local_namespace_` is being freed when `should_free_local_namespace_` is being set (which is part of this PR) this block should not leak memory.

The scope of `temp` doesn't affect when it is being cleaned up (since it is a bare pointer) and the variable is not used anywhere below. Therefore I think the scope of `temp` should be reverted to be local as it was before (keeping the new `deallocate` calls of course).

wjwwood:
You’re right, I’ll make a follow up pr ASAP.

PR #173, opened 2017-11-06
expand rcl_node_is_valid with allocator to standarize the node check …

jwang11:
…across RCL

Current RCL use very different way for node check in different components, which is very random. Actually, node.c already provide a standard way - rcl_node_is_valid. However, it just use
internal allocator, but some context has external allocator. The commit expand
rcl_node_is_valid with one more parameter for allocator. It can be used in all cases.

### DIFF
--- a/rcl/include/rcl/node.h
+++ b/rcl/include/rcl/node.h
@@ -234,11 +234,12 @@ rcl_node_get_default_options(void);
  * <i>[1] if `atomic_is_lock_free()` returns true for `atomic_uint_least64_t`</i>
  *
  * \param[in] node rcl_node_t to be validated
+ * \param[in] allocator external allocator pointer used for allocations.
  * \return `true` if the node is valid, otherwise `false`.
  */
 RCL_PUBLIC
 bool
-rcl_node_is_valid(const rcl_node_t * node);
+rcl_node_is_valid(const rcl_node_t * node, const rcl_allocator_t * allocator);
 
 /// Return the name of the node.
 /**
--- a/rcl/include/rcl/node.h
+++ b/rcl/include/rcl/node.h
@@ -205,6 +205,10 @@ rcl_node_get_default_options(void);
 /**
  * Also return `false` if the node pointer is `NULL` or the allocator is invalid.
  *
+ * The allocator needs to either be a valid allocator or `NULL`, in which case
+ * the default allocator will be used.
+ * The allocator is used when allocation is needed for an error message.
+ *
  * A node is invalid if:
  *   - the implementation is `NULL` (rcl_node_init not called or failed)
  *   - rcl_shutdown has been called since the node has been initialized
@@ -234,8 +238,7 @@ rcl_node_get_default_options(void);
  * <i>[1] if `atomic_is_lock_free()` returns true for `atomic_uint_least64_t`</i>
  *
  * \param[in] node rcl_node_t to be validated
- * \param[in] allocator external allocator pointer used for error messages,
- *            the default allocator will be used if `allocator` is `NULL`
+ * \param[in] allocator a valid allocator or `NULL`
  * \return `true` if the node and allocator are valid, otherwise `false`.
  */
 RCL_PUBLIC
--- a/rcl/include/rcl/node.h
+++ b/rcl/include/rcl/node.h
@@ -206,6 +206,8 @@ rcl_node_get_default_options(void);
  * Also return `false` if the node pointer is `NULL`.
  *
  * A node is invalid if:
+ *   - parameter node is NULL 
+ *   - allocator is invalid
  *   - the implementation is `NULL` (rcl_node_init not called or failed)
  *   - rcl_shutdown has been called since the node has been initialized
  *   - the node has been finalized with rcl_node_fini
@@ -215,7 +217,7 @@ rcl_node_get_default_options(void);
  * Consider:
  *
  * ```c
- * assert(rcl_node_is_valid(node));  // <-- thread 1
+ * assert(rcl_node_is_valid(node, NULL));  // <-- thread 1
  * rcl_shutdown();                   // <-- thread 2
  * // use node as if valid           // <-- thread 1
  * ```
--- a/rcl/include/rcl/node.h
+++ b/rcl/include/rcl/node.h
@@ -234,7 +234,8 @@ rcl_node_get_default_options(void);
  * <i>[1] if `atomic_is_lock_free()` returns true for `atomic_uint_least64_t`</i>
  *
  * \param[in] node rcl_node_t to be validated
- * \param[in] allocator external allocator pointer used for allocations
+ * \param[in] allocator external allocator pointer used for error messages,
+ *            the default allocator will be used if `allocator` is `NULL`
  * \return `true` if the node and allocator are valid, otherwise `false`.
  */
 RCL_PUBLIC
--- a/rcl/include/rcl/node.h
+++ b/rcl/include/rcl/node.h
@@ -203,11 +203,9 @@ rcl_node_get_default_options(void);
 
 /// Return `true` if the node is valid, else `false`.
 /**
- * Also return `false` if the node pointer is `NULL`.
+ * Also return `false` if the node pointer is `NULL` or the allocator is invalid.
  *
  * A node is invalid if:
- *   - parameter node is NULL 
- *   - allocator is invalid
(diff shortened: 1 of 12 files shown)


rcl/src/rcl/client.c
```diff
@@ -66,9 +66,7 @@ rcl_client_init(
     return RCL_RET_INVALID_ARGUMENT, rcl_get_default_allocator());
 
   RCL_CHECK_ARGUMENT_FOR_NULL(client, RCL_RET_INVALID_ARGUMENT, *allocator);
-  RCL_CHECK_ARGUMENT_FOR_NULL(node, RCL_RET_INVALID_ARGUMENT, *allocator);
-  if (!node->impl) {
-    RCL_SET_ERROR_MSG("invalid node", *allocator);
+  if (!rcl_node_is_valid(node, allocator)) {
```
dhood: this file is missing the equivalent change in the fini

jwang11:
Ok, I'll restore RCL_CHECK_ARGUMENT_FOR_NULL for node before rcl_node_is_valid and fix other issues you mentioned.

allenh1:
@ros2/team Should we maybe consider adding this parameter to the other `rcl_*_is_valid` functions?

rcl/test/rcl/test_node.cpp
```diff
@@ -118,16 +118,16 @@ TEST_F(CLASSNAME(TestNodeFixture, RMW_IMPLEMENTATION), test_rcl_node_accessors)
     });
   // Test rcl_node_is_valid().
   bool is_valid;
-  is_valid = rcl_node_is_valid(nullptr);
+  is_valid = rcl_node_is_valid(nullptr, NULL);
```
allenh1: nit: `nullptr`

jwang11: fixed

mikaelarguedas:
> Should we maybe consider adding this parameter to the other rcl_*_is_valid functions?

Yes we should, I wouldn't mind if it's done in a follow-up PR though given that this function is the only one used "widely" throughout the codebase.

---

@jwang11 Nitpick for the future: we (almost)always "squash and merge" our pull requests, please push new commits for your changes instead of force pushing a squashed version, that makes it easier to review the changes and also allows us to fix typos/nitpicks in the code without being overwritten. Thanks!

jwang11:
Ok, got it.

allenh1:
> Yes we should, I wouldn't mind if it's done in a follow-up PR though given that this function is the only one used "widely" throughout the codebase.

I figured. Since I wrote those functions, I think I'll do this when I have a few moments later today.

rcl/src/rcl/node.c
```diff
@@ -367,14 +366,15 @@ rcl_node_fini(rcl_node_t * node)
 }
 
 bool
-rcl_node_is_valid(const rcl_node_t * node)
+rcl_node_is_valid(const rcl_node_t * node, const rcl_allocator_t * allocator)
 {
-  RCL_CHECK_ARGUMENT_FOR_NULL(node, false, rcl_get_default_allocator());
+  rcl_allocator_t alloc = allocator ? *allocator : rcl_get_default_allocator();
+  RCL_CHECK_ARGUMENT_FOR_NULL(node, false, alloc);
```
mikaelarguedas: this should check that the provided allocator is valid before using it, we're in the process of exposing a macro to do so [here](https://github.com/ros2/rcl/pull/176)

jwang11: I think allocator check should be outside this function, that means we should check allocator valid before invoke the function. Otherwise, there maybe quite a bit duplicated check in some cases.

mikaelarguedas: As a user facing function, IMO it's hard to assume that every user will provide a valid argument or write an additional line for checking it passing it to the function. That's one of the reasons we check every argument that users provide to rcl functions

tfoote: We want all arguements to all functions to be checked. Just like we require all return values to be checked. If the checks become a performance problem they can be compiled out in those use cases if we use the macros.

jwang11: Ok, that is fair. Once allocator check PR is merged, I'll update this PR.

mikaelarguedas: @jwang11 FYI https://github.com/ros2/rcl/pull/176 has been merged

jwang11: @mikaelarguedas PR update done

rcl/src/rcl/node.c
```diff
@@ -363,14 +362,16 @@ rcl_node_fini(rcl_node_t * node)
...
+  RCL_CHECK_ALLOCATOR_WITH_MSG(allocator, "invalid allocator", return false);
```
mikaelarguedas: Sorry that was unclear in my previous comment: I meant that we should check that `alloc` is valid _before_ using it in `RCL_CHECK_ARGUMENT_FOR_NULL`:
something like:
```c++
rcl_allocator_t alloc = allocator ? *allocator : rcl_get_default_allocator();
RCL_CHECK_ALLOCATOR_WITH_MSG(alloc, "invalid allocator", return false);
RCL_CHECK_ARGUMENT_FOR_NULL(node, false, alloc);
```

dhood: please move this two lines up so it returns earlier

rcl/src/rcl/node.c
```diff
@@ -363,14 +362,16 @@ rcl_node_fini(rcl_node_t * node)
 }
 
 bool
-rcl_node_is_valid(const rcl_node_t * node)
+rcl_node_is_valid(const rcl_node_t * node, const rcl_allocator_t * allocator)
 {
-  RCL_CHECK_ARGUMENT_FOR_NULL(node, false, rcl_get_default_allocator());
+  rcl_allocator_t alloc = allocator ? *allocator : rcl_get_default_allocator();
+  RCL_CHECK_ALLOCATOR_WITH_MSG(&alloc, "invalid allocator", return false);
```
dhood: could you please update the documentation for this function to mention this new return case

rcl/include/rcl/node.h
```diff
@@ -236,8 +234,8 @@ rcl_node_get_default_options(void);
  * <i>[1] if `atomic_is_lock_free()` returns true for `atomic_uint_least64_t`</i>
  *
  * \param[in] node rcl_node_t to be validated
- * \param[in] allocator external allocator pointer used for allocations.
- * \return `true` if the node is valid, otherwise `false`.
+ * \param[in] allocator external allocator pointer used for allocations
+ * \return `true` if the node and allocator are valid, otherwise `false`.
```
mikaelarguedas: Not sure how to capture that passing a `NULL` allocator is valid too, maybe:
```
\param[in] allocator external allocator pointer used for allocations, must be a valid allocator or NULL
  a NULL allocator will result in using the default allocator
```

dhood: how about ca141f2d01d7741f2afbdfd72a8d406cd40feb32

wjwwood: If I need to put more extensive qualifications on a parameter, I do so in prose in the body of the doc block, and try to keep the parameter description succinct.

I would do something like this:

```
 * The allocator needs to either be a valid allocator or NULL, in which case
 * the default allocator will be used.
 * The allocator is used when allocation is needed (like for an error message)
 * but the node's allocator is invalid or cannot be checked.
...
 * \param[in] allocator a valid allocator or NULL
```

dhood: in cfde9b3a34df97ae2c595cae88717ad160483bab I mostly copied that @wjwwood but FWIU the node's allocator is never used in this function

wjwwood: Oh, well I just assumed it would try to use that if it were valid, but I guess that's the last case checked.

mikaelarguedas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3543/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/750/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2884/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3622/)

mikaelarguedas:
Having a hard time to get a MacOS job through.
I'm going to merge this and will follow up if the MacOS job complains about something

PR #175, opened 2017-11-07
re-reduce scope of temp variable

wjwwood:
follow up to #172


wjwwood:
Linux CI just to check for typos: [Build Status](http://ci.ros2.org/job/ci_linux/3450/)

PR #176, opened 2017-11-09
create a macro RCL_CHECK_ALLOCATOR

tfoote:
to check the required methods are set on the allocator

I found 2 instances of this already partially impelemented and was starting to write several more for the time PRs. 

Linux: [Build Status](http://ci.ros2.org/job/ci_linux/3481/)
Linux arm64: [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/718/)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/2808/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/3526/)

### DIFF
--- a/rcl/src/rcl/client.c
+++ b/rcl/src/rcl/client.c
@@ -58,13 +58,7 @@ rcl_client_init(
   // check the options and allocator first, so the allocator can be passed to errors
   RCL_CHECK_ARGUMENT_FOR_NULL(options, RCL_RET_INVALID_ARGUMENT, rcl_get_default_allocator());
   const rcl_allocator_t * allocator = &options->allocator;
-  RCL_CHECK_FOR_NULL_WITH_MSG(
-    allocator->allocate, "allocate not set",
-    return RCL_RET_INVALID_ARGUMENT, rcl_get_default_allocator());
-  RCL_CHECK_FOR_NULL_WITH_MSG(
-    allocator->deallocate, "deallocate not set",
-    return RCL_RET_INVALID_ARGUMENT, rcl_get_default_allocator());
-
+  RCL_CHECK_ALLOCATOR_WITH_MSG(allocator, "invalid allocator", return RCL_RET_INVALID_ARGUMENT);
   RCL_CHECK_ARGUMENT_FOR_NULL(client, RCL_RET_INVALID_ARGUMENT, *allocator);
   RCL_CHECK_ARGUMENT_FOR_NULL(node, RCL_RET_INVALID_ARGUMENT, *allocator);
   if (!node->impl) {
--- a/rcl/src/rcl/client.c
+++ b/rcl/src/rcl/client.c
@@ -22,6 +22,7 @@ extern "C"
 #include <stdio.h>
 #include <string.h>
 
+#include "rcl/error_handling.h"
 #include "rcl/expand_topic_name.h"
 #include "rcutils/logging_macros.h"
 #include "rmw/error_handling.h"
--- a/rcl/src/rcl/common.h
+++ b/rcl/src/rcl/common.h
@@ -23,16 +23,6 @@ extern "C"
 #include "rcl/error_handling.h"
 #include "rcl/types.h"
 
-#define RCL_CHECK_ARGUMENT_FOR_NULL(argument, error_return_type, allocator) \
-  RCL_CHECK_FOR_NULL_WITH_MSG(argument, #argument " argument is null", \
-    return error_return_type, allocator)
-
-#define RCL_CHECK_FOR_NULL_WITH_MSG(value, msg, error_statement, allocator) \
-  if (!(value)) { \
-    RCL_SET_ERROR_MSG(msg, allocator); \
-    error_statement; \
-  }
-
 /// Retrieve the value of the given environment variable if it exists, or "".
 /* The returned cstring is only valid until the next time this function is
  * called, because it is a direct pointer to the static storage.
--- a/rcl/src/rcl/common.h
+++ b/rcl/src/rcl/common.h
@@ -20,7 +20,6 @@ extern "C"
 {
 #endif
 
-#include "rcl/error_handling.h"
 #include "rcl/types.h"
 
 /// Retrieve the value of the given environment variable if it exists, or "".
--- a/rcl/src/rcl/publisher.c
+++ b/rcl/src/rcl/publisher.c
@@ -23,6 +23,7 @@ extern "C"
 #include <string.h>
 
 #include "./common.h"
+#include "rcl/allocator.h"
 #include "rcl/expand_topic_name.h"
 #include "rcutils/logging_macros.h"
 #include "rmw/error_handling.h"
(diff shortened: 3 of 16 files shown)


dhood:
rcutils has a macro for this already that we could re-use: https://github.com/ros2/rcutils/blob/00e2b9db9f4b71374f1388570be2de3edabd2e37/include/rcutils/allocator.h#L111 (it looks stricter though)

mikaelarguedas:
there is an equivalent one in rcutils that we could reuse https://github.com/ros2/rcutils/blob/00e2b9db9f4b71374f1388570be2de3edabd2e37/include/rcutils/allocator.h#L111
This basically checks that all fields are checked https://github.com/ros2/rcutils/blob/b9819514f058e32564ae6ece2da28ed64f71546d/src/allocator.c#L76-L88

tfoote:
That's useful, shall I #define an RCL alias and use that or just call the RCUTILS one? And should we push the null argument implementation up too?

dhood:
I'd say `#define` it rather than use the rcutils one directly, because at some point the `rcutils_allocator_t` and `rcl_allocator_t` may diverge.

Is your second question if we should de-duplicate the implementation of `RCUTILS_CHECK_ARGUMENT_FOR_NULL` and `RCL_CHECK_ARGUMENT_FOR_NULL`? I don't see a reason not to, I think the reason is just that the work pushing things to rcutils is still a bit WIP

tfoote:
Made public and deduplicated:

Linux: [Build Status](http://ci.ros2.org/job/ci_linux/3484/)
Linux arm64: [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/719/)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/2813/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/3527/)

tfoote:
I reverted the merge from @mikaelarguedas as it caused instability even with the compile fix. http://ci.ros2.org/job/ci_linux/3498/ I'd suggest we merge this and then we can address the bigger question of cleaning up all usages and homogenizing all the storage of the allocator_t and using the macro everywhere in a follow up. THis will also unblock #173 to use the macros.

With the roll back it's back to the state of the last CI above.

mikaelarguedas:
Sounds good to me, I'll work on another branch to get this consistently used by the various parts of rcl :+1:

rcl/src/rcl/client.c
```diff
@@ -58,13 +58,7 @@ rcl_client_init(
...
   RCL_CHECK_ARGUMENT_FOR_NULL(options, RCL_RET_INVALID_ARGUMENT, rcl_get_default_allocator());
   const rcl_allocator_t * allocator = &options->allocator;
-  RCL_CHECK_FOR_NULL_WITH_MSG(
-    allocator->allocate, "allocate not set",
-    return RCL_RET_INVALID_ARGUMENT, rcl_get_default_allocator());
-  RCL_CHECK_FOR_NULL_WITH_MSG(
-    allocator->deallocate, "deallocate not set",
-    return RCL_RET_INVALID_ARGUMENT, rcl_get_default_allocator());
-
+  RCL_CHECK_ALLOCATOR(allocator, return RCL_RET_INVALID_ARGUMENT);
```
dhood: This won't set the error message if the allocator is invalid. That's why the changes in https://github.com/ros2/rcl/pull/178
 caused the tests to fail, because they're [expecting the error message to be set](https://github.com/ros2/rcl/blob/72bcb6b5b648b75f0c2096607c9e2fe0266352c7/rcl/test/rcl/test_node.cpp#L346). Using `RCL_CHECK_ALLOCATOR_WITH_MSG` is closer to the existing behaviour.

tfoote:
Added msg as pointed out by @dhood and readded @mikaelarguedas 's extension.

Linux: [Build Status](http://ci.ros2.org/job/ci_linux/3520/)
Linux arm64: [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/740/) (rclpy callback count test failure believed unreleated)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/2854/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/3577/)

rcl/src/rcl/publisher.c
```diff
@@ -17,13 +17,13 @@ extern "C"
 {
 #endif
 
-#include "rcl/publisher.h"
```
dhood: the style of this repo has the "matching" include for the file at the top so it seems this change should be reverted

tfoote: Moved back up.

rcl/src/rcl/common.h
```diff
@@ -23,16 +23,6 @@ extern "C"
 #include "rcl/error_handling.h"
 #include "rcl/types.h"
 
-#define RCL_CHECK_ARGUMENT_FOR_NULL(argument, error_return_type, allocator) \
```
dhood: For my understanding, is it intentional that the files that use `RCL_CHECK_ARGUMENT_FOR_NULL` can only do so because it's implicitly included in `common.h`? FWIU we have an "include what you use" policy that might suggest those files (e.g. `client.c`) should now include `rcl/error_handling.h` explicitly, but the purpose of `common.h` might be to get around that rule in some circumstances.

tfoote: I don't think that was the point of common.h I removed the include from common.h and filled in every source file that had an issue, then added it back so common.c could compile and link correctly.

dhood: it should be fine to remove it from common.h now right? or could you give more info on why it's still needed?

mikaelarguedas: I believe it needs the types as it returns some `rcl_ret_t`

dhood: they should come from rcl/types.h no?

mikaelarguedas: yes exactly, all I meant is currently common.c doesnt include types.h

mikaelarguedas: Looks like I misunderstood what we're talking about here. If now all files include `rcl/error_handling.h` (including `common.c`) there should indeed be no need to include it in `common.h`. Not sure what the linking issue you faced was from

tfoote: The linking issue was due to one more missing include. I had to trace back to an `implicit declaration of function` that was caused by the missing macro. I cleared the include from `common.h` in 
a5e5825

tfoote:
Thanks for the reviews. Here's a sanity check that the header cleanup was fine. 

Linux: [Build Status](http://ci.ros2.org/job/ci_linux/3524/)

dhood:
Looks like there are some files (not all) that don't need to include common.h anymore because they only needed the argument checking macros: I added https://github.com/ros2/rcl/pull/176/commits/3fef7c147bcfb1ccd9f33d050549ddbc349ebf5b to this PR

tfoote:
Thanks one more CI on top of that: 

Linux: [Build Status](http://ci.ros2.org/job/ci_linux/3528/)

PR #180, opened 2017-11-11
Add allocator parameter to rcl_X_is_valid functions

allenh1:
This adds allocators to the `rcl_*_is_valid` functions, as discussed in #173.

This PR should be pending until #173 and #176 are merged.

### DIFF
--- a/rcl/include/rcl/client.h
+++ b/rcl/include/rcl/client.h
@@ -394,7 +394,7 @@ rcl_client_get_rmw_handle(const rcl_client_t * client);
  */
 RCL_PUBLIC
 bool
-rcl_client_is_valid(const rcl_client_t * client, const rcl_allocator_t * allocator);
+rcl_client_is_valid(const rcl_client_t * client, const rcl_allocator_t * error_msg_allocator);
 
 #if __cplusplus
 }
--- a/rcl/include/rcl/client.h
+++ b/rcl/include/rcl/client.h
@@ -394,7 +394,7 @@ rcl_client_get_rmw_handle(const rcl_client_t * client);
  */
 RCL_PUBLIC
 bool
-rcl_client_is_valid(const rcl_client_t * client);
+rcl_client_is_valid(const rcl_client_t * client, const rcl_allocator_t * allocator);
 
 #if __cplusplus
 }
--- a/rcl/include/rcl/client.h
+++ b/rcl/include/rcl/client.h
@@ -394,7 +394,7 @@ rcl_client_get_rmw_handle(const rcl_client_t * client);
  */
 RCL_PUBLIC
 bool
-rcl_client_is_valid(const rcl_client_t * client, const rcl_allocator_t * error_msg_allocator);
+rcl_client_is_valid(const rcl_client_t * client, rcl_allocator_t * error_msg_allocator);
 
 #if __cplusplus
 }
--- a/rcl/src/rcl/client.c
+++ b/rcl/src/rcl/client.c
@@ -292,11 +292,11 @@ rcl_take_response(
   return RCL_RET_OK;
 }
 
-bool rcl_client_is_valid(const rcl_client_t * client, const rcl_allocator_t * allocator)
+bool rcl_client_is_valid(const rcl_client_t * client, const rcl_allocator_t * error_msg_allocator)
 {
   const rcl_client_options_t * options;
-  const rcl_allocator_t alloc = allocator ? *allocator : rcl_get_default_allocator();
-  RCL_CHECK_ALLOCATOR_WITH_MSG(&alloc, "client's allocator is invalid", return false);
+  const rcl_allocator_t alloc = error_msg_allocator ? *error_msg_allocator : rcl_get_default_allocator();
+  RCL_CHECK_ALLOCATOR_WITH_MSG(&alloc, "error msg allocator is invalid", return false);
   RCL_CHECK_ARGUMENT_FOR_NULL(client, false, alloc);
   options = _client_get_options(client);
   RCL_CHECK_FOR_NULL_WITH_MSG(
--- a/rcl/src/rcl/client.c
+++ b/rcl/src/rcl/client.c
@@ -291,7 +291,7 @@ rcl_take_response(
 bool rcl_client_is_valid(const rcl_client_t * client, rcl_allocator_t * error_msg_allocator)
 {
   const rcl_client_options_t * options;
-  const rcl_allocator_t alloc =
+  rcl_allocator_t alloc =
     error_msg_allocator ? *error_msg_allocator : rcl_get_default_allocator();
   RCL_CHECK_ALLOCATOR_WITH_MSG(&alloc, "allocator is invalid", return false);
   RCL_CHECK_ARGUMENT_FOR_NULL(client, false, alloc);
--- a/rcl/src/rcl/client.c
+++ b/rcl/src/rcl/client.c
@@ -293,7 +293,7 @@ bool rcl_client_is_valid(const rcl_client_t * client, rcl_allocator_t * error_ms
   const rcl_client_options_t * options;
   const rcl_allocator_t alloc =
     error_msg_allocator ? *error_msg_allocator : rcl_get_default_allocator();
-  RCL_CHECK_ALLOCATOR_WITH_MSG(&alloc, "error msg allocator is invalid", return false);
+  RCL_CHECK_ALLOCATOR_WITH_MSG(&alloc, "allocator is invalid", return false);
   RCL_CHECK_ARGUMENT_FOR_NULL(client, false, alloc);
   RCL_CHECK_FOR_NULL_WITH_MSG(
     client->impl, "client's rmw implementation is invalid", return false, alloc);
--- a/rcl/src/rcl/client.c
+++ b/rcl/src/rcl/client.c
@@ -300,6 +300,8 @@ bool rcl_client_is_valid(const rcl_client_t * client, const rcl_allocator_t * er
   options = _client_get_options(client);
   RCL_CHECK_FOR_NULL_WITH_MSG(
     options, "client's options pointer is invalid", return false, alloc);
+  RCL_CHECK_FOR_NULL_WITH_MSG(
+    client->impl->rmw_handle, "client's rmw handle is invalid", return false, alloc);
   return true;
 }
 #if __cplusplus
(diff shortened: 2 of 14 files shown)


mikaelarguedas:
Labeling this PR as `in progress` as it will be updated once required PRs are merged

mikaelarguedas:
@allenh1 FYI: Both required PRs have been merged

rcl/src/rcl/client.c
```diff
@@ -278,14 +278,15 @@ rcl_take_response(
...
 }
 
-bool rcl_client_is_valid(const rcl_client_t * client)
+bool rcl_client_is_valid(const rcl_client_t * client, const rcl_allocator_t * allocator)
 {
   const rcl_client_options_t * options;
-  RCL_CHECK_ARGUMENT_FOR_NULL(
-    client, false, rcl_get_default_allocator());
+  const rcl_allocator_t alloc = allocator ? *allocator : rcl_get_default_allocator();
+  RCL_CHECK_ALLOCATOR_WITH_MSG(&alloc, "client's allocator is invalid", return false);
```
mikaelarguedas: Not sure how we should phrase it but when reading `client's allocator is invalid` in my console, I understand that the allocator tied to my client is invalid (`options->allocator`) and not the one passed to the validation function.
We could either call the macro without an error message or clarify the message to make it obvious which allocator is invalid

allenh1: @mikaelarguedas that's a good point. Should I just mimic the message from #173?

mikaelarguedas: Or something like "argument alloc is invalid" or "passed allocator is invalid", as "invalid allocator" doesn't really disambiguate IMO.

Karsten1987: But so why aren't we using the client's allocator?

mikaelarguedas: Yeah that's a bit hairy, here we need to confirm that the client and client->options structure are valid before using the allocator stored in it. The allocator passed to this function is only used to allocate error messages if something goes wrong when checking the validity of the fields of the client.
Maybe a less confusing way would be to rename the parameter to be "error_message_allocator" this way we know it's not the client's allocator and it won't be used to allocate memory within client's normal operation mode.

allenh1:
@mikaelarguedas Ok, rebased on top of master and ready for another look. I changed the parameter name to be as you suggested, and replicated the change for the parameter in `node.h`.

rcl/src/rcl/client.c
```diff
@@ -288,14 +288,15 @@ rcl_take_response(
...
-bool rcl_client_is_valid(const rcl_client_t * client)
+bool rcl_client_is_valid(const rcl_client_t * client, const rcl_allocator_t * error_msg_allocator)
 {
   const rcl_client_options_t * options;
-  RCL_CHECK_ARGUMENT_FOR_NULL(
-    client, false, rcl_get_default_allocator());
+  const rcl_allocator_t alloc = error_msg_allocator ? *error_msg_allocator : rcl_get_default_allocator();
+  RCL_CHECK_ALLOCATOR_WITH_MSG(&alloc, "error msg allocator is invalid", return false);
+  RCL_CHECK_ARGUMENT_FOR_NULL(client, false, alloc);
   options = _client_get_options(client);
```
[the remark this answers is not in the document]
allenh1: I'm a little unsure as to what you're saying here -- are you saying check `&client->impl` is not `NULL` before this line? If so, that certainly seems reasonable.

rcl/src/rcl/client.c
```diff
@@ -288,14 +288,15 @@ rcl_take_response(
...
   RCL_CHECK_FOR_NULL_WITH_MSG(
-    options, "client's options pointer is invalid", return false, rcl_get_default_allocator());
+    options, "client's options pointer is invalid", return false, alloc);
   RCL_CHECK_FOR_NULL_WITH_MSG(
     client->impl, "client's rmw implementation is invalid", return false, options->allocator);
```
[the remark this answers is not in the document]
allenh1: That certainly makes a lot of sense to me.

rcl/src/rcl/client.c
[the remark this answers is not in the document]
mikaelarguedas: > are you saying check &client->impl is not NULL before this line?

Correct, L300-301 should be moved before this line

allenh1:
How do you feel about checking the options pointer's allocator here as well?

mikaelarguedas:
> How do you feel about checking the options pointer's allocator here as well?

I don't feel very strongly one way or another. If we check that member, maybe we should check the other members as well. Though hat may be a lot of overhead in the end

allenh1:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3644/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/824/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2978/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3745/)

rcl/src/rcl/client.c
```diff
@@ -288,16 +288,18 @@ rcl_take_response(
...
+    error_msg_allocator ? *error_msg_allocator : rcl_get_default_allocator();
+  RCL_CHECK_ALLOCATOR_WITH_MSG(&alloc, "error msg allocator is invalid", return false);
+  RCL_CHECK_ARGUMENT_FOR_NULL(client, false, alloc);
   RCL_CHECK_FOR_NULL_WITH_MSG(
-    options, "client's options pointer is invalid", return false, rcl_get_default_allocator());
+    client->impl, "client's rmw implementation is invalid", return false, alloc);
+  options = _client_get_options(client);
   RCL_CHECK_FOR_NULL_WITH_MSG(
-    client->impl, "client's rmw implementation is invalid", return false, options->allocator);
+    options, "client's options pointer is invalid", return false, alloc);
```
mikaelarguedas: All the other functions check the x->impl->rmw_handle, I think the client should do the same

mikaelarguedas:
I rebased the branch and addressed the pending comments so that this API change can get in.

Can someone from @ros2/team give another :eyes: to this change please?

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3694/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/861/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3029/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3791/)

rcl/include/rcl/client.h
```diff
@@ -394,7 +394,7 @@ rcl_client_get_rmw_handle(const rcl_client_t * client);
  */
 RCL_PUBLIC
 bool
-rcl_client_is_valid(const rcl_client_t * client);
+rcl_client_is_valid(const rcl_client_t * client, const rcl_allocator_t * error_msg_allocator);
```
dirk-thomas: Why should the passed allocators be `const`?

Same for all signatures.

mikaelarguedas: My guess is that it's because that's what is used for the `node_is_valid` and that's just been propagated to those. I don't feel strongly about it though

dirk-thomas: Since the [internally called function](https://github.com/ros2/rcutils/blob/9fc616d389372d3bec080e8965e23f4acac7f29f/include/rcutils/error_handling.h#L101-L102) takes a non-const allocator this should neither be const.

mikaelarguedas: sounds good https://github.com/ros2/rcl/pull/180/commits/ec3a66f0ce62c6832575d55f9d9a0ab053bdd022

rcl/src/rcl/client.c
```diff
@@ -288,16 +288,20 @@ rcl_take_response(
...
 
-bool rcl_client_is_valid(const rcl_client_t * client)
+bool rcl_client_is_valid(const rcl_client_t * client, rcl_allocator_t * error_msg_allocator)
 {
   const rcl_client_options_t * options;
-  RCL_CHECK_ARGUMENT_FOR_NULL(
-    client, false, rcl_get_default_allocator());
+  const rcl_allocator_t alloc =
+    error_msg_allocator ? *error_msg_allocator : rcl_get_default_allocator();
+  RCL_CHECK_ALLOCATOR_WITH_MSG(&alloc, "error msg allocator is invalid", return false);
```
dirk-thomas: The error message seems to be misleading in case `error_msg_allocator` was a `nullptr`.

Same below.

mikaelarguedas: Please make a suggestion, or feel free to push an updated error message directly to this branch if it's faster

dirk-thomas: `allocator is invalid` is the simplest if we don't want to distinguish the two cases.

rcl/src/rcl/client.c
```diff
@@ -288,16 +288,20 @@ rcl_take_response(
   return RCL_RET_OK;
 }
 
-bool rcl_client_is_valid(const rcl_client_t * client)
+bool rcl_client_is_valid(const rcl_client_t * client, rcl_allocator_t * error_msg_allocator)
 {
   const rcl_client_options_t * options;
-  RCL_CHECK_ARGUMENT_FOR_NULL(
-    client, false, rcl_get_default_allocator());
+  const rcl_allocator_t alloc =
```
dirk-thomas: This is still `const`.

mikaelarguedas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3701/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/867/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3036/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3800/)

mikaelarguedas:
Well removing the const in the signature generated warnings as the X_init functions get a const options structure. https://github.com/ros2/rcl/pull/180/commits/b434d98a73e9ee70b4277a2eb6cbb1daf22efbd1 adds a cast to avoid the warnings

mikaelarguedas:
With last commit
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3704/)

mikaelarguedas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3709/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/870/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3042/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3808/)

mikaelarguedas:
thanks @allenh1 , ci is green , merging

PR #181, opened 2017-11-14
use ament_cmake_pytest instead of ament_cmake_nose

dirk-thomas:
Connect to ament/ament_cmake#115.

### DIFF
--- a/rcl/test/cmake/rcl_add_custom_launch_test.cmake
+++ b/rcl/test/cmake/rcl_add_custom_launch_test.cmake
@@ -32,6 +32,6 @@ macro(rcl_add_custom_launch_test test_name executable1 executable2)
     OUTPUT "test/${test_name}${target_suffix}_$<CONFIG>.py"
     INPUT "${CMAKE_CURRENT_BINARY_DIR}/${test_name}${target_suffix}.py.configure"
   )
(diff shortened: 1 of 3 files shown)


rcl/test/cmake/rcl_add_custom_launch_test.cmake
```diff
@@ -32,6 +32,6 @@ macro(rcl_add_custom_launch_test test_name executable1 executable2)
     OUTPUT "test/${test_name}${target_suffix}_$<CONFIG>.py"
     INPUT "${CMAKE_CURRENT_BINARY_DIR}/${test_name}${target_suffix}.py.configure"
   )
-  ament_add_nose_test(${test_name}${target_suffix} "${CMAKE_CURRENT_BINARY_DIR}/${test_name}${target_suffix}_$<CONFIG>.py" ${ARGN})
+  ament_add_pytest_test(${test_name}${target_suffix} "${CMAKE_CURRENT_BINARY_DIR}/${test_name}${target_suffix}_$<CONFIG>.py" ${ARGN})
```
mikaelarguedas: Nit: this line is getting pretty long

dirk-thomas: Since it only added two characters and our linters pass I would keep it as-is.

We might want to consider to reduce the maximum line length in our linter configuration?

mikaelarguedas: I think the issue is that the cmake linter doesnt catch them because this seems longer that our max length (and was already before this PR)

PR #194, opened 2017-11-28
[wip] C parameters mikael, redo of #54

mikaelarguedas:
This is a redo of #54  
I ported it manually to remove unrelated conflicting changes and avoid conflicts. The bulk of the code is the initial implementation from @jacquelinekay in #54.

The current prototype passes the tests for all rmw implementation except connext_dynamic (due to lack of Nested messages and service support).

What remains to do:
- move utilities to operate on rcl_interfaces C data structures to the `rcl_interfaces` package
- modify the current client/service functions to take a type support and thus be type-erased
  - remove internal storage for requests and responses from internal parameter_client
- address any interface changes that will come with feedback from https://github.com/ros2/ros2/issues/432
- enforce valid parameter names and "namespaces"
- polish and finish convenience functions

Bonus levels:
- provide a way for user to register parameter validation functions
- check/enforce uniqueness of parameter name
- provide more utilities / convenience functions for C array management
- add parameter array types
...

Connects to #28


ivanpauno:
Closing and nobody is working on this and it's outdated.
If in the future we want to move the parameters implementation to C this can be used as a reference.

PR #196, opened 2017-12-03
check if test exists before adding properties

dirk-thomas:
Connect to ros2/ros2#430.

### DIFF
--- a/rcl/test/cmake/rcl_add_custom_launch_test.cmake
+++ b/rcl/test/cmake/rcl_add_custom_launch_test.cmake
@@ -33,5 +33,7 @@ macro(rcl_add_custom_launch_test test_name executable1 executable2)
     INPUT "${CMAKE_CURRENT_BINARY_DIR}/${test_name}${target_suffix}.py.configure"
   )
   ament_add_pytest_test(${test_name}${target_suffix} "${CMAKE_CURRENT_BINARY_DIR}/test/${test_name}${target_suffix}_$<CONFIG>.py" ${ARGN})
-  set_tests_properties(${test_name}${target_suffix} PROPERTIES DEPENDS "${executable1}${target_suffix} ${executable2}${target_suffix}")
+  if(TEST ${test_name}${target_suffix})
+    set_tests_properties(${test_name}${target_suffix} PROPERTIES DEPENDS "${executable1}${target_suffix} ${executable2}${target_suffix}")
+  endif()
 endmacro()


rcl/test/cmake/rcl_add_custom_launch_test.cmake
```diff
@@ -33,5 +33,7 @@ macro(rcl_add_custom_launch_test test_name executable1 executable2)
     INPUT "${CMAKE_CURRENT_BINARY_DIR}/${test_name}${target_suffix}.py.configure"
   )
   ament_add_pytest_test(${test_name}${target_suffix} "${CMAKE_CURRENT_BINARY_DIR}/test/${test_name}${target_suffix}_$<CONFIG>.py" ${ARGN})
-  set_tests_properties(${test_name}${target_suffix} PROPERTIES DEPENDS "${executable1}${target_suffix} ${executable2}${target_suffix}")
+  if(TEST ${test_name}${target_suffix})
+    set_tests_properties(${test_name}${target_suffix} PROPERTIES DEPENDS "${executable1}${target_suffix} ${executable2}${target_suffix}")
+  endif()
```
mikaelarguedas: should ament_`add_*_test` notify the user (throw a warning or something?) if the test to register doesnt exist? it looks like here we would just skip it without notifying the developer..

dirk-thomas: That is already the case: https://github.com/ament/ament_cmake/blob/113915d0b8bd5e90e2649e9d39cf8cf5020a439f/ament_cmake_pytest/cmake/ament_add_pytest_test.cmake#L73-L75

mikaelarguedas: Cool, in that case could we just pass the properties to that function rather than having to add a check after calling ament_add_X_test ?

dirk-thomas: No, because we don't want to reimplement existing CMake functionality in "our" functions. Users should continue to use the CMake way to interact with the test target. We are already doing exactly the same for e.g. `gtest` test targets.

mikaelarguedas: Yeah I agree that's what we do elsewhere so I'm not against merging this as it is an incremental improvement. Though I do still think that we provide macros and helpers to make users' CMake code less verbose and that allowing them to pass the test properties to our test registration macros falls under that category

PR #197, opened 2017-12-05
remove const qualifier for transition states

Karsten1987:
Connects to ros2/rclcpp#418


Karsten1987:
@dirk-thomas addressed your comment. I amended the last commit.

PR #201, opened 2017-12-06
Explain error getting RMW impl id was because not installed

dhood:
[this is just changing the error message, not the functionality]

In the case of having one rmw impl installed, poco is not used, and `rmw_fastrtps_cpp` will be retrieved even if users specify `RMW_IMPLEMENTATION=garbage`. Then we say they don't match.

When multiple rmw impls are installed, poco is used, and nothing is retrieved for `rmw_get_implementation_identifier()` if users specify `RMW_IMPLEMENTATION=garbage`.

The current state since https://github.com/ros2/rcl/pull/198 with multiple rmw implementations installed:
```
$ RMW_IMPLEMENTATION=garbage ./build_isolated/rcl/test/test_rcl__rmw_fastrtps_cpp
[ERROR] [rcl]: Error getting RMW implementation identifier.
```
(it's not clear that the reason for the error was that the rmw implementation is not available)

With this proposal:
```
$ RMW_IMPLEMENTATION=garbage ./build_isolated/rcl/test/test_rcl__rmw_fastrtps_cpp
[ERROR] [rcl]: Error getting RMW implementation identifier/RMW implementation not installed (expected identifier of 'garbage'), exiting with 1.
```

### DIFF
--- a/rcl/src/rcl/rmw_implementation_identifier_check.c
+++ b/rcl/src/rcl/rmw_implementation_identifier_check.c
@@ -128,7 +128,7 @@ INITIALIZER(initialize) {
     if (!actual_rmw_impl_id) {
       RCUTILS_LOG_ERROR_NAMED(
         ROS_PACKAGE_NAME,
-        "Error getting RMW implementation identifier/RMW implementation not installed "
+        "Error getting RMW implementation identifier / RMW implementation not installed "
         "(expected identifier of '%s'), exiting with %d.",
         expected_rmw_impl,
         RCL_RET_ERROR
--- a/rcl/src/rcl/rmw_implementation_identifier_check.c
+++ b/rcl/src/rcl/rmw_implementation_identifier_check.c
@@ -63,7 +63,7 @@ INITIALIZER(initialize) {
   if (ret != RCL_RET_OK) {
     RCUTILS_LOG_ERROR_NAMED(
       ROS_PACKAGE_NAME,
-      "Error getting environement variable 'RMW_IMPLEMENTATION': %s",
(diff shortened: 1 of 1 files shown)


dhood:
As a note in case it's ever useful debugging older workspaces:

Prior to https://github.com/ros2/rcl/pull/198, if users requested garbage and had multiple rmw implementations installed, they'd see:
```
$ RMW_IMPLEMENTATION=garbage ros2 run demo_nodes_cpp talker

>>> [rcutils|error_handling.c:155] rcutils_set_error_state()

  'failed to find shared library of rmw implementation, at /Users/dhood/ros2_ardent/src/ros2/rmw_implementation/rmw_implementation/src/functions.cpp:127'

with this new error message:

  'failed to find shared library of rmw implementation, at /Users/dhood/ros2_ardent/src/ros2/rmw_implementation/rmw_implementation/src/functions.cpp:127'

...
many similar errors
...
<<<
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: failed to initialize rmw implementation: failed to find shared library of rmw implementation, at /Users/dhood/ros2_ardent/src/ros2/rmw_implementation/rmw_implementation/src/functions.cpp:127, at /Users/dhood/ros2_ardent/src/ros2/rcl/rcl/src/rcl/rcl.c:80
```
or 
```
$ RMW_IMPLEMENTATION=garbage RCL_ASSERT_RMW_ID_MATCHES=garbage ./build_isolated/rcl/test/test_rcl__rmw_fastrtps_cpp
Segmentation fault: 11
```

PR #202, opened 2017-12-11
segmentation fault for NULL dereference

gaoethan:
`rcl_lifecycle_get_state` may return NULL

`transition->goal` is checked for NULL but no return
while it's true and this may result the follow-up
NULL dereference

Signed-off-by: Ethan Gao <ethan.gao@linux.intel.com>

### DIFF
--- a/rcl/src/rcl/service.c
+++ b/rcl/src/rcl/service.c
@@ -248,7 +248,7 @@ rcl_take_request(
   RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME, "Service server taking service request")
   const rcl_service_options_t * options = rcl_service_get_options(service);
   RCL_CHECK_FOR_NULL_WITH_MSG(
-    options, "Failed to get service option", return RCL_RET_ERROR, rcl_get_default_allocator());
+    options, "Failed to get service options", return RCL_RET_ERROR, rcl_get_default_allocator());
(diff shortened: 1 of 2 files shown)


rcl_lifecycle/src/rcl_lifecycle.c
```diff
@@ -285,15 +285,22 @@ rcl_lifecycle_is_valid_transition(
   const rcl_lifecycle_state_t * current_state = rcl_lifecycle_get_state(
     &state_machine->transition_map, current_id);
 
+  if (NULL == current_state) {
```
Karsten1987: nitpick: I'd suggest either use `if (!current_state)` or also apply the `RCL_CHECK_FOR_NULL_WITH_MSG` macro here.

rcl/src/rcl/service.c
```diff
@@ -247,6 +247,8 @@ rcl_take_request(
 {
   RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME, "Service server taking service request")
   const rcl_service_options_t * options = rcl_service_get_options(service);
+  RCL_CHECK_FOR_NULL_WITH_MSG(
+    options, "Failed to get service option", return RCL_RET_ERROR, rcl_get_default_allocator());
```
dirk-thomas: Should the error message be "Failed to get service options" (note the trailing s)?

Same below.

Karsten1987:
thanks @gaoethan

PR #207, opened 2018-01-10
reset wait_set type index in SET_RESIZE

jwang11:
It is dangerous to keep type index float when do SET_RESIZE. It is
possible that index greater than new size. Thus, set index to 0.

### DIFF
--- a/rcl/src/rcl/wait.c
+++ b/rcl/src/rcl/wait.c
@@ -296,6 +296,7 @@ rcl_wait_set_get_allocator(const rcl_wait_set_t * wait_set, rcl_allocator_t * al
   } \
   rcl_allocator_t allocator = wait_set->impl->allocator; \
   wait_set->size_of_ ## Type ## s = 0; \
+  wait_set->impl->Type ## _index = 0; \
   if (size == 0) { \
     if (wait_set->Type ## s) { \
       allocator.deallocate((void *)wait_set->Type ## s, allocator.state); \


rcl/src/rcl/wait.c
```diff
@@ -296,6 +296,7 @@ rcl_wait_set_get_allocator(const rcl_wait_set_t * wait_set, rcl_allocator_t * al
   } \
   rcl_allocator_t allocator = wait_set->impl->allocator; \
   wait_set->size_of_ ## Type ## s = 0; \
+  wait_set->impl->Type ## _index = 0; \
```
wjwwood: Shouldn't this be set to the `min()` of the requested `size` and the `wait_set->impl->Type ## _index`? Otherwise this is also like a "clear" which is not the intention I think. If the size before was 4 and the index was 2, then the resize was to be 8, the index should still be 2 right?

jwang11: Look at the realloc code after that, it actually clear the new allocated memory. So, I think the old index is meaningless. If people want old index value, he should save it out before SET_RESIZE. 
  memset((void *)wait_set->Type ## s, 0, sizeof(rcl_ ## Type ## _t *) * size); \
  wait_set->size_of_ ## Type ## s = size; \

wjwwood: You're right, it does zero the memory. :+1:

wjwwood:
I'm combining the CI for this pr and https://github.com/ros2/rclcpp/pull/427:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3863/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/987/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3192/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3941/)

wjwwood:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3881/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1000/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3207/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3970/)

wjwwood:
Fixed an issue on https://github.com/ros2/rclcpp/pull/427, new CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3882/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1001/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3208/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3971/)

PR #208, opened 2018-01-13
change rcutils_time_point_value_t type from uint64_t to int64_t

serge-nikulin:
closes ros2/rcutils#79

@wjwwood, FYI

### DIFF
--- a/rcl/include/rcl/timer.h
+++ b/rcl/include/rcl/timer.h
@@ -50,7 +50,7 @@ typedef struct rcl_timer_t
  * was called, because that information is no longer accessible via the timer.
  * The time since the last callback call is given in nanoseconds.
  */
-typedef void (* rcl_timer_callback_t)(rcl_timer_t *, uint64_t);
+typedef void (* rcl_timer_callback_t)(rcl_timer_t *, int64_t);
 
 /// Return a zero initialized timer.
 RCL_PUBLIC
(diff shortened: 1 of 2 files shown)


wjwwood:
I rebased this pr.

rcl/include/rcl/timer.h
```diff
@@ -137,7 +137,7 @@ RCL_WARN_UNUSED
 rcl_ret_t
 rcl_timer_init(
   rcl_timer_t * timer,
-  uint64_t period,
+  int64_t period,
```
dirk-thomas: Should the passed period be checked to ensure it is >= 0?

wjwwood: I guess so, I'm not sure what a negative period would indicate. I suppose it could be lumped in with zero, but my instinct would be to only allow zero as a special case and raise an error when the period is zero.

dirk-thomas: > my instinct would be to only allow zero as a special case and raise an error when the period is zero.

Thanks, done in #295.

serge-nikulin: @wjwwood,

A negative period and frequency are well defined derived units: 
https://en.wikipedia.org/wiki/Negative_frequency

The problem with software timers is that we don't know in what direction the elctrons will flow ;)

PR #209, opened 2018-01-22
Reset RMWCount when DEALLOC rmw storage of wait set

jwang11:
It is safe to reset RMWCount when free rmw storage of wait set


dirk-thomas:
Thank you for the patch.

PR #210, opened 2018-01-23
Revert "Reset RMWCount when DEALLOC rmw storage of wait set"

wjwwood:
Reverts ros2/rcl#209

I believe this is a regression based on http://ci.ros2.org/job/ci_osx/3214/consoleFull. Since there was no CI (that I saw on the origin pr) I'll assume that's the case and revert until it can be tested separately.

### DIFF
--- a/rcl/src/rcl/wait.c
+++ b/rcl/src/rcl/wait.c
@@ -320,7 +320,6 @@ rcl_wait_set_get_allocator(const rcl_wait_set_t * wait_set, rcl_allocator_t * al
   if (wait_set->impl->RMWStorage) { \
     allocator.deallocate((void *)wait_set->impl->RMWStorage, allocator.state); \
     wait_set->impl->RMWStorage = NULL; \
-    wait_set->impl->RMWCount = 0; \
   }
 
 #define SET_RESIZE_RMW_REALLOC(Type, RMWStorage, RMWCount) \


wjwwood:
Also, nightlies: http://ci.ros2.org/view/nightly/job/nightly_linux_release/697/consoleFull

jwang11:
I know what issue is, RMWCount should be passed as parameter. I'll re-submit a PR

dirk-thomas:
@wjwwood Thank you for the quick revert.

@jwang11 Please do **not** make any pull requests **until** you have compiled the patch locally and ran the tests.

jwang11:
@dirk-thomas @wjwwood  Sorry about that, I should have run and test it locally, but somehow neglect. I'll be more careful later on. On the other hand, it remind me that if ROS2 CI can provide a quick pre-test (ament test) mechanism for each new PR?

wjwwood:
@jwang11 normally we run CI, but since it takes so long, we sometimes try to avoid it on trivial changes, but then again it bites us sometimes.

jwang11:
@wjwwood Formal CI is fine, which can be triggered manually when PR is approved, like you did now. what I mean is another kind of lightweight CI test, something like "ament test compoent_path". It can be auto-triggered when new PR arrive and be done in 10 minutes.

wjwwood:
Yeah, we're working on improving the CI, especially for external contributions. The problem is that if your change causes downstream changes it still bites us on our nightlies and larger changes which need "formal CI". It would help us with more trivial changes like this as a trip wire though.

PR #213, opened 2018-01-30
debug negative timeout in rcl_wait

astralien3000:
Hello !

If I understand well the `rcl_wait` documentation and [this line](https://github.com/astralien3000/rcl/blob/e5c5ee21278257a0df830508a623101d3d9db5df/rcl/src/rcl/wait.c#L561), the condition should be like this.

### DIFF
--- a/rcl/src/rcl/wait.c
+++ b/rcl/src/rcl/wait.c
@@ -557,7 +557,7 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
     temporary_timeout_storage.sec = 0;
     temporary_timeout_storage.nsec = 0;
     timeout_argument = &temporary_timeout_storage;
-  } else if (timeout > 0 || number_of_valid_timers > 0) {
+  } else if (timeout != 0 || number_of_valid_timers > 0) {
     int64_t min_timeout = timeout > 0 ? timeout : INT64_MAX;
     // Compare the timeout to the time until next callback for each timer.
     // Take the lowest and use that for the wait timeout.


rcl/src/rcl/wait.c
```diff
@@ -557,7 +557,7 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
     temporary_timeout_storage.sec = 0;
     temporary_timeout_storage.nsec = 0;
     timeout_argument = &temporary_timeout_storage;
-  } else if (timeout > 0 || number_of_valid_timers > 0) {
+  } else if (timeout != 0 || number_of_valid_timers > 0) {
```
dirk-thomas: `timeout != 0` seems to not make much sense when the above condition `timeout == 0` evaluated to `false` it will always be `true`?

astralien3000: ah, yes true, so removing this condition would be better I guess, and keeping only `number_of_valid_timers > 0`.
Well, actually I don't understand the `||`, should'nt it be a `&&` ?

mikaelarguedas: I think the logic here is that:
"if the timeout is null": wake up immediately

otherwise: "if the timeout provided is a duration to wait" (aka > 0) **or** if we know we have a timer running supposed to wake us up: do stuff

So I think the current behavior is correct

astralien3000: well, negativ values are not handled in this case

mikaelarguedas: Yes, as a negative timeout means "wait forever" there is no need to try to find the optimal timeout (if no timer is running).
So we go straight to [calling rmw_wait](https://github.com/ros2/rcl/pull/213/files#diff-8fda46206fc38bfdbcac0e5213bffddaL601) with a [NULL timeout argument](https://github.com/ros2/rcl/pull/213/files#diff-8fda46206fc38bfdbcac0e5213bffddaR530), that means wait until a condtion is ready: https://github.com/ros2/rmw/blob/master/rmw/include/rmw/rmw.h#L326

mikaelarguedas: Just saw the edit: you're right, "active" would be more appropriate than "valid" in that context

astralien3000: Ah, ok I did not understand that. So I know where my rmw implementation is wrong.
Thank you for your answer !
From my side you can close this PR.

mikaelarguedas: Thanks for the feedback, if you have suggestions to maker it clearer (adding comments ?), we'll happily review any follow-up PR to that effect

PR #214, opened 2018-02-02
update doc for rmw_get_node_names to mention potential null values

dirk-thomas:
Clarifying that each item in the populate array might be `NULL`.

Connect to ros2/ros2#438.

### DIFF
--- a/rcl/include/rcl/graph.h
+++ b/rcl/include/rcl/graph.h
@@ -166,6 +166,7 @@ rcl_names_and_types_fini(rcl_names_and_types_t * names_and_types);
  * The node_names parameter must be allocated and zero initialized.
  * The node_names is the output for this function, and contains
  * allocated memory.
+ * Note that entries in the array might contain `NULL` value.
  * Use rcutils_get_zero_initialized_string_array() for initializing an empty
  * rcutils_string_array_t struct.
  * This node_names struct should therefore be passed to rcutils_string_array_fini()


rcl/include/rcl/graph.h
```diff
@@ -158,6 +158,7 @@ rcl_names_and_types_fini(rcl_names_and_types_t * names_and_types);
  * The node_names parameter must be allocated and zero initialized.
  * The node_names is the output for this function, and contains
  * allocated memory.
+ * Note that entries in the array might contain `NULL` value.
```
sloretz: Just curious, what causes an entry to have a `NULL` value?

dirk-thomas: A participant without a "ROS node name" will result in a `NULL` value. That can e.g. happen when the DDS implementation has some entities running all the time or is there is a vanilla DDS participant which doesn't have a node name defined in ROS.

dhood:
I think this will cause [the `get_node_names` test](https://github.com/ros2/rcl/blob/6c5f98ed4c94d97fede05d5059394a0970b336dc/rcl/test/rcl/test_get_node_names.cpp#L91) to fail with opensplice now because of the nameless participants that opensplice creates.

Not an urgent issue because we don't run CI with opensplice often, and from discussion in https://github.com/ros2/rclcpp/pull/435 it sounds like we'll work towards only returning valid node names again in the future

dirk-thomas:
The parent ticket contains a list of Jenkins builds which were run with OpenSplice enabled. All cross vendor tests passed.

dhood:
maybe it's only cross-vendor `get_node_names` that causes the blank line output from opensplice? I don't remember anymore, but that would explain why it passes in CI

dirk-thomas:
> maybe it's only cross-vendor `get_node_names` that causes the blank line output from opensplice? I don't remember anymore, but that would explain why it passes in CI

Why would that explain that CI passes? The unit tests do test each and every vendor with each other (NxN tests) so they do cover [all these cases](https://ci.ros2.org/job/ci_linux/4159/testReport/test_rclcpp.build.test_rclcpp/).

dhood:
I am referring to RCL get node names test linked above, which is not tested cross-rmw

PR #215, opened 2018-02-28
update style

dirk-thomas:
Connect to ~~ament/uncrustify#20~~. Connect to ament/uncrustify#21.

Ready for review.

### DIFF
--- a/rcl/src/rcl/client.c
+++ b/rcl/src/rcl/client.c
@@ -211,9 +211,7 @@ rcl_client_get_service_name(const rcl_client_t * client)
   return client->impl->rmw_handle->service_name;
 }
 
-/* *INDENT-OFF* */
-#define _client_get_options(client) &client->impl->options;
-/* *INDENT-ON* */
+#define _client_get_options(client) & client->impl->options;
 
 const rcl_client_options_t *
 rcl_client_get_options(const rcl_client_t * client)
--- a/rcl/test/memory_tools/test_memory_tools.cpp
+++ b/rcl/test/memory_tools/test_memory_tools.cpp
@@ -20,22 +20,22 @@
  */
 TEST(TestMemoryTools, test_allocation_checking_tools) {
   size_t unexpected_mallocs = 0;
-  auto on_unexpected_malloc = (
+  auto on_unexpected_malloc =
     [&unexpected_mallocs]() {
       unexpected_mallocs++;
-    });
+    };
   set_on_unexpected_malloc_callback(on_unexpected_malloc);
   size_t unexpected_reallocs = 0;
-  auto on_unexpected_realloc = (
+  auto on_unexpected_realloc =
     [&unexpected_reallocs]() {
       unexpected_reallocs++;
-    });
+    };
   set_on_unexpected_realloc_callback(on_unexpected_realloc);
   size_t unexpected_frees = 0;
-  auto on_unexpected_free = (
+  auto on_unexpected_free =
     [&unexpected_frees]() {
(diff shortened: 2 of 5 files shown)


rcl/src/rcl/client.c
```diff
@@ -211,9 +211,7 @@ rcl_client_get_service_name(const rcl_client_t * client)
   return client->impl->rmw_handle->service_name;
 }
 
-/* *INDENT-OFF* */
-#define _client_get_options(client) &client->impl->options;
-/* *INDENT-ON* */
+#define _client_get_options(client) & client->impl->options;
```
wjwwood: This was on purpose. I don't think the new version handles this correctly still.

wjwwood: In each of the cases below as well where a reference is used in the macro.

rcl/test/memory_tools/test_memory_tools.cpp
```diff
@@ -20,22 +20,22 @@
  */
 TEST(TestMemoryTools, test_allocation_checking_tools) {
   size_t unexpected_mallocs = 0;
-  auto on_unexpected_malloc = (
-    [&unexpected_mallocs]() {
+  auto on_unexpected_malloc = [&unexpected_mallocs]()
+    {
       unexpected_mallocs++;
-    });
+    };
```
wjwwood: These look very strange, would this not work instead?

```c++
  auto on_unexpected_malloc = [&unexpected_mallocs]() {
    unexpected_mallocs++;
  };
```

Or:

```c++
  auto on_unexpected_malloc =
    [&unexpected_mallocs]() {
      unexpected_mallocs++;
    };
```

For me, this should look like an initializer list, so like this:

```c++
foo = {
  'a',
  'b',
}
```

And not:

```c++
foo =
  {
    'a',
    'b',
  };
```

dirk-thomas: Please try if the suggested alternative passes `uncrustify`. I really don't mind how it looks like.

rcl/src/rcl/client.c
dirk-thomas: > I don't think the new version handles this correctly still.

Can you clarify what you mean with this? The current patch passes fine with the current `uncrustify` version. That is why I removed the markers around the line.

rcl/test/memory_tools/test_memory_tools.cpp
dirk-thomas: Please try if the suggested alternative passes `uncrustify` (I think I recall it doesn't). I really don't mind how it looks like.

dirk-thomas: The first options doesn't pass. The second one does. Updated in https://github.com/ros2/rcl/pull/215/commits/dfd8ba25fcd9d440505e03b40161e264a9fd41e3#diff-b4937ca0f22682aded6f4aeaa9d0392bR23

wjwwood: Thanks, lgtm.

rcl/src/rcl/client.c
wjwwood: It's not very important, but there shouldn't be a space between the `&` and the code, that's why the `INDENT-OFF` was there before.

dirk-thomas: If you prefer that I can revert that change.

wjwwood: You can or not, like I said not that important.

PR #216, opened 2018-03-01
change export order for static linking

dirk-thomas:
Connect to ros2/ros2#468.

### DIFF
--- a/rcl/CMakeLists.txt
+++ b/rcl/CMakeLists.txt
@@ -45,11 +45,12 @@ set(${PROJECT_NAME}_sources
 )
 
 add_library(${PROJECT_NAME} ${${PROJECT_NAME}_sources})
+# specific order: dependents before dependencies
 ament_target_dependencies(${PROJECT_NAME}
   "rcl_interfaces"
-  "rcutils"
-  "rmw"
   "rmw_implementation"
+  "rmw"
+  "rcutils"
   "rosidl_generator_c"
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
@dirk-thomas Please open follow-up issues for the packages that can't build their tests with that option so that we don't lose track of them (at least `rcl` and `rclcpp` that I know of, there may be others)

dirk-thomas:
> Please open follow-up issues for the packages that can't build their tests with that option so that we don't lose track of them (at least rcl and rclcpp that I know of, there may be others)

Are you referring that something in the tests is not using that option but using a hard coded `SHARED`?
 In `rclcpp` I didn't see any library being built for the tests. In `rcl` I only found `rcl_memory_tools` and `rcl_memory_tools_interpose`. The PR for [rcutils](ros2/rcutils/pull/94) already addresses this for the memory tools in that repo. I though the consensus was that the memory tools should instead be made available for reuse (rather than each package having its own version). So I am not sure what exactly you would like to cover in the to be created ticket(s)?

mikaelarguedas:
Hmmm I though it had issues linking these packages' test executables when I tested these PRs. I just tried again and it seems to build and run fin :+1: sorry for the noise.

> The PR for rcutils already addresses this for the memory tools in that repo. I though the consensus was that the memory tools should instead be made available for reuse 

Agreed

Issue #219, opened 2018-03-15
handling of command line arguments


wjwwood:
In ROS 1, command line arguments specific to ROS were removed, see:

> argc and argv
> ROS uses these to parse remapping arguments from the command line. It also modifies them so that they no longer contain any remapping arguments, so that if you call ros::init() before processing your command line you will not need to skip those arguments yourself.

-- https://wiki.ros.org/roscpp/Overview/Initialization%20and%20Shutdown#Initializing_the_roscpp_Node

And the signature for `roscpp::init()` uses `int & argc, char ** argv`, allowing it technically to modify what ever C array of strings you give it. This is permissible according to my research, e.g.:

> > The strings are modifiable, although these modifications do not propagate back to the execution environment: they can be used, for example, with std::strtok. The size of the array pointed to by argv is at least argc+1, and the last element, argv[argc], is guaranteed to be a null pointer.

-- http://en.cppreference.com/w/cpp/language/main_function

But modifying the system provided `argv` automatically (implicitly) is a bit surprising for users.

Other places recommend `int main(int argc, char const * argv[])` as the signature to main, which I can't find a foundation for, but seems like a reasonable thing to do.

----

For ROS 2, I'd like to propose that we change our `rcl` interface and down to not directly modify `argv`, and instead provide a function that removes the arguments by returning a copy that is pruned (still not modifying the originals). In `rclcpp` we can do something like have `rclcpp::init()` return a `vector<string>` rather than mutate the `argc` and `argv` it was given. Which keeps the number of statements low, but again avoids mutating the arguments in place.

So that would mean:

- [x] `void init(int argc, char * argv[])` -> `void init(int argc, char const * const argv[])`
  - https://github.com/ros2/rclcpp/blob/168d75cf1ec29ac92ce0e1dfc224c6db96b67167/rclcpp/include/rclcpp/utilities.hpp#L55-L56
- [x] `rcl_ret_t rcl_init(int argc, char ** argv, rcl_allocator_t allocator)` -> `rcl_ret_t rcl_init(int argc, char const * const argv[], rcl_allocator_t allocator)`
  - https://github.com/ros2/rcl/blob/beee7c22ee2bf77c9dfd355ac51a4687109ac986/rcl/include/rcl/rcl.h#L123-L124
[...]

PR #223, opened 2018-03-21
Remove ros arguments

mjcarroll:
Addresses the rcl specific component of #219 

Connects to #219

### DIFF
--- a/rcl/include/rcl/arguments.h
+++ b/rcl/include/rcl/arguments.h
@@ -134,6 +134,20 @@ rcl_arguments_get_unparsed(
   rcl_allocator_t allocator,
   int ** output_unparsed_indices);
 
+/// Return a list of arguments with ROS-specific arguments removed.
+/**
+ *
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_remove_ros_arguments(
+    char const * const argv[],
+    rcl_arguments_t * args,
+    rcl_allocator_t allocator,
+    int* nonros_argc,
+    const char** nonros_argv[]);
+
 /// Reclaim resources held inside rcl_arguments_t structure.
 /**
  * <hr>
--- a/rcl/include/rcl/arguments.h
+++ b/rcl/include/rcl/arguments.h
@@ -42,8 +42,8 @@ rcl_get_zero_initialized_arguments(void);
 
 /// Parse command line arguments into a structure usable by code.
 /**
- * If an argument does not appear to be a valid ROS argument then it is skipped and parsing
- * continues with the next argument in `argv`.
+ * If an argument does not appear to be a valid ROS argument then it is skipped
+ * and parsing continues with the next argument in `argv`.
  * \sa rcl_arguments_get_count_unparsed()
  * \sa rcl_arguments_get_unparsed()
  *
@@ -136,9 +136,10 @@ rcl_arguments_get_unparsed(
 
 /// Return a list of arguments with ROS-specific arguments removed.
 /**
- * Some arguments may not have been intended as ROS arguments. This function
- * populates an array of the aruments in a new argv array. Since the first argument is
- * always assumed to be a process name, the list will always contain the process name.
+ * Some arguments may not have been intended as ROS arguments.
+ * This function populates an array of the aruments in a new argv array.
+ * Since the first argument is always assumed to be a process name, the list
+ * will always contain the first value from the argument vector.
  *
  * <hr>
  * Attribute          | Adherence
--- a/rcl/include/rcl/arguments.h
+++ b/rcl/include/rcl/arguments.h
@@ -164,11 +164,11 @@ RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_ret_t
 rcl_remove_ros_arguments(
-    char const * const argv[],
-    rcl_arguments_t * args,
-    rcl_allocator_t allocator,
-    int* nonros_argc,
-    const char** nonros_argv[]);
+  char const * const argv[],
+  rcl_arguments_t * args,
+  rcl_allocator_t allocator,
+  int * nonros_argc,
+  const char ** nonros_argv[]);
 
 /// Reclaim resources held inside rcl_arguments_t structure.
 /**
--- a/rcl/include/rcl/arguments.h
+++ b/rcl/include/rcl/arguments.h
@@ -136,7 +136,29 @@ rcl_arguments_get_unparsed(
 
 /// Return a list of arguments with ROS-specific arguments removed.
 /**
+ * Some arguments may not have been intended as ROS arguments. This function
+ * populates an array of the aruments in a new argv array. Since the first argument is
+ * always assumed to be a process name, the list will always contain the process name.
  *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | Yes
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] argv The argument vector
+ * \param[in] args An arguments structure that has been parsed.
+ * \param[in] allocator A valid allocator.
+ * \param[out] nonros_argc The count of arguments that aren't ROS-specific
+ * \param[out] nonros_argv An allocated array of arguments that aren't ROS-specific
+ *   This array must be deallocated by the caller using the given allocator.
+ *   If there are no non-ROS args, then the output will be set to NULL.
+ * \return `RCL_RET_OK` if everything goes correctly, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if any function arguments are invalid, or
+ * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
  */
 RCL_PUBLIC
 RCL_WARN_UNUSED
--- a/rcl/include/rcl/arguments.h
+++ b/rcl/include/rcl/arguments.h
@@ -99,7 +99,7 @@ RCL_PUBLIC
 RCL_WARN_UNUSED
 int
 rcl_arguments_get_count_unparsed(
-  rcl_arguments_t * args);
+  const rcl_arguments_t * args);
 
 /// Return a list of indexes that weren't successfully parsed.
 /**
@@ -130,7 +130,7 @@ RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_ret_t
 rcl_arguments_get_unparsed(
-  rcl_arguments_t * args,
+  const rcl_arguments_t * args,
   rcl_allocator_t allocator,
   int ** output_unparsed_indices);
 
@@ -166,7 +166,7 @@ RCL_WARN_UNUSED
 rcl_ret_t
 rcl_remove_ros_arguments(
   char const * const argv[],
-  rcl_arguments_t * args,
+  const rcl_arguments_t * args,
   rcl_allocator_t allocator,
   int * nonros_argc,
   const char ** nonros_argv[]);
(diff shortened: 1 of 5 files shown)


mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4127/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1208/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3410/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4234/)

rcl/src/rcl/arguments.c
```diff
@@ -354,6 +354,45 @@ rcl_get_zero_initialized_arguments(void)
...
+  char const * const argv[],
+  rcl_arguments_t * args,
+  rcl_allocator_t allocator,
+  int * nonros_argc,
+  const char ** nonros_argv[])
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(argv, RCL_RET_INVALID_ARGUMENT, allocator);
+  RCL_CHECK_ARGUMENT_FOR_NULL(nonros_argc, RCL_RET_INVALID_ARGUMENT, allocator);
+  RCL_CHECK_ARGUMENT_FOR_NULL(args, RCL_RET_INVALID_ARGUMENT, allocator);
+  RCL_CHECK_ALLOCATOR_WITH_MSG(&allocator, "invalid allocator", return RCL_RET_INVALID_ARGUMENT);
```
wjwwood: You should probably check the allocator first, otherwise you're passing a potentially invalid allocator to the other null check macros.

rcl/src/rcl/arguments.c
```diff
@@ -354,6 +354,45 @@ rcl_get_zero_initialized_arguments(void)
...
+
+  *nonros_argc = rcl_arguments_get_count_unparsed(args);
+  if (*nonros_argc > 0) {
+    int * unparsed_indices = NULL;
+    rcl_ret_t ret;
+    ret = rcl_arguments_get_unparsed(args, allocator, &unparsed_indices);
+
+    if (RCL_RET_OK != ret) {
+      allocator.deallocate(unparsed_indices, allocator.state);
```
wjwwood: My assumption was that if `rcl_arguments_get_unparsed()` fails it shouldn't store anything in `unparsed_indices`. But that might not be the case. I think this check is safe and defensive, since you initialized it to null.

The only corner case that could be dangerous is that if it does set `unparsed_indices`, at the very least that function should have already free'd the memory it allocated for that spot. So I think this can result in a double free.

I'd consult with @sloretz and see what the error post condition of this function is, and my expectation is that this deallocate when it fails should not happen.

rcl/src/rcl/arguments.c
```diff
@@ -354,6 +354,45 @@ rcl_get_zero_initialized_arguments(void)
...
+      unparsed_indices = NULL;
+      return ret;
+    }
+
+    *nonros_argv = allocator.allocate(sizeof(char *) * *nonros_argc, allocator.state);
```
wjwwood: This line is so confusing, lol. I don't have a recommendation on how to clarify it.

wjwwood: Also this can fail, you need to check if allocate returned null.

rcl/src/rcl/arguments.c
```diff
@@ -354,6 +354,45 @@ rcl_get_zero_initialized_arguments(void)
...
+    for (int i = 0; i < *nonros_argc; ++i) {
+      (*nonros_argv)[i] = argv[unparsed_indices[i]];
+    }
+
+    if (NULL != unparsed_indices) {
```
wjwwood: This check might be redundant, since it's either non-free'd memory or null, both of which are fine to pass to deallocate. Also I think it must be non-null at this point, since you access it in the loop above so long as `nonros_argc > 0`.

rcl/include/rcl/arguments.h
```diff
@@ -134,6 +134,42 @@ rcl_arguments_get_unparsed(
   rcl_allocator_t allocator,
   int ** output_unparsed_indices);
 
+/// Return a list of arguments with ROS-specific arguments removed.
+/**
+ * Some arguments may not have been intended as ROS arguments. This function
+ * populates an array of the aruments in a new argv array. Since the first argument is
+ * always assumed to be a process name, the list will always contain the process name.
```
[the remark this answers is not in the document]
mjcarroll: 80 or 100?  I found several examples in the same file that are wrapped at 100.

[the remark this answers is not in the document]
wjwwood: The hard limit for all lines is 100, I try to keep the doc blocks under 80.

My rationale was that people who want editors to be at 80, can turn on wrapping for code that's over it, but it looks especially bad with these doc blocks.

So, not a requirement, but a nice to have is to get as close to 80 in the paragraphs of doc blocks.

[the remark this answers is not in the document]
wjwwood: For reference, the developer guide for markdown applies here where it makes sense (despite not being 100% markdown):

https://github.com/ros2/ros2/wiki/Developer-Guide#markdown

rcl/src/rcl/arguments.c
mjcarroll: I reviewed the implementation of `rcl_arguments_get_unparsed` and you are correct.  It either passes with `unparsed_indices` set to non-NULL or it fails and `unparsed_indices` is NULL.

sloretz: The only error cases in `rcl_arguments_get_unparsed` happen either when one of the arguments is invalid or when allocating the output fails, do deallocating here isn't needed. I would also expect any new error cases in `rcl_arguments_get_unparsed` to cause it to clean up after itself before returning.

I can update `rcl_arguments_get_unparsed` documentation to say deallocating is only necessary if it returns `RCL_RET_OK`.

rcl/src/rcl/arguments.c
```diff
@@ -362,32 +362,34 @@ rcl_remove_ros_arguments(
...
-      allocator.deallocate(unparsed_indices, allocator.state);
-      unparsed_indices = NULL;
       return ret;
     }
 
-    *nonros_argv = allocator.allocate(sizeof(char *) * *nonros_argc, allocator.state);
+    size_t alloc_size = sizeof(char *) * *nonros_argc;
+    *nonros_argv = allocator.allocate(alloc_size, allocator.state);
+    if (NULL == *nonros_argv) {
+      return RCL_RET_BAD_ALLOC;
```
wjwwood: Do you leak `unparsed_indices` here?

mjcarroll: Yes.

Also as a question, does deallocate need to be checked as well?

rcl/include/rcl/arguments.h
```diff
@@ -64,7 +64,7 @@ rcl_get_zero_initialized_arguments(void);
  * Lock-Free          | Yes
  *
  * \param[in] argc The number of arguments in argv.
- * \param[in] argv Whe values of the arguments.
+ * \param[in] argv The values of the arguments.
```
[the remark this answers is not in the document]
mjcarroll: I'm going to assume that git can probably handle this.

mjcarroll:
@wjwwood Can you help me interpret the Windows failure here?

https://ci.ros2.org/job/ci_windows/4234/consoleFull

rcl/src/rcl/arguments.c
```diff
@@ -354,6 +354,47 @@ rcl_get_zero_initialized_arguments(void)
...
+{
+  RCL_CHECK_ALLOCATOR_WITH_MSG(&allocator, "invalid allocator", return RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(argv, RCL_RET_INVALID_ARGUMENT, allocator);
+  RCL_CHECK_ARGUMENT_FOR_NULL(nonros_argc, RCL_RET_INVALID_ARGUMENT, allocator);
+  RCL_CHECK_ARGUMENT_FOR_NULL(args, RCL_RET_INVALID_ARGUMENT, allocator);
+
+  *nonros_argc = rcl_arguments_get_count_unparsed(args);
+  *nonros_argv = NULL;
+
+  if (*nonros_argc > 0) {
```
sloretz: `nonros_argc` could be `-1` if `args` is zero-intialized (it would get past the `NULL` check above). Maybe this should return `RCL_RET_INVALID_ARGUMENT` in that case?

wjwwood: Maybe I missed why, but should `nonros_argc` just be unsigned, i.e. `size_t`?

rcl/src/rcl/arguments.c
wjwwood: > Also as a question, does deallocate need to be checked as well?

No, it doesn't have a return code. Things like double-free usually result in an abort.

rcl/src/rcl/arguments.c
mjcarroll: I treated it as signed because the `main` signature is also signed

wjwwood: Ah, you're right, never mind.

sloretz: `rcl_arguments_get_count_unparsed` returns `-1` when args is `NULL` or zero initialized. If that's undesirable a couple alternatives are returning an unsigned value that is `0` if the arguments are invalid, or returning `rcl_ret_t` and using an output parameter.

wjwwood: Nah, unless @mjcarroll thinks so, I'd say leave it and handle the negative case.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4128/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1209/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3411/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4235/)

rcl/test/rcl/test_arguments.cpp
```diff
@@ -189,3 +189,61 @@ TEST_F(CLASSNAME(TestArgumentsFixture, RMW_IMPLEMENTATION), test_fini_twice) {
...
+
+  ret = rcl_remove_ros_arguments(
+    argv,
+    &parsed_args,
+    alloc,
+    &nonros_argc,
+    &nonros_argv);
+
+  EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
+  EXPECT_EQ(nonros_argc, 4);
```
sloretz: `ASSERT_EQ` to avoid crash on the checks below if this check fails

rcl/include/rcl/arguments.h
```diff
@@ -134,6 +134,43 @@ rcl_arguments_get_unparsed(
...
+ * \return `RCL_RET_INVALID_ARGUMENT` if any function arguments are invalid, or
+ * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_remove_ros_arguments(
+  char const * const argv[],
+  rcl_arguments_t * args,
```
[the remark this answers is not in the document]
wjwwood: Follow up is fine by me too, but I agree we should do it at some point.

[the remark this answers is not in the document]
mjcarroll: @sloretz was there any reason that they were taking non-const currently?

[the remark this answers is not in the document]
sloretz: @mjcarroll No reason. I should have made them const.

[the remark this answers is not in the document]
mjcarroll: I went ahead and rolled the const in where appropriate.

PR #231, opened 2018-04-30
rcl_node_init() copies node options passed into it

sloretz:
From [this discussion](https://github.com/ros2/rclcpp/pull/467#discussion_r185078315) on ros2/rclcpp#467.

This adds 3 methods that do deep copies:
* `rcl_node_options_copy()`
* `rcl_arguments_copy()`
* `rcl_remap_copy()`

It uses these methods to make `rcl_node_init()` deep copy the options passed in. The documentation states that the caller is responsible for freeing any memory they allocated in the options.

### DIFF
--- a/rcl/src/rcl/node.c
+++ b/rcl/src/rcl/node.c
@@ -222,9 +222,12 @@ rcl_node_init(
   node->impl->rmw_node_handle = NULL;
   node->impl->graph_guard_condition = NULL;
   node->impl->logger_name = NULL;
+  node->impl->options = rcl_node_get_default_options();
   // Initialize node impl.
-  // node options (assume it is trivially copyable)
-  node->impl->options = *options;
+  ret = rcl_node_options_copy(*allocator, options, &(node->impl->options));
+  if (RCL_RET_OK != ret) {
+    goto fail;
+  }
 
   // Remap the node name and namespace if remap rules are given
   rcl_arguments_t * global_args = NULL;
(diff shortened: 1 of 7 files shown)


sloretz:
CI linux [Build Status](https://ci.ros2.org/job/ci_linux/4313/)

wjwwood:
Thanks for following up on it @sloretz!

rcl/src/rcl/node.c
```diff
@@ -383,6 +386,15 @@ rcl_node_init(
       }
       allocator->deallocate(node->impl->graph_guard_condition, allocator->state);
     }
+    if (NULL != node->impl->options.arguments.impl) {
```
Karsten1987: nitpick: You could adhere to the same style of the file/function
```
if (node->impl->options.arguments.impl) {
...
}
```

mikaelarguedas: As we're trying to move closer to MISRA, I encourage the opposite (keeping the current state of this PR) as all conditions need to evaluate explicitly as boolean

Karsten1987: i understand. Yet I believe we should either have a misra check on the complete file or not.
but as I said, just a nitpick. No show-stopper here from my side.

sloretz:
CI
 * Linux [Build Status](http://ci.ros2.org/job/ci_linux/4323/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1315/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3551/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4397/)

sloretz:
All PRs approved. Merging

PR #235, opened 2018-05-04
YAML parameter parser

anup-pem:
A basic yaml parameter parser to parse ros2 node parameters from a yaml file and populate a C data structure


anup-pem:
@dejanpan

anup-pem:
@mikaelarguedas @wjwwood Just wanted to touch base and see when you might be able to take a look at this PR? Thanks

rcl_yaml_param_parser/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
...
+ament_target_dependencies(${PROJECT_NAME} "yaml" "rcutils")
+
+install(TARGETS ${PROJECT_NAME}
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin)
+
+if(BUILD_TESTING)
+  if(NOT WIN32)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
```
mikaelarguedas: we recommend to pass -Wall -Wextra and -Wpedantic to gcc.
The C++ standard use should be defined using CMAKE_CXX_STANDARD (exanple [here](https://github.com/ros2/rclcpp/blob/1be4d2d91491509a692ae46bafc6968cb9b94bb8/rclcpp/CMakeLists.txt#L16-L18))

rcl_yaml_param_parser/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
...
+  endif()
+
+  find_package(ament_cmake_gmock REQUIRED)
+  find_package(ament_cmake_gtest REQUIRED)
+  find_package(ament_cmake_pytest REQUIRED)
```
mikaelarguedas: As this package doesnt have any pytest, this line can be removed

rcl_yaml_param_parser/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
...
+  find_package(ament_lint_auto REQUIRED)
+  ament_lint_auto_find_test_dependencies()
+
+  set(extra_test_libraries)
+  set(extra_test_env)
+  set(extra_lib_dirs)
```
mikaelarguedas: Are any of these variables used ?

rcl_yaml_param_parser/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
...
+
+  set(SKIP_TEST_IF_WIN32 "")
+  if(WIN32)  # (memory tools doesn't do anything on Windows)
+    set(SKIP_TEST_IF_WIN32 "SKIP_TEST")
+  endif()
```
mikaelarguedas: Same question here: is if used anywhere?

rcl_yaml_param_parser/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
...
+
+  macro(rcutils_custom_add_gtest target)
+    ament_add_gtest(${target} ${ARGN})
+  endmacro()
+
+  macro(rcutils_custom_add_gmock target)
+    ament_add_gmock(${target} ${ARGN})
+  endmacro()
```
mikaelarguedas: As this package registers a single gtest, these 2 marcos can be removed as well as `find_package(ament_cmake_gmock REQUIRED)` on line 39

rcl_yaml_param_parser/README.md
```diff
@@ -0,0 +1,25 @@
+**ROS2 rcl YAML paramter parser**
+
+Parse a YAML parameter file and populate the C data structure(params_st)
+
+The data structure params_st will then be used during node initialization
```
mikaelarguedas: params_st is just a variable name used in function definition, I recommend referring to the type here or generally to "a C datastructure"

rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
```diff
@@ -0,0 +1,46 @@
+// Copyright 2018 Open Source Robotics Foundation, Inc.
```
mikaelarguedas: Apex.ai (same in all files below)?

rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
```diff
@@ -0,0 +1,46 @@
...
+#include <stdlib.h>
+
+#include "rcl_yaml_param_parser/types.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+/// \brief Parse the YAML file and populate params_hdl
```
mikaelarguedas: What is `params_hdl` here?

rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
```diff
@@ -0,0 +1,46 @@
...
+/// \param[in] file_path is the path to the YAML file
+/// \param[out] params_st points to the populated paramter struct
+/// \return true on success and false on failure
+bool parse_yaml_file(const char * file_path, rcl_params_t * params_st);
```
mikaelarguedas: I think the functions should be prefixed with `rcl_` for clarity and avoid potential future collisions

rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
```diff
@@ -0,0 +1,46 @@
...
+
+/// \brief Free param structure
+/// \param[in] params_st points to the populated paramter struct
+/// \param[in] allocator memeory allocator to be used
+void free_node_struct(rcl_params_t * params_st, const rcutils_allocator_t allocator);
```
mikaelarguedas: We usually use functions named `_init` and `_fini` for structure allocation/deallocation

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
```diff
@@ -0,0 +1,80 @@
...
+/// \typedef variant_t
+/// \brief variant_t stores the value of a parameter
+/// Only one pointer in this struct will store the value
+typedef struct rcl_variant_s
+{
+  bool * bool_value;  ///< If bool, gets stored here
+  int64_t * integer_value;  ///< If integer, gets stored here
+  double * double_value;  ///< If double, gets stored here
+  char * string_value;  ///< If string, gets stored here
+  // byte_array_t * byte_array_value;  ///< If array of bytes
```
mikaelarguedas: Will this be populated and use in the future?
It should either have a "TODO" explaining what remains to be done or be deleted if this is not planned to be supported

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
```diff
@@ -0,0 +1,80 @@
...
+  rcutils_allocator_t allocator;  ///< Allocator used
+} rcl_variant_t;
+
+/// \typedef node_params_t
+/// \brief node_params_t stores all the parameters(key:value) of a single node
+typedef struct rcl_node_params_s
+{
+  char ** parameter_names;  ///< Array of parameter names (keys)
+  rcl_variant_t * parameter_values;  ///< Array of coressponding parameter values
+  uint32_t num_params;  ///< Number of parameters in the node
```
mikaelarguedas: if this represent the size of the arrays in the structure it should be a size type (`size_t`).
What is the motivation behind using `uint32_t` here? (same below)

rcl_yaml_param_parser/package.xml
```diff
@@ -0,0 +1,26 @@
...
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>rcl_yaml_param_parser</name>
+  <version>0.1.0</version>
+  <description>Package containing various utility types and functions for C</description>
+  <maintainer email="anup.pemmaiah@apex.ai">Anup Pemmaiah</maintainer>
+  <license>Apache License 2.0</license>
+
+  <buildtool_depend>ament_cmake_ros</buildtool_depend>
+  <buildtool_depend>python3-empy</buildtool_depend>
```
mikaelarguedas: It doesnt look like this package depends on / uses empy. This line can be removed

rcl_yaml_param_parser/package.xml
```diff
@@ -0,0 +1,26 @@
...
+
+  <build_depend>yaml</build_depend>
```
mikaelarguedas: Nitpick: If libyaml is needed at runtime as well, this should become `<depend>yaml</depend>`

rcl_yaml_param_parser/package.xml
```diff
@@ -0,0 +1,26 @@
...
+  <build_depend>rcutils</build_depend>
+
+  <test_depend>ament_cmake_gmock</test_depend>
+  <test_depend>ament_cmake_gtest</test_depend>
+  <test_depend>ament_cmake_pytest</test_depend>
```
mikaelarguedas: (Same as above) pytest and gmock can be removed from here

rcl_yaml_param_parser/package.xml
```diff
@@ -0,0 +1,26 @@
...
+  <test_depend>ament_lint_common</test_depend>
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>launch_testing</test_depend>
```
mikaelarguedas: this package doesn't use `launch_testing` so this line can be removed

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1164 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <stdlib.h>
+#include <stdint.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <errno.h>
+#include <yaml.h>
```
mikaelarguedas: Nitpick: include statements should be in alphabetical order

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1164 @@
...
+  DATA_TYPE_INT64 = 2U,
+  DATA_TYPE_DOUBLE = 3U,
+  DATA_TYPE_STRING = 4U
+} data_types_t;
+
+#define MAX_STRING_SZ 128U
+#define PARAMS_KEY "params"
+
+#define MAX_NUM_NODE_ENTRIES 256U
+#define MAX_NUM_PARAMS_PER_NODE 512U
```
mikaelarguedas: How were these numbers decided (not asking to change them just curious), we should document the values picked

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1164 @@
...
+
+static rcutils_ret_t node_params_create(
+  rcl_node_params_t * node_params,
+  const rcutils_allocator_t allocator);
+
+static rcutils_ret_t param_struct_initialize(
```
mikaelarguedas: do `create` and `initialize` have different meanings here ?
I recommend using `_init` or `get_zero_initialized_XX` to match the rest of the stack. Examples [here](https://github.com/ros2/rcl/blob/e8491ab8abba5c9653fdfc46064cc3b0d5ee18b9/rcl/include/rcl/wait.h#L118) and [here](https://github.com/ros2/rcl/blob/e8491ab8abba5c9653fdfc46064cc3b0d5ee18b9/rcl/include/rcl/wait.h#L63)

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1164 @@
...
+#include <stdint.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <errno.h>
+#include <yaml.h>
+
+#include "rcl_yaml_param_parser/parser.h"
+#include "rcl_yaml_param_parser/types.h"
+
+#include "rcutils/error_handling.h"
```
mikaelarguedas: All the functions in this file should use `rcl/error_handling`
`rcutils_ret_t` and `rcutils_allocators_t` should be replaced by their rcl equivalent where appropriate

anup-pem:
@mikaelarguedas 
- bytes structures are commented out.
    Have not provided the byte support in this phase. Was thinking of adding it later as it requires more discussion about identifying/differentiating a byte type. As per my understanding from the yaml doc (http://yaml.org/type/binary.html) an additional "!!binary" tag may have to be added before the byte value in the yaml file.

- At which point we'll have to look how to integrate libyaml to be able to test this code on all platforms in our CI
    For now I was thinking the user download the libyaml. Once we have the parser merged with all the supported API's can discuss about integrating libyaml into the CI. In my development, have a libyaml built as part of ament and since we will be modifying libyaml will have a copy of libyaml in apex repo. In case of ROS2 repo, might need some direction about is it better to download and compile on the fly before compiling the parser (like poco_vendor) or is it better to have a copy of libyaml in the CI infrastructure.

rcl_yaml_param_parser/src/parser.c
mikaelarguedas: yeah that's my bad I missed that during my first search / replace sweep

mikaelarguedas:
> For now I was thinking the user download the libyaml. Once we have the parser merged with all the supported API's can discuss about integrating libyaml into the CI.

Unfortunately merging this without having it working on CI means that our CI will fail for any new PR and nightly builds. If we merge this before, we need to add extra logic in CI to ignore this package.
This can be done quite easily but that means that people cloning our repository will not be able to build the workspace either. So I think this needs to build and pass on all platforms before being merged (that's our general policy for any new code).

> In case of ROS2 repo, might need some direction about is it better to download and compile on the fly before compiling the parser (like poco_vendor) or is it better to have a copy of libyaml in the CI infrastructure.

Yes, vendoring it ourselves in a first time is definitely a possibility :+1: . We can later look at how to avoid having to compile it every time.

mikaelarguedas:
> Have not provided the byte support in this phase. Was thinking of adding it later as it requires more discussion about identifying/differentiating a byte type.

Yeah that's fine, we should put a TODO in the code and open a follow-up ticket to make sure we come back to it after the first version is merged

rcl_yaml_param_parser/CMakeLists.txt
anup-pem: Good point. I have added those flags

rcl_yaml_param_parser/CMakeLists.txt
anup-pem: Removed it

rcl_yaml_param_parser/CMakeLists.txt
anup-pem: No, lingering from rcutils CMake. Removed

rcl_yaml_param_parser/CMakeLists.txt
anup-pem: No, lingering from rcutils CMake. Removed

rcl_yaml_param_parser/CMakeLists.txt
anup-pem: Ahhh ok. Removed

rcl_yaml_param_parser/README.md
anup-pem: Fixed it

rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
anup-pem: Good point. Fixed it

rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
anup-pem: Ahh... Initially I had named the parameter as params_hdl, where it is a handle. That makes more sense when the structure is hidden and the API just retrurns a void pointer. In this case since the structure is in the header, it does not make much sense. Anyways, have fixed it

rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
anup-pem: Agree! Fixed it

rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
anup-pem: Agree! Fixed it

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
anup-pem: Will not support for now so removed it. When we have some time, will discuss and implement it

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
anup-pem: I am not sure if I should use size_t here. Couple of things
1. Number of nodes and number of parameters describes the variable much better and would prefer uint32_t than size_t which is used to describe  just a set of bytes
1. I uses these parameters as indexes in the array and uint32_t makes more clearer that size_t

Just my 2 cents

rcl_yaml_param_parser/package.xml
anup-pem: Agree. Removed

rcl_yaml_param_parser/package.xml
anup-pem: Make sense. Fixed

rcl_yaml_param_parser/package.xml
anup-pem: Agree. Fixed

rcl_yaml_param_parser/package.xml
anup-pem: Removed it

rcl_yaml_param_parser/src/parser.c
anup-pem: Good point. Fixed it

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
mikaelarguedas: > Will not support for now so removed it. When we have some time, will discuss and implement it

If we agree that we should support for it. The field should be part of the datasrtucture.
It can always be NULL pending implementations but this allow us to have a well defined interface if people want to interact with these structures. Otherwise we may break when we add the field that is undesired in my opinion

rcl_yaml_param_parser/CMakeLists.txt
mikaelarguedas: Please also address the comment about the way to specify the C++ standard

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
mikaelarguedas: If you agree with the previous comment it would be great to add the field back

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
mikaelarguedas: As this field represent the size of the arrays, I think that this should have a size type

mikaelarguedas:
Ok I got this to compile on Linux with the linked PRs.

This generates some warnings on MacOS (https://ci.ros2.org/job/ci_osx/3592/warnings11Result/)

I didnt get this to compile on Windows so far. Weirdly Fast-RTPS is [failing to build](https://ci.ros2.org/job/ci_windows/4450/) (shouldn't have any interaction with this code...).

rcl_yaml_param_parser/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+  ament_lint_auto_find_test_dependencies()
+
+  macro(rcutils_custom_add_gtest target)
+    ament_add_gtest(${target} ${ARGN})
+  endmacro()
+
+  # Gtests
+  rcutils_custom_add_gtest(test_parse_yaml
+    test/test_parse_yaml.cpp
+  )
```
mikaelarguedas: can you please remove the custom `rcutils ...` macro and simply call
```
ament_add_gtest(test_parse_yaml
  test/test_parse_yaml.cpp
)
```

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1165 @@
...
+/// map level 5 : Params namespace mapping
+typedef enum yaml_map_lvl_e
+{
+  MAP_UNINIT_LVL = 0U,
+  MAP_TOP_LVL = 1U,
+  MAP_NODE_NS_LVL = 2U,
+  MAP_NODE_NAME_LVL = 3U,
+  MAP_PARAMS_LVL = 4U,
+  MAP_PARAMS_NS_LVL = 5U
+} yaml_map_lvl_t;
```
mikaelarguedas: Is this implementation robust to multiple namespacing levels? e.g.:

```
stereo_cameras_ns_top:
  stereo_cameras_ns_low:
    stereo_camera1:
      params:
        is_cam_on: [true, true, false, true, false, false]
        brand: Bosch
    stereo_camera2:
      params:
        stereo_camera_dr:
          stereo_camera_left:
            image_type: ir
          stereo_camera_right:
            image_type: rbg
```

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1165 @@
...
+{
+  DATA_TYPE_UNKNOWN = 0U,
+  DATA_TYPE_BOOL = 1U,
+  DATA_TYPE_INT64 = 2U,
+  DATA_TYPE_DOUBLE = 3U,
+  DATA_TYPE_STRING = 4U
+} data_types_t;
+
+#define MAX_STRING_SZ 128U
+#define PARAMS_KEY "params"
```
mikaelarguedas: After thinking about it a bit more. A name that cannot be a node name may be more appropriate for this key.
As we forbid double underscores in ROS node/topic names, `ros__parameters` may be a better candidate

anup-pem:
> Ok I got this to compile on Linux with the linked PRs.

Awesome! I saw you libyaml_vendor file

> This generates some warnings on MacOS (https://ci.ros2.org/job/ci_osx/3592/warnings11Result/)

Dont have access to Mac and windows build here. WIll look into it. Thanks for running.

> I didnt get this to compile on Windows so far. Weirdly Fast-RTPS is failing to build (shouldn't have any interaction with this code...).

Hmmm.... Might be can try re-start the build and see it.

rcl_yaml_param_parser/src/parser.c
anup-pem: Makes sense. WIll change it

rcl_yaml_param_parser/CMakeLists.txt
anup-pem: Good point!

rcl_yaml_param_parser/src/parser.c
mikaelarguedas: Oh I must have messed up something then. I tried just adding one level of namespacing to the correct_config.yaml file and it fails to parse on my machine. Do you have an idea of what I did wrong ?

```
diff --git a/rcl_yaml_param_parser/test/correct_config.yaml b/rcl_yaml_param_parser/test/correct_config.yaml
index 4b5e50d..51954f8 100644
--- a/rcl_yaml_param_parser/test/correct_config.yaml
+++ b/rcl_yaml_param_parser/test/correct_config.yaml
@@ -38,6 +38,7 @@ new_camera_ns:
       brand: Bosch
   new_camera2:
     params:
-      camera_dr:
-        dr_name: default
-      brand: Mobius 
+      camera_left:
+        camera_dr:
+          dr_name: default
+        brand: Mobius 
```

The test fail with the following message:
```
1: Will not support deeper mappings at line 43, at /home/mikael/work/ros2/bouncy_ws/src/ros2/rcl/rcl_yaml_param_parser/src/parser.c:1079
1: /home/mikael/work/ros2/bouncy_ws/src/ros2/rcl/rcl_yaml_param_parser/test/test_parse_yaml.cpp:37: Failure
1: Value of: res
1:   Actual: false
1: Expected: true

```

mikaelarguedas:
> > I didnt get this to compile on Windows so far. Weirdly Fast-RTPS is failing to build (shouldn't have any interaction with this code...).

> Hmmm.... Might be can try re-start the build and see it.

I'll look into it with more details but I ruled out the fact it could be host specific or due to internet connectivity as it failed on all our windows machines:
https://ci.ros2.org/job/ci_windows/4449/
https://ci.ros2.org/job/ci_windows/4450/
https://ci.ros2.org/job/ci_windows/4451/

But a build from master at the same time on the same machine succeeded: https://ci.ros2.org/job/ci_windows/4447/

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
```diff
@@ -0,0 +1,79 @@
...
+} rcl_node_params_t;
+
+/// \typedef params_t
+/// \brief params_t stores all the parameters of all nodes of a process
+typedef struct rcl_params_s
+{
+  char ** node_namespaces;  ///< List of namespaces the corresponding node belongs to
+  char ** node_names;  ///< List of names of the node
```
mikaelarguedas: Actually (sorry for going back and forth on this), there is no real benefit in keeping the node_namespaces and node_names separated so we could do the same as for `rcl_node_params_s` and have a single `char ** node_names` that contains `node_namespace/node_name` instead of 2 different fields

rcl_yaml_param_parser/src/parser.c
anup-pem: Sorry my earlier answer is not entirely true. I restrict it to one namespacing level after params but can have mutilpe namespaces under params.  Things can get crazy allowing multiple namespace levels. Looked at quite a few examples in the config files in various places like autoware stacks, our internal lidar stack, and did not find a need for many namespace levels. Trying to keep simple

mikaelarguedas: It is very common in ROS to have several levels of namespacing for parameters.
For example, a minimalistic pan-tilt camera controller:
```
gimbal_controller:
  pan_joint:
    pid: {p: 50.0, i: 0, d: 2}
  tilt_joint:
    pid: {p: 50.0, i: 0, d: 2}
```

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
anup-pem: I need to look into it.

anup-pem: Should not be bad

rcl_yaml_param_parser/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+
+project(rcl_yaml_param_parser)
+
+find_package(ament_cmake_ros REQUIRED)
+find_package(yaml REQUIRED)
+find_package(rcutils REQUIRED)
+find_package(rcl REQUIRED)
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wpedantic")
```
mikaelarguedas: Actually this is a C library so we need to set the flags for the C compiler.
A more generic way to do it is:
```
if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang")
  add_compile_options(-Wall -Wextra -Wpedantic)
```
This way the option is added for both C and C++.

Compiling with this generates warnings:
```
/home/mikael/work/ros2/bouncy_ws/src/ros2/rcl/rcl_yaml_param_parser/src/parser.c: In function ‘rcl_yaml_node_struct_print’:
/home/mikael/work/ros2/bouncy_ws/src/ros2/rcl/rcl_yaml_param_parser/src/parser.c:318:31: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
             for (int i = 0; i < param_var->bool_array_value->size; i++) {
                               ^
/home/mikael/work/ros2/bouncy_ws/src/ros2/rcl/rcl_yaml_param_parser/src/parser.c:327:31: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
             for (int i = 0; i < param_var->integer_array_value->size; i++) {
                               ^
/home/mikael/work/ros2/bouncy_ws/src/ros2/rcl/rcl_yaml_param_parser/src/parser.c:335:31: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
             for (int i = 0; i < param_var->double_array_value->size; i++) {
                               ^
/home/mikael/work/ros2/bouncy_ws/src/ros2/rcl/rcl_yaml_param_parser/src/parser.c:343:31: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
             for (int i = 0; i < param_var->string_array_value->size; i++) {
                               ^
/home/mikael/work/ros2/bouncy_ws/src/ros2/rcl/rcl_yaml_param_parser/src/parser.c: In function ‘parse_value’:
/home/mikael/work/ros2/bouncy_ws/src/ros2/rcl/rcl_yaml_param_parser/src/parser.c:619:24: warning: pointer targets in initialization differ in signedness [-Wpointer-sign]
   const char * value = event.data.scalar.value;
                        ^
/home/mikael/work/ros2/bouncy_ws/src/ros2/rcl/rcl_yaml_param_parser/src/parser.c: In function ‘parse_key’:
/home/mikael/work/ros2/bouncy_ws/src/ros2/rcl/rcl_yaml_param_parser/src/parser.c:797:24: warning: pointer targets in initialization differ in signedness [-Wpointer-sign]
   const char * value = event.data.scalar.value;
```

anup-pem: Ooops! you are right. Will do that

rcl_yaml_param_parser/src/parser.c
anup-pem: Yes, it supports just one mapping level(params namespace mapping) after params. In the above example it has two mapping levels after params `camera_left` and `camera_dr`. It can have 

```
   new_camera2:
     params:
      camera_left_dr:
        dr_name: default
        brand: Mobius 
      camera_right_dr:
        dr_name: default
        brand: Mobius 
```

rcl_yaml_param_parser/CMakeLists.txt
anup-pem: Fixed the CmakeLists.txt and the warnings

rcl_yaml_param_parser/src/parser.c
anup-pem: Changed it!

rcl_yaml_param_parser/CMakeLists.txt
anup-pem: Fixed it!

anup-pem:
> This generates some warnings on MacOS (https://ci.ros2.org/job/ci_osx/3592/warnings11Result/)

Changed to use PRId64 to print int64_t. Hopefully it is fixed in Mac

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
anup-pem: Have removed node_namespaces and just kept nodes names as discussed

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
anup-pem: Changed to size_t

rcl_yaml_param_parser/CMakeLists.txt
anup-pem: Fixed it

rcl_yaml_param_parser/src/parser.c
anup-pem: We in the design decided with the following format. A parameter_namespace_string and series of `<field_name> : <value>` pairs. 
```
<node_namespace_string>:  # optional
  <node1_name>:
    params:
      <field_name>: <field_value>
      <parameter_namespace_string>:   # optional
        <field1_name>: <field1_value>
        <field2_name>: <field2_value>
  <node2_name>:
    params:
      <field_name>: <field_value>
      <parameter_namespace_string>:   # optional
        <field1_name>: <field1_value>
        <field2_name>: <field2_value>
```
So the parameter name we decided was `parameter_namespace_string.field_name`.  If we want to change this behavior, we can put it as a TODO and add more support after the basic support is done.

Just to keep in mind, a libyaml C parser library is a very very minimalistic parser. It is more of a tokeniser. Just parses each word and returns it. It does not provide any more info and the rcl parser code has to keep track of everything else a yaml specification talks about. It does not even tell if a returned token is a key or a value.  Doing multiple levels of mappings, can make the rcl_yaml_parser code very cumbersome. That is why there are lot of other wrappers written around libyamlc to parse fancy things. The biggest advantage of using libyaml C is, it will be comparatively easier to get it through MISRA compared to other parsers. It still has around 3000 MISRA complaints. 

Again, if you think we should have mutiple namespace levels and mutiple layers of `key:value` pairs inside a value like  `pid: {p: 50.0, i: 0, d: 2}`, we can put as a TODO and support more adavanced features later on

anup-pem: Mikael, where can I find the code where mutiple level of parameter namespace gets used  (Eg: the gimbal_controller) . Just for me to get an understanding. Do you have a pointer to the code? Thanks

rcl_yaml_param_parser/CMakeLists.txt
```diff
@@ -8,6 +8,16 @@ find_package(rcl REQUIRED)
 find_package(libyaml_vendor REQUIRED)
 find_package(yaml REQUIRED)
 
+# Default to C11
+if(NOT CMAKE_C_STANDARD)
+  set(CMAKE_C_STANDARD 11)
```
mikaelarguedas: If this package doesnt require C11, let's default to C99 instead as this is the version we target by default: https://github.com/ros2/ros2/wiki/Developer-Guide#c

mikaelarguedas:
New round of CI with latest changes (to see if the warnings and failure on windows have been addressed)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4405/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1355/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3600/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4462/)

It looks like the Clang warnings are now fixed but Windows still fails.

It looks like an issue in the include paths when building in parallel.

I tried building them sequentially, and Fast-RTPS builds successfully but I face another failure, it looks like when passing `BUILD_SHARED_LIBS` to libyaml, the `yaml.dll` is installed but nor `yaml.lib`: https://ci.ros2.org/job/ci_windows/4466/console

anup-pem:
Hmmm..... was just curious if you try to build manually on windows, does yaml.lib get created and installed?

mikaelarguedas:
Yeah I'm currently doing so, and I could get it to compile and install both the lib and dll. Now it's rcl_yaml_parser that fails to build: https://ci.ros2.org/job/ci_windows/4472/console

Looking into it now

wjwwood:
Usually, in my experience, if there's no `.lib` but otherwise it builds ok, then there's missing dll exports on symbols. Maybe you need to pass something special to libyaml to get it to export symbols properly on Windows.

mikaelarguedas:
> Usually, in my experience, if there's no .lib but otherwise it builds ok, then there's missing dll exports on symbols. Maybe you need to pass something special to libyaml to get it to export symbols properly on Windows.

Yeah that is my understanding as well, but libyaml seems to do the right thing here: https://github.com/yaml/libyaml/blob/10c907871f1ccd779c7fccf6b81a62762b5c4e7b/CMakeLists.txt#L63-L65

I think I messed up something in my CMake arguments making it not export them anymore.

I'm currently adding visibility macros to this rcl_yaml_param_parser package as well so that it's usable by downstream packages and it currently doesnt export any symbol. I'll go back revisit my CMake args once I get that working

mikaelarguedas:
@anup-pem thanks for changing from `uint32_t` to `size_t`.
Apparently the variables that were using it did not get updated accordingly generating new warnings: https://ci.ros2.org/job/ci_windows/4477/

I can now get libyaml_vendor and this package to compile on Windows.
This still breaks the Fast-RTPS build when building in parallel. Still unclear to me why.

anup-pem:
>The main concern I have is the memory allocation strategy in this PR.
I think all data structures should provide an X_init and a X_fini function to allow user to pass an allocator and use that allocator for memory allocaiton/deallocation rather than using malloc without possibility to override it.

Agree. Since it was not in the API that was defined I did not add it, but I should have brought it up. I missed it. Have added below in the API. Let me know if this works, and then will change the code accordingly

```
/// \brief Parse the YAML file, initialize and populate params_st
/// \param[in] file_path is the path to the YAML file
/// \param[out] params_st points to the populated paramter struct
/// \return true on success and false on failure
RCL_YAML_PARAM_PARSER_PUBLIC
bool rcl_parse_yaml_file(
  const char * file_path,
  rcl_params_t * params_st,
  const rcutils_allocator_t allocator);  <<<<<<<<<<<<<<<<<<<<<<<<<<<<<
 
/// \brief Free param structure
/// \param[in] params_st points to the populated paramter struct
/// \param[in] allocator memeory allocator to be used
RCL_YAML_PARAM_PARSER_PUBLIC
void rcl_yaml_node_struct_fini(
  rcl_params_t * params_st, 
  const rcutils_allocator_t allocator);
```

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1336 @@
...
+  MAP_UNINIT_LVL = 0U,
+  MAP_NODE_NAME_LVL = 1U,
+  MAP_PARAMS_LVL = 2U,
+} yaml_map_lvl_t;
+
+/// Basic supported data types in the yaml file
+typedef enum data_types_e
+{
+  DATA_TYPE_UNKNOWN = 0U,
+  DATA_TYPE_BOOL = 1U,
```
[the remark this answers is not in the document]
anup-pem: Will add it when BYTE support is provided. As this is internal to the code and not public, it should be OK. The main reason I did not add is, if we add BYTE type here, then need to provide a empty switch statement for BYTE otherwise MISRA linters will complain

rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
```diff
@@ -0,0 +1,55 @@
...
+RCL_YAML_PARAM_PARSER_PUBLIC
+bool rcl_parse_yaml_file(
+  const char * file_path,
+  rcl_params_t * params_st);
+
+/// \brief Free param structure
+/// \param[in] params_st points to the populated paramter struct
+/// \param[in] allocator memeory allocator to be used
+RCL_YAML_PARAM_PARSER_PUBLIC
+void rcl_yaml_node_struct_fini(
```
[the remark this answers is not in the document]
anup-pem: Will be adding an allocator to rcl_parse_yaml_file as pointed by my comment in https://github.com/ros2/rcl/pull/235#issuecomment-390748012. That should solve the problem

rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
```diff
@@ -0,0 +1,55 @@
...
+  rcl_params_t * params_st,
+  const rcutils_allocator_t allocator);
+
+/// \brief Print the parameter structure to stdout
+/// \param[in] params_st points to the populated paramter struct
+RCL_YAML_PARAM_PARSER_PUBLIC
+void rcl_yaml_node_struct_print(
```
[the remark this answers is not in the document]
anup-pem: Makes sense. How about I file a ticket and fix it seperately?

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1336 @@
...
+{
+  char * cur_ns;
+  uint32_t * cur_count;
+  char sep_char;
+  size_t name_len;
+  size_t ns_len;
+  size_t tot_len;
+  rcl_ret_t res = RCL_RET_OK;
+
+  switch (ns_type) {
```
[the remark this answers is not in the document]
anup-pem: I agree with you. I went back and forth with this. Initially I had one function, with a flag indicating to add or remove but was not fully convinced with the clarity.

mikaelarguedas:
Yeah that looks good, I think we should also expose the initialization function as well as I expect people to:
1. allocate the structure
1. call rcl_parse_yaml_file
1. call rcl_yaml_node_struct_fini when they're done with the struct

So I think we chould expose the function to do 1.
Otherwise it's unclear how people should do 1.

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1336 @@
...
+#include "rcl/error_handling.h"
+#include "rcl/types.h"
+#include "rcutils/strdup.h"
+
+/// NOTE: Will allow a max YAML mapping depth of 5
+/// map level 1 : Top level mapping of the config file
+/// map level 2 : Node namespace mapping
+/// map level 3 : Node name mapping
+/// map level 4 : Params mapping
+/// map level 5 : Params namespace mapping
```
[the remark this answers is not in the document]
anup-pem: Oppps! Fixed it

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1336 @@
...
+  size_t name_len;
+  size_t ns_len;
+  size_t tot_len;
+  rcl_ret_t res = RCL_RET_OK;
+
+  switch (ns_type) {
+    case NS_TYPE_NODE:
+      cur_ns = ns_tracker->node_ns;
+      cur_count = &(ns_tracker->num_node_ns);
+      sep_char = '/';
```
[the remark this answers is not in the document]
anup-pem: Added support for string seperators.

rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
```diff
@@ -0,0 +1,55 @@
...
+/// \param[out] params_st points to the populated paramter struct
+/// \return true on success and false on failure
+RCL_YAML_PARAM_PARSER_PUBLIC
+bool rcl_parse_yaml_file(
+  const char * file_path,
+  rcl_params_t * params_st);
+
+/// \brief Free param structure
+/// \param[in] params_st points to the populated paramter struct
+/// \param[in] allocator memeory allocator to be used
```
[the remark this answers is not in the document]
anup-pem: Done!

rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
```diff
@@ -0,0 +1,55 @@
...
+#include "rcl_yaml_param_parser/visibility_control.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+/// \brief Parse the YAML file, initialize and populate params_st
+/// \param[in] file_path is the path to the YAML file
+/// \param[out] params_st points to the populated paramter struct
```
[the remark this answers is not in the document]
anup-pem: Done!

rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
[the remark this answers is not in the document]
anup-pem: Have provided a cl_yaml_node_struct_init() API

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
```diff
@@ -0,0 +1,87 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#ifndef RCL_YAML_PARAM_PARSER__TYPES_H_
+#define RCL_YAML_PARAM_PARSER__TYPES_H_
+
+#include "rcl/allocator.h"
+#include "rcutils/types/string_array.h"
+
+/// \typedef bool_array_t
```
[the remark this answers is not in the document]
anup-pem: Done!

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1336 @@
...
+    }
+    if (0U == *cur_count) {
+      cur_ns = rcutils_strdup(name, allocator);
+      if (NULL == cur_ns) {
+        return RCL_RET_BAD_ALLOC;
+      }
+    } else {
+      ns_len = strlen(cur_ns);
+      name_len = strlen(name);
+      tot_len = ns_len + name_len + 2U;
```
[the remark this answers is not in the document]
anup-pem: Got fixed as part of the string namespace seperator support

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1336 @@
...
+} ns_type_t;
+
+/// Keep track of node and parameter name spaces
+typedef struct name_space_tracker_s
+{
+  char * node_ns;
+  uint32_t num_node_ns;
+  char * param_ns;
+  uint32_t num_param_ns;
+} name_space_tracker_t;
```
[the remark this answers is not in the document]
anup-pem: Done!

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1336 @@
...
+  DATA_TYPE_INT64 = 2U,
+  DATA_TYPE_DOUBLE = 3U,
+  DATA_TYPE_STRING = 4U
+} data_types_t;
+
+typedef enum ns_type_e
+{
+  NS_TYPE_NODE = 1U,
+  NS_TYPE_PARAM = 2U
+} ns_type_t;
```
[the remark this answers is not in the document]
anup-pem: Done!

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1336 @@
...
+    char * node_name = params_st->node_names[nd_idx];
+    if (NULL != node_name) {
+      allocator.deallocate(node_name, NULL);
+    }
+
+    if (NULL != params_st->params) {
+      char * param_name;
+      rcl_variant_t * param_var;
+      for (pr_idx = 0; pr_idx < params_st->params[nd_idx].num_params; pr_idx++) {
+        param_name = params_st->params[nd_idx].parameter_names[pr_idx];
```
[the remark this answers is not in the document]
anup-pem: Good catch! Fixed!

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1336 @@
...
+}
+
+///
+/// Dump the param structure
+///
+void rcl_yaml_node_struct_print(
+  const rcl_params_t * const params_st)
+{
+  uint32_t nd_idx;
+  uint32_t pr_idx = 0U;
```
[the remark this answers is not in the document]
anup-pem: Done!

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1336 @@
...
+      cur_count = &(ns_tracker->num_node_ns);
+      sep_char = '/';
+      break;
+    case NS_TYPE_PARAM:
+      cur_ns = ns_tracker->param_ns;
+      cur_count = &(ns_tracker->num_param_ns);
+      sep_char = '.';
+      break;
+    default:
+      res = RCL_RET_ERROR;
```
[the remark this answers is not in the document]
anup-pem: pc_lint is not happy if returns are made there. I have seen pc_lint complaining. I will clean it up when I make pc_lint fixes

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1336 @@
...
+      if (1U == *cur_count) {
+        allocator.deallocate(cur_ns, NULL);
+        cur_ns = NULL;
+      } else {
+        char * last_name = cur_ns;
+        ns_len = strlen(cur_ns);
+        name_len = 0U;
+
+        last_name = (last_name + ns_len);
+        while ((NULL != last_name) && (sep_char != *(last_name))) {
```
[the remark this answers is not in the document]
anup-pem: Not required anymore and the line got removed

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1336 @@
...
+      break;
+    default:
+      res = RCL_RET_ERROR;
+      break;
+  }
+
+  if (RCL_RET_OK == res) {
+    /// Add a name to ns
+    if (NULL == name) {
+      return RCL_RET_INVALID_ARGUMENT;
```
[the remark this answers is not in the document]
anup-pem: Will keep it like that as a message is printed in the calling function

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1336 @@
...
+/// Keep track of node and parameter name spaces
+typedef struct name_space_tracker_s
+{
+  char * node_ns;
+  uint32_t num_node_ns;
+  char * param_ns;
+  uint32_t num_param_ns;
+} name_space_tracker_t;
+
+#define MAX_STRING_SZ 128U
```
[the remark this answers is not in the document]
anup-pem: Done!

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1336 @@
...
+    case DATA_TYPE_BOOL:
+      if (false == is_seq) {
+        param_value->bool_value = (bool *)ret_val;
+      } else {
+        if (DATA_TYPE_UNKNOWN == *seq_data_type) {
+          *seq_data_type = val_type;
+          param_value->bool_array_value =
+            allocator.zero_allocate(1U, sizeof(rcl_bool_array_t), NULL);
+          if (NULL == param_value->bool_array_value) {
+            RCL_SET_ERROR_MSG("Error allocating memory", allocator);
```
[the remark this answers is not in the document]
anup-pem: Good point! Done!

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1336 @@
...
+            return RCL_RET_BAD_ALLOC;
+          }
+        } else {
+          if (*seq_data_type != val_type) {
+            RCL_SET_ERROR_MSG_WITH_FORMAT_STRING(allocator, "Sequence should be of same"
+              " type. Value type bool do not belong at line_num %d", line_num);
```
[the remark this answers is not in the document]
anup-pem: Done!

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1336 @@
...
+      }
+      break;
+    case DATA_TYPE_STRING:
+      if (false == is_seq) {
+        param_value->string_value = (char *)ret_val;
+      } else {
+        if (DATA_TYPE_UNKNOWN == *seq_data_type) {
+          *seq_data_type = val_type;
+          param_value->string_array_value =
+            allocator.zero_allocate(1U, sizeof(rcutils_string_array_t), NULL);
```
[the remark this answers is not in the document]
anup-pem: Would prefer to keep as is because same way of allocation is done for other value types. It gives more clarity.

anup-pem:
> Yeah that looks good, I think we should also expose the initialization function as well as I expect people to:

> allocate the structure
call rcl_parse_yaml_file
call rcl_yaml_node_struct_fini when they're done with the struct
So I think we chould expose the function to do 1.
Otherwise it's unclear how people should do 1.

Agree. Fixed it! Added a new api `rcl_yaml_node_struct_init()`

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1362 @@
...
+              if (NULL != param_var->integer_array_value->values) {
+                allocator.deallocate(param_var->integer_array_value->values, NULL);
+              }
+              allocator.deallocate(param_var->integer_array_value, NULL);
+            } else if (NULL != param_var->double_array_value) {
+              if (NULL != param_var->double_array_value->values) {
+                allocator.deallocate(param_var->double_array_value->values, NULL);
+              }
+              allocator.deallocate(param_var->double_array_value, NULL);
+            } else if (NULL != param_var->string_array_value) {
```
sloretz: Missing dealloation of `param_var->string_array_value->data`. Recommend using `rcutils_string_array_fini()` instead.

sloretz: 337a755

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1362 @@
...
+
+  if ((NULL == parser) || (NULL == params_st)) {
+    return RCL_RET_INVALID_ARGUMENT;
+  }
+
+  while (0 == done_parsing) {
+    if (RCL_RET_OK != res) {
+      return res;
+    }
+    res = yaml_parser_parse(parser, &event);
```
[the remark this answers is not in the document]
sloretz: 20fa3a8

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1362 @@
...
+    return false;
+  }
+
+  res = yaml_parser_initialize(&parser);
+  if (0 == res) {
+    RCL_SET_ERROR_MSG("Could not initialize the parser", allocator);
+    return false;
+  }
+
+  yaml_file = fopen(file_path, "r");
```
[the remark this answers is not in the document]
sloretz: 953dc08

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1362 @@
...
+    RCL_SET_ERROR_MSG("YAML file path is NULL", allocator);
+    return false;
+  }
+
+  if (NULL == params_st) {
+    RCL_SET_ERROR_MSG("Pass a initialized paramter structure", allocator);
+    return false;
+  }
+
+  res = yaml_parser_initialize(&parser);
```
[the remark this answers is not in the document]
sloretz: 01e071e

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1362 @@
...
+          next_str = (next_str + strlen(sep_str));
+          next_str = strstr(next_str, sep_str);
+        }
+        if (NULL != last_idx) {
+          tot_len = (size_t)(last_idx - cur_ns);
+          cur_ns = allocator.reallocate(cur_ns, tot_len, NULL);
+          if (NULL == cur_ns) {
+            return RCL_RET_BAD_ALLOC;
+          }
+          cur_ns[tot_len] = '\0';
```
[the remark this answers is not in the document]
sloretz: 3a12493

rcl_yaml_param_parser/test/test_parse_yaml.cpp
```diff
@@ -0,0 +1,160 @@
...
+  bool res = rcl_parse_yaml_file(path, params_hdl, allocator);
+  fprintf(stderr, "%s\n", rcutils_get_error_string_safe());
+  EXPECT_FALSE(res);
+}
+
+TEST(test_file_parser, indented_ns) {
+  rcutils_reset_error();
+  EXPECT_TRUE(rcutils_get_cwd(cur_dir, 1024));
+  char * path = rcutils_join_path(cur_dir, "test");
+  path = rcutils_join_path(path, "indented_name_space.yaml");
```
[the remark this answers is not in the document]
sloretz: 33bffcd

rcl_yaml_param_parser/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+    test/test_parse_yaml.cpp
+  )
+  if(TARGET test_parse_yaml)
+    target_link_libraries(test_parse_yaml ${PROJECT_NAME})
+  endif()
+
+endif()
+
+ament_export_dependencies(ament_cmake)
+ament_export_include_directories(include)
```
sloretz: This line is good. Additionally the headers need to be installed.

```
install(
  DIRECTORY include/
  DESTINATION include
)
```

sloretz: 68724a860c0a18d1b76cd984d6294964d4ece1f0

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1362 @@
...
+///
+static rcl_ret_t parse_events(
+  yaml_parser_t * parser,
+  namespace_tracker_t * ns_tracker,
+  rcl_params_t * params_st,
+  const rcl_allocator_t allocator)
+{
+  int32_t done_parsing = 0;
+  yaml_event_t event;
+  int32_t res = RCL_RET_OK;
```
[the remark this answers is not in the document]
anup-pem: Fixed it!

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1362 @@
...
+
+    res = rcutils_string_array_init(val_array, 1, &allocator);
+    if (RCL_RET_OK != res) {
+      return res;
+    }
+    val_array->data[0U] = value;
+  } else {
+    /// Increase the array size by one and add the new value
+    char ** tmp_arr = val_array->data;
+    val_array->data = allocator.zero_allocate(val_array->size + 1U, sizeof(char *), NULL);
```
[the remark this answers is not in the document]
anup-pem: Agree!

rcl_yaml_param_parser/CMakeLists.txt
anup-pem: Fixed!

mikaelarguedas:
New round of CI on the current state of this and https://github.com/ros2/libyaml_vendor/pull/1

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4475/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1406/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3658/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4550/)

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1384 @@
...
+    if (0U == val_size) {
+      RCL_SET_ERROR_MSG_WITH_FORMAT_STRING(allocator, "No key at line %d", line_num);
+      return RCL_RET_ERROR;
+    }
+  }
+
+  if (NULL == value) {
+    return RCL_RET_INVALID_ARGUMENT;
+  }
+  num_nodes = params_st->num_nodes;  // New node index
```
sloretz: On windows I see a warning about a possible loss of data here because of the conversion from size_t to `uint32_t`

sloretz:
One more CI run now that #241 was merged
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4478/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1409/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3662/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4554/)

rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
```diff
@@ -0,0 +1,64 @@
...
+rcl_params_t * rcl_yaml_node_struct_init(
+  const rcutils_allocator_t allocator);
+
+/// \brief Free param structure
+/// \param[in] params_st points to the populated paramter struct
+/// \param[in] allocator memory allocator to be used
+RCL_YAML_PARAM_PARSER_PUBLIC
+void rcl_yaml_node_struct_fini(
+  rcl_params_t * params_st,
+  const rcutils_allocator_t allocator);
```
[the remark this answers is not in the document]
anup-pem: Agree. I think we can move the allocator field in rcl_variant_t into rcl_params_t and initialize it in rcl_yaml_node_struct_init(). Then remove "allocator" variable from other public API's except rcl_yaml_node_struct_init() and use the internal one. If you already have a fix can go ahead with it. If not I can make that change.

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1384 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <errno.h>
+#include <inttypes.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
```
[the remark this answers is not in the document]
anup-pem: I used it for some print formatting stuff but looks like it may not be needed.

rcl_yaml_param_parser/src/parser.c
anup-pem: Make sense

rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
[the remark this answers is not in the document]
sloretz: Sounds perfect. I don't have the fix already.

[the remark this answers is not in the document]
anup-pem: Thinking about it, because of the user specific state parameter in allocators, just passing the allocator in init() and cahing it in rcl_params_t and not able to pass the allocator in the fini() and other API's might defeat the purpose of the state field in the allocator. To overcome that, we can store the allocator in rcl_params_t, and let the user pass the allocator in other API's like fini and parse_yaml_file as it is currently implemented.. But, have to check if the incoming allocator is same as the stored allocator except the state value. Does it make sense?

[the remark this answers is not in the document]
sloretz: > Thinking about it, because of the user specific state parameter in allocators, just passing the allocator in init() and cahing it in rcl_params_t and not able to pass the allocator in the fini() and other API's might defeat the purpose of the state field in the allocator.

I think the expectation is that it's the allocator's state rather than a user's state. For example, someone might implement an allocator that only uses big chunk of memory at the start of the program. Calling `allocator.allocate(..., allocator.state)` would make the allocator record how much of that initial pool it's used. Since  `allocator.state` is a `void *` the state pointer itself must be the same when init, parse, and fini are called.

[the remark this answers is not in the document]
anup-pem: OK, so are you suggesting, allow allocator to be passed in init, parse and fini and  check all the fields of the allocator in the beginning of the API of parse and fini to that  stored in rcl_params_t? If so, is there a API in rcutils or rcl which checks if two allocators are same or should I just implement one in parser.c to compare two allocators. I could not find a API in rcutils to compare  two allocators.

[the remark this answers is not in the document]
sloretz: > so are you suggesting, allow allocator to be passed in init, parse and fini and check all the fields of the allocator in the beginning of the API of parse and fini to that stored in rcl_params_t?

I recommend only allowing the allocator to be passed into init.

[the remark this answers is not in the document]
anup-pem: Done!

sloretz:
CI

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4479/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1410/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3663/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4555/)

mikaelarguedas:
@sloretz If this now looks good to you,
Do you mind reviewing https://github.com/ros2/libyaml_vendor/pull/1 when you get a chance so that we can go ahead and merge the first version of this?

Thanks!

mikaelarguedas:
I opened a few follow-up tickets. @sloretz @anup-pem feel free to open tickets for the things I missed. Thanks!

rcl_yaml_param_parser/src/parser.c
```diff
@@ -0,0 +1,1384 @@
...
+        if (DATA_TYPE_UNKNOWN == *seq_data_type) {
+          *seq_data_type = val_type;
+          param_value->string_array_value =
+            allocator.zero_allocate(1U, sizeof(rcutils_string_array_t), allocator.state);
+          if (NULL == param_value->string_array_value) {
+            RCUTILS_SAFE_FWRITE_TO_STDERR("Error allocating mem");
+            return RCL_RET_BAD_ALLOC;
+          }
+        } else {
+          if (*seq_data_type != val_type) {
```
dirk-thomas: This line resulted in a new compiler warning in [this job](https://ci.ros2.org/view/packaging/job/packaging_linux-aarch64/451/warnings23Result/new/). Please address the new warning as soon as feasible. Thanks.

sloretz: Looking at it now

sloretz: #245

mikaelarguedas:
>  I have few more follow up tickets and will fix them in the next phase

@anup-pem Can you please open issues detailing the work to be done in these tickets ? This will allow us to be sure not to lose track of them as well as allowing other contributors to be aware of them / pick them up.

Thanks!

anup-pem:
@mikaelarguedas Have created new follow up tickets

rcl_yaml_param_parser/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+
+if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
+include_directories(include ${yaml_INCLUDE_DIRS} ${rcutils_INCLUDE_DIRS} ${rcl_INCLUDE_DIRS})
```
dirk-thomas, line 20: This variable `rcl_INCLUDE_DIRS` isn't defined. Removed in #620.

PR #237, opened 2018-05-15
[rcl_yaml_param_parser] fix export symbols

mikaelarguedas:
The first commit adds visibility macros to the public symbole of the library
The second commit remove the unused cmake macro as it was missed in the previous round of fixups

### DIFF
--- a/rcl_yaml_param_parser/CMakeLists.txt
+++ b/rcl_yaml_param_parser/CMakeLists.txt
@@ -28,6 +28,10 @@ add_library(
   ${rcl_yaml_parser_sources})
 ament_target_dependencies(${PROJECT_NAME} "yaml" "rcutils" "rcl")
 
+# Causes the visibility macros to use dllexport rather than dllimport,
+# which is appropriate when building the dll but not consuming it.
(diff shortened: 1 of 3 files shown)


rcl_yaml_param_parser/CMakeLists.txt
```diff
@@ -28,6 +28,10 @@ add_library(
   ${rcl_yaml_parser_sources})
 ament_target_dependencies(${PROJECT_NAME} "yaml" "rcutils" "rcl")
 
+# Causes the visibility macros to use dllexport rather than dllimport,
+# which is appropriate when building the dll but not consuming it.
+target_compile_definitions(${PROJECT_NAME} PRIVATE "RCL_YAML_PARAM_PARSER_BUILDING_DLL")
```
anup-pem: Should we put #if (WIN32) around this? I understand it may not be required as it is just defined in WIN32 in visibility_control.h but was curious.

mikaelarguedas: This will have no effect on the other platforms.
I didnt add the extra logic to match what is done in the other packages: e.g. https://github.com/ros2/rcl/blob/e8491ab8abba5c9653fdfc46064cc3b0d5ee18b9/rcl/CMakeLists.txt#L61-L63

PR #238, opened 2018-05-22
update to use new memory_tools from osrf_testing_tools_cpp

wjwwood:
Connects to ros2/rcutils#101


rcl/test/rcl/test_wait.cpp
```diff
@@ -143,15 +142,14 @@ TEST_F(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), zero_timeout) {
...
-    });
+  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT({
+    rcl_ret_t ret = rcl_guard_condition_fini(&guard_cond);
+    EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
+    ret = rcl_wait_set_fini(&wait_set);
+    EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
+  });
 
-  // Time spent during wait should be negligible.
+// Time spent during wait should be negligible.
```
dhood: these indents looked fine before, did uncrustify complain?

rcl/test/rcl/test_wait.cpp
```diff
@@ -265,11 +261,12 @@ TEST_F(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), multi_wait_set_threade
     size_t thread_id;
   };
   std::vector<TestSet> test_sets(number_of_threads);
+  // *INDENT-OFF* (prevent uncrustify from making unnecessary whitespace around [=])
   // Setup common function for waiting on the triggered guard conditions.
   auto wait_func_factory =
-    [count_target, retry_limit, wait_period](TestSet & test_set) {
+    [=](TestSet & test_set) {
```
dhood: was this change required with this PR?

rcl/test/rcl/test_rcl.cpp
```diff
@@ -117,9 +117,9 @@ TEST_F(CLASSNAME(TestRCLFixture, RMW_IMPLEMENTATION), test_rcl_init_and_ok_and_s
   {
     FakeTestArgv test_args;
     rcl_allocator_t failing_allocator = rcl_get_default_allocator();
-    failing_allocator.allocate = &failing_malloc;
-    failing_allocator.deallocate = failing_free;
```
dhood: what's the motivation for not having a failing `deallocate` anymore? and for only setting `failing_calloc` in some test files (accident?)?

rcl/test/rcl/test_timer.cpp
```diff
@@ -39,21 +36,10 @@ class TestTimerFixture : public ::testing::Test
     rcl_node_options_t node_options = rcl_node_get_default_options();
     ret = rcl_node_init(this->node_ptr, name, "", &node_options);
     ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
-    set_on_unexpected_malloc_callback([]() {ASSERT_FALSE(true) << "UNEXPECTED MALLOC";});
-    set_on_unexpected_realloc_callback([]() {ASSERT_FALSE(true) << "UNEXPECTED REALLOC";});
-    set_on_unexpected_free_callback([]() {ASSERT_FALSE(true) << "UNEXPECTED FREE";});
-    start_memory_checking();
```
dhood: Was the motivation for removing the use of memory tools from some tests because that info's captured elsewhere? Or something else?

rcl/test/rcl/test_wait.cpp
wjwwood: Yes without this, I get warnings on macOS about unnecessary captures, but if I remove them and do not replace it with `=` then it fails to compile on Windows. Something called `Wunused-lambda-capture`, though it might be a bug in clang, see:

https://stackoverflow.com/questions/41960347/why-am-i-getting-an-unused-lambda-capture-warning

But this works just as well and produces no warnings or errors on any platform.

rcl/test/rcl/test_rcl.cpp
wjwwood: Setting failing_calloc only in some files is a bug, I'll double check for those, but free doesn't return anything so it never fails. I suppose it could abort with double free, but it's not something we can easily test for nor would it likely increase our test coverage as no code in the functions we're testing would try to handle that case.

rcl/test/rcl/test_timer.cpp
wjwwood: Most of the time the check was meaningless or memory tools wasn't being used at all (it was setup, but never asserted that memory functions should not be called).

rcl/test/rcl/test_wait.cpp
wjwwood: Apparently just a typo, fixed it.

rcl/test/rcl/test_rcl.cpp
wjwwood: Some cases (like this one) don't need failing calloc.

wjwwood:
I think I addressed the comments (I didn't see any cases where calloc needed to be added, but I can add them later if identified). Since the changes are whitespace, I won't rerun CI.

PR #245, opened 2018-05-30
Fix maybe-uninitialized warning

sloretz:
This suppresses an occurrance of a `maybe-uninitialized` warning on gcc 4.x. I don't see a warning using gcc 5.4

It seems to be a false positive. Searching for false positive `maybe-uninitialized` gcc bugs returns [too many to reasonably search for a match](https://gcc.gnu.org/bugzilla/buglist.cgi?quicksearch=maybe-uninitialized). It's strange to me that the warning is not also present on lines 941, 921, 883, etc.

[Build Status](https://ci.ros2.org/view/packaging/job/packaging_linux-aarch64/452/)

In progress while CI runs

### DIFF
--- a/rcl_yaml_param_parser/src/parser.c
+++ b/rcl_yaml_param_parser/src/parser.c
@@ -1163,7 +1163,7 @@ static rcl_ret_t parse_events(
   bool is_key = true;
   bool is_seq = false;
   uint32_t line_num = 0;
-  data_types_t seq_data_type;
+  data_types_t seq_data_type = DATA_TYPE_UNKNOWN;
   uint32_t map_level = 1U;
   uint32_t map_depth = 0U;
   bool is_new_map = false;


rcl_yaml_param_parser/src/parser.c
```diff
@@ -967,7 +967,10 @@ static rcl_ret_t parse_value(
             return RCL_RET_BAD_ALLOC;
           }
         } else {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
```
wjwwood: This needs to be conditionally done, as it will not work on Windows. For example: https://github.com/ros2/rviz/blob/f594d9f00ecf87f75a0a3909f538bd463f7b7375/rviz_rendering/include/rviz_rendering/objects/point_cloud.hpp#L40-L63

mikaelarguedas:
> This suppresses an occurrance of a maybe-uninitialized warning on gcc 4.x

We target only platforms using gcc 5.4 at the moment.
I can reproduce the warning locally when building with the build type RelWithDebInfo, the same warning appears with gcc7.3.

> It's strange to me that the warning is not also present on lines 941, 921, 883, etc.

Actually it seem to show up only on the last comparison of `seq_data_type`. When commenting out the `case DATA_TYPE_STRING` bloc, the warning then appears line 941. so it may be more robust to ignore the warning in the multiple places this comparison happen or to spend more time investigating how to fix it without ignoring it.

Regarding the patch, this should be applied only if  _WIN32 is not defined like we do [in other places](https://github.com/ros2/rosidl/blob/ca8c8359a529a71a48c320d43b854e5205915c6f/rosidl_generator_cpp/test/test_msg_initialization.cpp#L257-L267)

CI without and with this patch:

Without: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/1417/)
With: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/1416/)

sloretz:
Thanks for the build type @mikaelarguedas. Fixed warning rather than suppressed it in e4d68e4

New CI (build up to and test only `rcl_yaml_param_parser` since this is a leaf package)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4495/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1421/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3684/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4573/)

mikaelarguedas:
thanks @sloretz, this is a better resolution indeed :+1:

Can you please run CI using the `RelWithDebInfo` build type instead of `None`?

sloretz:
Ooops. CI RelWithDebInfo

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4496/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1422/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3685/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4574/)

PR #251, opened 2018-06-05
Add failing test quoted number in string array

sloretz:
This adds a failing test and isn't intended to be merged.

Is there is another way to specify string array values in yaml? If a string array contains a quoted number then the parser complains about a type change.

```
1: [ RUN      ] test_file_parser.string_array_with_quoted_number
1: cur_path: /workspace/ros2-dev0/src/ros2/rcl/rcl_yaml_param_parser/test/string_array_with_quoted_number.yaml
1: Sequence should be of same type. Value type 'integer' do not belong at line_num 4, at /workspace/ros2-dev0/src/ros2/rcl/
rcl_yaml_param_parser/src/parser.c:914
1: /workspace/ros2-dev0/src/ros2/rcl/rcl_yaml_param_parser/test/test_parse_yaml.cpp:55: Failure
1: Value of: res
1:   Actual: false
1: Expected: true
```

### DIFF
--- a/rcl_yaml_param_parser/test/string_array_with_quoted_number.yaml
+++ b/rcl_yaml_param_parser/test/string_array_with_quoted_number.yaml
@@ -1,0 +1,4 @@
+initial_params_node:
+    ros__parameters:
+        sa1: ["Four", "score"]
+        sa2: ["and", "7"]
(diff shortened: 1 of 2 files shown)


sloretz:
Closing since it was replaced by #464 which has been merged :tada:

PR #261, opened 2018-06-15
changes to support memory audit

wjwwood:
This pull request makes sure that we're using `rcutils_allocator_t`, or the alias `rcl_allocator_t`, everywhere and not using `malloc`, `realloc`, `calloc`, or `free` anywhere at all.

Connects to ros2/rcutils#102

### DIFF
--- a/rcl/test/rcl/test_time.cpp
+++ b/rcl/test/rcl/test_time.cpp
@@ -193,14 +193,14 @@ TEST(CLASSNAME(rcl_time, RMW_IMPLEMENTATION), default_clock_instanciation) {
   EXPECT_EQ(retval, RCL_RET_OK) << rcl_get_error_string_safe();
   ASSERT_TRUE(rcl_clock_valid(&ros_clock));
 
-  rcl_clock_t * steady_clock =
-    reinterpret_cast<rcl_clock_t *>(calloc(1, sizeof(rcl_clock_t)));
+  rcl_clock_t * steady_clock = reinterpret_cast<rcl_clock_t *>(
+    allocator.zero_allocate(1, sizeof(rcl_clock_t), allocator.state));
   retval = rcl_steady_clock_init(steady_clock, &allocator);
   EXPECT_EQ(retval, RCL_RET_OK) << rcl_get_error_string_safe();
   ASSERT_TRUE(rcl_clock_valid(steady_clock));
 
-  rcl_clock_t * system_clock =
-    reinterpret_cast<rcl_clock_t *>(calloc(1, sizeof(rcl_clock_t)));
+  rcl_clock_t * system_clock = reinterpret_cast<rcl_clock_t *>(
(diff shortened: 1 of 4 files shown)


rcl/test/rcl/test_time.cpp
```diff
@@ -193,14 +193,14 @@ TEST(CLASSNAME(rcl_time, RMW_IMPLEMENTATION), default_clock_instanciation) {
   EXPECT_EQ(retval, RCL_RET_OK) << rcl_get_error_string_safe();
   ASSERT_TRUE(rcl_clock_valid(&ros_clock));
 
-  rcl_clock_t * steady_clock =
-    reinterpret_cast<rcl_clock_t *>(calloc(1, sizeof(rcl_clock_t)));
+  rcl_clock_t * steady_clock = reinterpret_cast<rcl_clock_t *>(
```
ubercool2: Why is this reinterpret_cast needed here? MISRA C++ guide on this is:
Approval required. Rationale: Reinterpreting bits into a different type should not be 
necessary in a high level programming language. Exceptions could be made when a wrapper 
to a third party library working with opaque data types needs to be created.

Also SEI Cert C++ is not fond of it: https://wiki.sei.cmu.edu/confluence/display/cplusplus/VOID+EXP13-CPP.+Prefer+dynamic_cast+over+static_cast+over+reinterpret_cast.

rcl/test/rcl/test_time.cpp
```diff
@@ -193,14 +193,14 @@ TEST(CLASSNAME(rcl_time, RMW_IMPLEMENTATION), default_clock_instanciation) {
...
+    allocator.zero_allocate(1, sizeof(rcl_clock_t), allocator.state));
```
dejanpan: Is there no way around `reinterpret_cast` here? MISRA C++ guide on this is: Approval required. Rationale: Reinterpreting bits into a different type should not be necessary in a high level programming language. Exceptions could be made when a wrapper to a third party library working with opaque data types needs to be created.

Also SEI Cert C++ is not fond of it: https://wiki.sei.cmu.edu/confluence/display/cplusplus/VOID+EXP13-CPP.+Prefer+dynamic_cast+over+static_cast+over+reinterpret_cast.

wjwwood: No, I believe it is required and I don't know any other way to do this when calling a C function. Malloc returns `void *` which must be cast to the pointer type you're using it as.

wjwwood: For example they use the C-style cast in the example:

http://en.cppreference.com/w/cpp/memory/c/malloc

But we have a rule to not use the C-style cast when we can use `reinterpret_cast` or `static_cast` instead.

dirk-thomas: Casting from a `void*` should be done with s `static_cast`.

PR #272, opened 2018-07-23
initialize timer with clock

dirk-thomas:
Required for ros2/rclcpp#523.

### DIFF
--- a/rcl/include/rcl/timer.h
+++ b/rcl/include/rcl/timer.h
@@ -137,6 +137,7 @@ RCL_WARN_UNUSED
 rcl_ret_t
 rcl_timer_init(
   rcl_timer_t * timer,
+  rcl_clock_t * clock,
   int64_t period,
   const rcl_timer_callback_t callback,
   rcl_allocator_t allocator);
--- a/rcl/src/rcl/timer.c
+++ b/rcl/src/rcl/timer.c
@@ -28,6 +28,8 @@ extern "C"
 
 typedef struct rcl_timer_impl_t
 {
+  // The clock providing time.
+  rcl_clock_t * clock;
   // The user supplied callback.
   atomic_uintptr_t callback;
   // This is a duration in nanoseconds.
@@ -50,26 +52,29 @@ rcl_get_zero_initialized_timer()
 rcl_ret_t
 rcl_timer_init(
   rcl_timer_t * timer,
+  rcl_clock_t * clock,
   int64_t period,
   const rcl_timer_callback_t callback,
   rcl_allocator_t allocator)
 {
   RCL_CHECK_ALLOCATOR_WITH_MSG(&allocator, "invalid allocator", return RCL_RET_INVALID_ARGUMENT);
   RCL_CHECK_ARGUMENT_FOR_NULL(timer, RCL_RET_INVALID_ARGUMENT, allocator);
+  RCL_CHECK_ARGUMENT_FOR_NULL(clock, RCL_RET_INVALID_ARGUMENT, allocator);
   RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME, "Initializing timer with period: %" PRIu64 "ns", period)
   if (timer->impl) {
     RCL_SET_ERROR_MSG("timer already initailized, or memory was uninitialized", allocator);
     return RCL_RET_ALREADY_INIT;
   }
-  rcl_time_point_value_t now_steady;
-  rcl_ret_t now_ret = rcutils_steady_time_now(&now_steady);
+  rcl_time_point_value_t now;
+  rcl_ret_t now_ret = clock->get_now(clock->data, &now);
   if (now_ret != RCL_RET_OK) {
     return now_ret;  // rcl error state should already be set.
   }
   rcl_timer_impl_t impl;
+  impl.clock = clock;
   atomic_init(&impl.callback, (uintptr_t)callback);
   atomic_init(&impl.period, period);
-  atomic_init(&impl.last_call_time, now_steady);
+  atomic_init(&impl.last_call_time, now);
   atomic_init(&impl.canceled, false);
   impl.allocator = allocator;
   timer->impl = (rcl_timer_impl_t *)allocator.allocate(sizeof(rcl_timer_impl_t), allocator.state);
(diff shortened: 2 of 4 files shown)


rcl/src/rcl/timer.c
```diff
@@ -50,26 +52,29 @@ rcl_get_zero_initialized_timer()
...
+  RCL_CHECK_ARGUMENT_FOR_NULL(clock, RCL_RET_INVALID_ARGUMENT, allocator);
   RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME, "Initializing timer with period: %" PRIu64 "ns", period)
   if (timer->impl) {
     RCL_SET_ERROR_MSG("timer already initailized, or memory was uninitialized", allocator);
     return RCL_RET_ALREADY_INIT;
   }
-  rcl_time_point_value_t now_steady;
-  rcl_ret_t now_ret = rcutils_steady_time_now(&now_steady);
+  rcl_time_point_value_t now;
+  rcl_ret_t now_ret = clock->get_now(clock->data, &now);
```
dhood: I think `rcl_clock_get_now` is the expected interface for getting a clock's time value

rcl/include/rcl/timer.h
```diff
@@ -137,6 +137,7 @@ RCL_WARN_UNUSED
 rcl_ret_t
 rcl_timer_init(
   rcl_timer_t * timer,
+  rcl_clock_t * clock,
```
dhood: docblock for this function needs to be updated

rcl/test/rcl/test_timer.cpp
```diff
@@ -50,13 +50,19 @@ class TestTimerFixture : public ::testing::Test
 
 TEST_F(TestTimerFixture, test_two_timers) {
   rcl_ret_t ret;
+
+  rcl_clock_t clock;
+  rcl_allocator_t allocator = rcl_get_default_allocator();
+  ret = rcl_clock_init(RCL_STEADY_TIME, &clock, &allocator);
```
sloretz: Recommend finishing tests with `rcl_*_clock_fini()` to reduce false-positive address sanitizer output.

rcl/src/rcl/timer.c
dirk-thomas: The commit e8f2e34b15951c86658a2ef3a0756059e3627f93 uses your proposed function.

It requires to get a `rcl_time_point_t` instead of a `rcl_time_point_value_t`. Since the code doesn't care / check the clock type of that time point wouldn't it be more efficient to just get the `rcl_time_point_value_t`? Not sure if that is a good idea...

rcl/test/rcl/test_timer.cpp
dirk-thomas: Absolutely - done in 0c972079e41d9a8730b1da4643aa0903dfaf2857.

rcl/include/rcl/timer.h
dirk-thomas: Absolutely, packages on-top of `rclcpp` also haven't been updated yet. Will put it back in review when that is the case and the docblocks are updated too.

rcl/src/rcl/timer.c
dirk-thomas: I raised the concern about the existing API in #273 in order to keep it separate from this patch.

rcl/include/rcl/timer.h
dirk-thomas: Updated the docblock in e1a400e6724de65cab6969c90fd798675b144fbb.

dirk-thomas:
I rolled the change related to #273 into this patch since both affect the same function signature: 28e6ba1d96fb78798d93719c02b78d49d797f36c

rcl/src/rcl/timer.c
dirk-thomas: I rolled the change related to #273 into this patch since both affect the same function signature: 28e6ba1d96fb78798d93719c02b78d49d797f36c

rcl/include/rcl/time.h
```diff
@@ -272,20 +272,20 @@ rcl_ret_t
...
+ * \param[out] time_point_value The time_point value to populate.
  * \return `RCL_RET_OK` if the last call time was retrieved successfully, or
  * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
  * \return `RCL_RET_ERROR` an unspecified error occur.
  */
 RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_ret_t
-rcl_clock_get_now(rcl_clock_t * clock, rcl_time_point_t * time_point);
+rcl_clock_get_now(rcl_clock_t * clock, rcl_time_point_value_t * time_point_value);
```
dhood: new functionality in rclpy is relying on the existing signature: https://github.com/ros2/rclpy/pull/209/files#diff-80dda03110b5be606b823d6b0558b5c2R3218 (CI wouldn't have picked up on that because the PR was merged after it). Sorry about that, I didn't realise that merge would cause conflict here (thought we were adding a new signature rather than modifying the existing one). I would make a PR for it but https://github.com/ros2/rclpy/pull/211 is already connected and needs to be rebased first

dirk-thomas: Thanks for pointing it out - no problem. I rebased the `rclpy` PR and updated it in https://github.com/ros2/rclpy/pull/211/commits/f938bac8f72d7ff0af79c0ba04abfd149ce68be6 to use the modified API.

rcl/test/rcl/test_wait.cpp
```diff
@@ -205,8 +210,14 @@ TEST_F(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), canceled_timer) {
   ret = rcl_wait_set_add_guard_condition(&wait_set, &guard_cond);
   EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
 
+  rcl_clock_t clock;
+  rcl_allocator_t allocator = rcl_get_default_allocator();
+  ret = rcl_clock_init(RCL_STEADY_TIME, &clock, &allocator);
```
sloretz: Missing `rcl_clock_fini()`, though I wouldn't block the PR for that in a test.

dirk-thomas: Added in 1c73377792db3b854a83584557cd92c53e9cdd07.

PR #276, opened 2018-08-09
Compile Flag Fixes

jpsamper2009:
## Description
- Courtesy of @serge-nikulin 
- This PR fixes warning generated when building with:

```
-Wall
-Wextra
-pedantic
-Wcast-align
-Wunused
-Wconversion
-Wsign-conversion
-Wdouble-promotion
-fvisibility=hidden
```

- Related to ros2/rclcpp#530 and ros2/rcutils#114

connects to ros2/rcutils#114

### DIFF
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -85,7 +85,7 @@ rcl_arguments_get_param_files(
   RCL_CHECK_ARGUMENT_FOR_NULL(arguments->impl, RCL_RET_INVALID_ARGUMENT, allocator);
   RCL_CHECK_ARGUMENT_FOR_NULL(parameter_files, RCL_RET_INVALID_ARGUMENT, allocator);
   *(parameter_files) = allocator.allocate(
-    sizeof(char *) * arguments->impl->num_param_files_args, allocator.state);
+    sizeof(char *) * (size_t)arguments->impl->num_param_files_args, allocator.state);
   if (NULL == *parameter_files) {
     return RCL_RET_BAD_ALLOC;
   }
@@ -175,17 +175,17 @@ rcl_parse_arguments(
   }
 
   // over-allocate arrays to match the number of arguments
-  args_impl->remap_rules = allocator.allocate(sizeof(rcl_remap_t) * argc, allocator.state);
+  args_impl->remap_rules = allocator.allocate(sizeof(rcl_remap_t) * (size_t)argc, allocator.state);
   if (NULL == args_impl->remap_rules) {
     ret = RCL_RET_BAD_ALLOC;
     goto fail;
   }
-  args_impl->unparsed_args = allocator.allocate(sizeof(int) * argc, allocator.state);
+  args_impl->unparsed_args = allocator.allocate(sizeof(int) * (size_t)argc, allocator.state);
   if (NULL == args_impl->unparsed_args) {
     ret = RCL_RET_BAD_ALLOC;
     goto fail;
   }
-  args_impl->parameter_files = allocator.allocate(sizeof(char *) * argc, allocator.state);
+  args_impl->parameter_files = allocator.allocate(sizeof(char *) * (size_t)argc, allocator.state);
   if (NULL == args_impl->parameter_files) {
     ret = RCL_RET_BAD_ALLOC;
     goto fail;
@@ -240,7 +240,7 @@ rcl_parse_arguments(
   // Shrink remap_rules array to match number of successfully parsed rules
   if (args_impl->num_remap_rules > 0) {
     args_impl->remap_rules = rcutils_reallocf(
-      args_impl->remap_rules, sizeof(rcl_remap_t) * args_impl->num_remap_rules, &allocator);
+      args_impl->remap_rules, sizeof(rcl_remap_t) * (size_t)args_impl->num_remap_rules, &allocator);
     if (NULL == args_impl->remap_rules) {
       ret = RCL_RET_BAD_ALLOC;
       goto fail;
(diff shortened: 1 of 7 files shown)


esteve:
I don't know why GitHub says I added the "in review" label, I don't recall doing that. Could it be something waffle did by itself? Anyway, I've removed the label from here and the other two pull requests.

mikaelarguedas:
Yes it looks like waffle picks your account sometimes to set labels.

All 3rd party contributions come `in review` by default to not get lost in the backlog. So I added back the label on all 3 PRs

Thanks for monitoring though !

rcl/src/rcl/arguments.c
```diff
@@ -85,7 +85,7 @@ rcl_arguments_get_param_files(
   RCL_CHECK_ARGUMENT_FOR_NULL(arguments->impl, RCL_RET_INVALID_ARGUMENT, allocator);
   RCL_CHECK_ARGUMENT_FOR_NULL(parameter_files, RCL_RET_INVALID_ARGUMENT, allocator);
   *(parameter_files) = allocator.allocate(
-    sizeof(char *) * arguments->impl->num_param_files_args, allocator.state);
+    sizeof(char *) * (size_t)arguments->impl->num_param_files_args, allocator.state);
```
clalancette: Hm, it looks like `num_params_files_args` (https://github.com/ros2/rcl/blob/master/rcl/src/rcl/arguments_impl.h#L38) is always used as an unsigned number (which makes sense, since it is the "number" of things).  Thus, I'd suggest that we actually change the type in the structure, rather than casting it here (in a quick test I did here, that seems to be OK).

rcl/src/rcl/arguments.c
```diff
@@ -240,7 +240,7 @@ rcl_parse_arguments(
   // Shrink remap_rules array to match number of successfully parsed rules
   if (args_impl->num_remap_rules > 0) {
     args_impl->remap_rules = rcutils_reallocf(
-      args_impl->remap_rules, sizeof(rcl_remap_t) * args_impl->num_remap_rules, &allocator);
+      args_impl->remap_rules, sizeof(rcl_remap_t) * (size_t)args_impl->num_remap_rules, &allocator);
```
clalancette: Similarly, it looks like `num_remap_rules` could probably be a `size_t` in the structure, rather than doing this cast.

rcl/src/rcl/arguments.c
```diff
@@ -312,7 +313,7 @@ rcl_arguments_get_unparsed(
   *output_unparsed_indices = NULL;
   if (args->impl->num_unparsed_args) {
     *output_unparsed_indices = allocator.allocate(
-      sizeof(int) * args->impl->num_unparsed_args, allocator.state);
+      sizeof(int) * (size_t)args->impl->num_unparsed_args, allocator.state);
```
clalancette: And the same for `num_unparsed_args`.

rcl/src/rcl/time.c
```diff
@@ -67,7 +67,7 @@ rcl_get_ros_time(void * data, rcl_time_point_value_t * current_time)
   if (!t->active) {
     return rcl_get_system_time(data, current_time);
   }
-  *current_time = rcl_atomic_load_uint64_t(&(t->current_time));
+  *current_time = (rcl_time_point_value_t)rcl_atomic_load_uint64_t(&(t->current_time));
```
clalancette: Hmpf, I can't say I love this.  If we look carefully at the types, `t->current_time` is of type `atomic_uint_least64_t` (https://github.com/ros2/rcl/blob/master/rcl/src/rcl/time.c#L30).  `rcl_atomic_load_uint64_t` expects a uint64 argument expects an `atomic_uint_least64_t` pointer as the input argument, so we are good so far.  Then `rcl_atomic_load_uint64_t` returns a `uint64_t` (https://github.com/ros2/rcl/blob/master/rcl/src/rcl/stdatomic_helper.h#L72), and here we get into trouble.  The next thing we are going to do is to store this into `*current_time` which is `rcl_time_point_value_t` -> `rcutils_time_point_value_t` -> `int64_t`.  So now we are storing an unsigned into a signed, potentially changing the interpretation.  I'm not sure how much it matters in this case since it is a 64-bit load, but this doesn't seem quite right.  @tfoote , any thoughts on this?

rcl/src/rcl/time.c
```diff
@@ -326,7 +326,7 @@ rcl_set_ros_time_override(
   if (storage->active && clock->pre_update) {
     clock->pre_update();
   }
-  rcl_atomic_store(&(storage->current_time), time_value);
+  rcl_atomic_store(&(storage->current_time), (uint64_t)time_value);
```
clalancette: And this is the same int64/uint64 mismatch.

rcl/src/rcl/timer.c
```diff
@@ -73,8 +73,8 @@ rcl_timer_init(
   rcl_timer_impl_t impl;
   impl.clock = clock;
   atomic_init(&impl.callback, (uintptr_t)callback);
-  atomic_init(&impl.period, period);
-  atomic_init(&impl.last_call_time, now);
+  atomic_init(&impl.period, (unint64_t)(period));
+  atomic_init(&impl.last_call_time, (unit64_t)(now));
```
clalancette: And again, these casts are needed because this is a type mismatch.  There may more down below, but I'm going to stop there.  In general I think it is a better idea to look at and think about why the cast is needed, and try to fix the root cause.  I know it is not always possible because some of the types are pre-defined for us, but for the types we do have control of it is better to use consistent types.

clalancette:
@jpsamper2009 Friendly ping.  Do you think you'll have time to pick this and https://github.com/ros2/rcutils/pull/114 back up?  Thanks.

jpsamper2009:
@clalancette This PR was a for a much older version of the code -- and it was done by me on behalf of someone else. I'll close it and readdress with a newer version (if the issues still exist).

PR #277, opened 2018-08-14
initialize memory before sending a message

dirk-thomas:
The memory of a C message must be initialized to avoid undefined behavior.

Before: [Build Status](https://ci.ros2.org/job/ci_linux/5041/)
After: [Build Status](https://ci.ros2.org/job/ci_linux/5043/)

### DIFF
--- a/rcl/test/rcl/client_fixture.cpp
+++ b/rcl/test/rcl/client_fixture.cpp
@@ -158,6 +158,7 @@ int main(int argc, char ** argv)
 
     // Initialize a request.
     test_msgs__srv__Primitives_Request client_request;
+    // TODO zero initialization necessary until https://github.com/ros2/ros2/issues/397 is implemented
     memset(&client_request, 0, sizeof(test_msgs__srv__Primitives_Request));
     test_msgs__srv__Primitives_Request__init(&client_request);
     client_request.uint8_value = 1;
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
If I understand correctly, this is a workaround pending https://github.com/ros2/ros2/issues/397 being implemented, is that right?

dhood:
Might it make sense to add the memset into the generated code for e.g. `test_msgs__srv__Primitives_Response__init`? (perhaps we have considered that already)

mikaelarguedas:
> Might it make sense to add the memset into the generated code for e.g. test_msgs__srv__Primitives_Response__init? (perhaps we have considered that already)

My understanding is that this is what https://github.com/ros2/ros2/issues/397 is supposed to cover (allocate and initialize all fields with their respective default)

rcl/test/rcl/client_fixture.cpp
```diff
@@ -158,6 +158,8 @@ int main(int argc, char ** argv)
 
     // Initialize a request.
     test_msgs__srv__Primitives_Request client_request;
+    // TODO zero initialization necessary until https://github.com/ros2/ros2/issues/397 is implemented
```
mikaelarguedas: arg this is both too long and missing a name attached to the TODO, I expect cpplint to not be happy :/

dirk-thomas: Fixed in e322f7daa8c2e74fc8402f45c1bc80aa790c3bd3: [Build Status](https://ci.ros2.org/job/ci_linux/5044/)

PR #280, opened 2018-08-19
Do not check environment variables on UWP

esteve:
UWP apps can't access environment variables, so this just skips that part.

### DIFF
--- a/rcl/CMakeLists.txt
+++ b/rcl/CMakeLists.txt
@@ -39,7 +39,6 @@ set(${PROJECT_NAME}_sources
   src/rcl/publisher.c
   src/rcl/rcl.c
   src/rcl/remap.c
-  src/rcl/rmw_implementation_identifier_check.c
   src/rcl/service.c
   src/rcl/subscription.c
   src/rcl/time.c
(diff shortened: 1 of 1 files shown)


wjwwood:
There are other places that `rcl` gets environment variables, don't you need to deal with those as well?

https://github.com/ros2/rcl/blob/master/rcl/src/rcl/node.c#L108

If there is more than one instance, I'd prefer to have a more generic "rcl supports environment variables" compiler option, which could be used in the files with `#if` directives. It could also be a CMake option with documentation and a default that's influenced by the target platform.

dirk-thomas:
@esteve Friendly ping.

esteve:
@wjwwood that's a good point, something like `-DROS2_SUPPORTS_ENV=OFF` that gets passed to CMake?

nuclearsandwich:
@esteve while I'm going around bumping other issues from this era. Is iterating on @wjwwood's comment something you'll be able to get to in the next couple of weeks?

esteve:
@nuclearsandwich @theseankelly has been lately porting this to Crystal. I haven't touched this code is a long time, @theseankelly is there any other place where this needs to be fixed to get ROS 2 running on the Hololens/UWP?

theseankelly:
@nuclearsandwich @esteve No I haven't encountered any other issues with environment variables thus far. Though, I do agree with sentiments from others that it might be a nice idea to handle environment limitations more broadly.

Without knowing most of the ROS2 code intimately, I would expect there could be some runtime implications of not having access to env variables? Or is this API flagged at compile time for UWP?

clalancette:
This has been open for a long time without any movement.  I'm going to go ahead and close it out, but do feel free to reopen if you are planning on working on it in the near future.

PR #299, opened 2018-09-15
[yaml parser] Fix FQN=//node_name when ns is /

dhood:
Consider the parameters file:

```
/:
    talker:
        ros__parameters:
            some_int: 42
```

Without this PR, the parameters are stored under the FQN of `//talker`, so passing the file to a node with `ros2 run demo_nodes_cpp talker __params:=demo_parameters.yaml` will not set the parameters on the node (`ros2 param list /talker` is empty).

With this PR, `ros2 param list /talker` returns the parameter `some_int` as expected.

---

Without this change, the parameter file will work if the root namespace is left out completely, but in some cases the user may want to specify explicitly that it is in the root namespace. (I am doing that in https://github.com/ros2/launch/pull/138/files#diff-d78062ac1f3a77c95190e470302b7e69R237 where I write the namespace to a file without considering if it is `/ns` or just `/`).

### DIFF
--- a/rcl_yaml_param_parser/src/parser.c
+++ b/rcl_yaml_param_parser/src/parser.c
@@ -187,11 +187,13 @@ static rcl_ret_t add_name_to_ns(
       ns_len = strlen(cur_ns);
       name_len = strlen(name);
       sep_len = strlen(sep_str);
-      if (ns_len == sep_len) {
-        // Current NS is only the separator: don't put another separator in.
+      // Check the last sep_len characters of the current NS against the separator string.
+      if (strcmp(cur_ns + ns_len - sep_len, sep_str) == 0) {
+        // Current NS already ends with the separator: don't put another separator in.
         sep_len = 0;
         sep_str = "";
       }
+
       tot_len = ns_len + sep_len + name_len + 1U;
 
       if (tot_len > MAX_STRING_SIZE) {
--- a/rcl_yaml_param_parser/src/parser.c
+++ b/rcl_yaml_param_parser/src/parser.c
@@ -187,7 +187,12 @@ static rcl_ret_t add_name_to_ns(
       ns_len = strlen(cur_ns);
       name_len = strlen(name);
       sep_len = strlen(sep_str);
-      tot_len = ns_len + name_len + sep_len + 1U;
+      if (ns_len == sep_len) {
+        // Current NS is only the separator: don't put another separator in.
+        sep_len = 0;
+        sep_str = "";
+      }
+      tot_len = ns_len + sep_len + name_len + 1U;
 
       if (tot_len > MAX_STRING_SIZE) {
         RCL_SET_ERROR_MSG("New namespace string is exceeding max string size",
--- a/rcl_yaml_param_parser/test/root_ns.yaml
+++ b/rcl_yaml_param_parser/test/root_ns.yaml
@@ -1,0 +1,7 @@
+# config/test_yaml
+---
+
+/:
+  my_node:
+    ros__parameters:
+      param_name: 'value'
--- a/rcl_yaml_param_parser/test/test_parse_yaml.cpp
+++ b/rcl_yaml_param_parser/test/test_parse_yaml.cpp
@@ -62,6 +62,31 @@ TEST(test_file_parser, multi_ns_correct_syntax) {
   allocator.deallocate(path, allocator.state);
 }
 
+TEST(test_file_parser, root_ns) {
+  rcutils_reset_error();
+  EXPECT_TRUE(rcutils_get_cwd(cur_dir, 1024));
+  rcutils_allocator_t allocator = rcutils_get_default_allocator();
+  char * test_path = rcutils_join_path(cur_dir, "test", allocator);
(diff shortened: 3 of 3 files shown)


dhood:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5195/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1961/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4315/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5175/)

rcl_yaml_param_parser/src/parser.c
```diff
@@ -187,7 +187,12 @@ static rcl_ret_t add_name_to_ns(
       ns_len = strlen(cur_ns);
       name_len = strlen(name);
       sep_len = strlen(sep_str);
-      tot_len = ns_len + name_len + sep_len + 1U;
+      if (ns_len == sep_len) {
```
dhood: self review: would be better to check the contents of the current ns, and skip appending if it ends with the separator already

dhood:
(newer CI):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5213/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1976/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4334/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5192/)

dhood:
looking for a review if anyone gets a chance

rcl_yaml_param_parser/src/parser.c
```diff
@@ -187,7 +187,14 @@ static rcl_ret_t add_name_to_ns(
       ns_len = strlen(cur_ns);
       name_len = strlen(name);
       sep_len = strlen(sep_str);
-      tot_len = ns_len + name_len + sep_len + 1U;
+      // Check the last sep_len characters of the current NS against the separator string.
+      if (strcmp(cur_ns + ns_len - sep_len, sep_str) == 0) {
+        // Current NS already ends with the separator: don't put another separator in.
```
sloretz: Looking at the tests I think the intent was that each token in the namespace would be it's own level in the config. Namespace `/foo/bar/baz` and node name `smith` would be described by:

```
foo:
  bar:
    baz:
      smith:
        ros__parameters:
```

I'm guessing it was not expected that the names could contain `/` and `.`. Allowing them does look more convenient though. To support that there is at least one more bug ros2/rcl#301.

dhood:
From what I've seen, the tests only check that files are parsed. I didn't add a test since the files are going to be parsed fine in either case, they just won't apply to the node correctly. I'll take another look if any tests check what would be passed to rcl.

mikaelarguedas:
The [design document](http://design.ros2.org/articles/topic_and_service_names.html#namespaces) seems to say that a name/namespace should not end with a slash. The implementation does assume that no namespace entry means root namespace. I'm not convinced specifying "/" as a namespace should be supported (at least in the current specification)

dhood:
can you clarify where you got that a _namespace_ shouldn't end with a slash? I see it for a topic name. I would expect that it's reasonable for someone to request that a node be put in the `/` namespace on the command line, which is why I think it should also be do-able from the yaml file

sloretz:
> The design document seems to say that a name/namespace should not end with a slash.

It does refer to `/` as a legal namespace later on

```
 In another example, the name /foo splits into one token, such that it is composed of a topic or service with the base name foo in the / namespace (the root namespace).
```

I think remapping allows `/` too, ex `__ns:=/`

rcl_yaml_param_parser/src/parser.c
dhood: by the way, if we're not sure if we want to support the case where there are trailing slashes _in general_ we can modify the change in this commit d311e0c55294bc4a7b1e2e8e621b2d0e886a5a78 and go back to just making the exception for the root namespace (maybe error in other cases).

mikaelarguedas:
My understanding was that this section was referring to namespaces (otherwise it would just be a duplicated version of the constraints on the topic names from the beginning of the document with a lot of constraints omitted)

```
Namespaces
Topic and service names:

may be split into tokens using a forward slash (/) as a delimiter

see the “Name Tokens” section for more details on tokens
must not end with a forward slash (/)
```

If not I think we should update that document so that the namespace section defines the constraints of a namespace and add to the list of valid examples the single `/` namespace as it currently only shows that namespace "None" would expand to `/`.

Not saying that it shouldn't be possible to pass the `/` namespace to nodes (via commandline or file) but that the design doc was not clear to me.

dhood:
regression test added in https://github.com/ros2/rcl/pull/299/commits/27fe5acc576c292b0ae69e4c7ab245a93111e7bf

dhood:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5227/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1988/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4346/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5205/)

PR #300, opened 2018-09-19
Expand node_secure_root using local_namespace

ruffsl:
Relates to: https://github.com/ros2/sros2/issues/68#issuecomment-422928655

### DIFF
--- a/rcl/src/rcl/node.c
+++ b/rcl/src/rcl/node.c
@@ -138,7 +138,7 @@ const char * rcl_get_secure_root(
     // Combine node namespace with node name
     node_fqn = rcutils_format_string(*allocator, "%s%s%s", node_namespace, "/", node_name);
     // Get normalized path, ignore the leading forward slash.
-    node_root_norm = rcutils_norm_path(node_fqn + 1, *allocator);
+    node_root_norm = rcutils_to_native_path(node_fqn + 1, *allocator);
     node_secure_root = rcutils_join_path(ros_secure_root_env, node_root_norm, *allocator);
     allocator->deallocate(node_fqn, allocator->state);
     allocator->deallocate(node_root_norm, allocator->state);
--- a/rcl/src/rcl/node.c
+++ b/rcl/src/rcl/node.c
@@ -129,6 +129,8 @@ const char * rcl_get_secure_root(
     return NULL;  // environment variable was empty
   }
   char * node_secure_root = NULL;
+  // TODO(ros2team): This make assumption on the value and length of the root namespace.
+  // This should likely come from another (rcl/rmw?) function for reuse.
   // If the namespace is the root namespace ("/"), the secure root is just the node name.
   if (strlen(node_namespace) == 1) {
     node_secure_root = rcutils_join_path(ros_secure_root_env, node_name, *allocator);
@@ -136,11 +138,10 @@ const char * rcl_get_secure_root(
     char * node_fqn = NULL;
     char * node_root_path = NULL;
     // Combine node namespace with node name
-    // TODO(ros2team): This make assumption on the value of the root namespace.
-    // This should likely come from another (rcl/rmw?) function for reuse.
+    // TODO(ros2team): remove the hard-coded value of the root namespace.
     node_fqn = rcutils_format_string(*allocator, "%s%s%s", node_namespace, "/", node_name);
     // Get native path, ignore the leading forward slash.
-    // TODO(ros2team): remove the hard-coded length, use the length of the root namespace instead
+    // TODO(ros2team): remove the hard-coded length, use the length of the root namespace instead.
     node_root_path = rcutils_to_native_path(node_fqn + 1, *allocator);
     node_secure_root = rcutils_join_path(ros_secure_root_env, node_root_path, *allocator);
     allocator->deallocate(node_fqn, allocator->state);
--- a/rcl/src/rcl/node.c
+++ b/rcl/src/rcl/node.c
@@ -99,7 +99,7 @@ const char * rcl_create_node_logger_name(
   return node_logger_name;
 }
 
-/// Return the secure root directory associated with a node given the validated node name and namespace.
+/// Return the secure root directory associated with a node given its validated name and namespace.
 /**
  * E.g. for a node named "c" in namespace "/a/b", the secure root path will be
  * "a/b/c", where the delimiter "/" is native for target file system (e.g. "\\" for _WIN32).
@@ -107,7 +107,7 @@ const char * rcl_create_node_logger_name(
  * \param[in] node_name validated node name (a single token)
  * \param[in] node_namespace validated, absolute namespace (starting with "/")
  * \param[in] allocator the allocator to use for allocation
- * \returns machine specific (absolute) node secure root path or null if there is an error
+ * \returns machine specific (absolute) node secure root path or NULL on failure
  */
 const char * rcl_get_secure_root(
   const char * node_name,
@@ -136,11 +136,11 @@ const char * rcl_get_secure_root(
     char * node_fqn = NULL;
     char * node_root_path = NULL;
     // Combine node namespace with node name
-    // TODO: This make assumption on the length of the root namespace as well as its value.
-    // We should likely come from another (rcl/rmw?) function for reuse.
+    // TODO(ros2team): This make assumption on the value of the root namespace.
+    // This should likely come from another (rcl/rmw?) function for reuse.
     node_fqn = rcutils_format_string(*allocator, "%s%s%s", node_namespace, "/", node_name);
     // Get native path, ignore the leading forward slash.
-    // TODO: The offset should be a define we can access, instead of hardcoding + 1
+    // TODO(ros2team): remove the hard-coded length, use the length of the root namespace instead
     node_root_path = rcutils_to_native_path(node_fqn + 1, *allocator);
     node_secure_root = rcutils_join_path(ros_secure_root_env, node_root_path, *allocator);
     allocator->deallocate(node_fqn, allocator->state);
--- a/rcl/src/rcl/node.c
+++ b/rcl/src/rcl/node.c
@@ -134,14 +134,14 @@ const char * rcl_get_secure_root(
     node_secure_root = rcutils_join_path(ros_secure_root_env, node_name, *allocator);
   } else {
     char * node_fqn = NULL;
-    char * node_root_norm = NULL;
+    char * node_root_path = NULL;
     // Combine node namespace with node name
     node_fqn = rcutils_format_string(*allocator, "%s%s%s", node_namespace, "/", node_name);
     // Get normalized path, ignore the leading forward slash.
-    node_root_norm = rcutils_to_native_path(node_fqn + 1, *allocator);
-    node_secure_root = rcutils_join_path(ros_secure_root_env, node_root_norm, *allocator);
+    node_root_path = rcutils_to_native_path(node_fqn + 1, *allocator);
+    node_secure_root = rcutils_join_path(ros_secure_root_env, node_root_path, *allocator);
     allocator->deallocate(node_fqn, allocator->state);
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
Thanks @ruffsl !

Do you happen to have a set of commands that can be reproduced to test this set of PRs?
Something like:
- generating artifacts for namespaced nodes
- show non namespaced nodes cant use them
- show that namespaced nodes can use them successfully

ruffsl:
> Do you happen to have a set of commands that can be reproduced to test this set of PRs?

#### Using Keymint

<details>

> save this file in ~/example:
##### talker_listener.xml
``` xml
<?xml version="1.0" encoding="UTF-8"?>

<profiles xmlns:xi="http://www.w3.org/2001/XInclude">
    <profile name="My Talker Profile">
        <attachments>
            <attachment>/*/talker</attachment>
        </attachments>
        <xi:include href="common/node.xml" parse="xml" />

        <ros_topic qualifier="ALLOW">
            <attachments>
                <attachment>/*/chatter</attachment>
            </attachments>
            <permissions>
                <ros_publish/>
            </permissions>
        </ros_topic>
    </profile>

    <profile name="My Listener Profile">
        <attachments>
            <attachment>/*/listener</attachment>
        </attachments>
        <xi:include href="common/node.xml" parse="xml" />

        <ros_topic qualifier="ALLOW">
            <attachments>
                <attachment>/*/chatter</attachment>
            </attachments>
            <permissions>
                <ros_subscribe/>
            </permissions>
        </ros_topic>
    </profile>
</profiles>
```

> run these commands from the same pwd:

``` shell
cd ~/example
docker run -it --rm -v=`pwd`:/root/keymint_ws keymint/keymint_tools:latest
keymint keystore init --bootstrap=keymint_ros
cp talker_listener.xml profile/comarmor.d/
keymint keystore create_pkg foo/bar/talker
keymint keystore create_pkg foo/bar/listener
keymint keystore build_pkg src/foo/bar/talker
keymint keystore build_pkg src/foo/bar/listener
exit

export ROS_SECURITY_ROOT_DIRECTORY=~/example/install
export ROS_SECURITY_ENABLE=true
export ROS_SECURITY_STRATEGY=Enforce
export ROS_NAMESPACE=/foo/bar

ros2 run demo_nodes_cpp talker
ros2 run demo_nodes_cpp listener
```

mikaelarguedas:
Please provide examples using `sros2` that can be tested on all platforms.
Ideally we would add test coverage for this in `test_security` so that we can run it on CI for all platforms, but a simple platform agnostic reproducible example using `sros2` would be enough to merge this IMO.

Can you clarify what is the `ROS_NAMESPACE` environment variable used for ? I don't think ROS 2 has such variable defined.

ruffsl:
I don't think the sros2 command is currently capable of using namespaces. When I get some time I can look into that.

> Can you clarify what is the ROS_NAMESPACE environment variable used for ? I don't think ROS 2 has such variable defined.

I updated my comment to use `__ns:`, I accidently was thinking of ROS1. But I just notice I have to patch how the common names are set in keymint, I'll patch that now.

rcl/src/rcl/node.c
```diff
@@ -99,7 +99,20 @@ const char * rcl_create_node_logger_name(
...
-const char * rcl_get_secure_root(const char * node_name, const rcl_allocator_t * allocator)
+/// Return the secure root dir associated with a node given the validated node name and namespace.
+/**
+ * E.g. for a node named "c" in namespace "/a/b", the secure root path will be
+ * "a/b/c", where the delimiter "/" is normlized for target file system (e.g. "\\" for _WIN32).
+ *
+ * \param[in] node_name validated node name (a single token)
+ * \param[in] node_namespace validated, absolute namespace (starting with "/")
+ * \param[in] allocator the allocator to use for allocation
+ * \returns duplicated string or null if there is an error
```
mikaelarguedas: what does `duplicated string` means in this context?
suggestions:
"machine specific node secure root path" would be more descriptive?
or simply "(absolute) node secure root path" maybe

rcl/src/rcl/node.c
```diff
@@ -99,7 +99,20 @@ const char * rcl_create_node_logger_name(
   return node_logger_name;
 }
 
-const char * rcl_get_secure_root(const char * node_name, const rcl_allocator_t * allocator)
```
mikaelarguedas: Thanks for adding the missing docblock :+1:

rcl/src/rcl/node.c
```diff
@@ -99,7 +99,20 @@ const char * rcl_create_node_logger_name(
...
+/// Return the secure root dir associated with a node given the validated node name and namespace.
```
mikaelarguedas: Nit: Please use full words rather than abbreviations: dir -> directory

rcl/src/rcl/node.c
```diff
@@ -99,7 +99,20 @@ const char * rcl_create_node_logger_name(
...
+/**
+ * E.g. for a node named "c" in namespace "/a/b", the secure root path will be
+ * "a/b/c", where the delimiter "/" is normlized for target file system (e.g. "\\" for _WIN32).
```
mikaelarguedas: typo: normlized -> normalized

rcl/src/rcl/node.c
ruffsl: Forgotten copypasta from `rcl_create_node_logger_name`; combined your phrasing for docblock.

rcl/src/rcl/node.c
```diff
@@ -115,7 +128,21 @@ const char * rcl_get_secure_root(const char * node_name, const rcl_allocator_t *
...
-  char * node_secure_root = rcutils_join_path(ros_secure_root_env, node_name, *allocator);
+  char * node_secure_root = NULL;
+  // If the namespace is the root namespace ("/"), the secure root is just the node name.
+  if (strlen(node_namespace) == 1) {
+    node_secure_root = rcutils_join_path(ros_secure_root_env, node_name, *allocator);
+  } else {
+    char * node_fqn = NULL;
+    char * node_root_norm = NULL;
+    // Combine node namespace with node name
+    node_fqn = rcutils_format_string(*allocator, "%s%s%s", node_namespace, "/", node_name);
```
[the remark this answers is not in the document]
ruffsl: TODOs added

mikaelarguedas:
@ruffsl FYI: both PRs fail the linters tests.

this was fixed in https://github.com/ros2/rcl/pull/300/commits/6cff52b011680f5743e83b23268017318ea669d3 and https://github.com/ros2/rcutils/pull/119/commits/5b691176abcc814bb50938ccff16b65248c2a848.

Tips for future PRs: 
- You can run tests locally using `colcon test --packages-select <PACKAGES_TO_TEST>`.
- Using the same branch name for PRs that need to be tested together allows to run CI by setting only a single parameter `CI_BRANCH_TO_TEST`

mikaelarguedas:
@ruffsl can you please comment here once keymint has been updated to generate files that can be used to test this PR?

ruffsl:
I have two PRs waiting for changes is ROS2 Crystal that deal with the CA changes:
https://github.com/keymint/keymint_profile/pull/1
https://github.com/keymint/keymint_package/pull/2

With those PRs, I've created and tested locally this keymint_ws that demonstrates the nested namespacing:
https://github.com/ruffsl/keymint_ws/pull/2

You could copy the `install` (secure root) directory from that branch into your tests (note the CAs are linked), but eventually it would be better to auto generate them given ROS2 substem mapping might change, like added parameter service calls. But it should give you a chance to sanity check the code.

mikaelarguedas:
Tested manually the rebased version of these branches on Linux and Windows.

It works on Linux but silently fails on my Windows VM.

As I don't think I'll have time to dig into it before ROSCon, it would be great to provide a PR to `test_security` with the proper security artifacts so that this can be tested on CI.

rcl/src/rcl/node.c
```diff
@@ -115,7 +128,25 @@ const char * rcl_get_secure_root(const char * node_name, const rcl_allocator_t *
...
+    // TODO(ros2team): remove the hard-coded value of the root namespace.
+    node_fqn = rcutils_format_string(*allocator, "%s%s%s", node_namespace, "/", node_name);
+    // Get native path, ignore the leading forward slash.
+    // TODO(ros2team): remove the hard-coded length, use the length of the root namespace instead.
+    node_root_path = rcutils_to_native_path(node_fqn + 1, *allocator);
+    node_secure_root = rcutils_join_path(ros_secure_root_env, node_root_path, *allocator);
+    allocator->deallocate(node_fqn, allocator->state);
+    allocator->deallocate(node_root_path, allocator->state);
+  }
   if (!rcutils_is_directory(node_secure_root)) {
```
clalancette: I know you didn't add this, but there is a problem with `NULL` here.  Assume that `rcutils_join_path` at line 136 returns NULL.  In that case, the `node_secure_root` will be NULL, which will be passed into `rcutils_is_directory`, which will be then passed directly to `stat`, which will most likely core dump.  A similar problem exists in the sequence in 138-148.  I think we need to check for `NULL` right after creating `node_secure_root`, and return `NULL` if it is (before the `rcutils_is_directory` check).

rcl/src/rcl/node.c
```diff
@@ -115,7 +128,25 @@ const char * rcl_get_secure_root(const char * node_name, const rcl_allocator_t *
   if (!ros_secure_root_size) {
     return NULL;  // environment variable was empty
   }
-  char * node_secure_root = rcutils_join_path(ros_secure_root_env, node_name, *allocator);
+  char * node_secure_root = NULL;
+  // TODO(ros2team): This make assumption on the value and length of the root namespace.
```
clalancette: `This make assumption` -> `This makes an assumption`

ruffsl:
I've added fcf0fa0 to provide way for users to exactly specify the target security directory to address ros2/sros2#69 . Let me know if you see a cleaner way to add this.

clalancette:
> > 
> 
> @clalancette , Do you mean that the inclusion of the TODO items are not warranted here, or that the potential issues they refer to should also be addressed in this PR?

The latter.  I don't have a lot of background to go on here, so I'm not sure if that limitation is fine for now.

ruffsl:
@clalancette , Looking at the rest of rcl, I'm seeing the string `"/"` used rather abundantly, making pervasive assumptions on the length of the root namespace as well as its value. About the only place I see the root namespace `# define`d is in `lexer.h`, but I wouldn't suspect it would make much sense to link against that header. I'd say resolving these todo items would be out scope for this small PR.

ruffsl:
ping @clalancette , I'd like to fit this in before the code freeze for crystal.  
Please let me know if you have any more changes in mind.

rcl/src/rcl/node.c
```diff
@@ -99,24 +100,80 @@ const char * rcl_create_node_logger_name(
...
 }
 
-const char * rcl_get_secure_root(const char * node_name, const rcl_allocator_t * allocator)
+/// Return the secure root directory associated with a node given its validated name and namespace.
+/**
+ * E.g. for a node named "c" in namespace "/a/b", the secure root path will be
+ * "a/b/c", where the delimiter "/" is native for target file system (e.g. "\\" for _WIN32).
+ * However, this expansion can be overridden by setting the secure node directory environment
+ * variable, allowing users to explicitly specify the exact secure root directory to be utilized.
+ * Such an override is useful for case where the FQN of a node is non-determinist before runtime,
```
clalancette: `determinist` -> `deterministic`

rcl/src/rcl/node.c
```diff
@@ -99,24 +100,80 @@ const char * rcl_create_node_logger_name(
...
+ * or when testing and using additional tools that may not otherwise not be easily provisioned
```
clalancette: Period at the end.

clalancette:
CI is being tracked here: https://github.com/ros2/rcutils/pull/119#issuecomment-436299019

Issue #306, opened 2018-10-15
Actions


jacobperron:
Implement actions API for use by client libraries.

Actions are a important feature/concept in ROS that allow for tracking and updates to long-running requests called "goals". Refer to the design doc for more information:  ros2/design#193

Prerequisites:

- [x] C typesupport for actions (rosidl_generator_c) ros2/rosidl#302

Tasks:

- [x] Propose header files for action server, action client, and other relevant entities #307 
- [ ] ~~Architecture doc~~ 
- Implementation:
  - [x] Goal state machine #311 
  - [x] Types implementation #312
  - [x] Goal handle implementation #320 
  - [x] Action client implementation #319 
  - [x] Action server implementation #323
  - [x] Wait set groups #308 
    - [ ] Integrate with rcl_action #337 

connects to ros2/ros2#583

PR #307, opened 2018-10-18
Add rcl_action package and headers

jacobperron:
Connects to #306 

This is a proposal of structs and functions for the core C implementation of ROS actions.
This is not a comprehensive set of function declarations, but I think there is enough of the core here for a review. In particular, it's missing utilities for adding actions to the wait set and generating UUIDs for action goals.

Let me know if there's any other core features that I have missed.


jacobperron:
@mkhansen-intel FYI

jacobperron:
Something I didn't consider until later (and reflect in TODOs) is to use generated definitions for the *cancel* service and *status* topics instead of defining structs for them:

https://github.com/ros2/rcl/blob/44210ebb723e6fa1433ba3dc60d530eb2c54e4f9/rcl_action/include/rcl_action/types.h#L83-L84

https://github.com/ros2/rcl/blob/44210ebb723e6fa1433ba3dc60d530eb2c54e4f9/rcl_action/include/rcl_action/types.h#L96-L97

https://github.com/ros2/rcl/blob/44210ebb723e6fa1433ba3dc60d530eb2c54e4f9/rcl_action/include/rcl_action/types.h#L106-L107

The main benefit of using generated code is to not have multiple definitions of the same structures in two places. Unless someone proposes a reason not to, I will update the functions related to canceling goals and taking/sending statuses to use types from `rcl_interfaces/srv` and `rcl_interfaces/msg`.

rcl_action/include/rcl_action/action_server.h
```diff
@@ -0,0 +1,667 @@
...
+ * \return `RCL_RET_ACTION_SERVER_INVALID` if the action server is invalid, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_action_send_goal_response(
+  const rcl_action_server_t * action_server,
+  const rcl_action_goal_info_t * goal_info,
+  const void * ros_goal,
```
sloretz: Is this the original request message for the goal? Is this needed? I would have expected the `rcl_action_goal_info_t` to be enough.

rcl_action/include/rcl_action/action_server.h
```diff
@@ -0,0 +1,667 @@
...
+/**
+ * This is a non-blocking call.
+ *
+ * It is the job of the caller to ensure that the type of the `ros_feedback`
+ * parameter and the type associate with the action server (via the type support)
+ * match.
+ * Passing a different type to publish produces undefined behavior and cannot
+ * be checked by this function and therefore no deliberate error will occur.
+ *
+ * This function acts like a ROS publisher and is potntially a blocking call.
```
sloretz: `potntially` -> `potentially`

rcl_action/include/rcl_action/action_server.h
```diff
@@ -0,0 +1,667 @@
...
+ * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return `RCL_RET_ACTION_SERVER_INVALID` if the action server is invalid, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_action_publish_status(
+  const rcl_action_server_t * action_server,
+  const rcl_action_status_array_t * status_mesage);
```
sloretz: I would have expected the `action_server` to internally know the status of each goal, and if so it wouldn't need to be passed in here. Where is the state for each goal tracked?

rcl_action/include/rcl_action/action_server.h
```diff
@@ -0,0 +1,667 @@
...
+ * \return `RCL_RET_ACTION_SERVER_TAKE_FAILED` if take failed but no error occurred
+ *         in the middleware, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_action_take_cancel_request(
+  const rcl_action_server_t * action_server,
+  rcl_action_goal_info_t * goal_info);
```
sloretz: Once a client library has a cancel request, where does the logic for telling if a goal is part of a cancel request live? Is that duplicated in `rclpy` and `rclcpp`?

rcl_action/include/rcl_action/action_server.h
```diff
@@ -0,0 +1,667 @@
...
+RCL_WARN_UNUSED
+const rcl_action_server_options_t *
+rcl_action_server_get_options(const rcl_action_server_t * action_server);
+
+/// Return the goal handles for all active or terminated goals.
+/**
+ * A pointer to the internally held array of goal handle structs is returned
+ * along with the number of items in the array.
+ * Goals that have terminated, successfully responded to a client with a
+ * result, and have expired (timed out) are not present in the array.
```
sloretz: Who controls the timeout. Is there a function to check if the goals are expired? Is there a timer part of the action server?

rcl_action/include/rcl_action/goal_handle.h
```diff
@@ -0,0 +1,347 @@
...
+ * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return `RCL_RET_ACTION_GOAL_HANDLE_INVALID` if the goal handle is invalid, or
+ * \return `RCL_RET_ACTION_SERVER_INVALID` if the action server is invalid, or
+ * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_action_goal_handle_init(
```
sloretz: Is it a correct assumption this function is only used internally by `rcl_action` since the goal handle is returned by `rcl_action_send_goal_response`?

rcl_action/include/rcl_action/goal_handle.h
```diff
@@ -0,0 +1,347 @@
...
+ * \return `RCL_RET_ACTION_GOAL_HANDLE_INVALID` if the goal handle is invalid, or
+ * \return `RCL_RET_ACTION_SERVER_INVALID` if the action server is invalid, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_action_goal_handle_fini(
+  rcl_action_goal_handle_t * goal_handle,
+  const rcl_action_server_t * action_server);
```
sloretz: I would expect `goal_handle` to have a reference to `action_server` already. That avoids accidentally passing the wrong action server in.

rcl_action/include/rcl_action/goal_handle.h
```diff
@@ -0,0 +1,347 @@
...
+ * \param[in] goal_event the event used to transition the goal state
+ * \return `RCL_RET_OK` if the goal state was updated successfully, or
+ * \return `RCL_RET_ACTION_GOAL_EVENT_INVALID` if the goal event is invalid, or
+ * \return `RCL_RET_ACTION_GOAL_HANDLE_INVALID` if the goal handle is invalid, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_action_update_goal_state(
```
sloretz: Does `rcl_action_update_goal_state` get called before or after `rcl_action_send_result_response`?

rcl_action/include/rcl_action/goal_handle.h
```diff
@@ -0,0 +1,347 @@
...
+ * \return `RCL_RET_ACTION_GOAL_HANDLE_INVALID` if the goal handle is invalid, or
+ * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_action_goal_handle_get_message(
+  const rcl_action_goal_handle_t * goal_handle,
+  void * ros_goal);
```
sloretz: Where is the original request message used later?

rcl_action/include/rcl_action/goal_state_machine.h
```diff
@@ -0,0 +1,130 @@
...
+#include "rcl_action/types.h"
+#include "rcl/visibility_control.h"
+
+
+typedef rcl_action_goal_state_t
+(* rcl_action_goal_event_handler)(rcl_action_goal_state_t, rcl_action_goal_event_t);
+
+// Transition event handlers
+rcl_action_goal_state_t
+_execute_event_handler(rcl_action_goal_state_t state, rcl_action_goal_event_t event)
```
sloretz: inline and/or `RCL_LOCAL`?

rcl_action/include/rcl_action/types.h
```diff
@@ -0,0 +1,119 @@
...
+typedef enum rcl_action_goal_state_t
+{
+  GOAL_STATE_ACCEPTED = 0,
+  GOAL_STATE_EXECUTING,
+  GOAL_STATE_CANCELING,
+  GOAL_STATE_SUCCEEDED,
+  GOAL_STATE_CANCELED,
+  GOAL_STATE_ABORTED,
+  GOAL_STATE_NUM_STATES,
+  GOAL_STATE_INVALID
```
sloretz: I would put invalid at 0, some compilers will zero initialize if building without optimizations, which could help catch bugs during development

jacobperron:
> Something I didn't consider until later (and reflect in TODOs) is to use generated definitions for the _cancel_ service and _status_ topic instead of defining structs for them:

Summarizing an offline discussion with @wjwwood:

- Using rcl structs for message passing would not allow client libraries in other languages to use their native service and topic facilities. This means we definitely should pass around `void *` types for the generating messages.
- Passing around generated types means that rcl will not have access to the messages unless the client libraries first convert the messages to rcl types. Ultimately, we could provide conversion functions alongside message generation, but as a stopgap write the conversions in the client libraries.
- If we want to keep logic (e.g. the cancel policy) in rcl, then it would be best to provide some utility functions in rcl for operating on requests/responses *after* they have been converted and passed back from the client library.

rcl_action/include/rcl_action/action_client.h
```diff
@@ -0,0 +1,672 @@
...
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] action_client handle to the client that will take the goal response
+ * \param[inout] goal_info pointer to the goal info
+ * \param[out] accepted set to `true` if the goal was accepted by the server, `false` otherwise
```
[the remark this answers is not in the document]
jacobperron: Yes, it was over from a previous iteration.

rcl_action/include/rcl_action/action_client.h
```diff
@@ -0,0 +1,672 @@
...
+ *   with a taken result is copied
+ * \param[out] terminal_state preallocated variable that is set to SUCCEEDED, ABORTED, or CANCELED
+ * \param[out] ros_result preallocated struct where the ROS result message is copied.
+ * \return `RCL_RET_OK` if the response was taken successfully, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return `RCL_RET_ACTION_CLIENT_INVALID` if the action client is invalid, or
+ * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
+ * \return `RCL_RET_ACTION_CLIENT_TAKE_FAILED` if take failed but no error occurred
+ *         in the middleware, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
```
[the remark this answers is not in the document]
jacobperron: `RCL_RET_ACTION_CLIENT_TAKE_FAILED`

rcl_action/include/rcl_action/action_client.h
```diff
@@ -0,0 +1,672 @@
...
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_action_take_cancel_response(
+  const rcl_action_client_t * action_client,
+  rcl_action_cancel_response_t * cancel_response);
+
+/// Get the name of the action for an action client.
+/**
+ * This function returns the action client's internal topic name string.
```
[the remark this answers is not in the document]
jacobperron: No, this is the name provided in the constructor without the leading `/_action/`.

rcl_action/include/rcl_action/action_server.h
```diff
@@ -0,0 +1,727 @@
...
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_action_server_t
+rcl_action_get_zero_initialized_server(void);
+
+/// Initialize a rcl_action_server_t.
+/**
+ * After calling this function on a rcl_action_server_t, it can be used to take
+ * goals of the given type for the given action name using rcl_action_take_goal_request()
+ * and take cancel requests with rcl_aciton_take_cancel_request().
```
hidmic: @jacobperron `rcl_aciton_take_cancel_request()` -> `rcl_action_take_cancel_request()`

rcl_action/include/rcl_action/action_server.h
jacobperron: I think the action server will track the state of each goal. I figured it might provide some more flexibility to separate the publish status and the getter for the array of goal handles. We can add an additional publish status function that does not take a status message for convenience.

rcl_action/include/rcl_action/action_server.h
jacobperron: I'll add a function for checking timeouts. I imagine it would be better to implement this logic in the action server so it isn't duplicated in the client libraries.

rcl_action/include/rcl_action/action_server.h
```diff
@@ -0,0 +1,727 @@
...
+ * \param[in] error_msg_allocator a valid allocator or `NULL`
+ * \return a pointer to a new goal handle representing the accepted goal, or
+ * \return `NULL` if a failure occured.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_action_goal_handle_t *
+rcl_action_accept_new_goal(
+  const rcl_action_server_t * action_server,
+  const rcl_action_goal_info_t * goal_info,
```
hidmic: @jacobperron so, if I follow the logic correctly, moments after a successful `rcl_action_take_goal_request()` call, `rcl_action` caller must populate an `rcl_action_goal_info_t` instance appropriately to let the server know of the goal. Is that right? If so, it'd be nice to be explicit about it.

rcl_action/include/rcl_action/action_server.h
```diff
@@ -0,0 +1,727 @@
...
+ *         in the middleware, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_action_process_cancel_request(
+  const rcl_action_server_t * action_server,
+  const rcl_action_cancel_request_t * cancel_request,
+  rcl_action_cancel_response_t * cancel_response);
```
hidmic: @jacobperron same as above, if I'm right about these base action types then I think it'd be nice to have it documented.

rcl_action/include/rcl_action/action_server.h
```diff
@@ -0,0 +1,727 @@
...
+ * The following cancel policy applies based on the goal ID and the timestamp
+ * contained in the cancel request:
+ *
+ * - If the goal ID is zero and timestamp is zero, cancel all goals.
+ * - If the goal ID is zero and timestamp is not zero, cancel all goals accepted
+ *   at or before the timestamp.
+ * - If the goal ID is not zero and timestamp is zero, cancel the goal with the
+ *   given ID regardless of the time it was accepted.
+ * - If the goal ID is not zero and timestamp is not zero, cancel the goal with the
+ *   given ID and all goals accepted at or before the timestamp.
```
hidmic: @jacobperron would it make sense to expose cancelling operations on a per goal handle basis too and build `rcl_action_process_cancel_request()`  on top of them instead?

rcl_action/include/rcl_action/goal_handle.h
jacobperron: Yes, that is correct.

rcl_action/include/rcl_action/goal_handle.h
jacobperron: I'm removing this to simplify implementation since this is the only place where typesupport for goal handles comes into play and it does not seem necessary.

rcl_action/include/rcl_action/action_client.h
```diff
@@ -0,0 +1,690 @@
...
+ * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return `RCL_RET_ACTION_CLIENT_INVALID` if the client is invalid, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_action_send_goal_request(
+  const rcl_action_client_t * action_client,
+  const void * ros_goal_request);
```
hidmic: @jacobperron as @sloretz suggested for `rcl_action_take_feedback()` below, may be useful to expose `rmw_request_id_t` if multiple goal requests are sent one after the other as in:

https://github.com/ros2/rcl/blob/5e3d4be72089a4c766b508877bc52a196bd7bead/rcl/include/rcl/service.h#L300-L303

rcl_action/include/rcl_action/action_client.h
```diff
@@ -0,0 +1,690 @@
...
+ * \param[in] ros_cancel_request pointer the ROS cancel request message
+ * \return `RCL_RET_OK` if the response was taken successfully, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return `RCL_RET_ACTION_CLIENT_INVALID` if the action client is invalid, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_action_send_cancel_request(
```
hidmic: @jacobperron same about exposing `rmw_request_id_t`

rcl_action/include/rcl_action/action_server.h
sloretz: +1 to implementing it in the action server

rcl_action/include/rcl_action/action_server.h
jacobperron: Yes, this is correct. I will makes sure this is documented. And add an init/fini for the `rcl_action_goal_info_t`

rcl_action/include/rcl_action/action_client.h
jacobperron: Perhaps, I was assuming that the goal ID would be present in the ROS message (`ros_goal_request`). The client could use the goal ID to distinguish between multiple responses.

rcl_action/include/rcl_action/action_client.h
jacobperron: Assuming we have a goal ID as part of the ROS request message, I'm not sure that we need to expose a `rmw_request_id_t`.

jacobperron:
> BTW where is blocking usually implemented? I've seen the same TODOs about it in the current rcl API for services.

I'm not sure myself, I've copied the TODOs over from the services implementation since actions are using services under the hood. I guess it is dependent on QoS settings configured in `rmw`.

rcl_action/include/rcl_action/action_server.h
jacobperron: The goal handle's state can be updated with [rcl_action_update_goal_state()](https://github.com/ros2/rcl/blob/48519db9c82a2b6f98b251779dd0732bef03a516/rcl_action/include/rcl_action/goal_handle.h#L194), which I plan to call from this  function.

jacobperron:
@sloretz @hidmic Thanks for the reviews. I believe I've address all of your comments and I think this is ready for another round of review.

rcl_action/include/rcl_action/wait.h
```diff
@@ -0,0 +1,100 @@
...
+ * \return `RCL_RET_OK` if added successfully, or
+ * \return `RCL_RET_WAIT_SET_INVALID` if the wait set is zero initialized, or
+ * \return `RCL_RET_WAIT_SET_FULL` if the subscription set is full, or
+ * \return `RCL_RET_ACTION_SERVER_INVALID` if the action server is invalid, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+*/
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_wait_set_add_action_server(
```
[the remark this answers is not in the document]
sloretz: Idea for waitset groups in #308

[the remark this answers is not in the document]
jacobperron: Yeah, I can see your point about providing a method for getting the number of services and subscriptions from an action server or action client. It's better than clients assuming the number.

Regarding "which services/subscribers?", I'm not sure how the client libraries will look when it comes to actions and wait sets (right now it's not clear to me without digging further), but I'll add some notify functions as you suggest. It could be the case that we come back and refactor to help with client library implementation.

jacobperron:
@sloretz 
Updated the wait set related functions. Rather than "notify" functions, since the client library is expect to take (and in some cases convert messages for rcl), I've added `rcl_action_server_wait_set_get_entities_ready()` and `rcl_action_client_wait_set_get_entities_ready()`. The idea is that the client library can call these functions after waiting so they know what `rcl_action_take_*` calls to make.

jacobperron:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5369/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2096/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4459/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5329/)

jacobperron:
Fixed windows build after adding an `inline` that should have been there:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5371/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2098/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4461/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5332/)

rcl_action/package.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>rcl_action</name>
+  <version>0.5.0</version>
+  <description>Package containing a C-based ROS action implementation</description>
+  <maintainer email="karsten@osrfoundation.org">Karsten Knese</maintainer>
```
dirk-thomas: @jacobperron Your name / email should go here.

Issue #309, opened 2018-10-27
Remove lifecycle_service_client_py.py; it is deprecated and no longer works correctly


fmrico:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04 
- Installation type:
  - binaries and source  
- Version or commit hash:
  - bouncy and 451bf4a1a41ad87920006a974a4d3b8ca528847b
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - N/A

#### Steps to reproduce issue

```
In terminal 1: 
$ ros2 run lifecycle lifecycle_talker

In terminal 2: 
$ ros2 run lifecycle lifecycle_service_client_py.py change_state --change-state-args configure lc_talker
$ ros2 run lifecycle lifecycle_service_client_py.py change_state --change-state-args activate lc_talker
$ ros2 run lifecycle lifecycle_service_client_py.py change_state --change-state-args shutdown lc_talker
```

#### Expected behavior
Transition from active to shuttingdown 
#### Actual behavior

```
[rcl_lifecycle]: No transition matching 5 found for current state unconfigured
[ERROR] []: Unable to start transition 5 from current state unconfigured: Transition is not registered., at /home/paco/ros2/ros2_ws/src/ros2/rcl/rcl_lifecycle/src/rcl_lifecycle.c:332

```
#### Additional information

These trensition messages are defined in rcl_interfaces/lifecycle_msgs/msg/Transition.msg:14 
```
uint8 TRANSITION_SHUTDOWN = 5
uint8 TRANSITION_DESTROY = 6
```

Actually, demos/lifecycle/src/lifecycle_service_client_py.py:40 uses it
```
    elif change_state_args == 'shutdown':
        req.transition.id = Transition.TRANSITION_SHUTDOWN
```

but these transitions are never registered. I have debugged this by adding these lines just at the end of function rcl_lifecycle_register_transition(..) at rcl/rcl_lifecycle/src/transition_map.c:154: 

```
  const char *lstart = rcl_lifecycle_get_state(transition_map, transition.start->id)->label;
  const char *lgoal = rcl_lifecycle_get_state(transition_map, transition.goal->id)->label;
  const char *ltrans = rcl_lifecycle_get_transitions(transition_map, transition.id)->label;

  RCUTILS_LOG_INFO_NAMED(ROS_PACKAGE_NAME,
    "registering transition %u (%s) state %u (%s) -> %u (%s)\n",
      transition.id, ltrans,
[...]

PR #310, opened 2018-10-27
[lifecycle] Add transitions from some states to shutting_down

fmrico:
Hi all,

If #309 is considered a bug, this PR fixed it. Otherwise, discard it.

TRANSITION_SHUTDOWN was specfied in messages, but this transition was not implemented.

http://design.ros2.org/articles/node_lifecycle.html specifies transitions from Unconfigured, Inactive and Active transitions to Shuttingdown state. This PR provides these transitions.

Best

### DIFF
--- a/rcl_lifecycle/src/default_state_machine.c
+++ b/rcl_lifecycle/src/default_state_machine.c
@@ -503,6 +503,54 @@ _register_transitions(
     }
   }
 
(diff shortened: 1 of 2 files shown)


fmrico:
Another solution could be just removing TRANSITION_SHUTDOWN from messages...

PR #311, opened 2018-10-29
Action goal state machine tests

jacobperron:
Connects to #306.

As a follow up, we might consider replacing the hard-coded transition map with the more general one [in rcl_lifecycle](https://github.com/ros2/rcl/blob/master/rcl_lifecycle/include/rcl_lifecycle/transition_map.h). This would involve moving the general transition map to a common package like [rcutils](https://github.com/ros2/rcutils).


jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5391/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2108/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4473/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5343/)

jacobperron:
Attempt to fix MSVC compiler warnings:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5399/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2115/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4482/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5349/)

rcl_action/include/rcl_action/goal_state_machine.h
```diff
@@ -83,23 +80,38 @@ _set_canceled_event_handler(rcl_action_goal_state_t state, rcl_action_goal_event
...
-    [GOAL_EVENT_SET_ABORTED] = _set_aborted_event_handler,
-  },
-  [GOAL_STATE_CANCELING] = {
-    [GOAL_EVENT_SET_SUCCEEDED] = _set_succeeded_event_handler,
-    [GOAL_EVENT_SET_ABORTED] = _set_aborted_event_handler,
-    [GOAL_EVENT_SET_CANCELED] = _set_canceled_event_handler,
-  },
-};
+static rcl_action_goal_event_handler
+  _goal_state_transition_map[GOAL_STATE_NUM_STATES][GOAL_EVENT_NUM_EVENTS];
```
hidmic: @jacobperron just out of curiosity, what  was wrong about previous direct initialization?

rcl_action/include/rcl_action/goal_state_machine.h
```diff
@@ -83,23 +80,38 @@ _set_canceled_event_handler(rcl_action_goal_state_t state, rcl_action_goal_event
...
+
+static inline void
+rcl_action_goal_transition_map_init()
```
hidmic: @jacobperron Why are we using static linkage for these? Why not defining them on goal_state_machine.c ?

clalancette: Yeah, agreed; for a large function like this, making it static inline is bloating the downstream users for not a lot of reason.  This is probably better as a function inside a C file.

rcl_action/include/rcl_action/goal_state_machine.h
```diff
@@ -83,23 +80,38 @@ _set_canceled_event_handler(rcl_action_goal_state_t state, rcl_action_goal_event
...
+{
+  // Only run this function once
+  static bool _is_goal_state_transition_map_init = false;
+  if (_is_goal_state_transition_map_init) {
```
clalancette: If this could be called from multiple threads, there is a TOCTTOU bug here; thread 1 could have read the state of `_is_goal_state_transition_map_init` as false then been swapped out, then thread 2 could read the state of `_is_goal_state_transition_map` as false, and do the initialization.  Then when thread 1 wakes back up, it still thinks that it needs to do the initialization, and does it again.

rcl_action/include/rcl_action/goal_state_machine.h
jacobperron: When the direct initialization gets included in C++ (e.g. gtest), then gcc complains about a sparsely populated 2d array. But we might be able to use null pointers in the direct initialization instead of this init function.

rcl_action/include/rcl_action/goal_state_machine.h
jacobperron: In retrospect, I think all of these static functions and the transition map can be moved to a C file.

wjwwood: +1, you should move as much out of the public interface as is possible/reasonable.

rcl_action/include/rcl_action/goal_state_machine.h
jacobperron: Moving the initialization to the C file let me do a direct initialization, so I believe this issue goes away.

jacobperron:
#315 should fix the unstable windows build. I'll rebase once it's merged.

rcl_action/src/rcl_action/goal_state_machine.c
```diff
@@ -20,12 +20,85 @@ extern "C"
...
+_set_canceled_event_handler(rcl_action_goal_state_t state, rcl_action_goal_event_t event)
+{
+  if (GOAL_STATE_CANCELING != state || GOAL_EVENT_SET_CANCELED != event) {
+    return GOAL_STATE_UNKNOWN;
+  }
+  return GOAL_STATE_CANCELED;
+}
+
+// Transition map
+rcl_action_goal_event_handler
```
hidmic: @jacobperron nit: this one could be `static`.

rcl_action/include/rcl_action/goal_state_machine.h
jacobperron: Moving the direct implementation to the C file got rid of the compiler warnings :)

jacobperron:
Reviews addressed and rebased on master:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5416/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2123/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4490/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5357/)

Issue #308, opened 2018-10-23
Wait set groups


sloretz:
## Feature request
This is a request for an API on top of the current waitset api that considers groups of waitable entities.

#### Feature description

An API that covers groups of entities that can be waited on would be immediately useful for actions. A server and client consists of several waitable entities (servers, clients, maybe a timer). The API would allow a client library to add a group of waitable entities to the wait set, and then notify the action server when one of them becomes available.

#### Implementation considerations

This could be implemented in a package `rcl_grouped_waitset` on top of the existing `rcl` API

Structs and methods to be added.
```C
struct rcl_grouped_waitset
{
  rcl_waitset_group_t * groups;
  size_t num_groups;

  rcl_waitset_group_impl_t * impl;
}

struct rcl_waitset_group_t
{
  rcl_subscription_t * subscriptions;
  rcl_guard_condition_t * guard_conditions;
  rcl_timer_t * timers;
  rcl_client_t * clients;
  rcl_service_t * servers;

  size_t num_subscriptions;
  size_t num_guard_conditions;
  size_t num_timers;
  size_t num_clients;
  size_t num_servers;
}
rcl_waitset_group_get_zero_initialized();

// User won't call these, but rcl_grouped_waitset_add and rcl_grouped_waitset_fini might
// rcl_waitset_group_init(rcl_waitset_group_t *, num_subs, subs, ..., rcl_allocator_t alloc);
// rcl_waitset_group_fini(rcl_waitset_group_t * g);

rcl_grouped_waitset_get_zero_initialized();
rcl_grouped_waitset_init(rcl_grouped_waitset_t * gws, size_t num_groups, rcl_allocator_t alloc);
rcl_grouped_waitset_fini(rcl_grouped_waitset_t * gws);

// Add a group waitable entities to the wait set
// This returns a waitset group that can be used to tell which entities in the group woke the wait set
// The returned group isn't usable until `rcl_grouped_waitset_build()` is called. The returned group never needs to be cleaned up by the user.
// subs, gcs, timers, etc pointers are borrowed and must be valid until `rcl_grouped_waitset_build()` is called
[...]

PR #319, opened 2018-11-02
Action client implementation

hidmic:
Connected to #306. This pull request introduces a first implementation of `rcl`'s action client interface. It depends on ros2/rosidl#310 for basic typesupport declarations to compile and on ros2/rcl_interfaces#47 for basic typesupport implementation to test and thus it's still lacking proper test coverage.

I'm pushing this early to trigger a follow up discussion on the pros and cons of the current fully type-erased action client API.

### DIFF
--- a/rcl_action/CMakeLists.txt
+++ b/rcl_action/CMakeLists.txt
@@ -70,20 +70,23 @@ install(TARGETS ${PROJECT_NAME}
 if(BUILD_TESTING)
   find_package(ament_cmake_gtest REQUIRED)
   find_package(ament_lint_auto REQUIRED)
+  find_package(test_msgs REQUIRED)
   ament_lint_auto_find_test_dependencies()
   ament_find_gtest()
   # Gtests
-  # ament_add_gtest(test_action_client
-  #   test/rcl_action/test_action_client.cpp
-  # )
-  # if(TARGET test_action_client)
-  #   target_include_directories(test_action_client PUBLIC
-  #     ${rcl_INCLUDE_DIRS}
-  #   )
-  #   target_link_libraries(test_action_client
-  #     ${PROJECT_NAME}
-  #   )
-  # endif()
+  ament_add_gtest(test_action_client
+    test/rcl_action/test_action_client.cpp
+  )
+  if(TARGET test_action_client)
+    target_include_directories(test_action_client PUBLIC
+      ${rcl_INCLUDE_DIRS}
+      ${test_msgs_INCLUDE_DIRS}
+    )
+    target_link_libraries(test_action_client
+      ${PROJECT_NAME}
+      ${test_msgs_LIBRARIES}
+    )
+  endif()
   ament_add_gtest(test_goal_handle
     test/rcl_action/test_goal_handle.cpp
   )
--- a/rcl_action/CMakeLists.txt
+++ b/rcl_action/CMakeLists.txt
@@ -32,6 +32,7 @@ add_executable(test_compile_headers
 )
 
 set(rcl_action_sources
+  src/${PROJECT_NAME}/action_client.c
   src/${PROJECT_NAME}/goal_handle.c
   src/${PROJECT_NAME}/goal_state_machine.c
   src/${PROJECT_NAME}/names.c
@@ -72,6 +73,17 @@ if(BUILD_TESTING)
   ament_lint_auto_find_test_dependencies()
   ament_find_gtest()
   # Gtests
+  # ament_add_gtest(test_action_client
+  #   test/rcl_action/test_action_client.cpp
+  # )
+  # if(TARGET test_action_client)
+  #   target_include_directories(test_action_client PUBLIC
+  #     ${rcl_INCLUDE_DIRS}
+  #   )
+  #   target_link_libraries(test_action_client
+  #     ${PROJECT_NAME}
+  #   )
+  # endif()
   ament_add_gtest(test_goal_handle
     test/rcl_action/test_goal_handle.cpp
   )
--- a/rcl_action/CMakeLists.txt
+++ b/rcl_action/CMakeLists.txt
@@ -79,13 +79,12 @@ if(BUILD_TESTING)
   )
   if(TARGET test_action_client)
     target_include_directories(test_action_client PUBLIC
-      ${rcl_INCLUDE_DIRS}
-      ${test_msgs_INCLUDE_DIRS}
+      include
     )
     target_link_libraries(test_action_client
       ${PROJECT_NAME}
-      ${test_msgs_LIBRARIES}
     )
+    ament_target_dependencies(test_action_client "rcl" "test_msgs")
   endif()
   ament_add_gtest(test_goal_handle
     test/rcl_action/test_goal_handle.cpp
--- a/rcl_action/include/rcl_action/action_client.h
+++ b/rcl_action/include/rcl_action/action_client.h
@@ -415,7 +415,6 @@ rcl_action_take_status(
   const rcl_action_client_t * action_client,
   void * ros_status_array);
 
-
 /// Send a request for the result of a completed goal associated with a rcl_action_client_t.
 /**
  * This is a non-blocking call.
--- a/rcl_action/include/rcl_action/action_client.h
+++ b/rcl_action/include/rcl_action/action_client.h
@@ -176,7 +176,7 @@ RCL_WARN_UNUSED
 rcl_ret_t
 rcl_action_client_init(
   rcl_action_client_t * action_client,
-  const rcl_node_t * node,
+  rcl_node_t * node,
   const rosidl_action_type_support_t * type_support,
   const char * action_name,
   const rcl_action_client_options_t * options);
@@ -355,9 +355,7 @@ rcl_action_take_goal_response(
  * Lock-Free          | Yes
  * <i>[1] only if required when filling the feedback message, avoided for fixed sizes</i>
  *
- * \param[in] action_client handle to the client that will take the goal response
- * \param[out] goal_info pointer to a struct for meta-data about the goal associated
- *   with taken feedback
+ * \param[in] action_client handle to the client that will take action feedback
  * \param[out] ros_feedback pointer to the ROS feedback message.
  * \return `RCL_RET_OK` if the response was taken successfully, or
  * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
@@ -419,6 +417,7 @@ rcl_ret_t
 rcl_action_take_status(
   const rcl_action_client_t * action_client,
   void * ros_status_array);
+
 
 /// Send a request for the result of a completed goal associated with a rcl_action_client_t.
 /**
@@ -661,7 +660,7 @@ RCL_WARN_UNUSED
 const rcl_action_client_options_t *
 rcl_action_client_get_options(const rcl_action_client_t * action_client);
 
-/// Check that a rcl_action_clieint_t is valid.
+/// Check that a rcl_action_client_t is valid.
 /**
  * The bool returned is `false` if `action_client` is invalid.
  * The bool returned is `true` otherwise.
@@ -677,14 +676,12 @@ rcl_action_client_get_options(const rcl_action_client_t * action_client);
  * Lock-Free          | Yes
  *
  * \param[in] action_client pointer to the rcl action client
- * \param[in] error_msg_allocator a valid allocator or `NULL`
  * \return `true` if `action_client` is valid, otherwise `false`
  */
 RCL_ACTION_PUBLIC
 bool
 rcl_action_client_is_valid(
-  const rcl_action_client_t * action_client,
-  rcl_allocator_t * error_msg_allocator);
+  const rcl_action_client_t * action_client);
 
 #ifdef __cplusplus
 }
--- a/rcl_action/include/rcl_action/action_client.h
+++ b/rcl_action/include/rcl_action/action_client.h
@@ -167,6 +167,7 @@ rcl_action_get_zero_initialized_client(void);
  * \return `RCL_RET_OK` if action_client was initialized successfully, or
  * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
  * \return `RCL_RET_NODE_INVALID` if the node is invalid, or
+ * \return `RCL_RET_ALREADY_INIT` if the action client is already initialized, or
  * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
  * \return `RCL_RET_ACTION_NAME_INVALID` if the given action name is invalid, or
  * \return `RCL_RET_ERROR` if an unspecified error occurs.
--- a/rcl_action/include/rcl_action/action_client.h
+++ b/rcl_action/include/rcl_action/action_client.h
@@ -20,10 +20,6 @@ extern "C"
 {
 #endif
 
-// TODO(jacobperron): replace type support typedef with one defined in rosdl_generator_c
-// #include "rosidl_generator_c/action_type_support_struct.h"
-typedef struct rosidl_action_type_support_t rosidl_action_type_support_t;
-
 #include "rcl_action/types.h"
 #include "rcl_action/visibility_control.h"
 #include "rcl/macros.h"
--- a/rcl_action/include/rcl_action/types.h
+++ b/rcl_action/include/rcl_action/types.h
@@ -31,6 +31,8 @@ extern "C"
 #include "rcl/macros.h"
 #include "rcl/types.h"
 
+#include "rosidl_generator_c/action_type_support_struct.h"
+typedef struct rosidl_action_type_support_t rosidl_action_type_support_t;
 
 // rcl action specific ret codes in 2XXX
 /// Action name does not pass validation return code.
--- a/rcl_action/include/rcl_action/types.h
+++ b/rcl_action/include/rcl_action/types.h
@@ -32,7 +32,6 @@ extern "C"
 #include "rcl/types.h"
 
 #include "rosidl_generator_c/action_type_support_struct.h"
-typedef struct rosidl_action_type_support_t rosidl_action_type_support_t;
 
 // rcl action specific ret codes in 2XXX
 /// Action name does not pass validation return code.
--- a/rcl_action/src/rcl_action/action_client.c
+++ b/rcl_action/src/rcl_action/action_client.c
@@ -55,244 +55,77 @@ rcl_action_get_zero_initialized_client(void)
   return null_action_client;
 }
 
-// \internal Initialize client for the given action's goal service.
-static rcl_ret_t rcl_action_goal_service_client_init(
-  rcl_client_t * goal_client,
-  const rcl_node_t * node,
-  const rosidl_service_type_support_t * goal_service_type_support,
-  const char * action_name,
-  const rcl_client_options_t * goal_client_options)
-{
-  assert(NULL != goal_client);
-  assert(NULL != node);
-  assert(NULL != goal_service_type_support);
-  assert(NULL != action_name);
-  assert(NULL != goal_client_options);
-
-  rcl_ret_t ret;
-  rcl_allocator_t allocator = goal_client_options->allocator;
-
-  char * goal_service_name = NULL;
-  ret = rcl_action_get_goal_service_name(action_name, allocator, &goal_service_name);
-  if (RCL_RET_OK != ret) {
-    if (RCL_RET_BAD_ALLOC == ret) {
-      return RCL_RET_BAD_ALLOC;
-    }
-    return RCL_RET_ERROR;
+// \internal Initializes an action client specific service client.
+#define CLIENT_INIT(Type) \
+  char * Type ## _service_name = NULL; \
+  ret = rcl_action_get_ ## Type ## _service_name(action_name, allocator, &Type ## _service_name); \
+  if (RCL_RET_OK != ret) { \
+    RCL_SET_ERROR_MSG("failed to get " #Type " service name"); \
+    if (RCL_RET_BAD_ALLOC == ret) { \
+      ret = RCL_RET_BAD_ALLOC; \
+    } else { \
+      ret = RCL_RET_ERROR; \
+    } \
+    goto fail; \
+  } \
+  rcl_client_options_t Type ## _service_client_options = { \
+    .qos = options->Type ## _service_qos, .allocator = allocator \
+  }; \
+  action_client->impl->Type ## _client = rcl_get_zero_initialized_client(); \
+  ret = rcl_client_init( \
+    &action_client->impl->Type ## _client, \
+    node, \
+    type_support->Type ## _service_type_support, \
+    Type ## _service_name, \
+    &Type ## _service_client_options); \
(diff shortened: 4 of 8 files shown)


rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,723 @@
...
+    goal_service_name,
+    goal_client_options);
+
+  allocator.deallocate(goal_service_name, allocator.state);
+
+  if (RCL_RET_OK != ret) {
+    if (RCL_RET_SERVICE_NAME_INVALID == ret) {
+      ret = RCL_RET_ACTION_NAME_INVALID;
+    } else if (RCL_RET_BAD_ALLOC != ret) {
+      ret = RCL_RET_ERROR;
```
sloretz: Same comment about keeping `RCL_RET_BAD_ALLOC` return code

rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,723 @@
...
+            .ignore_local_publications = false,
+            .allocator = *allocator
+          };
+          RCUTILS_LOG_DEBUG_NAMED(
+            ROS_PACKAGE_NAME, "Initializing action status subscription");
+          ret = rcl_action_status_subscription_init(
+            &impl->status_subscription, node,
+            &type_support->status_topic_type_support,
+            impl->action_name, &status_subscription_options);
+          if (RCL_RET_OK == ret) {
```
sloretz: I think in cases like this `goto error:` is ok to avoid nested `if` statements. No need to change what's here though.

rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,723 @@
...
+  RCL_CHECK_ARGUMENT_FOR_NULL(node, RCL_RET_INVALID_ARGUMENT);
+  if (!rcl_node_is_valid(node)) {
+    return RCL_RET_NODE_INVALID;
+  }
+  RCL_CHECK_ARGUMENT_FOR_NULL(type_support, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(action_name, RCL_RET_INVALID_ARGUMENT);
+  RCUTILS_LOG_DEBUG_NAMED(
+    ROS_PACKAGE_NAME, "Initializing client for action name '%s'", action_name);
+  // Allocate space for the implementation struct.
+  rcl_action_client_impl_t *impl = allocator->allocate(
```
sloretz: Should error if `NULL != action_client->impl` before allocating to avoid leaking memory if `action_client` is already initialized.

rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,723 @@
...
+      };
+      RCUTILS_LOG_DEBUG_NAMED(
+        ROS_PACKAGE_NAME, "Initializing action result client");
+      ret = rcl_action_result_client_init(
+        &impl->result_client, node,
+        &type_support->result_service_type_support,
+        impl->action_name, &result_client_options);
+      if (RCL_RET_OK == ret) {
+        RCUTILS_LOG_DEBUG_NAMED(
+          ROS_PACKAGE_NAME, "Action cancel result initialized");
```
sloretz: `cancel result` -> `get result`

rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,723 @@
...
+{
+  RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME, "Finalizing action client");
+  if (!rcl_action_client_is_valid(action_client)) {
+    return RCL_RET_ACTION_CLIENT_INVALID;
+  }
+  if (!rcl_node_is_valid(node)) {
+    return RCL_RET_NODE_INVALID;
+  }
+  rcl_ret_t ret = RCL_RET_OK;
+  if (action_client->impl) {
```
sloretz: I think the `rcl_action_client_is_valid()` check above already checked this

rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,723 @@
...
+    // TODO(hidmic): ideally we should rollback to a valid state if any
+    //               finalization failed, but it seems that's currently
+    //               not possible.
+    rcl_action_client_impl_t * impl = action_client->impl;
+    if (rcl_client_is_valid(&impl->goal_client)) {
+      ret = rcl_client_fini(&impl->goal_client, node);
+      if (RCL_RET_OK != ret) {
+        return RCL_RET_ERROR;
```
sloretz: Recommend returning the original error code here, same below

rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,723 @@
...
+  if (!rcl_node_is_valid(node)) {
+    return RCL_RET_NODE_INVALID;
+  }
+  rcl_ret_t ret = RCL_RET_OK;
+  if (action_client->impl) {
+    // TODO(hidmic): ideally we should rollback to a valid state if any
+    //               finalization failed, but it seems that's currently
+    //               not possible.
+    rcl_action_client_impl_t * impl = action_client->impl;
+    if (rcl_client_is_valid(&impl->goal_client)) {
```
sloretz: Probably don't need this, `rcl_client_fini` will check if the client is valid.

rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,723 @@
...
+  RCL_CHECK_ARGUMENT_FOR_NULL(ros_goal_response, RCL_RET_INVALID_ARGUMENT);
+  rmw_request_id_t ignored_request_header;
+  rcl_ret_t ret = rcl_take_response(
+    &action_client->impl->goal_client,
+    &ignored_request_header, ros_goal_response);
+  if (RCL_RET_OK != ret) {
+    if (RCL_RET_CLIENT_TAKE_FAILED == ret) {
+      ret = RCL_RET_ACTION_CLIENT_TAKE_FAILED;
+    } else {
+      ret = RCL_RET_ERROR;
```
sloretz: Same comment about returning original error code

rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,723 @@
...
+
+  rcl_ret_t ret;
+  rcl_allocator_t allocator = goal_client_options->allocator;
+
+  char * goal_service_name = NULL;
+  ret = rcl_action_get_goal_service_name(
+    action_name, allocator, &goal_service_name);
+  if (RCL_RET_OK != ret) {
+    if (RCL_RET_BAD_ALLOC != ret) {
+      ret = RCL_RET_ERROR;
```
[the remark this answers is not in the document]
hidmic: And it is! `ret` changes iff `ret` is NOT `RCL_RET_BAD_ALLOC` :)

rcl_action/src/rcl_action/action_client.c
hidmic: Same as above.

rcl_action/src/rcl_action/action_client.c
hidmic: I thought about using `goto`, but I went this route to simplify the rollback. Otherwise I would've needed a label for each failure stage (_which things have I initialized so far that need finalization to keep things consistent?_)

rcl_action/src/rcl_action/action_client.c
hidmic: So, I tried to be consistent with `rcl_action` documentation instead of propagating internal errors (that are usually hard to understand because of the lack of context). If we propagate, documentation needs an update.

rcl_action/src/rcl_action/action_client.c
hidmic: Yes, but there's an implicit purpose to this. If any of the five internal finalizations were to fail, some would end up finalized, some not. So this is an attempt to allow `rcl_action_client_fini` to be called once again and not fail unless your action client was completely finalized.

rcl_action/src/rcl_action/action_client.c
hidmic: Same about documentation. I'm fine either way, your call.

rcl_action/src/rcl_action/action_client.c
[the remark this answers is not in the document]
sloretz: Ah oops, My eyes missed the `!`

rcl_action/src/rcl_action/action_client.c
sloretz: I don't think a label is needed for each failure stage. It looks like `rcl_action_client_fini()` could be called instead of `fini`-ing everything individually

rcl_action/src/rcl_action/action_client.c
sloretz: It can stay the way it is. It looks like existing code in RCL also returns `RCL_RET_ERROR` instead of the original code.

https://github.com/ros2/rcl/blob/4d8cb487f86bc709c6c0988cdc3ce80544db4b31/rcl/src/rcl/client.c#L86

rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,723 @@
...
+} rcl_action_client_impl_t;
+
+rcl_action_client_t
+rcl_action_get_zero_initialized_client(void)
+{
+  static rcl_action_client_t null_action_client = {0};
+  return null_action_client;
+}
+
+static rcl_ret_t rcl_action_goal_service_client_init(
```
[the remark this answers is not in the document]
hidmic: Comments added in c5dabd05bc66f9c01431ec5cfe1b86815081c823, but I get warnings about attributes being ignored when using `RCL_ACTION_LOCAL` in combination with `static`. It kind of makes sense though if you think about it.

rcl_action/src/rcl_action/action_client.c
hidmic: Let's be consistent with rest of the codebase then :). Changed in c5dabd05bc66f9c01431ec5cfe1b86815081c823.

rcl_action/src/rcl_action/action_client.c
hidmic: Good catch. Done in c5dabd05bc66f9c01431ec5cfe1b86815081c823.

rcl_action/src/rcl_action/action_client.c
hidmic: Fixed in c5dabd05bc66f9c01431ec5cfe1b86815081c823.

rcl_action/src/rcl_action/action_client.c
hidmic: True that. Fixed in c5dabd05bc66f9c01431ec5cfe1b86815081c823.

rcl_action/src/rcl_action/action_client.c
hidmic: > then wouldn't the same failure happen again

That I don't think we can assume, some finalization errors come from deep into `rmw`. But I do like the idea of finalizing as much as possible. Changed in c5dabd05bc66f9c01431ec5cfe1b86815081c823.

rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,723 @@
...
+  assert(NULL != node);
+  assert(NULL != goal_service_type_support);
+  assert(NULL != action_name);
+  assert(NULL != goal_client_options);
+
+  rcl_ret_t ret;
+  rcl_allocator_t allocator = goal_client_options->allocator;
+
+  char * goal_service_name = NULL;
+  ret = rcl_action_get_goal_service_name(
```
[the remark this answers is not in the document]
hidmic: Fixed in c5dabd05bc66f9c01431ec5cfe1b86815081c823.

rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,723 @@
...
+  ret = rcl_subscription_init(
+    feedback_subscription, node,
+    feedback_message_type_support,
+    feedback_topic_name,
+    feedback_subscription_options);
+
+  allocator.deallocate(feedback_topic_name, allocator.state);
+
+  if (RCL_RET_OK != ret) {
+    if (RCL_RET_SERVICE_NAME_INVALID == ret) {
```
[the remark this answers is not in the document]
hidmic: Good catch! Fixed in c5dabd05bc66f9c01431ec5cfe1b86815081c823.

rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,723 @@
...
+  ret = rcl_subscription_init(
+    status_subscription, node,
+    status_message_type_support,
+    status_topic_name,
+    status_subscription_options);
+
+  allocator.deallocate(status_topic_name, allocator.state);
+
+  if (RCL_RET_OK != ret) {
+    if (RCL_RET_SERVICE_NAME_INVALID == ret) {
```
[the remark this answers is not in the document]
hidmic: Good catch! Fixed in c5dabd05bc66f9c01431ec5cfe1b86815081c823.

rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,723 @@
...
+  const rosidl_action_type_support_t * type_support,
+  const char * action_name,
+  const rcl_action_client_options_t * options)
+{
+  RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME, "Initializing action client");
+  RCL_CHECK_ARGUMENT_FOR_NULL(options, RCL_RET_INVALID_ARGUMENT);
+  rcl_allocator_t * allocator = (rcl_allocator_t *)&options->allocator;
+  RCL_CHECK_ALLOCATOR_WITH_MSG(
+    allocator, "invalid allocator", return RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(action_client, RCL_RET_INVALID_ARGUMENT);
```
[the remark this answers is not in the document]
hidmic: Done in c5dabd05bc66f9c01431ec5cfe1b86815081c823.

rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,723 @@
...
+
+rcl_ret_t
+rcl_action_client_init(
+  rcl_action_client_t * action_client,
+  rcl_node_t * node,
+  const rosidl_action_type_support_t * type_support,
+  const char * action_name,
+  const rcl_action_client_options_t * options)
+{
+  RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME, "Initializing action client");
```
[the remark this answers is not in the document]
hidmic: Yeap, the first one is unnecessary. Fixed in c5dabd05bc66f9c01431ec5cfe1b86815081c823.

rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,723 @@
...
+    ROS_PACKAGE_NAME, "Failed to initialize action client");
+  allocator->deallocate(impl->action_name, allocator->state);
+  allocator->deallocate(impl, allocator->state);
+  return ret;
+}
+
+rcl_ret_t
+rcl_action_client_fini(rcl_action_client_t * action_client, rcl_node_t * node)
+{
+  RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME, "Finalizing action client");
```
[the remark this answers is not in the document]
hidmic: I see, and it makes sense. Fixed in c5dabd05bc66f9c01431ec5cfe1b86815081c823.

hidmic:
So, about making the API thicker. Per offline discussion with @sloretz and @jacobperron, I think we can agree that we need a mechanism to retrieve the goal ID from the type-erased message _before_ we can start pushing more logic into this layer.

@sloretz I'll push some changes to ros2/rcl_interfaces#47 as a POC for a conversion function within generated `rosidl_action_type_support_t` instances. cc @apojomovsky as it impacts your ongoing work.

rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,706 @@
...
+  assert(NULL != action_name);
+  assert(NULL != result_client_options);
+
+  rcl_ret_t ret;
+  rcl_allocator_t allocator = result_client_options->allocator;
+
+  char * result_service_name = NULL;
+  ret = rcl_action_get_result_service_name(action_name, allocator, &result_service_name);
+  if (RCL_RET_OK != ret) {
+    if (RCL_RET_BAD_ALLOC != ret) {
```
jacobperron: Might be more clearly written as:

```c
if (RCL_RET_BAD_ALLOC == ret) {
  return RCL_RET_BAD_ALLOC;
}
return RCL_RET_ERROR;
```

hidmic: Makes sense.

hidmic:
On 011dc30212146e3c28541ac141494af2615af9b7, I assumed that:

1. the client has to be hooked to the status channel, unconditionally.
2. the wait set is used by the action client and the action client alone (same would go for the action server).

I'm not so sure about (1), API documentation was not consistent about the need of status. (2) seems restrictive to me but a reasonable trade off until [wait set groups](https://github.com/ros2/rcl/issues/308) are implemented.

hidmic:
@jacobperron I added basic tests. Let me know if/when we can move forward with having a single set of integration tests as suggested in https://github.com/ros2/rcl/pull/323/files#r232681489 and I'll add those.

hidmic:
This PR still doesn't include any logic as that requires typesupport we do not yet have. I'd like to know what the rest have to say in tomorrow's retrospective and how https://github.com/ros2/rcl/pull/323/files#r232674670 unfolds. May be a neat path forward.

rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,812 @@
...
+  rcl_wait_set_t * wait_set,
+  const rcl_action_client_t * action_client)
+{
+  rcl_ret_t ret;
+  RCL_CHECK_ARGUMENT_FOR_NULL(wait_set, RCL_RET_WAIT_SET_INVALID);
+  if (!rcl_action_client_is_valid(action_client)) {
+    return RCL_RET_ACTION_CLIENT_INVALID;  // error already set
+  }
+  // Wait on action goal service response messages.
+  ret = rcl_wait_set_add_client(
```
jacobperron: nit: fit on single line. Same for other lines below.

rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,812 @@
...
+rcl_ret_t
+rcl_action_client_init(
+  rcl_action_client_t * action_client,
+  rcl_node_t * node,
+  const rosidl_action_type_support_t * type_support,
+  const char * action_name,
+  const rcl_action_client_options_t * options)
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(action_client, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(node, RCL_RET_INVALID_ARGUMENT);
```
jacobperron: `rcl_node_is_valid()` already checks for a null pointer, so this line can be removed.

rcl_action/src/rcl_action/action_client.c
```diff
@@ -0,0 +1,812 @@
...
+      ret = RCL_RET_ERROR;
+    }
+  }
+  if (rcl_subscription_is_valid(&impl->status_subscription)) {
+    fini_ret = rcl_subscription_fini(&impl->status_subscription, node);
+    if (RCL_RET_OK != fini_ret) {
+      ret = RCL_RET_ERROR;
+    }
+  }
+  if (RCL_RET_OK != ret) {
```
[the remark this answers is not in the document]
hidmic: Alright, let's be consistent. Fixed in 0c313821659b90f03650dfcefcbff453acba82f6.

rcl_action/src/rcl_action/action_client.c
hidmic: Fixed in 0c313821659b90f03650dfcefcbff453acba82f6.

rcl_action/src/rcl_action/action_client.c
hidmic: Removed in 0c313821659b90f03650dfcefcbff453acba82f6.

hidmic:
@jacobperron since integration tests will likely need `rcl_wait` support for both client and server, let's move forward with this and get back to them when we get to refactor wait set logic for the client to use groups. What do you think?

jacobperron:
> since integration tests will likely need rcl_wait support for both client and server, let's move forward with this and get back to them when we get to refactor wait set logic for the client to use groups. What do you think?

@hidmic 
Sounds good. I've created a task for adding integration tests: #328

rcl_action/include/rcl_action/types.h
```diff
@@ -31,6 +31,8 @@ extern "C"
 #include "rcl/macros.h"
 #include "rcl/types.h"
 
+#include "rosidl_generator_c/action_type_support_struct.h"
+typedef struct rosidl_action_type_support_t rosidl_action_type_support_t;
```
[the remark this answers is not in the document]
hidmic: Done in 78168bb52afa82b2b16db36108fd8de254b74c01.

rcl_action/include/rcl_action/action_client.h
```diff
@@ -420,6 +415,7 @@ rcl_action_take_status(
   const rcl_action_client_t * action_client,
   void * ros_status_array);
 
+
```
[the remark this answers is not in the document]
hidmic: Argh, I forgot about the macros. Fixed in c7e61834b4ac29d1d087b773c9cc976b0f996f95.

hidmic:
@apojomovsky @jacobperron @sloretz may I ask one of you guys to start CI on this one for me (nope, I can't do it myself :( ) ?

sloretz:
CI up to rcl_action, testing only rcl_action

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5541/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2198/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4573/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5434/)

hidmic:
Alright! Let's get this one in.

rcl_action/CMakeLists.txt
```diff
@@ -69,9 +70,22 @@ install(TARGETS ${PROJECT_NAME}
 if(BUILD_TESTING)
   find_package(ament_cmake_gtest REQUIRED)
   find_package(ament_lint_auto REQUIRED)
+  find_package(test_msgs REQUIRED)
   ament_lint_auto_find_test_dependencies()
   ament_find_gtest()
   # Gtests
+  ament_add_gtest(test_action_client
+    test/rcl_action/test_action_client.cpp
+  )
```
dirk-thomas: This test is only covering the default rmw impl. It would be good if the test would be invoked for every rmw impl.

dirk-thomas:
@hidmic Is this branch still necessary or can it be deleted?

hidmic:
@dirk-thomas branch deleted, thanks for the FYI.

PR #321, opened 2018-11-05
[rcl] Refactor error checking

jacobperron:
- Remove redundant error checking
  - Checking for null pointers is delegated to the `_is_valid()` function (if such a function exists).
  - Removed null pointer check for "options" since it is not a pointer type.
- Change returned error codes to be more explicit
  - Replace `RCL_RET_INVALID_ARGUMENT` with `RCL_RET_*_INVALID` where possible.
  - Return `RCL_RET_BAD_ALLOC` where appropriate.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5466/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2151/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4523/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5385/)

### DIFF
--- a/rcl/src/rcl/client.c
+++ b/rcl/src/rcl/client.c
@@ -62,9 +62,8 @@ rcl_client_init(
   rcl_allocator_t * allocator = (rcl_allocator_t *)&options->allocator;
   RCL_CHECK_ALLOCATOR_WITH_MSG(allocator, "invalid allocator", return RCL_RET_INVALID_ARGUMENT);
   RCL_CHECK_ARGUMENT_FOR_NULL(client, RCL_RET_INVALID_ARGUMENT);
-  RCL_CHECK_ARGUMENT_FOR_NULL(node, RCL_RET_INVALID_ARGUMENT);
   if (!rcl_node_is_valid(node)) {
-    return RCL_RET_NODE_INVALID;
+    return RCL_RET_NODE_INVALID;  // error already set
   }
   RCL_CHECK_ARGUMENT_FOR_NULL(type_support, RCL_RET_INVALID_ARGUMENT);
   RCL_CHECK_ARGUMENT_FOR_NULL(service_name, RCL_RET_INVALID_ARGUMENT);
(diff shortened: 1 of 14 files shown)


hidmic:
@jacobperron wait sets still keep the fuzzy distinction between invalid argument and invalid entity, see [here](https://github.com/ros2/rcl/blob/master/rcl/include/rcl/wait.h#L181).

wjwwood:
Is this ready for review?

jacobperron:
> Is this ready for review?

Sure :)

rcl/src/rcl/client.c
```diff
@@ -316,13 +312,9 @@ rcl_take_response(
...
 rcl_client_is_valid(const rcl_client_t * client)
 {
-  const rcl_client_options_t * options;
-  RCL_CHECK_ARGUMENT_FOR_NULL(client, false);
+  RCL_CHECK_FOR_NULL_WITH_MSG(client, "client pointer is invalid", return false);
   RCL_CHECK_FOR_NULL_WITH_MSG(
     client->impl, "client's rmw implementation is invalid", return false);
-  options = _client_get_options(client);
-  RCL_CHECK_FOR_NULL_WITH_MSG(
-    options, "client's options pointer is invalid", return false);
```
wjwwood: What's the rationale for removing the check of the options? Does it prevent this functions reuse somewhere else or something?

jacobperron: I thought it was redundant since `options` is not a pointer in the underlying impl struct, which is already checked in the line above.
The private function `_client_get_options()` returns the address, and the user-facing `rcl_client_get_options()` calls `rcl_client_is_valid()` anyways.

jacobperron:
> wait sets still keep the fuzzy distinction between invalid argument and invalid entity, see [here](https://github.com/ros2/rcl/blob/master/rcl/include/rcl/wait.h#L181).

@hidmic I was considering tackling wait sets, timers, and guard conditions later since they do not have the same usage pattern of `_is_valid()`.

rcl/src/rcl/client.c
wjwwood: Makes sense, 👍

PR #323, opened 2018-11-08
Action server implementation

jacobperron:
First iteration of implementation.

Connects to #306 
Depends on ros2/rcl_interfaces#47
Depends on ros2/rosidl#310

### DIFF
--- a/rcl_action/CMakeLists.txt
+++ b/rcl_action/CMakeLists.txt
@@ -73,6 +73,7 @@ install(TARGETS ${PROJECT_NAME}
 )
 
 if(BUILD_TESTING)
+  find_package(test_msgs REQUIRED)
   find_package(ament_cmake_gtest REQUIRED)
   find_package(ament_lint_auto REQUIRED)
   find_package(test_msgs REQUIRED)
@@ -98,11 +99,13 @@ if(BUILD_TESTING)
     target_include_directories(test_action_server PUBLIC
       include
       ${rcl_INCLUDE_DIRS}
+      ${test_msgs_INCLUDE_DIRS}
     )
     target_link_libraries(test_action_server
       ${PROJECT_NAME}
       ${rcl_LIBRARIES}
       ${rcutils_LIBRARIES}
+      ${test_msgs_LIBRARIES}
     )
   endif()
   ament_add_gtest(test_goal_handle
--- a/rcl_action/CMakeLists.txt
+++ b/rcl_action/CMakeLists.txt
@@ -33,6 +33,7 @@ add_executable(test_compile_headers
 
 set(rcl_action_sources
   src/${PROJECT_NAME}/action_client.c
+  src/${PROJECT_NAME}/action_server.c
   src/${PROJECT_NAME}/goal_handle.c
   src/${PROJECT_NAME}/goal_state_machine.c
   src/${PROJECT_NAME}/names.c
@@ -85,6 +86,18 @@ if(BUILD_TESTING)
       ${PROJECT_NAME}
     )
     ament_target_dependencies(test_action_client "rcl" "test_msgs")
+  endif()
+  ament_add_gtest(test_action_server
+    test/rcl_action/test_action_server.cpp
+  )
+  if(TARGET test_action_server)
+    target_include_directories(test_action_server PUBLIC
+      include
+      ${rcl_INCLUDE_DIRS}
+    )
+    target_link_libraries(test_action_server
+      ${PROJECT_NAME}
+    )
   endif()
   ament_add_gtest(test_goal_handle
     test/rcl_action/test_goal_handle.cpp
--- a/rcl_action/CMakeLists.txt
+++ b/rcl_action/CMakeLists.txt
@@ -8,6 +8,7 @@ find_package(rosidl_generator_c REQUIRED)
 find_package(action_msgs REQUIRED)
 find_package(rcl REQUIRED)
 find_package(rcutils REQUIRED)
+find_package(rmw REQUIRED)
 
 include_directories(
   include
@@ -54,6 +55,7 @@ add_library(${PROJECT_NAME}
 ament_target_dependencies(${PROJECT_NAME}
   "rosidl_generator_c"
   "action_msgs"
+  "rmw"
   "rcutils"
   "rcl"
 )
@@ -159,6 +161,8 @@ ament_export_include_directories(include)
 ament_export_libraries(${PROJECT_NAME})
 ament_export_dependencies(ament_cmake)
 ament_export_dependencies(rcl)
+ament_export_dependencies(rcutils)
+ament_export_dependencies(rmw)
 ament_export_dependencies(action_msgs)
 ament_export_dependencies(rosidl_generator_c)
 ament_package()
--- a/rcl_action/CMakeLists.txt
+++ b/rcl_action/CMakeLists.txt
@@ -4,8 +4,10 @@ project(rcl_action)
 
 find_package(ament_cmake_ros REQUIRED)
 
+find_package(rosidl_generator_c REQUIRED)
 find_package(action_msgs REQUIRED)
 find_package(rcl REQUIRED)
+find_package(rcutils REQUIRED)
 
 include_directories(
   include
@@ -50,8 +52,10 @@ add_library(${PROJECT_NAME}
 )
 
 ament_target_dependencies(${PROJECT_NAME}
-  "rcl"
+  "rosidl_generator_c"
   "action_msgs"
+  "rcutils"
+  "rcl"
 )
 # Causes the visibility macros to use dllexport rather than dllimport,
 # which is appropriate when building the dll but not consuming it.
@@ -97,6 +101,8 @@ if(BUILD_TESTING)
     )
     target_link_libraries(test_action_server
       ${PROJECT_NAME}
+      ${rcl_LIBRARIES}
+      ${rcutils_LIBRARIES}
     )
   endif()
   ament_add_gtest(test_goal_handle
--- a/rcl_action/CMakeLists.txt
+++ b/rcl_action/CMakeLists.txt
@@ -103,10 +103,8 @@ if(BUILD_TESTING)
     )
     target_link_libraries(test_action_server
       ${PROJECT_NAME}
-      ${rcl_LIBRARIES}
-      ${rcutils_LIBRARIES}
-      ${test_msgs_LIBRARIES}
     )
+    ament_target_dependencies(test_action_server "test_msgs")
   endif()
   ament_add_gtest(test_goal_handle
     test/rcl_action/test_goal_handle.cpp
@@ -162,4 +160,5 @@ ament_export_libraries(${PROJECT_NAME})
 ament_export_dependencies(ament_cmake)
 ament_export_dependencies(rcl)
 ament_export_dependencies(action_msgs)
+ament_export_dependencies(rosidl_generator_c)
 ament_package()
--- a/rcl_action/CMakeLists.txt
+++ b/rcl_action/CMakeLists.txt
@@ -94,6 +94,20 @@ if(BUILD_TESTING)
     )
     ament_target_dependencies(test_action_client "rcl" "test_msgs")
   endif()
+  ament_add_gtest(test_action_communication
+    test/rcl_action/test_action_communication.cpp
+  )
+  if(TARGET test_action_communication)
+    target_include_directories(test_action_communication PUBLIC
+      include
+      ${rcl_INCLUDE_DIRS}
+      ${test_msgs_INCLUDE_DIRS}
+    )
+    target_link_libraries(test_action_communication
+      ${PROJECT_NAME}
+    )
+    ament_target_dependencies(test_action_communication "test_msgs")
+  endif()
   ament_add_gtest(test_action_server
     test/rcl_action/test_action_server.cpp
   )
--- a/rcl_action/include/rcl_action/action_server.h
+++ b/rcl_action/include/rcl_action/action_server.h
@@ -66,7 +66,7 @@ RCL_WARN_UNUSED
 rcl_action_server_t
 rcl_action_get_zero_initialized_server(void);
 
-/// Initialize a rcl_action_server_t.
+/// Initialize an action server.
 /**
  * After calling this function on a rcl_action_server_t, it can be used to take
  * goals of the given type for the given action name using rcl_action_take_goal_request()
@@ -150,9 +150,9 @@ rcl_action_get_zero_initialized_server(void);
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param[out] action_server a preallocated, zero-initialized action server structure
+ * \param[out] action_server handle to a preallocated, zero-initialized action server structure
  *   to be initialized.
- * \param[in] node valid rcl node handle
+ * \param[in] node valid node handle
  * \param[in] type_support type support object for the action's type
  * \param[in] action_name the name of the action
  * \param[in] options action_server options, including quality of service settings
@@ -173,9 +173,9 @@ rcl_action_server_init(
   const char * action_name,
   const rcl_action_server_options_t * options);
 
-/// Finalize a rcl_action_server_t.
+/// Finalize an action server.
 /**
- * After calling, the node will no longer listen for goals for this action server.
+ * After calling, the node will no longer listen to services and topics for this action server.
  * (assuming this is the only action server of this type in this node).
  *
  * After calling, calls to rcl_wait(), rcl_action_take_goal_request(),
@@ -223,7 +223,7 @@ RCL_WARN_UNUSED
 rcl_action_server_options_t
 rcl_action_server_get_default_options(void);
 
-/// Take a pending ROS goal using a rcl_action_server_t.
+/// Take a pending ROS goal using an action server.
 /**
  * \todo TODO(jacobperron) blocking of take?
  *
@@ -253,7 +253,7 @@ rcl_action_server_get_default_options(void);
  * Lock-Free          | Yes
  * <i>[1] only if required when filling the request, avoided for fixed sizes</i>
  *
- * \param[in] action_server the action server that will take the request
+ * \param[in] action_server handle to the action server that will take the request
  * \param[out] ros_goal_request a preallocated, zero-initialized, ROS goal request message
  *   where the request is copied
  * \return `RCL_RET_OK` if the request was taken, or
@@ -271,7 +271,7 @@ rcl_action_take_goal_request(
   const rcl_action_server_t * action_server,
   void * ros_goal_request);
 
-/// Send a response for a goal request to an action client using a rcl_action_server_t.
+/// Send a response for a goal request to an action client using an action server.
 /**
  * This is a non-blocking call.
  *
@@ -317,16 +317,24 @@ rcl_action_send_goal_response(
   const rcl_action_server_t * action_server,
   void * ros_goal_response);
 
-/// Accept a new goal using a rcl_action_server_t.
+/// Accept a new goal using an action server.
 /**
  * This is a non-blocking call.
  *
- * Creates and returns a pointer to a goal handle for a newly accepted goal.
+ * Creates and returns a new goal handle.
+ * The action server starts tracking it internally.
  * If a failure occurs, `NULL` is returned and an error message is set.
+ * Possible reasons for failure:
+ *   - action server is invalid
+ *   - goal info is invalid
+ *   - goal ID is already being tracked by the action server
+ *   - memory allocation failure
  *
  * This function should be called after receiving a new goal request with
  * rcl_action_take_goal_request() and before sending a response with
  * rcl_action_send_goal_response().
+ *
+ * After calling this function, the action server will start tracking the goal.
  *
  * Example usage:
  *
@@ -358,7 +366,7 @@ rcl_action_send_goal_response(
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param[in] action_server the action server that is accepting the goal
+ * \param[in] action_server handle to the action server that is accepting the goal
  * \param[in] goal_info a message containing info about the goal being accepted
  * \return a pointer to a new goal handle representing the accepted goal, or
  * \return `NULL` if a failure occured.
@@ -370,7 +378,7 @@ rcl_action_accept_new_goal(
   rcl_action_server_t * action_server,
   const rcl_action_goal_info_t * goal_info);
 
-/// Publish a ROS feedback message for an active goal using a rcl_action_server_t.
+/// Publish a ROS feedback message for an active goal using an action server.
 /**
  * The caller is responsible for ensuring that the type of `ros_feedback`
  * and the type associate with the client (via the type support) match.
@@ -418,7 +426,7 @@ rcl_action_publish_feedback(
   const rcl_action_server_t * action_server,
   void * ros_feedback);
 
-/// Get a status array message for accepted goals associated with a rcl_action_server_t.
+/// Get a status array message for accepted goals associated with an action server.
 /**
  * The provided `status_message` should be zero-initialized with
  * rcl_action_get_zero_initialized_goal_status_array() before calling this function.
@@ -445,7 +453,7 @@ rcl_action_get_goal_status_array(
   const rcl_action_server_t * action_server,
   rcl_action_goal_status_array_t * status_message);
 
-/// Publish a status array message for accepted goals associated with a rcl_action_server_t.
+/// Publish a status array message for accepted goals associated with an action server.
 /**
  * This function acts like a ROS publisher and is potentially a blocking call.
  * \see rcl_publish()
@@ -475,7 +483,7 @@ rcl_action_publish_status(
   const rcl_action_server_t * action_server,
   const void * status_message);
 
-/// Take a pending result request using a rcl_action_server_t.
+/// Take a pending result request using an action server.
 /**
  * \todo TODO(jacobperron) blocking of take?
  *
@@ -513,7 +521,7 @@ rcl_action_take_result_request(
   const rcl_action_server_t * action_server,
   void * ros_result_request);
 
-/// Send a result response using a rcl_action_server_t.
+/// Send a result response using an action server.
 /**
  * This is a non-blocking call.
  *
@@ -547,7 +555,7 @@ rcl_action_send_result_response(
   const rcl_action_server_t * action_server,
   void * ros_result_response);
 
-/// Clear all expired goals associated with a rcl_action_server_t.
+/// Clear all expired goals associated with an action server.
 /**
  * A goal is 'expired' if it has been in a terminal state (has a result) for more
  * than some duration.
@@ -581,7 +589,7 @@ rcl_action_clear_expired_goals(
   const rcl_action_server_t * action_server,
   size_t * num_expired);
 
-/// Take a pending cancel request using a rcl_action_server_t.
+/// Take a pending cancel request using an action server.
 /**
  * \todo TODO(jacobperron) blocking of take?
  *
@@ -622,7 +630,7 @@ rcl_action_take_cancel_request(
   const rcl_action_server_t * action_server,
   void * ros_cancel_request);
 
-/// Process a cancel request using a rcl_action_server_t.
+/// Process a cancel request using an action server.
 /**
  * This is a non-blocking call.
  *
@@ -666,7 +674,7 @@ rcl_action_process_cancel_request(
   const rcl_action_cancel_request_t * cancel_request,
   rcl_action_cancel_response_t * cancel_response);
 
-/// Send a cancel response using a rcl action server.
+/// Send a cancel response using an action server.
 /**
  * This is a non-blocking call.
  *
@@ -678,7 +686,7 @@ rcl_action_process_cancel_request(
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param[in] action_server the handle to the action server that will send the cancel response
+ * \param[in] action_server handle to the action server that will send the cancel response
  * \param[in] ros_cancel_response a ROS cancel response to send
  * \return `RCL_RET_OK` if the request was taken, or
  * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
@@ -692,12 +700,12 @@ rcl_action_send_cancel_response(
   const rcl_action_server_t * action_server,
   void * ros_cancel_response);
 
-/// Get the name of the action for a rcl_action_server_t.
+/// Get the action name for an action server.
 /**
  * This function returns the action server's internal topic name string.
  * This function can fail, and therefore return `NULL`, if the:
  *   - action server is `NULL`
- *   - action server is invalid (never called init, called fini, or invalid)
+ *   - action server is invalid (e.g. never called init or called fini)
  *
  * The returned string is only valid as long as the action server is valid.
  * The value of the string may change if the topic name changes, and therefore
@@ -711,20 +719,21 @@ rcl_action_send_cancel_response(
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param[in] action_server the pointer to the action server
- * \return name string if successful, otherwise `NULL`
+ * \param[in] action_server pointer to the action server
+ * \return name string if successful, or
+ * \return `NULL` otherwise.
  */
 RCL_ACTION_PUBLIC
 RCL_WARN_UNUSED
 const char *
 rcl_action_server_get_action_name(const rcl_action_server_t * action_server);
 
-/// Return the rcl_action_server_options_t for a rcl_action_server_t.
+/// Return the rcl_action_server_options_t for an action server.
 /**
  * This function returns the action server's internal options struct.
  * This function can fail, and therefore return `NULL`, if the:
  *   - action server is `NULL`
- *   - action server is invalid (never called init, called fini, or invalid)
+ *   - action server is invalid (e.g. never called init or called fini)
  *
  * The returned struct is only valid as long as the action server is valid.
  * The values in the struct may change if the action server's options change,
@@ -738,15 +747,16 @@ rcl_action_server_get_action_name(const rcl_action_server_t * action_server);
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param[in] action_server pointer to the action server
- * \return options struct if successful, otherwise `NULL`
+ * \param[in] action_server handle to the action server
+ * \return options struct if successful, or
+ * \return `NULL` otherwise.
  */
 RCL_ACTION_PUBLIC
 RCL_WARN_UNUSED
 const rcl_action_server_options_t *
 rcl_action_server_get_options(const rcl_action_server_t * action_server);
 
-/// Get the goal handles for all active or terminated goals.
+/// Get the goal handles for all goals an action server is tracking.
 /**
  * A pointer to the internally held array of pointers to goal handle structs is returned
  * along with the number of items in the array.
@@ -769,13 +779,14 @@ rcl_action_server_get_options(const rcl_action_server_t * action_server);
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param[in] action_server pointer to the rcl action server
+ * \param[in] action_server handle to the action server
  * \param[out] goal_handles is set to the array of pointers to goal handles if successful.
  * \param[out] num_goals is set to the number of goals in the returned array if successful,
  *   not set otherwise.
- * \return RCL_RET_OK if successful, or
- * \return RCL_RET_ACTION_SERVER_INVALID if the action server is invalid, or
- * \return RCL_RET_INVALID_ARGUMENT if any arguments are invalid
+ * \return `RCL_RET_OK` if successful, or
+ * \return `RCL_RET_ACTION_SERVER_INVALID` if the action server is invalid, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
  */
 RCL_ACTION_PUBLIC
 RCL_WARN_UNUSED
@@ -785,9 +796,23 @@ rcl_action_server_get_goal_handles(
   rcl_action_goal_handle_t *** goal_handles,
   size_t * num_goals);
 
-/// Check if a goal is already being tracked by an action server
+/// Check if a goal is already being tracked by an action server.
 /**
- * TODO(jacobperron): Document
+ * Checks whether or not a goal is being tracked in the internal goal array.
+ * The goal state has no effect on the return value.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] action_server handle to the action server
+ * \param[in] goal_info handle to a struct containing the goal ID to check for
+ * \return `true` if `action_server` is currently tracking a goal with the provided goal ID, or
+ * \return `false` otherwise.
  */
 RCL_ACTION_PUBLIC
 RCL_WARN_UNUSED
@@ -796,12 +821,11 @@ rcl_action_server_goal_exists(
   const rcl_action_server_t * action_server,
   const rcl_action_goal_info_t * goal_info);
 
-/// Check that the action server is valid.
+/// Check if an action server is valid.
 /**
- * The bool returned is `false` if `action_server` is invalid.
- * The bool returned is `true` otherwise.
- * In the case where `false` is to be returned, an
- * error message is set.
+ * In the case where `false` is returned (ie. the action server is invalid),
+ * an error message is set.
+ *
  * This function cannot fail.
  *
  * <hr>
@@ -812,8 +836,9 @@ rcl_action_server_goal_exists(
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param[in] action_server pointer to the rcl action server
- * \return `true` if `action_server` is valid, otherwise `false`
+ * \param[in] action_server handle to the action server
+ * \return `true` if `action_server` is valid, or
+ * \return `false` otherwise.
  */
 RCL_ACTION_PUBLIC
 RCL_WARN_UNUSED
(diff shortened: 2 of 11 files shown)


jacobperron:
Intermediate CI to find any issues:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5505/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2173/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4545/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5407/)

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,644 @@
...
+      &action_server->impl->goal_handles[i], &status_message->msg.status_list.data[i].status);
+    if (RCL_RET_OK != ret) {
+      return RCL_RET_ERROR;
+    }
+  }
+  return RCL_RET_OK;
+}
+
+rcl_ret_t
+rcl_action_publish_status(
```
hidmic: @jacobperron IIUC here we're passing around pointers to `rcl_action_goal_status_array_t` structs instead of their client library specific representation to free said library from having to deal with statuses in any way and just `rcl_action_get_goal_status_array(); rcl_action_publish_status();` them. But I wonder, will it work? Isn't the status publisher initialized with client library specific type support? 

---- 

Digging into rosidl* packages, it seems it _might_ actually be **currently** safe if `struct` layouts in C and C++ are fully compatible (which I'd expect). Which opens up very interesting possibilities, since AFAIK the following type cast in C is safe:

```c
struct A {
  int i;
};

struct B {
  struct A a;
  int j;
};

int main() {
  struct B b = {{0}, 0};
  struct A * a = (struct A *)&a;
  printf("%d", a->i);
}
```

We could exploit this to peek into otherwise type-erased generated messages safely from within `rcl_action`.

rcl_action/test/rcl_action/test_action_server.cpp
```diff
@@ -0,0 +1,738 @@
...
+  EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+
+  ret = rcl_shutdown();
+  EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+}
+
+class TestActionServer : public ::testing::Test
+{
+protected:
+  void SetUp()
```
hidmic: @jacobperron missing trailing `override`.

rcl_action/test/rcl_action/test_action_server.cpp
```diff
@@ -0,0 +1,738 @@
...
+    const rosidl_action_type_support_t * ts = ROSIDL_GET_ACTION_TYPE_SUPPORT(
+      test_msgs, Fibonacci);
+    const rcl_action_server_options_t options = rcl_action_server_get_default_options();
+    const char * action_name = "test_action_server_name";
+    this->action_server = rcl_action_get_zero_initialized_server();
+    ret = rcl_action_server_init(&this->action_server, &this->node, ts, action_name, &options);
+    ASSERT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
+  }
+
+  void TearDown()
```
hidmic: @jacobperron missing trailing `override`.

rcl_action/test/rcl_action/test_action_server.cpp
```diff
@@ -0,0 +1,738 @@
...
+  rcl_action_cancel_request_t cancel_request = rcl_action_get_zero_initialized_cancel_request();
+  cancel_request.goal_info.stamp.sec = 21;
+  rcl_action_cancel_response_t cancel_response = rcl_action_get_zero_initialized_cancel_response();
+  rcl_ret_t ret = rcl_action_process_cancel_request(
+    &this->action_server, &cancel_request, &cancel_response);
+  EXPECT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
+  EXPECT_NE(cancel_response.msg.goals_canceling.data, nullptr);
+  ASSERT_EQ(cancel_response.msg.goals_canceling.size, 22u);
+  rcl_action_goal_info_t * goal_info_out;
+  for (int i = 0; i < 22; ++i) {
```
hidmic: @jacobperron here and below, consider using `cancel_response.msg.goals_canceling.size` instead of a literal.

rcl_action/include/rcl_action/types.h
```diff
@@ -52,6 +51,11 @@ extern "C"
 /// Action invalid event return code.
 #define RCL_RET_ACTION_GOAL_EVENT_INVALID 2301
 
+// TODO(jacobperron): Move these to a common place for UUIDs
+#define uuidcmp(uuid0, uuid1) (0 == memcmp(uuid0, uuid1, 16))
+#define zerouuid (uint8_t[16]) {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
+#define uuidcmpzero(uuid) uuidcmp(uuid, zerouuid)
```
hidmic: @jacobperron +1, I'd also add a `#define` for UUID size, just to weed out `16` from the code.

rcl_action/test/rcl_action/test_action_server.cpp
```diff
@@ -0,0 +1,738 @@
...
+      EXPECT_EQ(goal_info_out->uuid[j], static_cast<uint8_t>(i + j));
+    }
+  }
+  goal_info_out = &cancel_response.msg.goals_canceling.data[11];
+  for (int i = 0; i < 16; ++i) {
+    EXPECT_EQ(goal_info_out->uuid[i], static_cast<uint8_t>(i + 40));
+  }
+}
+
+#ifdef RMW_IMPLEMENTATION
```
hidmic: @jacobperron here and below, these tests look more like integration tests to me. We'd be testing both server and client. What do you think about splitting them into a separate source file?

hidmic:
Does this PR depend on #325 too?

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,644 @@
...
+  }
+
+  // Allocate for action server
+  action_server->impl = (rcl_action_server_impl_t *)allocator.allocate(
+    sizeof(rcl_action_server_impl_t), allocator.state);
+  RCL_CHECK_FOR_NULL_WITH_MSG(
+    action_server->impl, "allocating memory failed", return RCL_RET_BAD_ALLOC);
+
+  // TODO(jacobperron): Expand the given action name
+  // Test if name is valid
```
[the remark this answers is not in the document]
jacobperron: Yes, I think that is correct. I have a test case locally for this case that is failing atm. I'll add a check when initializing the services/publishers.

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,644 @@
...
+
+  // Initialize services
+  rcl_ret_t ret;
+  SERVICE_INIT(goal);
```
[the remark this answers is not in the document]
jacobperron: I agree that it is better to be consistent. I went with macros because it made for more concise code, cutting down on code duplication, but I could be convinced to use `static` functions instead.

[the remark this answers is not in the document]
hidmic: Not arguing in favor or against. Either way is fine for me as long as we use just one.

rcl_action/test/rcl_action/test_action_server.cpp
jacobperron: Yeah, that sounds good to me.

rcl_action/include/rcl_action/types.h
jacobperron: Added in 69c0287

rcl_action/test/rcl_action/test_action_server.cpp
jacobperron: Done 69c0287

rcl_action/test/rcl_action/test_action_server.cpp
jacobperron: Added in 69c0287

rcl_action/test/rcl_action/test_action_server.cpp
jacobperron: Added in 69c0287

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,644 @@
...
+rcl_ret_t
+rcl_action_take_cancel_request(
+  const rcl_action_server_t * action_server,
+  void * ros_cancel_request)
+{
+  TAKE_SERVICE_REQUEST(cancel);
+}
+
+uint64_t
+_goal_info_stamp_to_nanosec(const rcl_action_goal_info_t * goal_info)
```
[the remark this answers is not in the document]
jacobperron: Done 69c0287

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,644 @@
...
+rcl_action_send_goal_response(
+  const rcl_action_server_t * action_server,
+  void * ros_goal_response)
+{
+  SEND_SERVICE_RESPONSE(goal);
+}
+
+rcl_action_goal_handle_t *
+rcl_action_accept_new_goal(
+  const rcl_action_server_t * action_server,
```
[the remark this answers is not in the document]
jacobperron: Done 69c0287

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,644 @@
...
+  rcl_action_goal_handle_t * goal_handles_to_cancel[total_num_goals];
+  size_t num_goals_to_cancel = 0u;
+
+  // Request data
+  const rcl_action_goal_info_t * request_goal_info = &cancel_request->goal_info;
+  const uint8_t * request_uuid = request_goal_info->uuid;
+  uint64_t request_nanosec = _goal_info_stamp_to_nanosec(request_goal_info);
+
+  // Determine how many goals should transition to canceling
+  if (!uuidcmpzero(request_uuid) && (0u == request_nanosec)) {
```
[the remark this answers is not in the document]
hidmic: Right, the no-zero-timestamp assumption is there. Fine to leave it as is if we think that's a scenario we want to deal with, I had not seen the unit tests.

rcl_action/src/rcl_action/action_server.c
jacobperron: >  Isn't the status publisher initialized with client library specific type support?

I completely overlooked this, I don't think it will work.

I'm not sure if I understand what you are proposing, sounds like a potential solution or alternative to ros2/rosidl#314 ?
In the meantime, I'll update this function to take a type-erased pointer.

rcl_action/test/rcl_action/test_action_server.cpp
jacobperron: Moved to new file `test_action_communication.cpp` in 69c0287

jacobperron:
> Is the calling library responsible for updating the goal handle (i.e. to transition to execute, to succeeded, etc.)?

In the current implementation, yes. I was hoping to build on top of this (or refactor) after ros2/rosidl#314 in order to encapsulate more logic.

> Does this PR depend on #325 too?

Yes, those commits are included in this PR for testing purposes, but I'll rebase after #325 is merged.

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,644 @@
...
+    if (rcl_publisher_fini(&action_server->impl->feedback_publisher, node) != RCL_RET_OK) {
+      ret = RCL_RET_ERROR;
+    }
+    if (rcl_publisher_fini(&action_server->impl->status_publisher, node) != RCL_RET_OK) {
+      ret = RCL_RET_ERROR;
+    }
+    // Deallocate struct
+    rcl_allocator_t allocator = action_server->impl->options.allocator;
+    allocator.deallocate(action_server->impl, allocator.state);
+    action_server->impl = NULL;
```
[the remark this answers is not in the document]
jacobperron: See https://github.com/ros2/rcl/pull/319#discussion_r232764209

rcl_action/src/rcl_action/action_server.c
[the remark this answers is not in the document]
jacobperron: I'm currently preferring the macro route because of its conciseness, if you don't mind updating #319 for consistency.

rcl_action/test/rcl_action/test_action_server.cpp
```diff
@@ -0,0 +1,526 @@
...
+
+  // Try to initialize again
+  ret = rcl_action_server_init(&action_server, &node, ts, action_name, &options);
+  EXPECT_EQ(ret, RCL_RET_ALREADY_INIT) << rcl_get_error_string().str;
+  rcl_reset_error();
+
+  // Try to initialize another action server with the same name
+  // FIXME(jacobperron): test failing
+  // rcl_action_server_t same_name_action_server = rcl_action_get_zero_initialized_server();
+  // ret = rcl_action_server_init(&same_name_action_server, &node, ts, action_name, &options);
```
hidmic: @jacobperron I presume we're aiming to catch this by means of invalid service name checks, but it doesn't seem like that outcome means the service name is repeated (see [here](https://github.com/ros2/rcl/blob/master/rcl/src/rcl/service.c#L117))?

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,759 @@
...
+  const rcl_action_server_t * action_server,
+  rcl_action_goal_handle_t *** goal_handles,
+  size_t * num_goals)
+{
+  if (!rcl_action_server_is_valid(action_server)) {
+    return RCL_RET_ACTION_SERVER_INVALID;  // error already set
+  }
+  RCL_CHECK_ARGUMENT_FOR_NULL(goal_handles, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(num_goals, RCL_RET_INVALID_ARGUMENT);
+  *goal_handles = action_server->impl->goal_handles;
```
hidmic: @jacobperron should we check for `NULL == *goal_handles`?

rcl_action/include/rcl_action/action_server.h
```diff
@@ -771,16 +770,31 @@ rcl_action_server_get_options(const rcl_action_server_t * action_server);
...
+rcl_ret_t
 rcl_action_server_get_goal_handles(
   const rcl_action_server_t * action_server,
-  uint32_t * num_goals);
+  rcl_action_goal_handle_t *** goal_handles,
+  size_t * num_goals);
+
+/// Check if a goal is already being tracked by an action server
+/**
+ * TODO(jacobperron): Document
```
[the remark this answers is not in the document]
jacobperron: Yes :grin:

rcl_action/test/rcl_action/test_action_server.cpp
hidmic: Oh, that's a bug in `rcl_action/names.cc`.

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,759 @@
...
+      // Fill in any gaps left in the array with pointers from the end
+      action_server->impl->goal_handles[i] = action_server->impl->goal_handles[num_goal_handles];
+      action_server->impl->goal_handles[num_goal_handles--] = NULL;
+
+      ++(*num_expired);
+    }
+  }
+
+  // Shrink goal handle array if some goals expired
+  if (*num_expired > 0u) {
```
sloretz: This should deallocate if the new size is zero because `allocator.reallocate()` [is supposed to behave like realloc()](https://github.com/ros2/rcutils/blob/3c982d048914e1c30002d8ea69d86ac1cc6976fd/include/rcutils/allocator.h#L58-L62) which has [undefined behavior when the new size is zero](https://en.cppreference.com/w/c/memory/realloc).

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,759 @@
...
+
+  // Determine how many goals should transition to canceling
+  if (!uuidcmpzero(request_uuid) && (0u == request_nanosec)) {
+    // UUID is not zero and timestamp is zero; cancel exactly one goal (if it exists)
+    rcl_action_goal_info_t goal_info = rcl_action_get_zero_initialized_goal_info();
+    rcl_action_goal_handle_t * goal_handle;
+    for (size_t i = 0; i < total_num_goals; ++i) {
+      goal_handle = action_server->impl->goal_handles[i];
+      rcl_ret_t ret = rcl_action_goal_handle_get_info(goal_handle, &goal_info);
+      assert(RCL_RET_OK == ret);
```
sloretz: IIUC assert is compiled out in release builds. Is this an `assert()` because its calling out an assumption that the above call will always succeed? Is it possible the above call could fail in normal use?

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,759 @@
...
+
+      if (uuidcmp(request_uuid, goal_info.uuid)) {
+        if (rcl_action_goal_handle_is_cancelable(goal_handle)) {
+          goal_handles_to_cancel[num_goals_to_cancel++] = goal_handle;
```
sloretz: Intentional postfix increment? I think the style guide says to always use prefix increment.

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,759 @@
...
+    }
+
+    // Cancel all active goals at or before the timestamp
+    // Also cancel any goal matching the UUID in the cancel request
+    rcl_action_goal_info_t goal_info = rcl_action_get_zero_initialized_goal_info();
+    rcl_action_goal_handle_t * goal_handle;
+    for (size_t i = 0; i < total_num_goals; ++i) {
+      goal_handle = action_server->impl->goal_handles[i];
+      rcl_ret_t ret = rcl_action_goal_handle_get_info(goal_handle, &goal_info);
+      assert(RCL_RET_OK == ret);
```
sloretz: same comment about assert

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,759 @@
...
+
+      const int64_t goal_nanosec = _goal_info_stamp_to_nanosec(&goal_info);
+      if (rcl_action_goal_handle_is_cancelable(goal_handle) &&
+        ((goal_nanosec <= request_nanosec) || uuidcmp(request_uuid, goal_info.uuid)))
+      {
+        goal_handles_to_cancel[num_goals_to_cancel++] = goal_handle;
```
sloretz: comment++

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,759 @@
...
+  rcl_action_goal_info_t gh_goal_info = rcl_action_get_zero_initialized_goal_info();
+  rcl_ret_t ret;
+  for (size_t i = 0; i < action_server->impl->num_goal_handles; ++i) {
+    ret = rcl_action_goal_handle_get_info(action_server->impl->goal_handles[i], &gh_goal_info);
+    if (RCL_RET_OK != ret) {
+      RCL_SET_ERROR_MSG("failed to get info for goal handle");
+      return false;
+    }
+    // Compare UUIDs
+    if (uuidcmp(gh_goal_info.uuid, goal_info->uuid)) {
```
sloretz: nit could return this directly

rcl_action/src/rcl_action/action_server.c
jacobperron: Yeah, this is intentional. Assigning the goal handle to index `num_goals_to_cancel` and then increment.

rcl_action/src/rcl_action/action_server.c
jacobperron: ditto

rcl_action/src/rcl_action/action_server.c
jacobperron: I don't think so. If it is false, there still might be a match in a future iteration of the for-loop.

rcl_action/include/rcl_action/action_server.h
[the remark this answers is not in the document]
jacobperron: Addressed in 7c4f579

rcl_action/package.xml
```diff
@@ -18,6 +18,7 @@
   <test_depend>ament_cmake_gtest</test_depend>
   <test_depend>ament_lint_common</test_depend>
   <test_depend>ament_lint_auto</test_depend>
+  <test_depend>test_msgs</test_depend>
```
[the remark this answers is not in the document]
jacobperron: Addressed in 00ecdf0

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,759 @@
...
+  // Allocate for action server
+  action_server->impl = (rcl_action_server_impl_t *)allocator.allocate(
+    sizeof(rcl_action_server_impl_t), allocator.state);
+  RCL_CHECK_FOR_NULL_WITH_MSG(
+    action_server->impl, "allocating memory failed", return RCL_RET_BAD_ALLOC);
+
+  // Initialize clock
+  rcl_ret_t ret = rcl_clock_init(options->clock_type, &action_server->impl->clock, &allocator);
+  if (RCL_RET_OK != ret) {
+    return RCL_RET_ERROR;
```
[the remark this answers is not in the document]
jacobperron: Addressed in c6b436e

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,759 @@
...
+  // !!! MAKE SURE THAT CHANGES TO THESE DEFAULTS ARE REFLECTED IN THE HEADER DOC STRING
+  static rcl_action_server_options_t default_options;
+  default_options.goal_service_qos = rmw_qos_profile_services_default;
+  default_options.cancel_service_qos = rmw_qos_profile_services_default;
+  default_options.result_service_qos = rmw_qos_profile_services_default;
+  default_options.feedback_topic_qos = rmw_qos_profile_default;
+  default_options.status_topic_qos = rcl_action_qos_profile_status_default;
+  default_options.allocator = rcl_get_default_allocator();
+  default_options.clock_type = RCL_ROS_TIME;
+  default_options.result_timeout.nanoseconds = (rcl_duration_value_t)9e11;  // 15 minutes
```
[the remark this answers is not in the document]
jacobperron: Addressed in c6b436e

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,759 @@
...
+{
+  SEND_SERVICE_RESPONSE(goal);
+}
+
+// Implementation only
+static int64_t
+_goal_info_stamp_to_nanosec(const rcl_action_goal_info_t * goal_info)
+{
+  assert(goal_info);
+  return (int64_t)(goal_info->stamp.sec * 1e9) + (int64_t)goal_info->stamp.nanosec;
```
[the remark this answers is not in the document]
jacobperron: Addressed in c6b436e

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,759 @@
...
+}
+
+// Implementation only
+static void
+_nanosec_to_goal_info_stamp(const int64_t * nanosec, rcl_action_goal_info_t * goal_info)
+{
+  assert(nanosec);
+  assert(goal_info);
+  goal_info->stamp.sec = (int32_t) (*nanosec / (int64_t)1e9);
```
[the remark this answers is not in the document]
jacobperron: Addressed in c6b436e

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,759 @@
...
+  goal_info->stamp.nanosec = (uint32_t)(*nanosec - (int64_t)(goal_info->stamp.sec * 1e9));
```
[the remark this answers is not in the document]
jacobperron: Addressed in c6b436e

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,759 @@
...
+    }
+    return RCL_RET_ERROR;
+  }
+
+  // Populate status array
+  for (size_t i = 0; i < num_goals; ++i) {
+    ret = rcl_action_goal_handle_get_info(
+      action_server->impl->goal_handles[i], &status_message->msg.status_list.data[i].goal_info);
+    if (RCL_RET_OK != ret) {
+      return RCL_RET_ERROR;
```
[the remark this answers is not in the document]
jacobperron: Addressed in c6b436e

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,759 @@
...
+    }
+    ret = rcl_action_goal_handle_get_status(
+      action_server->impl->goal_handles[i], &status_message->msg.status_list.data[i].status);
+    if (RCL_RET_OK != ret) {
+      return RCL_RET_ERROR;
```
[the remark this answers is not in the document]
jacobperron: Addressed in c6b436e

rcl_action/src/rcl_action/action_server.c
jacobperron: Addressed in c6b436e

hidmic:
Looks like this PR needs a rebase.

rcl_action/src/rcl_action/action_server.c
sloretz: Ooops

jacobperron:
Rebased.

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,759 @@
...
+  }
+
+  // Allocate for action server
+  action_server->impl = (rcl_action_server_impl_t *)allocator.allocate(
+    sizeof(rcl_action_server_impl_t), allocator.state);
+  RCL_CHECK_FOR_NULL_WITH_MSG(
+    action_server->impl, "allocating memory failed", return RCL_RET_BAD_ALLOC);
+
+  // Initialize clock
+  rcl_ret_t ret = rcl_clock_init(options->clock_type, &action_server->impl->clock, &allocator);
```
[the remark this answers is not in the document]
jacobperron: Okay. Then I suppose it will not be the action servers responsibility to finalize the clock and we hope that it stays valid as long as the action server is valid. Does this sound right?

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,759 @@
...
+  void * tmp_ptr = allocator.reallocate(
+    goal_handles, new_num_goal_handles * sizeof(rcl_action_goal_handle_t *), allocator.state);
+  if (!tmp_ptr) {
+    RCL_SET_ERROR_MSG("memory allocation failed for goal handle pointer");
+    return NULL;
+  }
+  goal_handles = (rcl_action_goal_handle_t **)tmp_ptr;
+
+  // Allocate space for a new goal handle
+  tmp_ptr = allocator.allocate(sizeof(rcl_action_goal_handle_t), allocator.state);
```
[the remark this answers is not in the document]
jacobperron: True, the main reason I went with an array of pointers was that is made it easier for removing expired goals from the middle of the list:

https://github.com/ros2/rcl/blob/06df1fdf3cd1789fbf6ae186784bbfde18e7d9f6/rcl_action/src/rcl_action/action_server.c#L533-L535

But I guess it wouldn't bee too much overhead to copy the entire goal handle in this case (instead of just the pointer).

rcl_action/src/rcl_action/action_server.c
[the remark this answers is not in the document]
sloretz: > Does this sound right?

I think so

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,759 @@
...
+  const rcl_action_server_t * action_server,
+  void * ros_result_response)
+{
+  SEND_SERVICE_RESPONSE(result);
+}
+
+rcl_ret_t
+rcl_action_clear_expired_goals(
+  const rcl_action_server_t * action_server,
+  size_t * num_expired)
```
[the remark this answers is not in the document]
jacobperron: My original thinking was that since the action server is the entity creating the goal handles, it should be the one responsible for destroying them.

In a rcl goal handle is finalized, then the goal handle becomes invalid (ie. `rcl_action_goal_handle_is_valid()` returns false) and so, I'm pretty sure, any rcl operations involving the goal handle should fail with an error code. So at least, the C++ wrapper of a goal handle wouldn't be using a free'd goal handle without getting an explicit error.

Is it enough for the client libraries to check their goal handles with `rcl_action_goal_handle_is_valid()` and/or get a fresh list of goal handles with `rcl_action_server_get_goal_handles()` ?

rcl_action/src/rcl_action/action_server.c
jacobperron: I thought it should always succeed, but after thinking about it some more, this can fail if the goal handle is finalized outside of the action server's control.

rcl_action/src/rcl_action/action_server.c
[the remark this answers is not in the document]
sloretz: >Is it enough for the client libraries to check their goal handles with `rcl_action_goal_handle_is_valid()`

I guess the question is what does a C++ goal handle wrapper have a reference to? If it has an `rcl_action_goal_handle_t * rcl_handle_;` pointing to storage owned by [`rcl_action_server_impl_t`](https://github.com/ros2/rcl/pull/323/files#diff-d6a1cca7ac458b893d4cd7204485edf2R47), then `rcl_action_goal_handle_is_valid(rcl_handle_)` isn't safe because the memory it's pointing to could change any time `impl->goal_handles` is resized. It seems like it needs to be the other way around where the `C++` goal handle wrapper has an `rcl_action_goal_handle_t` and `rcl_action_server_t` is the one with a pointer to it.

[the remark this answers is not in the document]
jacobperron: I see; makes sense now.

rcl_action/src/rcl_action/action_server.c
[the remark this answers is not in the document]
jacobperron: Addressed in 4eb7fdb

rcl_action/src/rcl_action/action_server.c
jacobperron: Addressed in 8f9041e

rcl_action/src/rcl_action/action_server.c
jacobperron: No. We can set `*goal_handles` to a valid pointer, regardless of it's initial value.

rcl_action/test/rcl_action/test_action_server.cpp
```diff
@@ -0,0 +1,526 @@
...
+  ASSERT_EQ(ret, RCL_RET_OK);
+  EXPECT_EQ(num_goals, 2u);
+  EXPECT_NE(goal_handle_array, nullptr) << rcl_get_error_string().str;
+  EXPECT_NE(goal_handle_array[0], nullptr) << rcl_get_error_string().str;
+  EXPECT_NE(goal_handle_array[1], nullptr) << rcl_get_error_string().str;
+}
+
+TEST_F(TestActionServer, test_action_clear_expired_goals)
+{
+  size_t num_expired;
```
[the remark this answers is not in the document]
jacobperron: Addressed in bc9a5df

rcl_action/test/rcl_action/test_action_server.cpp
```diff
@@ -0,0 +1,526 @@
...
+
+  // Accept with valid arguments
+  goal_handle = rcl_action_accept_new_goal(&this->action_server, &goal_info_in);
+  EXPECT_NE(goal_handle, nullptr) << rcl_get_error_string().str;
+  rcl_action_goal_info_t goal_info_out = rcl_action_get_zero_initialized_goal_info();
+  rcl_ret_t ret = rcl_action_goal_handle_get_info(goal_handle, &goal_info_out);
+  ASSERT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
+  EXPECT_TRUE(uuidcmp(goal_info_out.uuid, goal_info_in.uuid));
+  size_t num_goals;
+  rcl_action_goal_handle_t ** goal_handle_array;
```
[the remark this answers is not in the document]
jacobperron: Addressed in bc9a5df

rcl_action/test/rcl_action/test_action_server.cpp
```diff
@@ -0,0 +1,546 @@
...
+  }
+}
+
+TEST_F(TestActionServerCancelPolicy, test_action_process_cancel_request_single_goal)
+{
+  // Request to cancel a specific goal
+  rcl_action_cancel_request_t cancel_request = rcl_action_get_zero_initialized_cancel_request();
+  for (int i = 0; i < UUID_SIZE; ++i) {
+    cancel_request.goal_info.uuid[i] = static_cast<uint8_t>(i + 2);
+  }
```
hidmic: @jacobperron just curious, why not reusing one of the `init_test_uuid*()` functions?

rcl_action/test/rcl_action/test_action_server.cpp
```diff
@@ -0,0 +1,546 @@
...
+  rcl_action_cancel_response_t cancel_response = rcl_action_get_zero_initialized_cancel_response();
+  rcl_ret_t ret = rcl_action_process_cancel_request(
+    &this->action_server, &cancel_request, &cancel_response);
+  EXPECT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
+  EXPECT_NE(cancel_response.msg.goals_canceling.data, nullptr);
+  ASSERT_EQ(cancel_response.msg.goals_canceling.size, 1u);
+  rcl_action_goal_info_t * goal_info = &cancel_response.msg.goals_canceling.data[0];
+  for (int i = 0; i < UUID_SIZE; ++i) {
+    EXPECT_EQ(goal_info->uuid[i], static_cast<uint8_t>(i + 2));
+  }
```
hidmic: @jacobperron could be `EXPECT_TRUE(uuidcmp(goal_info->uuid, cancel_request.goal_info.uuid));` as well.

rcl_action/test/rcl_action/test_action_server.cpp
```diff
@@ -0,0 +1,546 @@
...
+    &this->action_server, &cancel_request, &cancel_response);
+  EXPECT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
+  EXPECT_NE(cancel_response.msg.goals_canceling.data, nullptr);
+  ASSERT_EQ(cancel_response.msg.goals_canceling.size, time_index + 1);  // goals at indices [0, 7]
+  rcl_action_goal_info_t * goal_info_out;
+  for (size_t i = 0; i < cancel_response.msg.goals_canceling.size; ++i) {
+    goal_info_out = &cancel_response.msg.goals_canceling.data[i];
+    for (size_t j = 0; j < UUID_SIZE; ++j) {
+      EXPECT_EQ(goal_info_out->uuid[j], static_cast<uint8_t>(i + j));
+    }
```
hidmic: @jacobperron same as above but using `goals_info_out` instead.

rcl_action/test/rcl_action/test_action_server.cpp
```diff
@@ -0,0 +1,546 @@
...
+  for (size_t i = 0; i < num_goals_canceling - 1; ++i) {
+    goal_info_out = &cancel_response.msg.goals_canceling.data[i];
+    for (size_t j = 0; j < UUID_SIZE; ++j) {
+      EXPECT_EQ(goal_info_out->uuid[j], static_cast<uint8_t>(i + j));
+    }
+  }
+  goal_info_out = &cancel_response.msg.goals_canceling.data[num_goals_canceling - 1];
+  for (int i = 0; i < UUID_SIZE; ++i) {
+    EXPECT_EQ(goal_info_out->uuid[i], static_cast<uint8_t>(i + goal_index));
+  }
```
hidmic: @jacobperron same as above.

rcl_action/src/rcl_action/action_server.c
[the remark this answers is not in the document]
hidmic: Hmm, so just to be clear, rcl* will handle all incoming result requests by first checking for goal handle validity, rejecting the request once the goal handle is no longer valid and only then deallocating it, right? I'd be somewhat more comfortable with just marking the goal handle as expired to keep finalization and deallocation close together, but this will do to.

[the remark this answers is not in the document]
jacobperron: @hidmic I was going to refactor rcl action server so that goal handle finalization does not happen here. It will be the responsibility of the caller to finalize and deallocate goal handles. rcl will only resize the internal array as goals become expired.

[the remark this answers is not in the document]
hidmic: Hmm, then how does the calling library knows when a goal handle has expired?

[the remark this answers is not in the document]
jacobperron: > Hmm, then how does the calling library know when a goal handle has expired?
 
What about if we return a list of goal handles as part of the expire function.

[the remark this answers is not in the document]
hidmic: Sounds like an option. Though judging by @sloretz rclcpp API skeleton, I'd imagine that having the client library wrappers checking if the goal handle they own has expired or not may be simpler. But we can somehow reverse the mapping to know which wrappers have to go away when goal handles expire.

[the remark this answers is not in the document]
jacobperron: The client library can call `rcl_action_server_goal_exists()` to check if a previously accepted goal has expired.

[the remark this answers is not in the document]
jacobperron: Removed finalize of expired goal handles in 1140112.
It is expected that the client library will finalize goal handles either after they expire (which can be checked with `rcl_action_server_goal_exists()`) or after the action server is finalized.

Note, the rcl goal handles are still initialized in `rcl_action_accept_new_goal()`. I was going to update the API to take an already initialized goal handle as an argument, but then we lose the time stamp created when the goal was accepted:

https://github.com/ros2/rcl/blob/11401120af69d5652ffb7a2ad46502159941039b/rcl_action/src/rcl_action/action_server.c#L362-L370

If we want the client library to pass the goal handle to rcl, then we can

1. Create a setter function for the goal handle's info and override the info in `rcl_action_accept_goal_info`, or
2. Leave the responsibility to the client library to use the same clock and stamp the goal handles properly.

rcl_action/test/rcl_action/test_action_server.cpp
jacobperron: I can reuse `init_test_uui0()`, but the `init_test_uuid1()` wouldn't do because the ID doesn't exist in this test case.

rcl_action/test/rcl_action/test_action_server.cpp
jacobperron: In this case, we would need to reconstruct the uuid in a similar loop before comparing anyways.

rcl_action/test/rcl_action/test_action_server.cpp
jacobperron: Addressed 7aadb55

rcl_action/test/rcl_action/test_action_server.cpp
jacobperron: Addressed 7aadb55

rcl_action/test/rcl_action/test_action_server.cpp
jacobperron: Addressed 7aadb55

rcl_action/src/rcl_action/action_server.c
```diff
@@ -0,0 +1,782 @@
...
+    }
+  }
+
+  // Shrink goal handle array if some goals expired
+  if (*num_expired > 0u) {
+    if (0 == num_goal_handles) {
+      allocator.deallocate(action_server->impl->goal_handles, allocator.state);
+    } else {
+      void * tmp_ptr = allocator.reallocate(
+        action_server->impl->goal_handles, num_goal_handles, allocator.state);
```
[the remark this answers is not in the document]
jacobperron: Addressed 7aadb55

hidmic:
@jacobperron One more CI run and I think this one is ready to go.

jacobperron:
Up to rcl_action:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5593/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2227/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4603/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5472/)

rcl_action/CMakeLists.txt
```diff
@@ -49,8 +53,11 @@ add_library(${PROJECT_NAME}
 )
 
 ament_target_dependencies(${PROJECT_NAME}
-  "rcl"
+  "rosidl_generator_c"
```
sloretz: nit, alphabetical order

rcl_action/CMakeLists.txt
```diff
@@ -86,6 +94,34 @@ if(BUILD_TESTING)
...
     ament_target_dependencies(test_action_client "rcl" "test_msgs")
   endif()
+  ament_add_gtest(test_action_communication
+    test/rcl_action/test_action_communication.cpp
+  )
+  if(TARGET test_action_communication)
+    target_include_directories(test_action_communication PUBLIC
+      include
+      ${rcl_INCLUDE_DIRS}
+      ${test_msgs_INCLUDE_DIRS}
```
sloretz: nit, This is taken care of by `ament_target_dependencies()` below

rcl_action/CMakeLists.txt
```diff
@@ -86,6 +94,34 @@ if(BUILD_TESTING)
...
+    ament_target_dependencies(test_action_communication "test_msgs")
+  endif()
+  ament_add_gtest(test_action_server
+    test/rcl_action/test_action_server.cpp
+  )
+  if(TARGET test_action_server)
+    target_include_directories(test_action_server PUBLIC
+      include
+      ${rcl_INCLUDE_DIRS}
+      ${test_msgs_INCLUDE_DIRS}
```
sloretz: nit, This is taken care of by ament_target_dependencies() below

jacobperron:
@sloretz Thanks, I'll take note and include fixes for the small things in the next PR.

rcl_action/CMakeLists.txt
```diff
@@ -86,6 +94,34 @@ if(BUILD_TESTING)
     )
     ament_target_dependencies(test_action_client "rcl" "test_msgs")
   endif()
+  ament_add_gtest(test_action_communication
+    test/rcl_action/test_action_communication.cpp
+  )
```
dirk-thomas: This test (as well as the one below) is only covering the default rmw impl. It would be good if they test would be invoked for every rmw impl.

PR #327, opened 2018-11-13
Adding new cli parameters for configuring the logging.

nburek:
## Summary
This change is part of a set of changes for new logging features. These changes include the following features:

### rcutils
- Slight refactor to the buffer used in logging to make it more generic and reusable
- A new output handler that allows multiple output handlers to be set (default is still only the console output function). This includes the following three output handlers: 1) The console output handler 2) The external logging lib output handler 3) The rosout output handler
- Change to the output format to accept only the log string instead of the format and vargs. The log will be formatted before being sent to the output so that there will be a consistent format across all the outputs
- Added an interface for an external logging library, an output handler function for sending logs to this library, and cmake changes so that the external logging lib can be specified (default is rc_logging_log4cxx)

### rcl
- Added new command line parameters for disabling various logging output handlers
- Added new command line parameter for passing in a config file to the external logging library
- Called the new configuration function in the rcutils logging to configure the logging during the ROS init process

### rclcpp
- Fixing a broken unit test

### rc_logging_log4cxx
- An external logging implementation that uses log4cxx under the hood
- Sets up a default FileAppender  logger to forward logs to files
- Uses the process Id as the filename by default

## Remaining Feature Work
- The rosout handlers is stubbed out in rcutils right now, but more work will need to be done to actually support sending logs to a rosout topic on a node. This will likely include moving the rosout handler up to rcl so that it can use the node interfaces to publish. 

## Links
https://discourse.ros.org/t/ros2-logging/6469
https://github.com/nburek/rc_logging_log4cxx

### Pull requests for change
rclcpp: https://github.com/ros2/rclcpp/pull/582
rcl: https://github.com/ros2/rcl/pull/327
rcutils: https://github.com/ros2/rcutils/pull/127

### DIFF
--- a/rcl/include/rcl/arguments.h
+++ b/rcl/include/rcl/arguments.h
@@ -34,8 +34,12 @@ typedef struct rcl_arguments_t
   struct rcl_arguments_impl_t * impl;
 } rcl_arguments_t;
 
-#define RCL_LOG_LEVEL_ARG_RULE "__log_level:="
-#define RCL_PARAM_FILE_ARG_RULE "__params:="
+#define RCL_LOG_LEVEL_ARG_RULE              "__log_level:="
+#define RCL_EXTERNAL_LOG_CONFIG_ARG_RULE    "__log_config_file:="
+#define RCL_LOG_DISABLE_STDOUT_ARG_RULE     "__log_disable_stdout:="
+#define RCL_LOG_DISABLE_ROSOUT_ARG_RULE     "__log_disable_rosout:="
+#define RCL_LOG_DISABLE_EXT_LIB_ARG_RULE    "__log_disable_external_lib:="
+#define RCL_PARAM_FILE_ARG_RULE             "__params:="
 
 /// Return a rcl_node_t struct with members initialized to `NULL`.
 RCL_PUBLIC
--- a/rcl/include/rcl/arguments.h
+++ b/rcl/include/rcl/arguments.h
@@ -34,12 +34,12 @@ typedef struct rcl_arguments_t
   struct rcl_arguments_impl_t * impl;
 } rcl_arguments_t;
 
-#define RCL_LOG_LEVEL_ARG_RULE              "__log_level:="
-#define RCL_EXTERNAL_LOG_CONFIG_ARG_RULE    "__log_config_file:="
-#define RCL_LOG_DISABLE_STDOUT_ARG_RULE     "__log_disable_stdout:="
-#define RCL_LOG_DISABLE_ROSOUT_ARG_RULE     "__log_disable_rosout:="
-#define RCL_LOG_DISABLE_EXT_LIB_ARG_RULE    "__log_disable_external_lib:="
-#define RCL_PARAM_FILE_ARG_RULE             "__params:="
+#define RCL_LOG_LEVEL_ARG_RULE "__log_level:="
+#define RCL_EXTERNAL_LOG_CONFIG_ARG_RULE "__log_config_file:="
+#define RCL_LOG_DISABLE_STDOUT_ARG_RULE "__log_disable_stdout:="
+#define RCL_LOG_DISABLE_ROSOUT_ARG_RULE "__log_disable_rosout:="
+#define RCL_LOG_DISABLE_EXT_LIB_ARG_RULE "__log_disable_external_lib:="
+#define RCL_PARAM_FILE_ARG_RULE "__params:="
 
 /// Return a rcl_node_t struct with members initialized to `NULL`.
 RCL_PUBLIC
--- a/rcl/package.xml
+++ b/rcl/package.xml
@@ -37,7 +37,7 @@
   <test_depend>osrf_testing_tools_cpp</test_depend>
   <test_depend>test_msgs</test_depend>
 
-  <group_depend>rc_logging_packages</group_depend>
+  <group_depend>rcl_logging_packages</group_depend>
 
   <export>
     <build_type>ament_cmake</build_type>
--- a/rcl/package.xml
+++ b/rcl/package.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
-<package format="2">
+<package format="3">
   <name>rcl</name>
   <version>0.6.0</version>
   <description>The ROS client library common implementation.
@@ -26,6 +26,8 @@
   <exec_depend>rosidl_default_runtime</exec_depend>
 
   <depend>rmw_implementation</depend>
+
+  <group_depend>rc_logging_packages</group_depend>
 
   <test_depend>ament_cmake_gtest</test_depend>
   <test_depend>ament_cmake_pytest</test_depend>
--- a/rcl/package.xml
+++ b/rcl/package.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="3">
   <name>rcl</name>
   <version>0.6.0</version>
@@ -27,8 +27,6 @@
 
   <depend>rmw_implementation</depend>
 
-  <group_depend>rc_logging_packages</group_depend>
-
   <test_depend>ament_cmake_gtest</test_depend>
   <test_depend>ament_cmake_pytest</test_depend>
   <test_depend>ament_lint_auto</test_depend>
@@ -38,6 +36,8 @@
   <test_depend>launch</test_depend>
   <test_depend>osrf_testing_tools_cpp</test_depend>
   <test_depend>test_msgs</test_depend>
+
+  <group_depend>rc_logging_packages</group_depend>
 
   <export>
     <build_type>ament_cmake</build_type>
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -25,6 +25,8 @@
 #include "rcl/validate_topic_name.h"
 #include "rcutils/allocator.h"
 #include "rcutils/error_handling.h"
+#include "rcutils/format_string.h"
+#include "rcutils/logging.h"
 #include "rcutils/logging_macros.h"
 #include "rcutils/strdup.h"
 #include "rmw/validate_namespace.h"
@@ -150,9 +152,8 @@ _rcl_parse_external_log_config_file(
 /**
  * \param[in] arg the argument to parse
  * \param[in] key the key for the argument to parse. Should be a null terminated string
- * \param[in] allocator an allocator to use
  * \param[in,out] value parsed boolean value
- * \return RCL_RET_OK if a valid log level was parsed, or
+ * \return RCL_RET_OK if the bool argument was parsed successfully, or
  * \return RLC_RET_ERROR if an unspecified error occurred.
  */
 RCL_LOCAL
@@ -162,15 +163,18 @@ _rcl_parse_bool_arg(
   const char * key,
   bool * value);
 
-/// Parse a null terminated string to a boolean value
+/// Parse a null terminated string to a boolean value.
 /**
- *  The case sensitive values "T", "t", "True", "true", "Y", "y", "Yes", "yes", and "1" will all map to true
- *  The case sensitive values "F", "f", "False", "false", "N", "n", "No", "no", and "0" will all map to false
- *  \param[in] str a null terminated string to be parsed into a boolean
- *  \param[in,out] val the boolean value parsed from the string. Left unchanged if string cannot be parsed to a valid bool
- *  \param[in] allocator an allocator to use
- *  \return RCL_RET_OK if a valid boolean parsed, or
- *  \return RLC_RET_ERROR if an unspecified error occurred.
+ * The case sensitive values: "T", "t", "True", "true", "Y", "y", "Yes", "yes",
+ * and "1" will all map to true.
+ * The case sensitive values: "F", "f", "False", "false", "N", "n", "No", "no",
+ * and "0" will all map to false.
+ *
+ * \param[in] str a null terminated string to be parsed into a boolean
+ * \param[in,out] val the boolean value parsed from the string.
+ *   Left unchanged if string cannot be parsed to a valid bool.
+ * \return RCL_RET_OK if a valid boolean parsed, or
+ * \return RLC_RET_ERROR if an unspecified error occurred.
  */
 RCL_LOCAL
 rcl_ret_t
@@ -285,8 +289,8 @@ rcl_parse_arguments(
     rcl_reset_error();
 
     // Attempt to parse argument as log configuration file
-    rcl_ret_t ret = _rcl_parse_external_log_config_file(argv[i], allocator,
-        &args_impl->external_log_config_file);
+    rcl_ret_t ret = _rcl_parse_external_log_config_file(
+      argv[i], allocator, &args_impl->external_log_config_file);
     if (RCL_RET_OK == ret) {
       continue;
     }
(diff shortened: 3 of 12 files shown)


rcl/src/rcl/arguments.c
```diff
@@ -134,6 +133,54 @@ _rcl_parse_log_level_rule(
...
+  const char * arg,
+  const char * key,
+  bool * value);
+
+/// Parse a null terminated string to a boolean value
+/**
+ *  The case sensitive values "T", "t", "True", "true", "Y", "y", "Yes", "yes", and "1" will all map to true
+ *  The case sensitive values "F", "f", "False", "false", "N", "n", "No", "no", and "0" will all map to false
+ *  \param[in] str a null terminated string to be parsed into a boolean
+ *  \param[in,out] val the boolean value parsed from the string. Left unchanged if string cannot be parsed to a valid bool
```
wjwwood: nitpick: This line looks like it needs to be wrapped (our line length, even for docs, is 100). You can have the linter do it for you, with `ament_uncrustify . --reformat` in the directory of your project, in a shell with `install/setup.bash` sourced.

rcl/src/rcl/arguments.c
```diff
@@ -235,6 +286,43 @@ rcl_parse_arguments(
       rcl_get_error_string());
     rcl_reset_error();
 
+    // Attempt to parse argument as log configuration file
+    if (RCL_RET_OK == _rcl_parse_external_log_config_file(argv[i], allocator, &args_impl->external_log_config_file)) {
```
wjwwood: nitpick: again, these look too long, I recommend using something like this:

```c++
rcl_ret_t ret = ...;
if (RCL_RET_OK == ret) { ...
```

rcl/src/rcl/arguments.c
```diff
@@ -235,6 +286,43 @@ rcl_parse_arguments(
...
+      continue;
+    }
+    RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME,
+      "Couldn't parse arg %d (%s) as log config rule. Error: %s", i, argv[i],
+      rcl_get_error_string());
```
wjwwood: I think this is out-of-date (I recommend updating your branches to make sure you're using master), it should be `rcl_get_error_string().str`, here and below.

rcl/src/rcl/arguments.c
```diff
@@ -1074,6 +1162,77 @@ _rcl_parse_param_file_rule(
...
+  const char * arg,
+  rcl_allocator_t allocator,
+  char ** log_config_file)
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(arg, RCL_RET_INVALID_ARGUMENT);
+
+  const size_t param_prefix_len = sizeof(RCL_EXTERNAL_LOG_CONFIG_ARG_RULE);
+  if (strncmp(RCL_EXTERNAL_LOG_CONFIG_ARG_RULE, arg, param_prefix_len) == 0) {
+    size_t outlen = strlen(arg) - param_prefix_len;
+    log_config_file = allocator.allocate(sizeof(char) * (outlen + 1), allocator.state);
```
wjwwood: Is `log_config_file` allowed to be `NULL`?

Also, shouldn't it be `*log_config_file = ...`? Unless I miss understand, isn't this supposed to be an output variable? Changing the value of `log_config_file` will not affect the the value passed into the function, which is an address in a struct anyways. Sorry if I'm "off-by-one" dereferencing there, this code is always tricky.

rcl/src/rcl/arguments.c
```diff
@@ -1074,6 +1162,77 @@ _rcl_parse_param_file_rule(
...
+  const char * arg,
+  const char * key,
+  bool * value)
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(arg, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(key, RCL_RET_INVALID_ARGUMENT);
+
+  const size_t param_prefix_len = strlen(key);
+  if (strncmp(key, arg, param_prefix_len) == 0) {
+    return _atob(arg + param_prefix_len, value);
```
wjwwood: Is value allowed to be `NULL`? If not should it be checked on the input at the beginning of the function.

wjwwood: Oh I guess it's checked in `_atob` anyways.

rcl/src/rcl/arguments_impl.h
```diff
@@ -44,6 +44,11 @@ typedef struct rcl_arguments_impl_t
 
   /// Default log level (represented by `RCUTILS_LOG_SEVERITY` enum) or -1 if not specified.
   int log_level;
+  /// A file used to configure the external logging library
+  char * external_log_config_file;
+  bool log_stdout_disabled;
+  bool log_rosout_disabled;
+  bool log_ext_lib_disabled;
```
wjwwood: nitpick: Even if a trivial sentence, it would be nice to have a doc brief for each of the new options.

rcl/src/rcl/rcl.c
```diff
@@ -118,10 +119,16 @@ rcl_init(int argc, char const * const * argv, rcl_allocator_t allocator)
...
   }
 
-  // Update the default log level if specified in arguments.
-  if (global_args->impl->log_level >= 0) {
-    rcutils_logging_set_default_logger_level(global_args->impl->log_level);
+  fail_ret = rcutils_logging_configure(global_args->impl->log_level,
+                                      global_args->impl->external_log_config_file,
+                                      !global_args->impl->log_stdout_disabled,
+                                      !global_args->impl->log_rosout_disabled,
+                                      !global_args->impl->log_ext_lib_disabled);
```
wjwwood: nitpick: This doesn't follow our style I think, but again, you can use our linters to see the diff it thinks should be made or use the `--reformat` option to have it do it for you (in the case of `ament_uncrustify`) at least.

The linters are also run as automated tests (via CTest).

tfoote:
Running CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5604/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2236/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4617/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5482/)

wjwwood:
@tfoote this will definitely fail CI due to style changes, and I suspect the compiler failures are due to upstream changes which will require this to be rebased to fix.

rcl/include/rcl/arguments.h
```diff
@@ -34,8 +34,12 @@ typedef struct rcl_arguments_t
...
 } rcl_arguments_t;
 
-#define RCL_LOG_LEVEL_ARG_RULE "__log_level:="
-#define RCL_PARAM_FILE_ARG_RULE "__params:="
+#define RCL_LOG_LEVEL_ARG_RULE              "__log_level:="
+#define RCL_EXTERNAL_LOG_CONFIG_ARG_RULE    "__log_config_file:="
+#define RCL_LOG_DISABLE_STDOUT_ARG_RULE     "__log_disable_stdout:="
+#define RCL_LOG_DISABLE_ROSOUT_ARG_RULE     "__log_disable_rosout:="
+#define RCL_LOG_DISABLE_EXT_LIB_ARG_RULE    "__log_disable_external_lib:="
+#define RCL_PARAM_FILE_ARG_RULE             "__params:="
```
dirk-thomas: Please don't align the values but use a single space between the macro name and the string literal.

rcl/src/rcl/arguments.c
nburek: The `log_config_file` input should be allowed to be NULL or an empty string if no config file is provided. 

You are correct about the pointer. A last minute change didn't get fully tested and this slipped by. I'll correct it.

rcl/src/rcl/arguments.c
nburek: Added the `.str` to the end.

rcl/src/rcl/arguments.c
nburek: Fixed the pointer dereference.

rcl/src/rcl/rcl.c
nburek: Fixed with ament_uncrustify.

rcl/include/rcl/arguments.h
nburek: Reverted to using a single space.

nburek:
I ran ament_uncrustify and had it fix all the formatting for the whole repository. I addressed all the formatting issues as well and all unit tests were passing locally on a Docker Linux image. 

The major change in this revision was to move the external logger, rosout, and multiple output log handlers into rcl instead of having them in rcutils. This was in response to the comments on the rcutils review suggesting that they not remain in rcutils. If you have a different place you'd prefer these to live I'd be more than happy to move them there instead.

rcl/src/rcl/arguments_impl.h
nburek: Added comments.

rcl/src/rcl/graph.c
```diff
@@ -52,7 +52,7 @@ rcl_get_topic_names_and_types(
     &rcutils_allocator,
     no_demangle,
     topic_names_and_types
-  );
+    );
```
dirk-thomas: Please avoid changes to code not touched by this PR.

nburek:
It looks like the automated build for this is failing because of the new package that I added not being available. Currently I've just posted this package in a repository on my personal account (here: https://github.com/nburek/rc_logging_log4cxx) because I think it should live in a new repository under the ros2 organization. Let me know how you would like to handle creating/reviewing this code.

rcl/src/rcl/graph.c
nburek: Removed the files that weren't part of my changes.

nburek:
Rebased on top of master and migrated changes made to rcl.c into init.c

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5763/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2348/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4764/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5665/)

mjcarroll:
CI Failures because I didn't include rc_logging_log4cxx, rerunning:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5765/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2350/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4766/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5667/)

mjcarroll:
It's going to fail due to log4cxx dependency not being on the CI at the moment.

mjcarroll:
@nburek I commented over on the repo, but I wanted to note it here as well.  In order for this to make it in, we'll need a way to satisfy the `log4cxx` dependency on all of our target platforms.  I couldn't find a readily available `choco` package or vendor package for Windows. https://github.com/nburek/rc_logging_log4cxx/issues/1

nburek:
Hey Michael, sorry I missed the log4cxx dependency issue. Since it was available on Linux via rosdep I took for granted that it would be available on other platforms as well. I'm actively working on trying to get this unblocked. If you have examples of other vendor packages I could reference that would be useful. 

In the mean time, if it would help unblock these Pull Requests and the new Pull Request I just opened for the rosout topic feature, I can swap out the rc_logger_log4cxx implementation that we're defaulting to with a no-op implementation then follow up with the rc_logger_log4cxx once that's actually working on all platforms.

nburek:
Michael, I've added a noop logger to the rc_logger_log4cxx repository and put a COLCON_IGNORE in the rc_logger_log4cxx package. If you guys are okay with that we can move forward with this PR and the rosout PR while we work on solving the dependency issue for Windows and MAC.

mjcarroll:
Here are a few of our `vendor_` packages that we've had to make in the past.  In general, they default to using the system package if available, otherwise they are built using some variant of CMake's `ExternalProject`: https://github.com/ros2?utf8=%E2%9C%93&q=vendor&type=&language=

mjcarroll:
CI with `noop`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5775/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2354/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4770/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5671/)

nburek:
Latest build failuire on Linux and macOS are because I need to rebase to pick up a change in `rcl_actions` that's breaking due to a change fields in a struct from another repository. 

The Windows failure is because it's failing to import the library exported by rc_logger_noop. I'm just finishing provisioning a Windows machine and should be able to start troubleshooting that as well.

mjcarroll:
With the rebase: 

Linux: [Build Status](https://ci.ros2.org/job/ci_linux/5778/)

macOS: [Build Status](https://ci.ros2.org/job/ci_osx/4775/)

nburek:
I've pushed a slight change to rc_logging_noop (now with the renamed github repository `rc_logging`) that forces the Windows build to compile rc_logging_noop as a static library as that seemed to be what it wanted. This unblocks the build, but I am still working on getting the unit tests to run as my windows machine doesn't seem to be able to link in fastrtps during runtime. 

I've also submitted an additional PR and linked it to this one in order to add nburek/rc_logging to ros2.repos file in ros2/ros2

nburek:
Looks like the unstable build on Linux and Mac is caused by two issues:
1) The xmllint is failing for `rcl` because I had to increase the package.xml version to 3 in order to add the group_depend and it doesn't seem to know how to deal with a version 3 file. 
2) There are uncrustify issues with rclcpp. Unless I missed something though, these issues appear to be with files that I did not touch in my PR.

rcl/package.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
```
sloretz: > The xmllint is failing for rcl because I had to increase the package.xml version to 3 in order to add the group_depend and it doesn't seem to know how to deal with a version 3 file.

I think the xml schema referenced here `package_format2.xsd` will have to be changed or removed if the package format is bumped. Looks like there is a version 3 schema http://download.ros.org/schema/package_format3.xsd

dirk-thomas: And the `group_depend` needs to be listed after the `test_depend`.

nburek: Changed to format3 and moved the group_depend after the test_depend. Ran colcon test locally on Linux and it appears everything in `rcl` is passing.

nburek:
Newest update here fixes the xmllint issue with rcl. We were also able to build rcl with the `rc_logging_noop` package from the `rc_logging` repository on Windows and were able to run some quick manually testing that it was working.

rcl/src/rcl/logging.c
```diff
@@ -0,0 +1,186 @@
...
+void
+rcl_logging_rosout_output_handler(
+  const rcutils_log_location_t * location,
+  int severity, const char * name, rcutils_time_point_value_t timestamp,
+  const char * format, va_list * args);
+
+rcl_ret_t
+rcl_logging_configure(const rcl_arguments_t * global_args, const rcl_allocator_t * allocator)
+{
+  // TODO(wjwwood): to be used in a future pull request
```
wjwwood:
```suggestion
```

rcl/src/rcl/logging.c
```diff
@@ -0,0 +1,186 @@
...
+
+  va_list args_clone;
+  va_copy(args_clone, *args);
+  status = rcutils_char_array_vsprintf(&msg_array, format, args_clone);
+  va_end(args_clone);
+
+  if (RCL_RET_OK == status) {
+    status = rcutils_logging_format_message(
+      location, severity, name, timestamp, msg_array.buffer, &output_array);
+  }
```
wjwwood: Similar comment here that we should print to the screen or something at least when a function like this fails. For `rcutils_logging_format_message()`, which probably sets the error state, so the printed message should probably contain `rcl_get_error_string().str` and do `rcl_reset_error()` afterwards.

rcl/src/rcl/logging.c
```diff
@@ -0,0 +1,186 @@
...
+  rcl_logging_external_log(severity, name, output_array.buffer);
+  status = rcutils_char_array_fini(&msg_array);
+  RCL_UNUSED(status);
```
wjwwood: Again, still useful to see something when these fails. Even though we cannot propagate this error or mitigate it, if these fail something is probably wrong.

PR #330, opened 2018-11-14
Add function for waiting on multiple wait sets together

### LINKED ISSUE #308: Wait set groups

jacobperron:
Internally, the function collates all wait sets into a single wait set that is waited on.

Resolves #308

### DIFF
--- a/rcl/include/rcl/wait.h
+++ b/rcl/include/rcl/wait.h
@@ -406,7 +406,6 @@ rcl_wait_set_add_service(
  * If the timeout is greater than 0 then this function will return after
  * that period of time has elapsed or the wait set becomes ready, which ever
  * comes first.
- * Passing a timeout struct with uninitialized memory is undefined behavior.
  *
  * This function is thread-safe for unique wait sets with unique contents.
  * This function cannot operate on the same wait set in multiple threads, and
@@ -427,6 +426,74 @@ RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_ret_t
 rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout);
+
+/// Block until multiple wait sets are ready or until the timeout has been exceeded.
+/**
+ * This function has similar behavior as rcl_wait(), but acts on multiple wait sets.
+ *
+ * This function will collect the items from the array of rcl_wait_set_t's and pass them
+ * to the underlying rmw_wait function.
+ *
+ * The items in the wait sets will be either left untouched or set to `NULL` after
+ * this function returns.
+ * Items that are not `NULL` are ready, where ready means different things based
+ * on the type of the item.
+ * For subscriptions this means there are messages that can be taken.
+ * For guard conditions this means the guard condition was triggered.
+ *
+ * Expected usage:
+ *
+ * TODO(jacobperron): Usage example
+ *
+ * The wait sets must be allocated, initialized, and should have been
+ * cleared and then filled with items, e.g. subscriptions and guard conditions.
+ * If there are no wait-able items in all wait sets passed, this function will fail.
+ * `NULL` items in the sets are ignored, e.g. it is valid to have as input:
+ *  - `subscriptions[0]` = valid pointer
+ *  - `subscriptions[1]` = `NULL`
+ *  - `subscriptions[2]` = valid pointer
+ *  - `size_of_subscriptions` = 3
+ * Passing an uninitialized (zero initialized) wait set struct will fail.
+ * Passing a wait set struct with uninitialized memory is undefined behavior.
+ *
+ * The unit of timeout is nanoseconds.
+ * If the timeout is negative then this function will block indefinitely until
+ * something in the wait set is valid or it is interrupted.
+ * If the timeout is 0 then this function will be non-blocking; checking what's
+ * ready now, but not waiting if nothing is ready yet.
+ * If the timeout is greater than 0 then this function will return after
+ * that period of time has elapsed or the wait set becomes ready, which ever
+ * comes first.
+ *
+ * This function is thread-safe for unique wait sets with unique contents.
+ * This function cannot operate on the same wait set in multiple threads, and
+ * the wait sets may not share content.
+ * For example, calling rcl_wait_multiple() in two threads on two different wait sets
+ * that both contain a single, shared guard condition is undefined behavior.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[inout] wait_sets pointer to an array of wait sets to be waited on and to be pruned
+ *   if not ready
+ * \param[in] num_wait_sets the number of wait sets in the input array
+ * \param[in] timeout the duration to wait for the wait set to be ready, in nanoseconds
+ * \return `RCL_RET_OK` if cleared successfully, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return `RCL_RET_WAIT_SET_INVALID` if any wait sets are zero initialized, or
+ * \return `RCL_RET_WAIT_SET_EMPTY` if all wait sets are empty, or
+ * \return `RCL_RET_TIMEOUT` if the timeout expired before something was ready, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_wait_multiple(rcl_wait_set_t ** wait_sets, const size_t num_wait_sets, int64_t timeout);
 
 #ifdef __cplusplus
 }
--- a/rcl/include/rcl/wait.h
+++ b/rcl/include/rcl/wait.h
@@ -378,7 +378,7 @@ rcl_wait_set_add_service(
  *   }
  *   for (int i = 0; i < wait_set.size_of_guard_conditions; ++i) {
  *     if (wait_set.guard_conditions[i]) {
- *       // The subscription is ready...
+ *       // The guard condition is ready...
  *     }
  *   }
  * } while(check_some_condition());
@@ -443,7 +443,52 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout);
  *
  * Expected usage:
  *
- * TODO(jacobperron): Usage example
+ * ```c
+ * #include <rcl/rcl.h>
+ *
+ * // rcl_init() called successfully before here...
+ * rcl_node_t node;  // initialize this, see rcl_node_init()
+ * rcl_subscription_t sub1;  // initialize this, see rcl_subscription_init()
+ * rcl_subscription_t sub2;  // initialize this, see rcl_subscription_init()
+ * rcl_guard_condition_t gc1;  // initialize this, see rcl_guard_condition_init()
+ * rcl_wait_set_t wait_sets[2];  // two wait sets in this example
+ * wait_sets[0] = rcl_get_zero_initialized_wait_set();
+ * wait_sets[1] = rcl_get_zero_initialized_wait_set();
+ * rcl_ret_t ret = rcl_wait_set_init(&wait_sets[0] 1, 0, 0, 0, 0, rcl_get_default_allocator());
+ * // ... error handling
+ * ret = rcl_wait_set_init(&wait_sets[1] 1, 1, 0, 0, 0, rcl_get_default_allocator());
+ * // ... error handling
+ * do {
+ *   ret = rcl_wait_set_clear(&wait_sets[0]);
+ *   // ... error handling
+ *   ret = rcl_wait_set_add_subscription(&wait_sets[0], &sub1);
+ *   // ... error handling
+ *   ret = rcl_wait_set_clear(&wait_sets[1]);
+ *   // ... error handling
+ *   ret = rcl_wait_set_add_subscription(&wait_sets[1], &sub2);
+ *   // ... error handling
+ *   ret = rcl_wait_set_add_guard_condition(&wait_sets[1], &gc1);
+ *   // ... error handling
+ *   ret = rcl_wait_multiple(wait_sets, RCL_MS_TO_NS(1000));  // 1000ms == 1s, passed as ns
+ *   if (ret == RCL_RET_TIMEOUT) {
+ *     continue;
+ *   }
+ *   if (wait_sets[0].subscriptions[0]) {
+ *     // The subscription in the first wait set is ready...
+ *   }
+ *   if (wait_sets[1].subscriptions[0]) {
+ *     // The subscription in the second wait set is ready...
+ *   }
+ *   if (wait_sets[1].guard_conditions[0]) {
+ *     // The guard condition in the first wait set is ready...
+ *   }
+ * } while(check_some_condition());
+ * // ... fini node, and subscriptions and guard conditions...
+ * ret = rcl_wait_set_fini(&wait_sets[0]);
+ * // ... error handling
+ * ret = rcl_wait_set_fini(&wait_sets[1]);
+ * // ... error handling
+ * ```
  *
  * The wait sets must be allocated, initialized, and should have been
  * cleared and then filled with items, e.g. subscriptions and guard conditions.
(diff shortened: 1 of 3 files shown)


jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5546/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2200/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4575/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5436/)

rcl/include/rcl/wait.h
```diff
@@ -428,6 +427,119 @@ RCL_WARN_UNUSED
 rcl_ret_t
 rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout);
 
+/// Block until multiple wait sets are ready or until the timeout has been exceeded.
+/**
+ * This function has similar behavior as rcl_wait(), but acts on multiple wait sets.
```
hidmic: @jacobperron nit: similar behavior _to_?

rcl/test/rcl/test_wait.cpp
```diff
@@ -435,3 +435,28 @@ TEST_F(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), guard_condition) {
   EXPECT_EQ(RCL_RET_OK, f.get());
   EXPECT_LE(std::abs(diff - trigger_diff.count()), TOLERANCE);
 }
+
+// Test rcl_wait_multiple with a positive finite timeout value (1ms)
+TEST_F(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), finite_timeout_multiple) {
```
hidmic: @jacobperron shouldn't we add a few more tests? Like, for negative timeout, for zero timeout, for invalid arguments, for invalid wait sets, with a dummy guard and a dummy timer in different wait sets to make sure things are being collated and then updated back properly, and so on. Not to repeat `rcl_wait()` tests, just to verify the new functionality is correct.

rcl/src/rcl/wait.c
```diff
@@ -614,6 +614,154 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
...
+  size_t i;
+  size_t j;
+  rcl_wait_set_t * wait_set;
+  // Count number of entities in wait sets while checking for invalid wait sets
+  size_t number_of_subscriptions = 0u;
+  size_t number_of_guard_conditions = 0u;
+  size_t number_of_timers = 0u;
+  size_t number_of_clients = 0u;
+  size_t number_of_services = 0u;
+  for (i = 0u; i < num_wait_sets; ++i) {
```
hidmic: @jacobperron it could be that no wait set was provided, should we have a different rcl error message for that case (right now it'll say that wait sets are empty)?

rcl/src/rcl/wait.c
```diff
@@ -614,6 +614,154 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
...
+    number_of_services += wait_set->size_of_services;
+  }
+  if (
+    number_of_subscriptions == 0u &&
+    number_of_guard_conditions == 0u &&
+    number_of_timers == 0u &&
+    number_of_clients == 0u &&
+    number_of_services == 0u)
+  {
+    RCL_SET_ERROR_MSG("wait set is empty");
```
hidmic: @jacobperron consider `"wait sets are empty"` instead.

rcl/src/rcl/wait.c
```diff
@@ -614,6 +614,154 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
   return RCL_RET_OK;
 }
 
+rcl_ret_t
+rcl_wait_multiple(rcl_wait_set_t wait_sets[], const size_t num_wait_sets, int64_t timeout)
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(wait_sets, RCL_RET_INVALID_ARGUMENT);
+  size_t i;
+  size_t j;
```
hidmic: @jacobperron nit: I have a slight preference for declaring loop variables within the `for` statement. Won't block if you prefer to keep things C89-ish.

rcl/src/rcl/wait.c
```diff
@@ -614,6 +614,154 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
...
+  // Collate all wait sets into a single wait set
+  for (i = 0u; i < num_wait_sets; ++i) {
+    wait_set = &wait_sets[i];
+    for (j = 0u; j < wait_set->size_of_timers; ++j) {
+      const size_t current_index = collated_wait_set.impl->timer_index++;
+      collated_wait_set.timers[current_index] = wait_set->timers[j];
+      rcl_guard_condition_t * guard_condition = rcl_timer_get_guard_condition(wait_set->timers[j]);
+      if (NULL != guard_condition) {
+        // rcl_wait() will take care of moving these backwards and setting guard_condition_count.
+        const size_t index = wait_set->size_of_guard_conditions + (wait_set->impl->timer_index - 1);
```
hidmic: @jacobperron not sure I follow how `index` would take different values across multiple iterations for the same `wait_set`.

rcl/src/rcl/wait.c
```diff
@@ -614,6 +614,154 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
...
+
+  // Update individual wait sets
+  size_t collated_subscription_index = 0u;
+  size_t collated_guard_condition_index = 0u;
+  size_t collated_timer_index = 0u;
+  size_t collated_client_index = 0u;
+  size_t collated_service_index = 0u;
+  for (i = 0u; i < num_wait_sets; ++i) {
+    wait_set = &wait_sets[i];
+    for (j = 0u; j < wait_set->size_of_timers; ++j) {
```
hidmic: @jacobperron nit: fine to use `for` loops, but we could use `memcpy` too (that depending on the compiler and target architecture may be more performant).

rcl/src/rcl/wait.c
jacobperron: My intention was to return `RCL_RET_WAIT_SET_EMPTY` thinking it is intuitive, but I can add a new return code e.g. `RCL_RET_WAIT_SET_NONE`.

hidmic: Either that or `RCL_RET_INVALID_ARGUMENT` plus `RCL_SET_ERROR_MSG()` with a nice message.

rcl/src/rcl/wait.c
jacobperron: I guess my head was in C89 mode. Since we're targeting C99, I'm happy to change it.

rcl/src/rcl/wait.c
jacobperron: Sure, I can do that. Do you think there is any issues with readability (and correctness)?

```c
    memcpy(
      wait_set->timers,
      collated_wait_set.timers + collated_timer_index,
      sizeof(rcl_timer_t *) * wait_set->size_of_timers);
```

hidmic: Ugh, that doesn't look nearly as clean as your current code. Nevermind, let's trust loop unrolling will work :).

jacobperron: :+1:

rcl/src/rcl/wait.c
jacobperron: It won't, this is a copy-paste bug.

rcl/test/rcl/test_wait.cpp
jacobperron: Yes, I meant to, but forgot.

jacobperron:
@hidmic Thanks for the review! I think I've address all of your comments. Ready for another round of review.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5561/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2207/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4582/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5450/)

rcl/src/rcl/wait.c
```diff
@@ -614,6 +610,157 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
...
 }
 
+rcl_ret_t
+rcl_wait_multiple(rcl_wait_set_t wait_sets[], const size_t num_wait_sets, int64_t timeout)
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(wait_sets, RCL_RET_INVALID_ARGUMENT);
+  if (0u == num_wait_sets) {
+    RCL_SET_ERROR_MSG("number of wait sets must be greater than zero");
+    return RCL_RET_INVALID_ARGUMENT;
+  }
```
sloretz: Could take a short cut and return `rcl_wait()` if `1u == num_wait_sets`

rcl/src/rcl/wait.c
```diff
@@ -614,6 +610,157 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
...
+  // Collate all wait sets into a single wait set
+  for (size_t i = 0u; i < num_wait_sets; ++i) {
+    rcl_wait_set_t * wait_set = &wait_sets[i];
+    for (size_t j = 0u; j < wait_set->size_of_timers; ++j) {
+      const size_t current_index = collated_wait_set.impl->timer_index++;
+      collated_wait_set.timers[current_index] = wait_set->timers[j];
+      rcl_guard_condition_t * guard_condition = rcl_timer_get_guard_condition(wait_set->timers[j]);
+      if (NULL != guard_condition) {
+        // rcl_wait() will take care of moving these backwards and setting guard_condition_count.
+        const size_t offset_rmw_index = wait_set->size_of_guard_conditions + current_index;
```
sloretz: `offset_rmw_index` is in the combined wait set, so I think this should be `collated_wait_set->size_of_guard_conditions ...`

rcl/test/rcl/test_wait.cpp
```diff
@@ -435,3 +435,256 @@ TEST_F(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), guard_condition) {
...
+  const size_t kNumWaitSets = 3u;
+  rcl_wait_set_t wait_sets[kNumWaitSets];
+  rcl_guard_condition_t guard_conds[kNumWaitSets];
+  rcl_timer_t timers[kNumWaitSets];
+  for (size_t i = 0u; i < kNumWaitSets; ++i) {
+    wait_sets[i] = rcl_get_zero_initialized_wait_set();
+    ret = rcl_wait_set_init(&wait_sets[i], 0, 1, 1, 0, 0, rcl_get_default_allocator());
+    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+
+    // Add a dummy guard condition to avoid an error
```
sloretz: What error is the dummy guard condition avoiding?

rcl/src/rcl/wait.c
jacobperron: Yeah, good catch!

rcl/test/rcl/test_wait.cpp
jacobperron: That is a good question. I copied this over from the existing test involving a timer, without it the test segfaults :fearful:

I dug into it a bit more, and it seems like the rmw guard condition count isn't being incremented in [rcl_wait_set_add_timer()](https://github.com/ros2/rcl/blob/f256da0a86e20c72c6babd394bb34dea697ecbbe/rcl/src/rcl/wait.c#L427). Adding a dummy guard condition appears to be a hack to get the count to increment...

I think it can be fixed easily by ensuring the rmw guard condition count is incremented when adding timers, but without understanding the reason for the current implementation I am hesitant to change it. @wjwwood, is this a bug or intentional?

rcl/src/rcl/wait.c
jacobperron: Fixed in b58dfa9

rcl/test/rcl/test_wait.cpp
jacobperron: Nevermind, I see that the rmw guard condition count is increment for timers in `rcl_wait` as the comment says (my bad):

https://github.com/ros2/rcl/blob/f256da0a86e20c72c6babd394bb34dea697ecbbe/rcl/src/rcl/wait.c#L502

Setting the number of guard conditions to zero resolves the issue. I've also updated  the tests for `rcl_wait()` that originally confused me. See e40a26b

jacobperron:
Reviews addressed.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5567/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2212/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4587/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5457/)

hidmic:
@jacobperron are we pushing this one as-is or are you considering a re-implementation in terms of the feature introduced by #335 ?

jacobperron:
>are we pushing this one as-is or are you considering a re-implementation in terms of the feature introduced by #335 ?

@hidmic I think this feature is not necessary anymore in favor of #335. If you agree, I will close this PR.

hidmic:
Makes total sense to me, let's close it.

PR #285, opened 2018-08-21
consolidate wait set functions

sloretz:
This consolidates `rcl_wait_set_clear_*()` and `rcl_wait_set_reset_*()` into `rcl_wait_set_clear()` and `rcl_wait_set_reset()`. The reason for doing this is in another PR (todo create PR and link it) I need to create an RMW guard condition for each timer, and I found it easier to write the logic to resize and clear timers and guard conditions if all the info is in the same place at the same time.

PR #334, opened 2018-11-17
Prepare 0.6.0

nuclearsandwich:
I generated and edited changelogs for this repository using my best judgement but I think it would be optimal of folks with more context also gave them a once-over.

I may release this as-is over the weekend to unblock further crystal releases but changelog modifications can come after-the-fact too.

### DIFF
--- a/rcl/CHANGELOG.rst
+++ b/rcl/CHANGELOG.rst
@@ -2,6 +2,48 @@
 Changelog for package rcl
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
+0.6.0 (2018-11-16)
+------------------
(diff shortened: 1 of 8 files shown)


jacobperron:
The changelogs look okay to me. It looks like there were some API changes that are not backwards compatible (e.g. #285). Does this constitute a major version bump?

dirk-thomas:
A minor version bump pre 1.0 is sensually a "major" bump allowing any kind of breakage.

nuclearsandwich:
There was a PR that went in after I'd set this up so I pulled in the latest master updated changelogs and released that instead of this as 0.6.0.

PR #336, opened 2018-11-19
refactor init to not be global

wjwwood:
There are few important things to point here:

- init now takes `init_options` as input and outputs into a `context` type
- the `context` is passed to "top level" things like nodes, timers, and guard conditions
- some global state is now no longer global and therefore some functions are no longer required, like `rcl_get_global_arguments()` (now arguments are in the context object)
- the `context` has four states: uninitialized (undefined behavior when used), zero-initialized, initialized and valid, and initialized and invalid (post shutdown, pre-fini)
  - look at the context header for more details
  - this extra "invalid but still initialized" state allows you to shutdown while "children" like nodes and their children still use the context to cleanup

Connects to ros2/rmw#154


rcl/include/rcl/context.h
```diff
@@ -0,0 +1,251 @@
...
+ * |                                                                                 |
+ * |               +-----------------------------------------------------------------+
+ * |               |
+ * |  +------------v------------+
+ * |  |                         |
+ * |  | initialized but invalid +---> finalize all entities, then rcl_context_fini() +
+ * |  |                         |                                                    |
+ * |  +-------------------------+                                                    |
+ * |                                                                                 |
+ * +---------------------------------------------------------------------------------+
```
dirk-thomas: Maybe using `\dot` or `\msc` would look nicer than ASCII art.

rcl/include/rcl/context.h
```diff
@@ -0,0 +1,251 @@
...
+   * A value of `0` indicates that it is either invalid or zero-initialized.
+   *
+   * Accessing the instance id should still be done using the function
+   * `rcl_context_get_instance_id()` because the instance id's type is an
+   * atomic and needs to be accessed properly to ensure safety.
+   *
+   * The instance id should not be changed manually - doing so is undefined
+   * behavior.
+   *
+   * The instance id cannot be protected within the `impl` pointer's type
```
jacobperron: Could we instead update `rcl_context_get_instance_id()` to return `0` if the context is zero initialized? Then we could push the instance ID into the impl.

rcl/include/rcl/context.h
wjwwood: dot is not very easy to understand as text, and I don't know what msc is. I won't spend more time on this diagram unless there's some other reason.

rcl/test/test_namespace.cpp
```diff
@@ -33,17 +33,28 @@ using namespace std::chrono_literals;
...
-    ret = rcl_init(0, nullptr, rcl_get_default_allocator());
-    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+    {
+      rcl_init_options_t init_options = rcl_get_zero_initialized_init_options();
+      ret = rcl_init_options_init(&init_options, rcl_get_default_allocator());
+      ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+      OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT({
+        EXPECT_EQ(RCL_RET_OK, rcl_init_options_fini(&init_options)) << rcl_get_error_string().str;
+      });
+      this->context_ptr = new rcl_context_t;
```
jacobperron: Is there a particular reason to use `new`? We could also declare as a non-pointer type and pass a reference below.

The same for the node handle as well.

rcl/include/rcl/context.h
wjwwood: I just looked up `msc`, which could be a useful way to display it in the rendered view, but as I said, I would prefer something that can be understood easily in plain text. If the diagram were more complex I'd give up on that and just use `dot` or something.

rcl/include/rcl/context.h
wjwwood: The reason I wanted to avoid that was this case:

- thread 1: call rcl_context_get_instance_id()
- thread 1: if (context->impl)
- thread 2: rcl_context_fini(context) ... until complete
- thread 1: return context->impl->instance_id <--- nullptr access

rcl/test/test_namespace.cpp
wjwwood: Mostly because that's how it was before (node was a pointer setup with new/delete). I think there might have been a reason for it, but I don't remember now.

rcl/include/rcl/context.h
jacobperron: Okay, I see. Maybe the address of the impl should should double as the ID :stuck_out_tongue_winking_eye:

wjwwood: That would be neat, but unfortunately once memory is free'd it can be reused (so a future init can get the same pointer as the current init if it is ever finalized).

wjwwood: I pushed to address this by hiding it behind "storage" (i.e. uint8_t array), but it still cannot be put into `impl` due to the race condition: 9e7ce7a

wjwwood: Also ros2/rcutils@fb42961.

jacobperron: Well, at least people should be less tempted to modify a byte array than a single integer :smile:

wjwwood:
> Might need a rebase if I get #323 merged first 😉

Definitely did :) But that's ok, I'm almost there...

gonzodepedro:
LGTM

rcl/src/rcl/init.c
```diff
@@ -0,0 +1,172 @@
...
+
+  // Zero initialize global arguments.
+  context->global_arguments = rcl_get_zero_initialized_arguments();
+
+  // Setup impl for context.
+  context->impl = allocator.allocate(sizeof(rcl_context_impl_t), allocator.state);
+  RCL_CHECK_FOR_NULL_WITH_MSG(
+    context->impl, "failed to allocate memory for context impl", return RCL_RET_BAD_ALLOC);
+  // memset to 0 so the cleanup function will not try to clean up uninitialized parts later
+  memset(context->impl, 0x0, sizeof(rcl_context_impl_t));
```
sloretz: nit, could use [allocator.zero_allocate(...)](https://github.com/ros2/rcutils/blob/670a3e302be214f3f2e7550ab7594ecfae0a99da/include/rcutils/allocator.h#L67)

rcl/src/rcl/init.c
```diff
@@ -0,0 +1,172 @@
...
+  // Copy the argc and argv into the context, if argc >= 0.
+  context->impl->argc = argc;
+  context->impl->argv = NULL;
+  if (0 != argc && argv != NULL) {
+    context->impl->argv = (char **)allocator.allocate(sizeof(char *) * argc, allocator.state);
+    RCL_CHECK_FOR_NULL_WITH_MSG(
+      context->impl->argv,
+      "failed to allocate memory for argv",
+      fail_ret = RCL_RET_BAD_ALLOC; goto fail);
+    memset(context->impl->argv, 0, sizeof(char *) * argc);
```
sloretz: nit, zero_allocate

rcl/src/rcl/init.c
```diff
@@ -0,0 +1,172 @@
...
+    ROS_PACKAGE_NAME,
+    "Shutting down ROS client library, for context at address: %p", context);
+  RCL_CHECK_ARGUMENT_FOR_NULL(context, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_FOR_NULL_WITH_MSG(
+    context->impl, "context is zero-initialized", return RCL_RET_INVALID_ARGUMENT);
+  if (!rcl_context_is_valid(context)) {
+    RCL_SET_ERROR_MSG("rcl_shutdown already called on the given context");
+    return RCL_RET_ALREADY_SHUTDOWN;
+  }
+
```
sloretz: also call `rmw_shutdown()`?

rcl/src/rcl/init_options.c
```diff
@@ -0,0 +1,122 @@
...
+  init_options->impl = allocator.allocate(sizeof(rcl_init_options_impl_t), allocator.state);
+  RCL_CHECK_FOR_NULL_WITH_MSG(
+    init_options->impl,
+    "failed to allocate memory for init options impl",
+    return RCL_RET_BAD_ALLOC);
+  init_options->impl->allocator = allocator;
+  init_options->impl->rmw_init_options = rmw_get_zero_initialized_init_options();
+  rmw_ret_t rmw_ret = rmw_init_options_init(&(init_options->impl->rmw_init_options), allocator);
+  if (RMW_RET_OK != rmw_ret) {
+    RCL_SET_ERROR_MSG(rmw_get_error_string().str);
```
sloretz: should also deallocate `init_options->impl` before returning

rcl/src/rcl/init_options.c
```diff
@@ -0,0 +1,122 @@
...
+    return ret;  // error already set
+  }
+
+  // copy src information into dst
+  dst->impl->allocator = src->impl->allocator;
+  // first zero-initialize rmw init options
+  rmw_ret_t rmw_ret = rmw_init_options_fini(&(dst->impl->rmw_init_options));
+  if (RMW_RET_OK != rmw_ret) {
+    RCL_SET_ERROR_MSG(rmw_get_error_string().str);
+    return rcl_convert_rmw_ret_to_rcl_ret(rmw_ret);
```
sloretz: fini `dst` before returning?

rcl/src/rcl/init_options.c
```diff
@@ -0,0 +1,122 @@
...
+  }
+  // then copy
+  dst->impl->rmw_init_options = rmw_get_zero_initialized_init_options();
+  rmw_ret =
+    rmw_init_options_copy(&(src->impl->rmw_init_options), &(dst->impl->rmw_init_options));
+  if (RMW_RET_OK != rmw_ret) {
+    RCL_SET_ERROR_MSG(rmw_get_error_string().str);
+    return rcl_convert_rmw_ret_to_rcl_ret(rmw_ret);
```
sloretz: fini `dst` before returning?

rcl/test/rcl/test_publisher.cpp
```diff
@@ -53,7 +64,8 @@ class CLASSNAME (TestPublisherFixture, RMW_IMPLEMENTATION) : public ::testing::T
     rcl_ret_t ret = rcl_node_fini(this->node_ptr);
     delete this->node_ptr;
     EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
-    ret = rcl_shutdown();
+    ret = rcl_shutdown(this->context_ptr);
+    delete this->context_ptr;
```
sloretz: `rcl_context_fini()` after `rcl_shutdown()`?

rcl/test/rcl/test_service.cpp
```diff
@@ -56,7 +67,8 @@ class CLASSNAME (TestServiceFixture, RMW_IMPLEMENTATION) : public ::testing::Tes
     rcl_ret_t ret = rcl_node_fini(this->node_ptr);
     delete this->node_ptr;
     EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
-    ret = rcl_shutdown();
+    ret = rcl_shutdown(this->context_ptr);
+    delete this->context_ptr;
```
sloretz: same comment about `rcl_context_fini()`

rcl/test/rcl/test_subscription.cpp
```diff
@@ -56,7 +67,8 @@ class CLASSNAME (TestSubscriptionFixture, RMW_IMPLEMENTATION) : public ::testing
     rcl_ret_t ret = rcl_node_fini(this->node_ptr);
     delete this->node_ptr;
     EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
-    ret = rcl_shutdown();
+    ret = rcl_shutdown(this->context_ptr);
+    delete this->context_ptr;
```
sloretz: same comment about `rcl_context_fini()`

rcl/test/rcl/test_timer.cpp
```diff
@@ -45,7 +56,8 @@ class TestTimerFixture : public ::testing::Test
     rcl_ret_t ret = rcl_node_fini(this->node_ptr);
     delete this->node_ptr;
     EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
-    ret = rcl_shutdown();
+    ret = rcl_shutdown(this->context_ptr);
+    delete this->context_ptr;
```
sloretz: same comment about `rcl_context_fini()`

rcl/test/rcl/test_wait.cpp
```diff
@@ -41,18 +41,29 @@
...
+    ret = rcl_init(0, nullptr, &init_options, this->context_ptr);
     ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
   }
 
   void TearDown()
   {
-    rcl_ret_t ret;
-    ret = rcl_shutdown();
-    EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+    OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT({
```
sloretz: nit, `std::shared_ptr` instead of scope exit

rcl/test/test_namespace.cpp
```diff
@@ -52,7 +63,8 @@ class TestNamespaceFixture : public ::testing::Test
     rcl_ret_t ret = rcl_node_fini(this->node_ptr);
     delete this->node_ptr;
     EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
-    ret = rcl_shutdown();
+    ret = rcl_shutdown(this->context_ptr);
+    delete this->context_ptr;
```
sloretz: same comment about `rcl_context_fini()`

rcl_lifecycle/test/test_default_state_machine.cpp
```diff
@@ -56,7 +68,8 @@ class TestDefaultStateMachine : public ::testing::Test
     rcl_ret_t ret = rcl_node_fini(this->node_ptr);
     delete this->node_ptr;
     EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
-    ret = rcl_shutdown();
+    ret = rcl_shutdown(this->context_ptr);
+    delete this->context_ptr;
```
sloretz: same comment about rcl_context_fini

rcl_lifecycle/test/test_multiple_instances.cpp
```diff
@@ -53,7 +66,8 @@ class TestMultipleInstances : public ::testing::Test
     rcl_ret_t ret = rcl_node_fini(this->node_ptr);
     delete this->node_ptr;
     EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
-    ret = rcl_shutdown();
+    ret = rcl_shutdown(this->context_ptr);
+    delete this->context_ptr;
```
sloretz: same comment about rcl_context_fini

rcl/src/rcl/init.c
wjwwood: Definitely, fixed that in latest commits.

PR #340, opened 2018-11-24
update tests to use separated action types

dirk-thomas:
Connected to ros2/rosidl#334.

### DIFF
--- a/rcl_action/test/rcl_action/test_action_communication.cpp
+++ b/rcl_action/test/rcl_action/test_action_communication.cpp
@@ -156,14 +156,14 @@ protected:
 
 TEST_F(CLASSNAME(TestActionCommunication, RMW_IMPLEMENTATION), test_valid_goal_comm)
 {
-  test_msgs__action__Fibonacci_Goal_Request outgoing_goal_request;
-  test_msgs__action__Fibonacci_Goal_Request incoming_goal_request;
-  test_msgs__action__Fibonacci_Goal_Response outgoing_goal_response;
-  test_msgs__action__Fibonacci_Goal_Response incoming_goal_response;
-  test_msgs__action__Fibonacci_Goal_Request__init(&outgoing_goal_request);
-  test_msgs__action__Fibonacci_Goal_Request__init(&incoming_goal_request);
-  test_msgs__action__Fibonacci_Goal_Response__init(&outgoing_goal_response);
-  test_msgs__action__Fibonacci_Goal_Response__init(&incoming_goal_response);
+  test_msgs__action__Fibonacci_Action_Goal_Request outgoing_goal_request;
+  test_msgs__action__Fibonacci_Action_Goal_Request incoming_goal_request;
+  test_msgs__action__Fibonacci_Action_Goal_Response outgoing_goal_response;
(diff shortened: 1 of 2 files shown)


rcl_action/test/rcl_action/test_action_communication.cpp
```diff
@@ -156,18 +156,18 @@ class CLASSNAME (TestActionCommunication, RMW_IMPLEMENTATION) : public ::testing
...
+  test_msgs__action__Fibonacci_SendGoal_Response incoming_goal_response;
+  test_msgs__action__Fibonacci_SendGoal_Request__init(&outgoing_goal_request);
+  test_msgs__action__Fibonacci_SendGoal_Request__init(&incoming_goal_request);
+  test_msgs__action__Fibonacci_SendGoal_Response__init(&outgoing_goal_response);
+  test_msgs__action__Fibonacci_SendGoal_Response__init(&incoming_goal_response);
 
   // Initialize goal request
-  init_test_uuid0(outgoing_goal_request.action_goal_id.uuid);
-  outgoing_goal_request.order = 10;
+  init_test_uuid0(outgoing_goal_request.goal_id);
```
jacobperron: I would have expected this to be `outgoing_goal_request.goal_id.uuid`, as `outgoing_goal_request.action_goal_id` was of type [unique_identifer_msgs/UUID](https://github.com/ros2/unique_identifier_msgs/blob/master/msg/UUID.msg).

https://github.com/ros2/rosidl/blob/7c5c7d3af98fb8f5fd8ae9112ee278d21f899df1/rosidl_adapter/rosidl_adapter/parser.py#L888

dirk-thomas: The referenced change has landed a long time after the IDL branches were created. So atm this branch uses a `uint8[16]` (see https://github.com/ros2/rosidl/blob/6c44fd023b5503e174bafaeee2544b063fd654e6/rosidl_parser/rosidl_parser/definition.py#L505).

This certainly needs to be updated to keep using the `uuid` message.

hidmic: Sorry for jumping into the discussion, and this is absolutely tangential to the IDL work, but IMHO only a small portion of the implementation should need to know about the `uuid` field in a `goal_id`, while the rest could just use `goal_id` as an opaque entity.

jacobperron: I agree. But I believe in this case the code should know about the UUID (or `uint8[16]`) since it is setting it to a known value.

Currently, client libraries rely on the knowledge that a goal ID is a UUID for tracking goal handles, e.g.

https://github.com/ros2/rclcpp/blob/10d7b7c72bc243e4e6da0546f9d17e39460df55c/rclcpp_action/include/rclcpp_action/types.hpp#L32

https://github.com/ros2/rclcpp/blob/10d7b7c72bc243e4e6da0546f9d17e39460df55c/rclcpp_action/include/rclcpp_action/client.hpp#L321

IMO ideally, we could limit the heavy usage of UUIDs to `rcl_action` and try to keep most of the goal tracking logic abstracted there. I've created an feature request as a better venue for this discussion: #389

PR #341, opened 2018-11-27
rcl_action_process_cancel_request no longer changes goal state

sloretz:
`rcl_action_process_cancel_request()` both calculates which goals are being cancelled and transitions them to the `CANCELLING` state. This PR makes it only calculate the list of goals.

The c++ action server needs the list first so it can call a user's cancel callback. The goal should be transitioned to the `CANCELLING` state after the user callback indicates the goal can be cancelled.

### DIFF
--- a/rcl_action/include/rcl_action/action_server.h
+++ b/rcl_action/include/rcl_action/action_server.h
@@ -659,7 +659,8 @@ rcl_action_take_cancel_request(
 /**
  * This is a non-blocking call.
  *
- * This function will transition one or more active goals to the state CANCELING.
+ * This function will compute a list of goals that a cancelation request is attempting to cancel.
+ * It does not change the state of any goal.
  * The following cancel policy applies based on the goal ID and the timestamp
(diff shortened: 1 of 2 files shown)


sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5665/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2274/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4684/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5557/)

sloretz:
> Okay, so it will be the callers responsibility to cancel, correct?

Yup. If the user's [`handle_cancel`](https://github.com/ros2/examples/blob/426c1dbbfb6a754701c07891311d64c357ef7be9/rclcpp/minimal_action_server/not_composable.cpp#L38-L42) returns `ACCEPT` then `rclcpp_action` will transition the goal.

PR #342, opened 2018-11-28
rcl_action_expire_goals() outputs goals that expire

sloretz:
The C++ action server needs to keep around the action result message until the goals expire. In order to know when it can throw away the result it needs to know when the `rcl_action_server_t` decided to expire the goal. This makes the API `rcl_action_expire_goals()` output the list of goals that expired.

PR #343, opened 2018-11-29
Add timer to action server to check expired goals

sloretz:
This adds a timer to `rcl_action_server_t` that fires when an inactive goal handle expires and should be forgotten by the server.

This also includes commits from #342 only because I need both for ros2/rclcpp#593. #342 should be merged first.


sloretz:
This one might conflict with #336 (will need to pass context to `rcl_timer_init()`) so I'll wait until that one is merged before fixing it here and running CI.

sloretz:
CI

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5707/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2306/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4717/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5620/)

rcl_action/src/rcl_action/action_server.c
```diff
@@ -395,6 +421,66 @@ rcl_action_accept_new_goal(
...
+  rcl_clock_t * clock)
+{
+  size_t num_inactive_goals = 0u;
+  int64_t minimum_period = 0;
+
+  // Get current time (nanosec)
+  int64_t current_time;
+  rcl_ret_t ret = rcl_clock_get_now(clock, &current_time);
+  if (RCL_RET_OK != ret) {
+    return RCL_RET_ERROR;
```
wjwwood:
```suggestion
    return ret;
```

return `ret` instead?

rcl_action/src/rcl_action/action_server.c
```diff
@@ -395,6 +421,66 @@ rcl_action_accept_new_goal(
...
+
+  for (size_t i = 0; i < num_goal_handles; ++i) {
+    rcl_action_goal_handle_t * goal_handle = goal_handles[i];
+    if (!rcl_action_goal_handle_is_active(goal_handle)) {
+      ++num_inactive_goals;
+
+      rcl_action_goal_info_t goal_info;
+      ret = rcl_action_goal_handle_get_info(goal_handle, &goal_info);
+      if (RCL_RET_OK != ret) {
+        return RCL_RET_ERROR;
```
wjwwood:
```suggestion
        return ret;
```

If you want, I'm honestly not sure what's appropriate or not.

rcl_action/src/rcl_action/action_server.c
```diff
@@ -395,6 +421,66 @@ rcl_action_accept_new_goal(
...
+    return RCL_RET_ERROR;
+  }
+
+  for (size_t i = 0; i < num_goal_handles; ++i) {
+    rcl_action_goal_handle_t * goal_handle = goal_handles[i];
+    if (!rcl_action_goal_handle_is_active(goal_handle)) {
+      ++num_inactive_goals;
+
+      rcl_action_goal_info_t goal_info;
+      ret = rcl_action_goal_handle_get_info(goal_handle, &goal_info);
```
wjwwood: I don't know what's in `goal_info`, but does it need to be `fini` at all?

rcl_action/src/rcl_action/action_server.c
```diff
@@ -554,20 +640,23 @@ rcl_action_expire_goals(
       continue;
     }
     goal_time = _goal_info_stamp_to_nanosec(info_ptr);
-    assert(current_time > goal_time);
     if ((current_time - goal_time) > timeout) {
       // Stop tracking goal handle
       // Fill in any gaps left in the array with pointers from the end
+      --num_goal_handles;
       action_server->impl->goal_handles[i] = action_server->impl->goal_handles[num_goal_handles];
-      action_server->impl->goal_handles[num_goal_handles--] = NULL;
```
wjwwood: Is this change desired? You previously accessed, then decremented but now you decrement and the access `num_goal_handles`. I guessing maybe, if you're removing the last element in the list of goal handles, so `num_goal_handles - 1` is both the last index and the new length, whereas before you were probably accessing out of bounds (`[len]` rather than `[len - 1]`)?

rcl_action/src/rcl_action/action_server.c
```diff
@@ -554,20 +640,23 @@ rcl_action_expire_goals(
...
 
-  // Shrink goal handle array if some goals expired
   if (num_goals_expired > 0u) {
+    // Shrink goal handle array if some goals expired
     if (0u == num_goal_handles) {
       allocator.deallocate(action_server->impl->goal_handles, allocator.state);
+      action_server->impl->goal_handles = NULL;
+      action_server->impl->num_goal_handles = num_goal_handles;
     } else {
       void * tmp_ptr = allocator.reallocate(
```
wjwwood: Unrelated to this change, but there's `rcl_reallocf()`/`rcutils_reallocf()` for this pattern: http://docs.ros2.org/bouncy/api/rcutils/allocator_8h.html#ac159eba8f5cde86f891d28c7be7693bc

rcl_action/src/rcl_action/action_server.c
```diff
@@ -590,6 +685,20 @@ rcl_action_expire_goals(
   return ret_final;
 }
 
+rcl_ret_t
+rcl_action_notify_goal_done(
+  const rcl_action_server_t * action_server,
+  rcl_action_goal_handle_t * goal_handle)
+{
+  (void)goal_handle;
```
wjwwood: Why is this unused here? Are you expecting it to be expired and remove from the `goal_handles` automatically in `_recalculate_expire_timer()`? If so would it make sense to use `goal_handle` to ensure that did happen before returning?

rcl_action/src/rcl_action/action_server.c
sloretz: I'm not sure if it's convention to hide the error code to match the documentation or return the more specific one. It's inconsistent in `rcl`, but `rcl_action` is consistent about doing the former. :man_shrugging:

rcl_action/src/rcl_action/action_server.c
sloretz: You're right this should have a fini function and it should be used. It's a typedef to a generated `action_msgs` msg which currently doesn't allocate anything on the heap, but there's no guarantee that won't change in the future.

rcl_action/src/rcl_action/action_server.c
sloretz: This one is a fix for an out of bounds error. I think the change is right, but maybe I haven't looked at this closely enough.

In the previous code when there is just one goal handle and it is being expired it would:

```
// num_goal_handles = 1, size of goal_handles = 1
...goal_handles[0] = ...goal_handles[1];  // read out of bounds
...goal_handles[1] = NULL;  // write out of bounds
```

If there were two goal handles and the first is being expired then it would
```
// num_goal_handles = 2, size of goal_handles = 2
// First iteration i = 0
...goal_handles[0] = ...goal_handles[2];  // read out of bounds
...goal_handles[2] = NULL; num_goal_handles -= 1;  // write out of bounds
```


The same two cases in the new code
```
// num_goal_handles = 1, size of goal_handles = 1
num_goal_handles -= 1;
...goal_handles[0] = ...goal_handles[0];  // noop
...goal_handles[0] = NULL;
```

```
// num_goal_handles = 2, size of goal_handles = 2
// First iteration i = 0
num_goal_handles -= 1;
...goal_handles[0] = ...goal_handles[1];
...goal_handles[1] = NULL;
```

rcl_action/src/rcl_action/action_server.c
sloretz: I don't have a good reason for passing it in. I'll remove the `goal_handle` as a parameter.

 It felt like the action server should have the goal handle in case it wanted to do something with it in the future, but the only thing that needs to happen when a goal reaches a terminal state is to check if the expire timer period needs to be set.

rcl_action/src/rcl_action/action_server.c
sloretz: There's more to think about here, and I think the solution is to make `rcl_action_goal_info_t` a struct instead of reusing the c struct from action messages, but that's a lot of places to change it so if it's alright I'd like to do this in a follow up PR.

rcl_action/src/rcl_action/action_server.c
sloretz: Removed parameter in 2aea0e3db0be5fcb002e5d61f2cc5efb747a44a1

sloretz:
CI with 2aea0e3db0be5fcb002e5d61f2cc5efb747a44a1
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5724/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2316/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4731/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5632/)

PR #348, opened 2018-12-01
add rcl_action_server_is_valid_except_context

sloretz:
This adds a function to check if the action server is valid except for the rcl context and uses it in the wait set apis. This gets rid of an exception thrown by the C++ action server when SIGINT invalidates the context but the wait set tries to check the action server entities and finds that the server has reported itself as invalid.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5728/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2317/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4734/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5633/)

### DIFF
--- a/rcl_action/src/rcl_action/action_server.c
+++ b/rcl_action/src/rcl_action/action_server.c
@@ -916,6 +916,35 @@ rcl_action_server_is_valid(const rcl_action_server_t * action_server)
   return true;
 }
 
+bool
+rcl_action_server_is_valid_except_context(const rcl_action_server_t * action_server)
+{
+  RCL_CHECK_FOR_NULL_WITH_MSG(action_server, "action server pointer is invalid", return false);
+  RCL_CHECK_FOR_NULL_WITH_MSG(
+    action_server->impl, "action server implementation is invalid", return false);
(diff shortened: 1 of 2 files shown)


rcl_action/src/rcl_action/action_server.c
```diff
@@ -916,14 +916,43 @@ rcl_action_server_is_valid(const rcl_action_server_t * action_server)
...
+  }
+  if (!rcl_publisher_is_valid_except_context(&action_server->impl->feedback_publisher)) {
+    RCL_SET_ERROR_MSG("feedback publisher is invalid");
+    return false;
+  }
+  if (!rcl_publisher_is_valid_except_context(&action_server->impl->status_publisher)) {
+    RCL_SET_ERROR_MSG("status publisher is invalid");
+    return false;
+  }
+  return true;
```
wjwwood: Could deduplicate with `rcl_action_server_is_valid()`.

sloretz: I'll do this on a follow up PR

PR #350, opened 2018-12-05
Publish logs to Rosout

nburek:
This pull request adds the rosout logging feature, which will forward logs from node loggers to the rosout topic on that node. This change is on top of other commits being reviewed in ros2/rcl#327. This PR will currently contain the changes from both that review and this review. After that review is merged in this one should just show the last commit adding the rosout feature.

### DIFF
--- a/rcl/include/rcl/logging_rosout.h
+++ b/rcl/include/rcl/logging_rosout.h
@@ -121,7 +121,6 @@ rcl_logging_rosout_init_publisher_for_node(
  *
  * \param[in] node a valid rcl_node_t that the publisher will be created on
  * \return `RCL_RET_OK` if the publisher was created successfully, or
- * \return `RCL_RET_NOT_INIT` if the publisher does not exist for the provided node, or
  * \return `RCL_RET_NODE_INVALID` if any arguments are invalid, or
  * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
  * \return `RCL_RET_ERROR` if an unspecified error occurs.
--- a/rcl/include/rcl/logging_rosout.h
+++ b/rcl/include/rcl/logging_rosout.h
@@ -1,0 +1,169 @@
+// Copyright 2017 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCL__LOGGING_ROSOUT_H_
+#define RCL__LOGGING_ROSOUT_H_
+
+#include "rcl/allocator.h"
+#include "rcl/error_handling.h"
+#include "rcl/node.h"
+#include "rcl/types.h"
+#include "rcl/visibility_control.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+/// Initializes the rcl_logging_rosout features
+/**
+ * Calling this will initialize the rcl_logging_rosout features. This function must be called
+ * before any other rcl_logging_rosout_* functions can be called.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] allocator The allocator used for metadata related to the rcl_logging_rosout features
+ * \return `RCL_RET_OK` if the rcl_logging_rosout features are successfully initialized, or
+ * \return `RCL_RET_ALREADY_INIT` if the feature has already initialized, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_LOCAL
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_logging_rosout_init(
+  const rcl_allocator_t * allocator);
+
+/// Uninitializes the rcl_logging_rosout features
+/**
+ * Calling this will set the rcl_logging_rosout features into the an unitialized state that is
+ * functionally the same as before rcl_logging_rosout_init was called.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \return `RCL_RET_OK` if the rcl_logging_rosout feature was successfully unitialized, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_LOCAL
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_logging_rosout_fini();
+
+/// Creates a rosout publisher for a node and registers to be used by the logging system
+/**
+ * Calling this for an rcl_node_t will create a new publisher on that node that will be
+ * used by the logging system to publish all log messages from that Node's logger.
+ *
+ * If a publisher already exists for this node then a new publisher will NOT be created.
+ *
+ * It is expected that after creating a rosout publisher with this function that
+ * rcl_logging_destroy_rosout_publisher_for_node() will be called for the node to cleanup
+ * the publisher while the Node is still valid.
+ *
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] node a valid rcl_node_t that the publisher will be created on
+ * \return `RCL_RET_OK` if the publisher was created successfully, or
+ * \return `RCL_RET_ALREADY_INIT` if the publisher has already exists, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_LOCAL
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_logging_rosout_init_publisher_for_node(
+  rcl_node_t * node);
+
+/// Deregisters a rosout publisher for a node and cleans up allocated resources
+/**
+ * Calling this for an rcl_node_t will destroy the rosout publisher on that node and remove it from
+ * the logging system so that no more Log messages are published to this function.
+ *
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] node a valid rcl_node_t that the publisher will be created on
+ * \return `RCL_RET_OK` if the publisher was created successfully, or
+ * \return `RCL_RET_NOT_INIT` if the publisher does not exist for the provided node, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_LOCAL
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_logging_rosout_fini_publisher_for_node(
+  rcl_node_t * node);
+
+/// The output handler outputs log messages to rosout topics.
+/**
+ * When called with a logger name and log message this function will attempt to
+ * find a rosout publisher correlated with the logger name and publish a Log
+ * message out via that publisher. If there is no publisher directly correlated
+ * with the logger then nothing will be done.
+ *
+ * This function is meant to be registered with the logging functions for rcutils
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param location The pointer to the location struct or NULL
+ * \param severity The severity level
+ * \param name The name of the logger, must be null terminated c string
+ * \param timestamp The timestamp for when the log message was made
+ * \param log_str The string to be logged
+ */
+RCL_PUBLIC
+void rcl_logging_rosout_output_handler(
+  const rcutils_log_location_t * location,
+  int severity, const char * name, rcutils_time_point_value_t timestamp,
+  const char * format, va_list * args);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif  // RCL__LOGGING_ROSOUT_H_
--- a/rcl/include/rcl/logging_rosout.h
+++ b/rcl/include/rcl/logging_rosout.h
@@ -41,7 +41,6 @@ extern "C"
  *
  * \param[in] allocator The allocator used for metadata related to the rcl_logging_rosout features
  * \return `RCL_RET_OK` if the rcl_logging_rosout features are successfully initialized, or
- * \return `RCL_RET_ALREADY_INIT` if the feature has already initialized, or
  * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
  * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
  * \return `RCL_RET_ERROR` if an unspecified error occurs.
--- a/rcl/include/rcl/logging_rosout.h
+++ b/rcl/include/rcl/logging_rosout.h
@@ -1,4 +1,4 @@
-// Copyright 2017 Open Source Robotics Foundation, Inc.
+// Copyright 2018-2019 Open Source Robotics Foundation, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -73,14 +73,14 @@ RCL_WARN_UNUSED
 rcl_ret_t
 rcl_logging_rosout_fini();
 
-/// Creates a rosout publisher for a node and registers to be used by the logging system
+/// Creates a rosout publisher for a node and registers it to be used by the logging system
 /**
  * Calling this for an rcl_node_t will create a new publisher on that node that will be
  * used by the logging system to publish all log messages from that Node's logger.
  *
  * If a publisher already exists for this node then a new publisher will NOT be created.
  *
- * It is expected that after creating a rosout publisher with this function that
+ * It is expected that after creating a rosout publisher with this function
  * rcl_logging_destroy_rosout_publisher_for_node() will be called for the node to cleanup
  * the publisher while the Node is still valid.
  *
@@ -96,7 +96,7 @@ rcl_logging_rosout_fini();
  * \param[in] node a valid rcl_node_t that the publisher will be created on
  * \return `RCL_RET_OK` if the publisher was created successfully, or
  * \return `RCL_RET_ALREADY_INIT` if the publisher has already exists, or
- * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return `RCL_RET_NODE_INVALID` if any arguments are invalid, or
  * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
  * \return `RCL_RET_ERROR` if an unspecified error occurs.
  */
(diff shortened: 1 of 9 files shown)


rcl/include/rcl/logging_rosout.h
```diff
@@ -0,0 +1,169 @@
...
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] node a valid rcl_node_t that the publisher will be created on
+ * \return `RCL_RET_OK` if the publisher was created successfully, or
+ * \return `RCL_RET_NOT_INIT` if the publisher does not exist for the provided node, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
```
jacobperron: Suggest `RCL_RET_NODE_INVALID` instead.

rcl/src/rcl/logging_rosout.c
```diff
@@ -0,0 +1,236 @@
...
+
+    va_list args_clone;
+    va_copy(args_clone, *args);
+    status = rcutils_char_array_vsprintf(&msg_array, format, args_clone);
+    va_end(args_clone);
+
+
+    rcl_interfaces__msg__Log * log_message = rcl_interfaces__msg__Log__create();
+    if (NULL != log_message) {
+      log_message->stamp.sec = (timestamp / 1000000000);
```
jacobperron:
```suggestion
      log_message->stamp.sec = RCL_NS_TO_S(timestamp);
```

rcl/src/rcl/logging_rosout.c
```diff
@@ -0,0 +1,236 @@
...
+      log_message->stamp.nanosec = (timestamp % 1000000000);
```
jacobperron:
```suggestion
      log_message->stamp.nanosec = (timestamp % RCL_S_TO_NS(1));
```

rcl/src/rcl/logging_rosout.c
```diff
@@ -0,0 +1,236 @@
...
+      .buffer = msg_buf,
+      .owns_buffer = false,
+      .buffer_length = 0u,
+      .buffer_capacity = sizeof(msg_buf),
+      .allocator = __rosout_allocator
+    };
+
+    va_list args_clone;
+    va_copy(args_clone, *args);
+    status = rcutils_char_array_vsprintf(&msg_array, format, args_clone);
```
jacobperron: Maybe we should check the return value. If there is an error we can print an error and return early.

rcl/src/rcl/logging_rosout.c
```diff
@@ -0,0 +1,236 @@
...
+    if (NULL != log_message) {
+      log_message->stamp.sec = (timestamp / 1000000000);
+      log_message->stamp.nanosec = (timestamp % 1000000000);
+      log_message->level = severity;
+      log_message->line = location->line_number;
+      rosidl_generator_c__String__assign(&log_message->name, name);
+      rosidl_generator_c__String__assign(&log_message->msg, msg_array.buffer);
+      rosidl_generator_c__String__assign(&log_message->file, location->file_name);
+      rosidl_generator_c__String__assign(&log_message->function, location->function_name);
+      status = rcl_publish(&entry.publisher, log_message);
```
jacobperron: Check the return value?

rcl/src/rcl/logging_rosout.c
```diff
@@ -0,0 +1,236 @@
...
+static bool __is_initialized = false;
+static rcl_allocator_t __rosout_allocator;
+
+rcl_ret_t rcl_logging_rosout_init(
+  const rcl_allocator_t * allocator)
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(allocator, RCL_RET_INVALID_ARGUMENT);
+  rcl_ret_t status = RCL_RET_OK;
+  if (__is_initialized) {
+    return RCL_RET_OK;
```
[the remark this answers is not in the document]
nburek: I think I had to change this in order to fix unit tests because they were causing initialize to be called multiple times. I've corrected the documentation in the header to match.

rcl/src/rcl/logging_rosout.c
nburek: Checked for error and printing, but didn't return early so that fini could be called on resources just in case.

rcl/src/rcl/logging_rosout.c
nburek: Done

rcl/src/rcl/logging_rosout.c
nburek: Done

rcl/include/rcl/logging_rosout.h
nburek: Done

rcl/src/rcl/logging_rosout.c
```diff
@@ -0,0 +1,236 @@
...
+rcl_ret_t rcl_logging_rosout_fini()
+{
+  RCL_LOGGING_ROSOUT_VERIFY_INITIALIZED
+  rcl_ret_t status = RCL_RET_OK;
+  const char * key = NULL;
+  rosout_map_entry_t entry;
+
+  // fini all the outstanding publishers
+  status = rcutils_hash_map_get_next_key_and_data(&__logger_map, NULL, &key, &entry);
+  while (RCUTILS_RET_OK == status) {
```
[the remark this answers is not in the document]
nburek: The reason I did it this way is that so it is explicit when something fails and it doesn't mask issues by combining errors. If multiple things fail then what is the return status and if there is an error message what is it set to?

rcl/src/rcl/logging_rosout.c
nburek: Done.

rcl/include/rcl/logging_rosout.h
```diff
@@ -0,0 +1,169 @@
...
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param location The pointer to the location struct or NULL
+ * \param severity The severity level
+ * \param name The name of the logger, must be null terminated c string
+ * \param timestamp The timestamp for when the log message was made
+ * \param log_str The string to be logged
```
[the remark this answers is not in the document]
nburek: Done.

rcl/include/rcl/logging_rosout.h
```diff
@@ -0,0 +1,169 @@
...
+rcl_logging_rosout_fini();
+
+/// Creates a rosout publisher for a node and registers to be used by the logging system
+/**
+ * Calling this for an rcl_node_t will create a new publisher on that node that will be
+ * used by the logging system to publish all log messages from that Node's logger.
+ *
+ * If a publisher already exists for this node then a new publisher will NOT be created.
+ *
+ * It is expected that after creating a rosout publisher with this function that
```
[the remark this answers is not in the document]
nburek: Removed extra that

rcl/include/rcl/logging_rosout.h
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * \return `RCL_RET_OK` if the rcl_logging_rosout feature was successfully unitialized, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_LOCAL
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_logging_rosout_fini();
+
+/// Creates a rosout publisher for a node and registers to be used by the logging system
```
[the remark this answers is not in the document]
nburek: Added 'it'

rcl/include/rcl/logging_rosout.h
```diff
@@ -0,0 +1,169 @@
...
+ * \return `RCL_RET_ALREADY_INIT` if the feature has already initialized, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_LOCAL
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_logging_rosout_init(
+  const rcl_allocator_t * allocator);
```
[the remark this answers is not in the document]
jacobperron: Okay, if the linters are okay with it.

rcl/src/rcl/logging_rosout.c
```diff
@@ -0,0 +1,236 @@
...
+rcl_ret_t rcl_logging_rosout_init(
+  const rcl_allocator_t * allocator)
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(allocator, RCL_RET_INVALID_ARGUMENT);
+  rcl_ret_t status = RCL_RET_OK;
+  if (__is_initialized) {
+    return RCL_RET_OK;
+  }
+  __logger_map = rcutils_get_zero_initialized_hash_map();
+  status = rcutils_hash_map_init(&__logger_map, 2, sizeof(const char *), sizeof(rosout_map_entry_t),
```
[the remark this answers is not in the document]
nburek: I've added a macro in logging_rosout to do this conversion and switched to using it everywhere I was previously comparing an RCUTILS error code to an RCL error code. Will update the PR shortly with the change.

rcl/src/rcl/logging_rosout.c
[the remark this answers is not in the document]
nburek: Let me know how strongly you feel about this one. I just fixed some unit tests that were failing due to the hashmap not being cleaned out after finalization. It doesn't feel likely to run into the situation of only one publisher failing to fini, but it's definitely way easier to troubleshoot when you don't have compounded failures, so I'm inclined to leave it if you don't feel strongly it should be changed.

nburek:
Thanks for all the feedback. I've finished making updates based on your comments. There was only one that I pushed back on, so let me know what you think after you read my reasoning above.

rcl/src/rcl/logging_rosout.c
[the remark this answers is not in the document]
jacobperron: I don't feel strongly about it. I think your justification makes sense and we can leave it.

hidmic:
Running CI, including this PR and ros2/rcutils#131:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5994/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2503/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4947/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5870/)

nuclearsandwich:
@nburek following up on @chapulina's comment https://github.com/ros2/rcutils/pull/131#issuecomment-452919632 output from the Linux CI machines includes:

```
double free or corruption (fasttop)
```
with [this test](https://ci.ros2.org/job/ci_linux/5994/testReport/junit/test_rclcpp.build.test_rclcpp/test_node_name__rmw_connext_cpp_/test_node_name/) being one example of the above.

I'm less savvy with how Mac and Windows but I believe the test failures there stem from the same problem with differing system-specific output.

nburek:
I'm digging into this now. It looks like something is not getting torn down correctly, which may be a result of the refactor to rcl that occurred while I was developing this that converted some portions of the rcl from being a single instance to being a context that gets initialized and torn down separate. While adjusting to the new lifecycle I probably missed a case in how something could be torn down. That's just my initial assumption at this point though, it still needs more investigation into what's going on.

nburek:
This issue should now be fixed. A bug in the last commit I made introduced the bug that caused it to try to cleanup the rosout topic multiple times. This should now be fixed and all the tests from "test_rclcpp" were passing locally.

nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6019/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2525/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4969/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5893/)

nburek:
It looks like the Mac and Windows builds here failed because of a Jenkins issue:
```
07:05:00 FATAL: command execution failed
07:05:00 java.nio.channels.ClosedChannelException
07:05:00 	at org.jenkinsci.remoting.protocol.impl.ChannelApplicationLayer.onReadClosed(ChannelApplicationLayer.java:209)
07:05:00 	at org.jenkinsci.remoting.protocol.ApplicationLayer.onRecvClosed(ApplicationLayer.java:222)
07:05:00 	at org.jenkinsci.remoting.protocol.ProtocolStack$Ptr.onRecvClosed(ProtocolStack.java:816)
```

From the CI build that was run on the rcutils PR that includes this PR, the only issue is with a couple of Warnings on Windows. I'll work on getting those warnings fixed.
https://github.com/ros2/rcutils/pull/131#issuecomment-453327534

nuclearsandwich:
> It looks like the Mac and Windows builds here failed because of a Jenkins issue:

Yep. Re-triggered those although we're backed up on Mac hosts so that one will take some time to show up.

[Build Status](https://ci.ros2.org/job/ci_windows/5897/)
[Build Status](https://ci.ros2.org/job/ci_osx/4973/)

nburek:
Can we try to get this PR merged into the next Crystal patch release? 
https://github.com/ros2/ros2/issues/647

nuclearsandwich:
This is queued for patch release 2.

chapulina:
It looks like this pull request broke backwards compatibility by [throwing an error](https://github.com/ros2/rcl/pull/350/files#diff-bf8aead5abb376e0ca7e38051163fb9dR156) if there are multiple nodes with the same name. This is affecting downstream packages like navigation and rosbag (see issues referenced above) and has the potential to break many more packages.

@nburek, just a heads up that we're revisiting whether this will in fact go into patch 2, or if it should be retargeted for Dashing given all the downstream consequences. If your team can think of a more gentle transition, like printing a warning on Crystal and throwing the error on Dashing, that would make it easier to release this into Crystal.

nburek:
The code could be modified to just print a warning, but it could explicitly break functionality because it would essentially cause two different node instances to share the same publisher attached to just one of the nodes. Destroying the one with the publisher would cause the other to break without additional complexity for maintaining both publishers and uniquely identifying which node they belong to so that they're torn down with the right node. 

I'm curious what the long term stance is going to be on having multiple nodes with the same name. My understanding was that ROS1 explicitly disallows this and while ROS2 didn't explicitly disallow it until now that the recommendation was not to rely on it. If it's going to be allowed then this needs to be fixed regardless and if it's not going to be allowed then it feels like we're just adding an exception for a use case that we say shouldn't be supported but unofficially is. My worry in this case is that it won't be a temporary patch and I'd almost prefer to miss patch 2 and break the interface with Dashing.

chapulina:
> I'm curious what the long term stance is going to be on having multiple nodes with the same name.

I believe the long term plan is to explicitly disallow duplicate names. 

> the recommendation was not to rely on it

Correct, and the fact that it hasn't been enforced so far was just an oversight. However, there's a big difference between recommending against it and enforcing it.

> we're just adding an exception for a use case that we say shouldn't be supported but unofficially is

There's nothing preventing downstream users from creating multiple nodes with the same name right now, and it looks like it's been working for them. So even though it is not recommended, we shouldn't break those users' code without a strong reason, like a security risk or a critical bug. A new feature is not a reason to break existing code in a patch release.

> The code could be modified to just print a warning, but it could explicitly break functionality 

It would just break the new functionality being added by this PR, right? Which is something that the preexisting packages and users wouldn't be counting on anyway. A warning would raise awareness without breaking existing code. It would be up to downstream developers whether they want to address the warnings on a future Crystal patch (as I believe navigation and rosbag are planning to do) or they could wait until Dashing to adapt to the new strict requirements.

>  I'd almost prefer to miss patch 2 and break the interface with Dashing

I've talked to the team and we agree that we shouldn't break existing functionality. So I'll leave it up to you if you want to follow the warning route, or just keep the functionality on master so it goes into Dashing.

dirk-thomas:
> I believe the long term plan is to explicitly disallow duplicate names.

That has not been decided yet.

nburek:
> A new feature is not a reason to break existing code in a patch release.

I totally agree. I don't think it can go into the patch as is since it breaks the interface by adding the explicit exclusion. 



> > I believe the long term plan is to explicitly disallow duplicate names.
> 
> That has not been decided yet.

I was hoping that this decision had already been made. While just making this exception a warning right now would mean that logging has weird behavior in the case where you violate the policy it's at least telling people that they shouldn't be doing that. It would be something in ROS2 core that is explicitly telling people not to use the same name inside the same process, even if it's not enforced yet. If it's not going to be the policy that unique names are enforced within a process then making it a warning becomes a bit of a contradiction in my opinion between what a feature in ROS core is telling people and what the design says is allowed. In that case I'd rather fix the rosout feature to work with non-unique names in a way that makes sense, though that is probably more work and means it may slip past patch 2.

chapulina:
> That has not been decided yet.

I wasn't aware of that. @nburek is right that in that case, a warning may be misleading.

dirk-thomas:
> I'd rather fix the rosout feature to work with non-unique names in a way that makes sense

That sounds like the best alternative at the moment :+1:

nburek:
@dirk-thomas Do we have any line of site on when/if a decision is going to be made on unique naming within a process? Is there any previous conversation around it I could read up on?

dirk-thomas:
> Do we have any line of site on when/if a decision is going to be made on unique naming within a process?

It is currently not planned to work on that conceptional question within this release cycle.

> Is there any previous conversation around it I could read up on?

I am not aware of a specific thread about the topic but it appear in multiple discussions where identifying nodes is a topic: e.g. http://design.ros2.org/articles/static_remapping.html

wjwwood:
> > I believe the long term plan is to explicitly disallow duplicate names.

> That has not been decided yet.

But we have many places where that's the assumption. My understanding is that unique names is the preferred route, but we're just missing the enforcement. If we were to go the other way and allow duplicate node names we'd have lots of refactoring to do.

> > I'd rather fix the rosout feature to work with non-unique names in a way that makes sense

> That sounds like the best alternative at the moment 👍

Again, I disagree. I think the simplest and safest approach is to assume node names need to be unique, but add a check if duplicates are detected and warn against them.

> > Is there any previous conversation around it I could read up on?

> I am not aware of a specific thread about the topic but it appear in multiple discussions where identifying nodes is a topic: e.g. http://design.ros2.org/articles/static_remapping.html

https://github.com/ros2/design/issues/187

dirk-thomas:
Since the description in that design ticket says explicitly:

> We need a design document that states node names need to be unique (or not) with rationale why and if unique then how to enforce it and if not unique how to deal with existing features that assume and rely on unique node names.

I would highly recommend to not go further down a path which we have not settled on. We might lean towards requiring unique node names but if we decide not to for whatever reason this would add one more thing which needs to be revisited.

As long as there is no technical reason why this specific issue can't be resolved without relying on unique node names I think implementing it in a way that it does support non-unique names is the more flexible way to go.

wjwwood:
> As long as there is no technical reason why this specific issue can't be resolved without relying on unique node names I think implementing it in a way that it does support non-unique names is the more flexible way to go.

If it is easy to make it work without assuming node names are unique then that's fine, but we shouldn't go out of our way to make this feature work with non-unique node names. Because in my experience it is often harder to assume they are not unique, e.g. setting parameter values with services which are namespaced by node name.

I'm not up to speed on this issue enough to say which is the case this time, but if it is at all difficult to implement with non-unique node names, then I stand by my assertion that detecting and warning is a better route for now, because I would wager we end up going with enforced unique node names eventually, mostly because there are cases were we need the names to be unique and I don't see a reasonable way to implement those cases without unique names.

nuclearsandwich:
@nburek from the discussion above it seems that our immediate path forward will be informed by your estimate of the work involved to allow logging to function without node-name uniqueness. If you have any initial estimates on what's involved that would help us come to a decision for how to handle the upcoming Crystal patch release.

nburek:
To just switch this over to a warning I would estimate it would take someone on my team 4-6 hours to make the change, thoroughly test it, and send a PR for it. 

To add some support for non-unique node names the simplest solution I can think of for now would be to change the map to allow for a list of publishers for a node name instead of just a single one. It would have to rely on the node->impl pointer value for uniqueness as I don't know of any other unique identifiers for a node. This would at least allow the rosout code to properly track and deconstruct publishers as nodes are created/destroyed. This would not solve the issue of actually sending the logs for different nodes with the same name over their distinct publishers. It would just have to pick one and all logs generated by node loggers with that name would go out over the first publisher in the list. That may be slightly odd behavior if you're looking at composition use case, because the code for one node could impact another based on the amount of logging it is doing. For a random developer from my team to pick this up I would estimate roughly 5 days to implement, test, and post the PR. 

Any solution to actually keep track of which node loggers should use which node publishers when there is no unique naming would require changes to the interfaces for logging and how rclcpp/rclpy loggers track information. That would almost require another design review for the changes. 

I think it would be really valuable to get the rosout feature into the crystal patch because it'll allow people to start using it and we'll actually get some good feedback on how it's working before Dashing. However, I think the only way our team would be able to get it ready in time for the second patch is to accept either a warning or a silent assumption with undefined behavior like many other parts of the code. Then we could have a discussion around what the right solution is and get it implemented for Dashing.

adam-dunc:
The way I see it we have 3 options forward, please correct me:
1. Hold this feature until Dashing, after we have a decision on node uniqueness
2. Modify the node to throw a warning and include it in this patch
3. Modify this to work with non-unique nodes and push in a future patch

 My recommendation would be #2 at this stage. We discussed Crystal not being LTS so we could get feedback on all the new features from users. I'm glad we are dealing with this discussion now and not right in front of the Dashing release or we may have ended up without rosout logging in Dashing at all. If we hold this off until the full decision is made we'll realistically get no functional feedback from users on it until Dashing which defeats some of the purpose that we discussed with Crystal not being LTS. By logging warning it would work in most cases and can be modified once a decision is made on node uniqueness.

I'm happy to have Amazon help facilitate the long term approach discussions  if it offloads some of the program management, design and decision matrix burden. If an owner isn't needed, we'd also be happy to participate.

-Adam

nuclearsandwich:
@wjwwood, @dirk-thomas, @chapulina and I just had an offline discussion. We also agree that a issuing a warning is the preferred path forward. And we're going to determine at our ROS 2 meeting next Tuesday which future release we want to have settled the node name uniqueness constraints (https://github.com/ros2/design/issues/187) for.

Two technical notes that came up for the immediate PR:
1. Please target master with the pull request as we'd like Crystal and master to behave consistently until we have an explicit design goal stating otherwise.
2. When submitting the PR please describe the side effects of non-unique node names and summarize the description in the printed warning so that those who receive the warning know what to expect.

@nburek thanks very much for the estimates they were a big help in informing our decision making. Thanks also to @Karsten1987 and @chapulina for finding and reporting this issue before it got into a release and to everyone for their input. :yellow_heart:.

nburek:
Followup PR for changing this to a warning: https://github.com/ros2/rcl/pull/384

PR #355, opened 2018-12-08
Fix high CPU usage due to rcl_wait waking quickly

### LINKED ISSUE #354: 100% Action Server CPU usage after goal is done until it is expired
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu bionic
- Installation type:
  - source
- Version or commit hash:
  - https://gist.github.com/sloretz/6e5d7324b115da574d79c7e785d8968e
- DDS implementation:
  - Fast-RTPS
  - rclcpp

#### Steps to reproduce issue

1. Start `top` in a terminal
1. Start an action server using `rmw_fastrtps_cpp`
   ```
   RMW_IMPLEMENTATION=rmw_fastrtps_cpp ./build/test_communication/test_action_server_cpp Fibonacci /foo/bar
   ```
1. Start an action client using `rmw_fastrtps_cpp`
   ```
   RMW_IMPLEMENTATION=rmw_fastrtps_cpp ./build/test_communication/test_action_client_cpp Fibonacci /foo/bar
   ```

#### Expected behavior

I would expect negligible CPU usage after the goal has finished processing

#### Actual behavior

Server using 100% CPU after the goal has finished processing

#### Additional information
It does not happen using RMW connext.

If I debug the server, it seems that there is a loop inside `rclcpp::spin()` where `rcl_wait()` wakes indicating a timeout, the executor checks and nothing is ready so it goes back into `rcl_wait()` which wakes immediately with a timeout. The timeout given to `rcl_wait()` is `-1` which becomes `INT_MAX` when passed to `rmw_wait()`. The 100% CPU usage ends when the action server expires the old goal (after 15 minutes by default). 

Maybe there's an issue with how `rcl_action_server` is using the timer api (ros2/rcl#343)?

jacobperron:
Resolves #354 

Passing a max integer value to `rmw_wait()` was leading to integer overflow in the rmw layer causing a quick timeout. Instead, pass `NULL` to block in the case "block until ready".

Also fixed another small bug in `rcl_action` found along the way.

Big thanks to @sloretz for helping find the issue!

### DIFF
--- a/rcl/src/rcl/wait.c
+++ b/rcl/src/rcl/wait.c
@@ -484,13 +484,10 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
 
   bool is_timer_timeout = false;
   int64_t min_timeout = timeout > 0 ? timeout : INT64_MAX;
-  // calculate the number of valid (non-NULL and non-canceled) timers
-  size_t number_of_valid_timers = wait_set->size_of_timers;
   {  // scope to prevent i from colliding below
     uint64_t i = 0;
     for (i = 0; i < wait_set->impl->timer_index; ++i) {
       if (!wait_set->timers[i]) {
-        number_of_valid_timers--;
         continue;  // Skip NULL timers.
       }
       bool is_canceled = false;
@@ -499,7 +496,6 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
         return ret;  // The rcl error state should already be set.
       }
       if (is_canceled) {
-        number_of_valid_timers--;
         wait_set->timers[i] = NULL;
         continue;
       }
@@ -530,7 +526,7 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
     temporary_timeout_storage.sec = 0;
     temporary_timeout_storage.nsec = 0;
     timeout_argument = &temporary_timeout_storage;
-  } else if (timeout > 0 || number_of_valid_timers > 0) {
+  } else if (timeout > 0 || is_timer_timeout) {
     // If min_timeout was negative, we need to wake up immediately.
     if (min_timeout < 0) {
       min_timeout = 0;
--- a/rcl_action/src/rcl_action/action_server.c
+++ b/rcl_action/src/rcl_action/action_server.c
@@ -451,7 +451,7 @@ _recalculate_expire_timer(
         return RCL_RET_ERROR;
       }
 
-      int64_t delta = timeout - current_time - _goal_info_stamp_to_nanosec(&goal_info);
+      int64_t delta = timeout - (current_time - _goal_info_stamp_to_nanosec(&goal_info));
       if (delta < minimum_period) {
         minimum_period = delta;
       }


rcl_action/src/rcl_action/action_server.c
```diff
@@ -228,6 +228,7 @@ rcl_action_server_init(
 rcl_ret_t
 rcl_action_server_fini(rcl_action_server_t * action_server, rcl_node_t * node)
 {
+
```
mjcarroll: beep boop extra newline :robot:

jacobperron: hehe, I saw that and tried to hide it before anyone would notice

nuclearsandwich:
Changes had been made since the last CI run so I optimistically started fresh builds

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5842/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2406/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4835/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5738/)

rcl/src/rcl/wait.c
```diff
@@ -525,7 +525,11 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
     }
   }
 
-  if (timeout == 0) {
+  if (INT64_MAX == min_timeout) {
+    // Then timeout < 0 and no valid timers have duration less than INT64_MAX
+    // To avoid integer overflow in rmw layer, we pass NULL (block until something is ready)
```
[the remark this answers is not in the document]
jacobperron: You're right, I'll refactor the logic so it accounts for the case where `timeout == 0`.

jacobperron:
@wjwwood @mjcarroll Please take a look again. I realized the problem was that `number_of_valid_timers` was greater than zero even though `min_timeout` was not set in this block:

https://github.com/ros2/rcl/blob/7f31f7e200afbfe39c2d087c6b6c5fd655563111/rcl/src/rcl/wait.c#L504-L508

I've replaced it's usage with a check if `min_timeout` was set. Of course, if someone sets a timer to be `INT64_MAX` then we'll still have the same issue. What do you think about changing the default timeout value to something smaller (like `INT64_MAX / 2`) ?

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5844/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2408/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4840/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5740/)

wjwwood:
> Of course, if someone sets a timer to be INT64_MAX then we'll still have the same issue. What do you think about changing the default timeout value to something smaller (like INT64_MAX / 2) ?

Could this also be solved with an extra boolean that is set by checking the timers before (or during) calculation of the `min_timeout`? If so, then I wouldn't change the default timer (implication is that `INT64_MAX` is no longer a valid timer period) just for that. If not (I honestly can't say off-hand), then changing the timer default is ok, but I'd do `INT64_MAX - 1` and then document that `INT64_MAX` is reserved and should not be used with the timer period (and then put code in the timer init to check that it is not).

jacobperron:
> Could this also be solved with an extra boolean that is set by checking the timers before (or during) calculation of the min_timeout?

@wjwwood Done.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5849/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2413/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4845/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5747/)

jacobperron:
Rebuild Windows: [Build Status](https://ci.ros2.org/job/ci_windows/5755/)

PR #373, opened 2019-01-10
pass context to wait set, and fini rmw context

wjwwood:
Connects to ros2/rmw#163

### DIFF
--- a/rcl/src/rcl/context.c
+++ b/rcl/src/rcl/context.c
@@ -127,7 +127,7 @@ __cleanup_context(rcl_context_t * context)
     }
 
     // clean up rmw_context
-    if (NULL != context->impl->rmw_context.impl) {
+    if (NULL != context->impl->rmw_context.implementation_identifier) {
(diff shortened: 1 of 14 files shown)


rcl/src/rcl/context.c
```diff
@@ -126,6 +126,19 @@ __cleanup_context(rcl_context_t * context)
       }
     }
 
+    // clean up rmw_context
+    if (NULL != context->impl->rmw_context.impl) {
```
sloretz: IIUC this check means `rmw_context_fini()` won't be called for any current rmw implementation because they intentionally set `impl` to `NULL`. Maybe this could be a problem if an rmw_implementation uses global state instead of the `impl` pointer, but also expected cleanup code in `rmw_context_fini` to run?

wjwwood: This is a valid point, I'll have to rethink how I avoid calling fini when the case is that init failed rather than a normal init-fini cycle.

sloretz: Makes sense to me

PR #383, opened 2019-02-07
Proposal of changes for RMW_Preallocate. 

gonzodepedro:
The idea is to have a bool value in init_options to enable the `rmw_preallocate` feature.

Connects to  ros2/rmw#159

### DIFF
--- a/rcl/src/rcl/publisher.c
+++ b/rcl/src/rcl/publisher.c
@@ -194,9 +194,9 @@ rcl_publisher_init(
   // context
   publisher->impl->context = node->context;
 
-goto cleanup;
+  goto cleanup;
 fail:
- if (publisher->impl) {
+  if (publisher->impl) {
     allocator->deallocate(publisher->impl, allocator->state);
   }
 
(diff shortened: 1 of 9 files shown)


rcl/src/rcl/publisher.c
```diff
@@ -37,6 +37,7 @@ typedef struct rcl_publisher_impl_t
   rcl_publisher_options_t options;
   rcl_context_t * context;
   rmw_publisher_t * rmw_handle;
+  rmw_publisher_allocation_t * allocation;
```
wjwwood: You cannot just store a publish preallocation in the publisher because it is not thread safe. The preallocation needs to be created by the user or by rclcpp in order to ensure it is used in a thread safe manner.

gonzodepedro: I agree. In that case the RCL should be changed or augmented to allow the user to pass the allocation to RCL.

gonzodepedro:
Currently the test_memory_prealloc tests are built with the LD_BIND_NOW=1 option. This is because without that option the tests fail. Apparently the dynamic linker is allocating memory on run-time and that makes the memory tests fail.

wjwwood:
> Apparently the dynamic linker is allocating memory on run-time and that makes the memory tests fail.

At what step? publish? The library should already be loaded at that point. If it is TLS, we have functions to address specific cases of that, e.g.: https://github.com/ros2/rcutils/blob/fe82622f3c9d76400ad6065229777aa013fb8628/include/rcutils/error_handling.h#L101-L141

jacobperron:
I think this is connected to the wrong card. The description should be updated  to connect to ros2/rmw#159, right?

gonzodepedro:
> I think this is connected to the wrong card. The description should be updated to connect to [ros2/rmw#159](https://github.com/ros2/rmw/issues/159), right?

You are right. It is now connected to the right card. Thanks

Issue #375, opened 2019-01-23
Exception being thrown: failed to initialize rcl node: Logger already initialized for node


mkhansen-intel:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - source
- Version or commit hash:
  - c6788e4e54843d769cd4515c6f5e4ec6bb1a1cd0
master latest after "Publish logs to Rosout (#350)"

- DDS implementation:
  - Default - FastRTPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

Install navigation2
```
ros2 launch nav2_bringup nav2_bringup_1st_launch.py map:=<path/to/map>
```

#### Expected behavior
Should launch fine without issue.

#### Actual behavior
Exception occurs, and crashes amcl node:
```
ros2 launch nav2_bringup nav2_bringup_1st_launch.py map:=/home/mkhansen/ros2_dev/navigation2_ws/src/navigation2/nav2_system_tests/maps/map_circular.yaml
[INFO] [launch]: process[amcl-1]: started with pid [14806]
[INFO] [amcl]: Initializing AMCL
[INFO] [amcl]: Sensor model type is: "likelihood_field"
[INFO] [amcl]: Robot motion model is differential
terminate called after throwing an instance of 'rclcpp::exceptions::RCLError'
  what():  failed to initialize rcl node: Logger already initialized for node., at /home/mkhansen/ros2_dev/ros2_ws/src/ros2/rcl/rcl/src/rcl/logging_rosout.c:156
[ERROR] [launch]: process[amcl-1] process has died [pid 14806, exit code -6, cmd '/home/mkhansen/ros2_dev/navigation2_ws/install/nav2_amcl/lib/nav2_amcl/amcl __node:=amcl'].
```

#### Additional information
SO, the launch file by default has this for the amcl portion:
```
        launch_ros.actions.Node(
            package='nav2_amcl',
            node_executable='amcl',
            node_name='amcl',
            output='screen'),
            parameters=[{ 'use_sim_time': use_sim_time}]),
```
When I comment out these 2 lines, setting the node_name and parameters, it works fine, but use_sim_time isn't set (which is needed).
```
        launch_ros.actions.Node(
            package='nav2_amcl',
            node_executable='amcl',
#            node_name='amcl',
            output='screen'),
#            parameters=[{ 'use_sim_time': use_sim_time}]),

```
[...]

PR #384, opened 2019-02-07
Add change error to warn for multiple loggers

ross-desmond:
**Reproduced issue with the following:**
```
#include "rclcpp/rclcpp.hpp"
  
using namespace std;

int main(int argc, char ** argv)
{
  rclcpp::init(argc, argv);
  auto node1 = rclcpp::Node::make_shared("foo");
  {
    auto node2 = rclcpp::Node::make_shared("bar");
  }
  RCLCPP_INFO(node1->get_logger(), "Bar test complete");
  // rclcpp::spin(node);
  rclcpp::shutdown();

  return 0;
}
```
**Ran with:**
`ros2 run examples_rclcpp_minimal_subscriber multinode __node:=abc`

**Result:**
Before
```
terminate called after throwing an instance of 'rclcpp::exceptions::RCLError'
  what():  failed to initialize rcl node: Logger already initialized for node., at /root/ros2_ws/src/ros2/rcl/rcl/src/rcl/logging_rosout.c:156
```
After
```
[WARN] [log_infrastructure]: Logger already initialized for node
[INFO] [abc]: Bar test complete
```

### DIFF
--- a/rcl/src/rcl/logging_rosout.c
+++ b/rcl/src/rcl/logging_rosout.c
@@ -154,7 +154,8 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(
     return RCL_RET_ERROR;
   }
   if (rcutils_hash_map_key_exists(&__logger_map, &logger_name)) {
-    // @todo: nburek enfornce node name unique
+    // @todo: nburek Update behavior to either enforce unique names or work with non-unique 
+    // names based on the outcome here: https://github.com/ros2/design/issues/187
     RCUTILS_LOG_WARN_NAMED("log_infrastructure", "Publisher already registered for provided node name. If this is due to multiple nodes with the same name then all logs for that logger name will go out over the existing publisher. As soon as any node with that name is destructed it will unregister the publisher, preventing any further logs for that name from going out over the rosout topic.");
     // RCL_SET_ERROR_MSG("Logger already initialized for node.");
     // return RCL_RET_ALREADY_INIT;
--- a/rcl/src/rcl/logging_rosout.c
+++ b/rcl/src/rcl/logging_rosout.c
@@ -157,7 +157,11 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(
     // @TODO(nburek) Update behavior to either enforce unique names or work with non-unique
     // names based on the outcome here: https://github.com/ros2/design/issues/187
     RCUTILS_LOG_WARN_NAMED("rcl.logging_rosout",
-      "Publisher already registered for provided node name. If this is due to multiple nodes with the same name then all logs for that logger name will go out over the existing publisher. As soon as any node with that name is destructed it will unregister the publisher, preventing any further logs for that name from being published on the rosout topic.");
+      "Publisher already registered for provided node name. If this is due to multiple nodes "
+      "with the same name then all logs for that logger name will go out over the existing "
+      "publisher. As soon as any node with that name is destructed it will unregister the "
+      "publisher, preventing any further logs for that name from being published on the rosout "
+      "topic.");
     // RCL_SET_ERROR_MSG("Logger already initialized for node.");
     // return RCL_RET_ALREADY_INIT;
     return RCL_RET_OK;
--- a/rcl/src/rcl/logging_rosout.c
+++ b/rcl/src/rcl/logging_rosout.c
@@ -162,8 +162,6 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(
       "publisher. As soon as any node with that name is destructed it will unregister the "
       "publisher, preventing any further logs for that name from being published on the rosout "
       "topic.");
-    // RCL_SET_ERROR_MSG("Logger already initialized for node.");
-    // return RCL_RET_ALREADY_INIT;
     return RCL_RET_OK;
   }
 
--- a/rcl/src/rcl/logging_rosout.c
+++ b/rcl/src/rcl/logging_rosout.c
@@ -155,7 +155,7 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(
   }
   if (rcutils_hash_map_key_exists(&__logger_map, &logger_name)) {
     // @todo: nburek enfornce node name unique
-    RCUTILS_LOG_WARN_NAMED("log_infrastructure", "Logger already initialized for node");
+    RCUTILS_LOG_WARN_NAMED("log_infrastructure", "Publisher already registered for provided node name. If this is due to multiple nodes with the same name then all logs for that logger name will go out over the existing publisher. As soon as any node with that name is destructed it will unregister the publisher, preventing any further logs for that name from going out over the rosout topic.");
     // RCL_SET_ERROR_MSG("Logger already initialized for node.");
     // return RCL_RET_ALREADY_INIT;
   }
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6140/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2612/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5066/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6007/)

rcl/src/rcl/logging_rosout.c
```diff
@@ -153,8 +154,10 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(
     return RCL_RET_ERROR;
   }
   if (rcutils_hash_map_key_exists(&__logger_map, &logger_name)) {
-    RCL_SET_ERROR_MSG("Logger already initialized for node.");
-    return RCL_RET_ALREADY_INIT;
+    // @todo: nburek enfornce node name unique
+    RCUTILS_LOG_WARN_NAMED("log_infrastructure", "Logger already initialized for node");
```
nburek:
```suggestion
    RCUTILS_LOG_WARN_NAMED("log_infrastructure", "Publisher already registered for provided node name. If this is due to multiple nodes with the same name then all logs for that logger name will go out over the existing publisher and as soon any node with that name is destructed it will unregister the publisher, preventing any further logs for that node name from going out over the rosout topic.");
```

nburek:
Thank you Ross for making and testing this change. For more context, this PR is fixing Issue https://github.com/ros2/rcl/issues/375, which is discussed in more detail in this PR https://github.com/ros2/rcl/pull/350. 

Changing this from an exception to a warning will have the following effects: 
1) A node being created with an already existing name will no longer fail with an error message. 
2) All log messages coming from Node Loggers of the same name will go out over a single publisher that was registered to the first node that was created. 
3) The first node that is torn down with the name that matches the logger will cause the publisher to become unregistered. At that point any log messages from other Node Loggers of the same name will not be published on the rosout topic.

rcl/src/rcl/logging_rosout.c
```diff
@@ -153,8 +154,10 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(
     return RCL_RET_ERROR;
   }
   if (rcutils_hash_map_key_exists(&__logger_map, &logger_name)) {
-    RCL_SET_ERROR_MSG("Logger already initialized for node.");
-    return RCL_RET_ALREADY_INIT;
+    // @todo: nburek enfornce node name unique
```
nuclearsandwich: Aside from the typo, which I was instinctively fixing when I stopped to think, is this TODO for enforcing node-name uniqueness categorically or following up on the node-name uniqueness design question (https://github.com/ros2/design/issues/187)

I think the TODO would be clearer if it included a link to that issue, if indeed that's what it is intended to regard.

dirk-thomas: Beside the text of the todo the annotation used in the rest of code base is: `TODO(name) ...` instead of `@todo: name ....`

rcl/src/rcl/logging_rosout.c
```diff
@@ -153,8 +154,10 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(
...
+    RCUTILS_LOG_WARN_NAMED("log_infrastructure", "Publisher already registered for provided node name. If this is due to multiple nodes with the same name then all logs for that logger name will go out over the existing publisher. As soon as any node with that name is destructed it will unregister the publisher, preventing any further logs for that name from going out over the rosout topic.");
```
dirk-thomas: Since `log_infrastructure` is not a term anywhere atm I would suggest to change it to: `rcl.logging_rosout`

And for the message instead of

> ... from going out over the rosout topic.

I would suggest

> ... from being published on the rosout topic.

rcl/src/rcl/logging_rosout.c
```diff
@@ -153,8 +154,11 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(
     return RCL_RET_ERROR;
   }
   if (rcutils_hash_map_key_exists(&__logger_map, &logger_name)) {
-    RCL_SET_ERROR_MSG("Logger already initialized for node.");
-    return RCL_RET_ALREADY_INIT;
+    // @todo: nburek Update behavior to either enforce unique names or work with non-unique 
+    // names based on the outcome here: https://github.com/ros2/design/issues/187
+    RCUTILS_LOG_WARN_NAMED("log_infrastructure", "Publisher already registered for provided node name. If this is due to multiple nodes with the same name then all logs for that logger name will go out over the existing publisher. As soon as any node with that name is destructed it will unregister the publisher, preventing any further logs for that name from going out over the rosout topic.");
+    // RCL_SET_ERROR_MSG("Logger already initialized for node.");
+    // return RCL_RET_ALREADY_INIT;
```
jacobperron: I think we want to return here to avoid a possible memory leak.
Otherwise, we replace an reference to an existing `rcl_publisher_t` without finalizing it.

```suggestion
    return RCL_RET_OK;
```

jacobperron: I think this will also have the desired behavior that the first node created for a given name will be the one associated with the publisher.

nburek:
- Updated the TODO message with the desired format and referenced the github issue for unique name designs. 
- Ran uncrustify to fix the line length issue. 
- Added the early return back in

nuclearsandwich:
@ross-desmond, @nburek I took the liberty of making some of the discussed fixups directly as well as added one of my own in db9954b. With the approvals from Jacob and William I'll go ahead and merge this once we have green CI.

As you might have noticed we've introduced a [Developer Certificate of Origin](https://developercertificate.org/) check on the ROS 2 repositories. I am going to override that for today but it will be good to bear in mind that this check will be required in the future. More info about it can be found in [our developer guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#pull-requests).

rcl/src/rcl/logging_rosout.c
```diff
@@ -153,8 +154,13 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(
     return RCL_RET_ERROR;
   }
   if (rcutils_hash_map_key_exists(&__logger_map, &logger_name)) {
-    RCL_SET_ERROR_MSG("Logger already initialized for node.");
-    return RCL_RET_ALREADY_INIT;
+    // @TODO(nburek) Update behavior to either enforce unique names or work with non-unique
+    // names based on the outcome here: https://github.com/ros2/design/issues/187
+    RCUTILS_LOG_WARN_NAMED("rcl.logging_rosout",
+      "Publisher already registered for provided node name. If this is due to multiple nodes with the same name then all logs for that logger name will go out over the existing publisher. As soon as any node with that name is destructed it will unregister the publisher, preventing any further logs for that name from being published on the rosout topic.");
```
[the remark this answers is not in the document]
wjwwood: I think you're thinking of something like Python, but in C++ you don't need backslashes unless it's a macro definition. This is valid C++ code:

```c++
#include <cstdio>

int main()
{
  const char * foo =
    "This "
    "is "
    "a "
    "sentance.";
  printf("%s\n", foo);
  printf(
    "This "
    "is "
    "a "
    "sentance.\n");
  return 0;
}
```

PR #387, opened 2019-02-08
Drops legacy launch API usage.

hidmic:
Connected to ros2/launch#159.

### DIFF
--- a/rcl/package.xml
+++ b/rcl/package.xml
@@ -34,7 +34,7 @@
   <test_depend>ament_lint_common</test_depend>
   <test_depend>rmw</test_depend>
   <test_depend>rmw_implementation_cmake</test_depend>
-  <test_depend>launch</test_depend>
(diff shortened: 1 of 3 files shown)


rcl/package.xml
```diff
@@ -35,6 +35,7 @@
   <test_depend>rmw</test_depend>
   <test_depend>rmw_implementation_cmake</test_depend>
   <test_depend>launch</test_depend>
+  <test_depend>launch_testing</test_depend>
```
dirk-thomas: Just a note: this adds another case of a circular repository dependency (same as for the test dependency on launch).

wjwwood: This doesn't make that case any worse, but it is just an actual circular dependency. The testing code will likely need to be split into ROS and non-ROS parts like launch itself is, otherwise anything at or below rclcpp/rclpy will not be able to use any of it.

PR #396, opened 2019-03-08
Add Deadline, Liveliness, and Lifespan QoS Event (rcl_event_t type) and their rcl_take_* Function

mm318:
Implementation for Deadline, Liveliness, and Lifespan QoS. The design is at ros2/design#212

This change introduces a new `rcl_event_t` type as well as its respective `rcl_take_event()` function.


mm318:
Hi @wjwwood, can you take a look at this PR?

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6414/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2797/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5272/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6246/)

rcl/src/rcl/types_impl.h
```diff
@@ -0,0 +1,60 @@
...
+typedef struct rcl_service_impl_t
+{
+  rcl_service_options_t options;
+  rmw_service_t * rmw_handle;
+} rcl_service_impl_t;
+
+typedef struct rcl_event_impl_t
+{
+  rmw_event_t * rmw_handle;
+} rcl_event_impl_t;
```
wjwwood: Please don't put all of these in the same file, keep them separated. If you must have a `X_impl.h` do one for each type as needed.

rcl/test/rcl/test_timer.cpp
```diff
@@ -459,7 +459,8 @@ TEST_F(TestTimerFixture, test_ros_time_wakes_wait) {
 
   std::thread wait_thr([&](void) {
       rcl_wait_set_t wait_set = rcl_get_zero_initialized_wait_set();
-      ret = rcl_wait_set_init(&wait_set, 0, 0, 1, 0, 0, context_ptr, rcl_get_default_allocator());
+      ret =
+      rcl_wait_set_init(&wait_set, 0, 0, 1, 0, 0, 0, context_ptr, rcl_get_default_allocator());
```
wjwwood: nitpick: you have to indent here, if that makes it too long, then you need to break the line in another way

rcl/include/rcl/event.h
```diff
@@ -0,0 +1,149 @@
...
+  RCL_PUBLISHER_DEADLINE,
+  RCL_PUBLISHER_LIVELINESS
+} rcl_publisher_event_type_t;
+
+
+typedef enum rcl_subscription_event_type_t
+{
+  RCL_SUBSCRIPTION_DEADLINE,
+  RCL_SUBSCRIPTION_LIVELINESS
+} rcl_subscription_event_type_t;
```
wjwwood: Why are these defined here, but not stored in the event type directly? It seems to be in the impl or the rmw version of the event? If in the impl, then why not in the non-impl part of the struct, and if in rmw, why redefine the enum's here?

mm318:
Sorry that this PR has gone kind of stale. We will update as soon as possible.

The reason for separate init function for each "kind" of event is that the event needs to access the `void * data` of its respective `rmw_publisher_t`/`rmw_subscription_t`/etc.

There are no separate take functions for each "kind" of event.

rcl/include/rcl/event.h
mm318: `rclcpp` needs to pass in these types during `rcl_*_event_init()`. My understanding is that if these enums are strictly in the impl, then they will be unavailable to `rclcpp`.

wjwwood:
> There are no separate take functions for each "kind" of event.

You're right, sorry, I must have misread it.

rcl/include/rcl/event.h
wjwwood: Ok I understand they are passed as arguments to the init functions, but why is the storage for them in the impl though? Why not as members of `rcl_event_t`. If it must stay in the `impl`, then why not have accessors functions for the event types?

If I have an event, it would be nice to know what kind by introspection. This migh imply that the event type enums need to be merged, into just "event types" and not subscription/publisher event types.

rcl/test/rcl/test_timer.cpp
mm318: If I remember correctly, I introduced the line break with an indent, but then uncrustify removed the indent.

I will break the line in a different way.

rcl/include/rcl/event.h
mm318: These types (technically a translation of it to a `rmw_event_type_t` enum) are stored in `rmw_event_t`.

The `rclcpp` layer is aware of the event type because it is the layer setting it. The only other layer that actually needs to know the event type of a `rcl_event_t`/`rmw_event_t` is `rmw_fastrtps`/`rmw_connext`/`rmw_opensplice`. The event type just passes through the `rcl` and `rmw` layers.

You are right that the event type enums should be merged at some point. A merge does occur when translating `rcl_subscription_event_type_t` or `rcl_publisher_event_type_t` to `rmw_event_type_t`. We just thought that having separate `rcl_subscription_event_type_t`/`rcl_publisher_event_type_t` types is better for type-safety when calling the `rcl_*_event_init()` functions.

rcl/src/rcl/types_impl.h
mm318: Will do.

rcl/include/rcl/event.h
wjwwood: Ok, that makes more sense to me now. I'll defer to you guys on what to change if anything. You don't have to merge them at my request, I didn't consider everything until you put it into context there.

rcl/include/rcl/event.h
```diff
@@ -0,0 +1,133 @@
+// Copyright 2015 Open Source Robotics Foundation, Inc.
```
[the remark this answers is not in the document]
mm318: Will do.

mm318:
This PR is being replaced by https://github.com/ros2/rcl/pull/408. Closing this PR.

PR #403, opened 2019-03-18
Add function rcl_names_and_types_init

jacobperron:
This wraps the rmw function just like the finalize function.

---

I plan to use this function in the action graph API ([for example](https://github.com/ros2/rcl/blob/ba83518bf381139938d7341a1120890cf2594a94/rcl_action/src/rcl_action/graph.c#L88)).

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6447/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2819/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5294/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6268/)

### DIFF
--- a/rcl/include/rcl/graph.h
+++ b/rcl/include/rcl/graph.h
@@ -296,7 +296,7 @@ rcl_get_service_names_and_types(
 /**
  * This function initializes the string array for the names and allocates space
  * for all the string arrays for the types according to the given size, but
- * it does not initialize the string array for each setup of types.
+ * it does not initialize the string array for each set of types.
  * However, the string arrays for each set of types is zero initialized.
  *
  * <hr>
@@ -311,7 +311,7 @@ rcl_get_service_names_and_types(
  * \param[in] size the number of names and sets of types to be stored
  * \param[in] allocator to be used to allocate and deallocate memory
  * \returns `RCL_RET_OK` on success, or
- * \returns `RCL_RET_INVALID_ARGUMENT` if `names_and_types` is `NULL`, or
+ * \returns `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
  * \returns `RCL_BAD_ALLOC` if memory allocation fails, or
  * \returns `RCL_RET_ERROR` when an unspecified error occurs.
  */
--- a/rcl/include/rcl/graph.h
+++ b/rcl/include/rcl/graph.h
@@ -292,6 +292,37 @@ rcl_get_service_names_and_types(
   rcl_allocator_t * allocator,
   rcl_names_and_types_t * service_names_and_types);
 
+/// Initialize a rcl_names_and_types_t object.
+/**
+ * This function initializes the string array for the names and allocates space
+ * for all the string arrays for the types according to the given size, but
+ * it does not initialize the string array for each setup of types.
+ * However, the string arrays for each set of types is zero initialized.
+ *
+ * <hr>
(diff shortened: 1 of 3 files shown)


jacobperron:
Actions could use the `rmw` equivalent, `rmw_names_and_types_init`, like `rcl_get_topic_names_and_types` and `rcl_get_service_names_and_types` do, but I thought it would be nicer to encapsulate the `rmw` call.

It's almost possible to remove the `rmw` dependency from `rcl_action`. Currently, it needs the QoS profiles defined in `rmw`, but I thought it would be nice to break that in the future (e.g. with `rcl` aliases).

wjwwood:
That's fine, but could the others use the new function or no? If yes, then it might be nice to deduplicate the code. Maybe it doesn't matter though, I was just curious.

rcl/include/rcl/graph.h
```diff
@@ -292,6 +292,37 @@ rcl_get_service_names_and_types(
   rcl_allocator_t * allocator,
   rcl_names_and_types_t * service_names_and_types);
 
+/// Initialize a rcl_names_and_types_t object.
+/**
+ * This function initializes the string array for the names and allocates space
+ * for all the string arrays for the types according to the given size, but
+ * it does not initialize the string array for each setup of types.
```
wjwwood:
```suggestion
 * it does not initialize the string array for each set of types.
```

rcl/include/rcl/graph.h
```diff
@@ -292,6 +292,37 @@ rcl_get_service_names_and_types(
...
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[inout] names_and_types object to be initialized
+ * \param[in] size the number of names and sets of types to be stored
+ * \param[in] allocator to be used to allocate and deallocate memory
+ * \returns `RCL_RET_OK` on success, or
+ * \returns `RCL_RET_INVALID_ARGUMENT` if `names_and_types` is `NULL`, or
```
wjwwood: Could also be if `allocator` is null too right? I'd just leave it as "if any arguments are invalid" personally.

jacobperron:
> That's fine, but could the others use the new function or no? If yes, then it might be nice to deduplicate the code. Maybe it doesn't matter though, I was just curious.

Actually, I was mistaken. The `rcl` functions do not make any calls to `rmw_names_and_types_init`. They let the rmw layer handle initialization when calling `rmw_get_*_names_and_types`. So there is nothing to deduplicate.

The reason I want to use this function for actions is because there is no concept of actions in `rmw`. My strategy is to parse service names and count the number of actions, then initialize a `rcl_names_and_types_t` to return:

https://github.com/ros2/rcl/blob/ba83518bf381139938d7341a1120890cf2594a94/rcl_action/src/rcl_action/graph.c#L73-L91

rcl/include/rcl/graph.h
jacobperron: Yes, updated: 1fe86a4

rcl/include/rcl/graph.h
jacobperron: Fixed: 1fe86a4

jacobperron:
Test failures are unrelated.

PR #407, opened 2019-03-26
Update graph test for change to rmw names and types struct

jacobperron:
Connects to ros2/ros2#677


rcl_action/test/rcl_action/test_graph.cpp
```diff
@@ -35,6 +36,9 @@
 # define CLASSNAME(NAME, SUFFIX) NAME
 #endif
 
+bool is_opensplice =
```
sloretz: nit, `const bool`

jacobperron: 7492bd8

PR #408, opened 2019-03-29
New interfaces and their implementations for QoS features

mm318:
Implementation for Deadline, Liveliness, and Lifespan QoS. The design is at ros2/design#212

This change introduces a new `rcl_event_t` type as well as its respective `rcl_take_event()` function.

Connects to ros2/rmw#171

### DIFF
--- a/rcl/include/rcl/event.h
+++ b/rcl/include/rcl/event.h
@@ -1,0 +1,171 @@
+// Copyright 2019 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCL__EVENT_H_
+#define RCL__EVENT_H_
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#include "rcl/client.h"
+#include "rcl/macros.h"
+#include "rcl/publisher.h"
+#include "rcl/service.h"
+#include "rcl/subscription.h"
+#include "rcl/visibility_control.h"
+
+
+typedef enum rcl_publisher_event_type_t
+{
+  RCL_PUBLISHER_OFFERED_DEADLINE_MISSED,
+  RCL_PUBLISHER_LIVELINESS_LOST
+} rcl_publisher_event_type_t;
+
+typedef enum rcl_subscription_event_type_t
+{
+  RCL_SUBSCRIPTION_REQUESTED_DEADLINE_MISSED,
+  RCL_SUBSCRIPTION_LIVELINESS_CHANGED
+} rcl_subscription_event_type_t;
+
+/// rmw struct.
+typedef struct rmw_event_t rmw_event_t;
+
+/// Internal rcl implementation struct.
+struct rcl_event_impl_t;
+
+/// Structure which encapsulates a ROS QoS event handle.
+typedef struct rcl_event_t
+{
+  struct rcl_event_impl_t * impl;
+} rcl_event_t;
+
+
+/// Return a rcl_event_t struct with members set to `NULL`.
+/**
+ * Should be called to get a null rcl_event_t before passing to
+ * rcl_event_init().
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_event_t
+rcl_get_zero_initialized_event(void);
+
+/// Initialize an rcl_event_t with a publisher.
+/**
+ * Fill the rcl_event_t with the publisher and desired event_type.
+ *
+ * \param[in,out] event pointer to fill
+ * \param[in] publisher to get events from
+ * \param[in] event_type to listen for
+ * \return `RCL_RET_OK` if the rcl_event_t is filled, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_publisher_event_init(
+  rcl_event_t * event,
+  const rcl_publisher_t * publisher,
+  const rcl_publisher_event_type_t event_type);
+
+/// Initialize an rcl_event_t with a subscription.
+/**
+ * Fill the rcl_event_t with the subscription and desired event_type.
+ *
+ * \param[in,out] event pointer to fill
+ * \param[in] subscription to get events from
+ * \param[in] event_type to listen for
+ * \return `RCL_RET_OK` if the rcl_event_t is filled, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_subscription_event_init(
+  rcl_event_t * event,
+  const rcl_subscription_t * subscription,
+  const rcl_subscription_event_type_t event_type);
+
+// Take event using the event handle.
+/**
+ * Take an event from the event handle.
+ *
+ * \param[in] event_handle event object to take from
+ * \param[in, out] event_info event info object to write taken data into
+ * \param[in, out] taken boolean flag indicating if an event was taken or not
+ * \return `RCL_RET_OK` if successful, or
+ * \return `RCL_RET_BAD_ALLOC` if memory allocation failed, or
+ * \return `RCL_RET_ERROR` if an unexpected error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_take_event(
+  const rcl_event_t * event,
+  void * event_info);
+
+// Finalize an event.
+/**
+ * Finalize an event.
+ *
+ * \param[in] event to finalize
+ * \return `RCL_RET_OK` if successful, or
+ * \return `RCL_RET_EVENT_INVALID` if event is null, or
+ * \return `RCL_RET_ERROR` if an unexpected error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_event_fini(rcl_event_t * event);
+
+/// Return the rmw event handle.
+/**
+ * The handle returned is a pointer to the internally held rmw handle.
+ * This function can fail, and therefore return `NULL`, if the:
+ *   - event is `NULL`
+ *   - event is invalid (never called init, called fini, or invalid node)
+ *
+ * The returned handle is made invalid if the event is finalized or if
+ * rcl_shutdown() is called.
+ * The returned handle is not guaranteed to be valid for the life time of the
+ * event as it may be finalized and recreated itself.
+ * Therefore it is recommended to get the handle from the event using
+ * this function each time it is needed and avoid use of the handle
+ * concurrently with functions that might change it.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] event pointer to the rcl event
+ * \return rmw event handle if successful, otherwise `NULL`
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rmw_event_t *
+rcl_event_get_rmw_handle(const rcl_event_t * event);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif  // RCL__EVENT_H_
--- a/rcl/include/rcl/event.h
+++ b/rcl/include/rcl/event.h
@@ -27,7 +27,6 @@ extern "C"
 #include "rcl/subscription.h"
 #include "rcl/visibility_control.h"
 
-
 typedef enum rcl_publisher_event_type_t
 {
   RCL_PUBLISHER_OFFERED_DEADLINE_MISSED,
@@ -51,7 +50,6 @@ typedef struct rcl_event_t
 {
   struct rcl_event_impl_t * impl;
 } rcl_event_t;
-
 
 /// Return a rcl_event_t struct with members set to `NULL`.
 /**
--- a/rcl/include/rcl/node.h
+++ b/rcl/include/rcl/node.h
@@ -361,6 +361,32 @@ RCL_WARN_UNUSED
 rcl_ret_t
 rcl_node_get_domain_id(const rcl_node_t * node, size_t * domain_id);
 
+/// Manually assert that this node is alive (for RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_NODE)
+/**
+ * If the rmw Liveliness policy is set to RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_NODE, the creator of
+ * this node may manually call `assert_liveliness` at some point in time to signal to the rest
+ * of the system that this Node is still alive.
+ * This function must be called at least as often as the qos_profile's liveliness_lease_duration
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | Yes
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] node handle to the node that needs liveliness to be asserted
+ * \return `RCL_RET_OK` if the liveliness assertion was completed successfully, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
+ * \return `RCL_RET_NODE_INVALID` if the node is invalid, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_node_assert_liveliness(const rcl_node_t * node);
+
 /// Return the rmw node handle.
 /**
  * The handle returned is a pointer to the internally held rmw handle.
--- a/rcl/include/rcl/node.h
+++ b/rcl/include/rcl/node.h
@@ -378,7 +378,6 @@ rcl_node_get_domain_id(const rcl_node_t * node, size_t * domain_id);
  *
  * \param[in] node handle to the node that needs liveliness to be asserted
  * \return `RCL_RET_OK` if the liveliness assertion was completed successfully, or
- * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
  * \return `RCL_RET_NODE_INVALID` if the node is invalid, or
  * \return `RCL_RET_ERROR` if an unspecified error occurs.
  */
--- a/rcl/include/rcl/types.h
+++ b/rcl/include/rcl/types.h
@@ -99,6 +99,13 @@ typedef rmw_ret_t rcl_ret_t;
 /// Argument is not a valid log level rule
 #define RCL_RET_INVALID_LOG_LEVEL_RULE 1020
 
+// rcl event specific ret codes in 20XX
+/// Invalid rcl_event_t given return code.
+#define RCL_RET_EVENT_INVALID 2000
+/// Failed to take an event from the event handle
+#define RCL_RET_EVENT_TAKE_FAILED 2001
+
+
 /// typedef for rmw_serialized_message_t;
 typedef rmw_serialized_message_t rcl_serialized_message_t;
 
--- a/rcl/include/rcl/types.h
+++ b/rcl/include/rcl/types.h
@@ -105,7 +105,6 @@ typedef rmw_ret_t rcl_ret_t;
 /// Failed to take an event from the event handle
 #define RCL_RET_EVENT_TAKE_FAILED 2001
 
-
 /// typedef for rmw_serialized_message_t;
 typedef rmw_serialized_message_t rcl_serialized_message_t;
 
--- a/rcl/src/rcl/event.c
+++ b/rcl/src/rcl/event.c
@@ -53,6 +53,7 @@ rcl_publisher_event_init(
   const rcl_publisher_event_type_t event_type)
 {
   rcl_ret_t ret = RCL_RET_OK;
+  RCL_CHECK_ARGUMENT_FOR_NULL(event, RCL_RET_EVENT_INVALID);
   // Check publisher and allocator first, so allocator can be used with errors.
   RCL_CHECK_ARGUMENT_FOR_NULL(publisher, RCL_RET_INVALID_ARGUMENT);
   rcl_allocator_t * allocator = &publisher->impl->options.allocator;
@@ -92,6 +93,7 @@ rcl_subscription_event_init(
   const rcl_subscription_event_type_t event_type)
 {
   rcl_ret_t ret = RCL_RET_OK;
+  RCL_CHECK_ARGUMENT_FOR_NULL(event, RCL_RET_EVENT_INVALID);
   // Check subscription and allocator first, so allocator can be used with errors.
   RCL_CHECK_ARGUMENT_FOR_NULL(subscription, RCL_RET_INVALID_ARGUMENT);
   rcl_allocator_t * allocator = &subscription->impl->options.allocator;
@@ -130,7 +132,7 @@ rcl_take_event(
   void * event_info)
 {
   bool taken = false;
-  RCL_CHECK_ARGUMENT_FOR_NULL(event, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(event, RCL_RET_EVENT_INVALID);
   RCL_CHECK_ARGUMENT_FOR_NULL(event_info, RCL_RET_INVALID_ARGUMENT);
   rmw_ret_t ret = rmw_take_event(&event->impl->rmw_handle, event_info, &taken);
   if (RMW_RET_OK != ret) {
--- a/rcl/src/rcl/event.c
+++ b/rcl/src/rcl/event.c
@@ -1,0 +1,184 @@
+// Copyright 2019 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#include "rcl/event.h"
+
+#include <stdio.h>
+
+#include "rcl/error_handling.h"
+#include "rcl/expand_topic_name.h"
+#include "rcl/remap.h"
+#include "rcutils/logging_macros.h"
+#include "rmw/error_handling.h"
+#include "rmw/validate_full_topic_name.h"
+#include "rmw/event.h"
+
+#include "./common.h"
+#include "./publisher_impl.h"
+#include "./subscription_impl.h"
+
+typedef struct rcl_event_impl_t
+{
+  rmw_event_t rmw_handle;
+  rcl_allocator_t allocator;
+} rcl_event_impl_t;
+
+rcl_event_t
+rcl_get_zero_initialized_event()
+{
+  static rcl_event_t null_event = {0};
+  return null_event;
+}
+
+rcl_ret_t
+rcl_publisher_event_init(
+  rcl_event_t * event,
+  const rcl_publisher_t * publisher,
+  const rcl_publisher_event_type_t event_type)
+{
+  rcl_ret_t ret = RCL_RET_OK;
+  // Check publisher and allocator first, so allocator can be used with errors.
+  RCL_CHECK_ARGUMENT_FOR_NULL(publisher, RCL_RET_INVALID_ARGUMENT);
+  rcl_allocator_t * allocator = &publisher->impl->options.allocator;
+  RCL_CHECK_ALLOCATOR_WITH_MSG(allocator, "invalid allocator", return RCL_RET_INVALID_ARGUMENT);
+
+  // Allocate space for the implementation struct.
+  event->impl = (rcl_event_impl_t *) allocator->allocate(
+    sizeof(rcl_event_impl_t), allocator->state);
+  RCL_CHECK_FOR_NULL_WITH_MSG(
+    event->impl, "allocating memory failed", ret = RCL_RET_BAD_ALLOC; return ret);
+
+  event->impl->rmw_handle = rmw_get_zero_initialized_event();
+  event->impl->allocator = *allocator;
+
+  rmw_event_type_t rmw_event_type = RMW_EVENT_INVALID;
+  switch (event_type) {
+    case RCL_PUBLISHER_OFFERED_DEADLINE_MISSED:
+      rmw_event_type = RMW_EVENT_OFFERED_DEADLINE_MISSED;
+      break;
+    case RCL_PUBLISHER_LIVELINESS_LOST:
+      rmw_event_type = RMW_EVENT_LIVELINESS_LOST;
+      break;
+    default:
+      RCL_SET_ERROR_MSG("Event type for publisher not supported");
+      return RCL_RET_INVALID_ARGUMENT;
+  }
+  return rmw_publisher_event_init(
+    &event->impl->rmw_handle,
+    publisher->impl->rmw_handle,
+    rmw_event_type);
+}
+
+rcl_ret_t
+rcl_subscription_event_init(
+  rcl_event_t * event,
+  const rcl_subscription_t * subscription,
+  const rcl_subscription_event_type_t event_type)
+{
+  rcl_ret_t ret = RCL_RET_OK;
+  // Check subscription and allocator first, so allocator can be used with errors.
+  RCL_CHECK_ARGUMENT_FOR_NULL(subscription, RCL_RET_INVALID_ARGUMENT);
+  rcl_allocator_t * allocator = &subscription->impl->options.allocator;
+  RCL_CHECK_ALLOCATOR_WITH_MSG(allocator, "invalid allocator", return RCL_RET_INVALID_ARGUMENT);
+
+  // Allocate space for the implementation struct.
+  event->impl = (rcl_event_impl_t *) allocator->allocate(
+    sizeof(rcl_event_impl_t), allocator->state);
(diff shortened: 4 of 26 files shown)


chapulina:
If this PR is ready to be reviewed please replace the `in progress` label with `in review`. Otherwise, please list the tickets it's blocked by. Thanks!

mm318:
This PR is ready for review.

rcl/src/rcl/event.c
```diff
@@ -0,0 +1,182 @@
...
+#include "rcutils/logging_macros.h"
+#include "rmw/error_handling.h"
+#include "rmw/validate_full_topic_name.h"
+#include "rmw/event.h"
+
+#include "./common.h"
+#include "./publisher_impl.h"
+#include "./subscription_impl.h"
+
+
```
wjwwood: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

rcl/src/rcl/event.c
```diff
@@ -0,0 +1,182 @@
...
+typedef struct rcl_event_impl_t
+{
+  rmw_event_t rmw_handle;
+  rcl_allocator_t allocator;
+} rcl_event_impl_t;
+
+
```
wjwwood: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

Here and below, I won't comment on each line since there are several.

rcl/src/rcl/event.c
```diff
@@ -0,0 +1,182 @@
...
+  rmw_event_type_t rmw_event_type = RMW_EVENT_INVALID;
+  switch (event_type) {
+    case RCL_PUBLISHER_OFFERED_DEADLINE_MISSED:
+      rmw_event_type = RMW_EVENT_OFFERED_DEADLINE_MISSED;
+      break;
+    case RCL_PUBLISHER_LIVELINESS_LOST:
+      rmw_event_type = RMW_EVENT_LIVELINESS_LOST;
+      break;
+    default:
+      return RCL_RET_INVALID_ARGUMENT;
```
wjwwood: You should set the error message in this case to something specific about the event type being unknown.

rcl/src/rcl/event.c
```diff
@@ -0,0 +1,182 @@
...
+  rmw_event_type_t rmw_event_type = RMW_EVENT_INVALID;
+  switch (event_type) {
+    case RCL_SUBSCRIPTION_REQUESTED_DEADLINE_MISSED:
+      rmw_event_type = RMW_EVENT_REQUESTED_DEADLINE_MISSED;
+      break;
+    case RCL_SUBSCRIPTION_LIVELINESS_CHANGED:
+      rmw_event_type = RMW_EVENT_LIVELINESS_CHANGED;
+      break;
+    default:
+      return RCL_RET_INVALID_ARGUMENT;
```
wjwwood: Same here.

rcl/src/rcl/event.c
```diff
@@ -0,0 +1,182 @@
...
+  bool taken = false;
+  RCL_CHECK_ARGUMENT_FOR_NULL(event, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(event_info, RCL_RET_INVALID_ARGUMENT);
+  rmw_ret_t ret = rmw_take_event(&event->impl->rmw_handle, event_info, &taken);
+  if (RMW_RET_OK != ret) {
+    RCL_SET_ERROR_MSG(rmw_get_error_string().str);
+    return rcl_convert_rmw_ret_to_rcl_ret(ret);
+  }
+  RCUTILS_LOG_DEBUG_NAMED(
+    ROS_PACKAGE_NAME, "Event take request succeeded: %s", taken ? "true" : "false");
```
wjwwood: Please try to avoid memory allocation where possible (especially in commonly called functions like publish and take, as opposed to init and fini), instead of doing a formatted string here, just use a generic message, or and if/else to log two different messages.

rcl/include/rcl/event.h
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * Should be called to get a null rcl_event_t before passing to
+ * rcl_event_init().
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_event_t
+rcl_get_zero_initialized_event(void);
+
+RCL_PUBLIC
```
wjwwood: All public methods need documentation.

rcl/src/rcl/publisher.c
```diff
@@ -22,23 +22,17 @@ extern "C"
...
-typedef struct rcl_publisher_impl_t
-{
-  rcl_publisher_options_t options;
-  rmw_qos_profile_t actual_qos;
-  rcl_context_t * context;
-  rmw_publisher_t * rmw_handle;
-} rcl_publisher_impl_t;
+#include "./common.h"
+#include "./publisher_impl.h"
+
```
wjwwood: sneaky extra line so I'll mention it...

nitpick: minimize vertical whitespace

There's another like it below.

rcl/test/rcl/test_events.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+    ret = rcl_wait(&wait_set, RCL_MS_TO_NS(period_ms));
+    if (ret == RCL_RET_TIMEOUT) {
+      continue;
+    }
+    EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+
+    for (size_t i = 0; i < wait_set.size_of_subscriptions; ++i) {
+      if (wait_set.subscriptions[i] && wait_set.subscriptions[i] == subscription) {
+        msg_ready = true;
```
wjwwood: Is it intentional that this function will wait the max number of iterations every time? It could return here or at the end of the do-while loop if not.

rcl/test/rcl/test_events.cpp
```diff
@@ -0,0 +1,386 @@
...
+    }
+    this->node_ptr = new rcl_node_t;
+    *this->node_ptr = rcl_get_zero_initialized_node();
+    const char * name = "test_event_node";
+    rcl_node_options_t node_options = rcl_node_get_default_options();
+    ret = rcl_node_init(this->node_ptr, name, "", this->context_ptr, &node_options);
+    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  }
+
+  void SetUpPubSub(
```
wjwwood: nitpick: Method names should be `snake_case`, even if the other methods on this class are not (they are part of the GTest API) and cannot be changed.

rcl/test/rcl/test_events.cpp
```diff
@@ -0,0 +1,386 @@
...
+}
+
+/*
+ * Basic test of publisher and subscriber liveliness events
+ */
+TEST_F(CLASSNAME(TestEventFixture, RMW_IMPLEMENTATION), test_pubsub_liveliness)
+{
+  SetUpPubSub(RCL_PUBLISHER_LIVELINESS_LOST, RCL_SUBSCRIPTION_LIVELINESS_CHANGED);
+
+  // TODO(wjwwood): add logic to wait for the connection to be established
```
wjwwood: Why have the tools to do this properly now, let's not build in this issue from the get go. You can use `rcl_subscription_get_publisher_count` (https://github.com/ros2/rcl/blob/64a1412594b147bc363146e02d5ebc065c0d8a59/rcl/include/rcl/subscription.h#L415-L440) and/or `rcl_publisher_get_subscription_count` (https://github.com/ros2/rcl/blob/64a1412594b147bc363146e02d5ebc065c0d8a59/rcl/include/rcl/publisher.h#L441-L466) to check in a busy loop, allowing you to wait a long time if needed, but a very short time most of the time.

Something like:

```c++
// total wait time of 10 seconds, if never ready
size_t max_iterations = 100;
std::milliseconds wait_period(10);
size_t iteration = 0;
do {
  iteration++;
  size_t count = 0;
  rcl_ret_t ret = rcl_subscription_get_publisher_count(&subscription, &count);
  ASSERT_EQ(RCL_RET_OK, ret);
  if (count > 0) {
    break;
  }
  std::this_thread::sleep_for(wait_period);
} while (iteration < max_iterations);
```

rcl/src/rcl/event.c
ross-desmond: Removed

rcl/src/rcl/event.c
ross-desmond: Removed

rcl/src/rcl/event.c
ross-desmond: Good point, added

rcl/src/rcl/event.c
ross-desmond: Added

rcl/src/rcl/event.c
ross-desmond: Refactored

rcl/include/rcl/event.h
ross-desmond: Added

rcl/src/rcl/publisher.c
ross-desmond: Removed

rcl/test/rcl/test_events.cpp
ross-desmond: Unintentional, good catch!

rcl/test/rcl/test_events.cpp
ross-desmond: Done

rcl/test/rcl/test_events.cpp
ross-desmond: Added

rcl/test/rcl/test_get_actual_qos.cpp
```diff
@@ -182,34 +192,42 @@ get_parameters()
...
       rmw_qos_profile_system_default,
       expected_system_default_qos,
       "publisher_system_default_qos"});
   } else {
-    if (!rmw_implementation_str.compare("rmw_opensplice_cpp") ||
-      !rmw_implementation_str.compare("rmw_connext_cpp") ||
-      !rmw_implementation_str.compare("rmw_connext_dynamic_cpp"))
+    if (rmw_implementation_str == "rmw_connext_cpp" ||
+        rmw_implementation_str == "rmw_connext_dynamic_cpp" ||
+        rmw_implementation_str == "rmw_opensplice_cpp")
```
wjwwood: Please either undo this change, or change it to be like this:

```c++
    if (
      rmw_implementation_str == "rmw_connext_cpp" ||
      rmw_implementation_str == "rmw_connext_dynamic_cpp" ||
      rmw_implementation_str == "rmw_opensplice_cpp")
```

We avoid indenting more than one level between any two lines, or indenting non-multiples of the indentation (e.g. 3 or 5 spaces), including doing so in order to vertically align the indention like this.

rcl/test/rcl/test_events.cpp
```diff
@@ -0,0 +1,546 @@
...
+  msg_ready = false;
+  subscription_event_ready = false;
+  publisher_event_ready = false;
+
+  int num_subscriptions = (nullptr == subscription ? 0 : 1);
+  int num_events = (nullptr == subscription_event ? 0 : 1) + (nullptr == publisher_event ? 0 : 1);
+
+  rcl_wait_set_t wait_set = rcl_get_zero_initialized_wait_set();
+  rcl_ret_t ret = rcl_wait_set_init(&wait_set, num_subscriptions, 0, 0, 0, 0, num_events,
+      context, rcl_get_default_allocator());
```
wjwwood: nitpick: please do not indent more than one level (in this case two spaces)

rcl/test/rcl/test_get_actual_qos.cpp
ross-desmond: Sure, I will change the formatting.

mm318: Removed the extra indentation, but no longer vertically aligned. Is that okay?

rcl/test/rcl/test_events.cpp
mm318: Fixed.

ross-desmond: Done

rcl/test/rcl/test_get_actual_qos.cpp
wjwwood: Yes, as-is is fine. I normally would place the first part of condition on its own line too, so they are vertically aligned without extra indention, e.g.:

```c++
    if (
      rmw_implementation_str == "rmw_connext_cpp" ||
      rmw_implementation_str == "rmw_connext_dynamic_cpp" ||
      rmw_implementation_str == "rmw_opensplice_cpp")
    {
      // ...
    }
```

Or if you prefer:

```c++
    if (
      rmw_implementation_str == "rmw_connext_cpp" ||
      rmw_implementation_str == "rmw_connext_dynamic_cpp" ||
      rmw_implementation_str == "rmw_opensplice_cpp"
    ) {
      // ...
    }
```

The most important thing is that if you're scanning the code you can see where different scopes start/end, e.g.:

```
// bad, is first statement an indention typo or a new scope (at first glance)?
if statement starts -------> | if (condition1 &&
                             |     condition2 &&
hard to see end here ------> |     condition3) {
hard to see body start ----> |   statement1;
                             |   statement2;
                             |   statement3;
                             | }

// good, start and end for scopes are in the same column
if statement starts -------> | if (
                             |   condition1 &&
                             |   condition2 &&
                             |   condition3)
easy to see body start ----> | {
                             |   statement1;
                             |   statement2;
                             |   statement3;
                             | }
```

I think about it like as if it were the "same as":

```c++
bool condition = (
  rmw_implementation_str == "rmw_connext_cpp" ||
  rmw_implementation_str == "rmw_connext_dynamic_cpp" ||
  rmw_implementation_str == "rmw_opensplice_cpp"
);

if (condition) {
  // ...
}
```

Which matches our style with initializer lists in c++ and dictionaries in Python, e.g.:

```
... = {
  "item1",
  "item2,
  // ...
}
```

rcl/test/rcl/test_events.cpp
```diff
@@ -0,0 +1,546 @@
...
+  return RCL_RET_OK;
+}
+
+TEST_F(CLASSNAME(TestEventFixture, RMW_IMPLEMENTATION), test_unsupported_lifespan) {
+  if (is_unsupported) {
+    rmw_time_t deadline {0, 0};
+    rmw_time_t lifespan {1, 0};
+    rmw_time_t lease_duration {1, 0};
+    rmw_qos_liveliness_policy_t liveliness_policy = RMW_QOS_POLICY_LIVELINESS_AUTOMATIC;
+    EXPECT_EQ(RMW_RET_ERROR,
```
wjwwood: Wasn't there supposed to be an `RMW_RET_UNSUPPORTED` or something similar? Would that not be used here?

rcl/test/rcl/test_events.cpp
```diff
@@ -0,0 +1,546 @@
...
+      setup_subscriber(deadline, lifespan, lease_duration,
```
wjwwood: Doesn't this return an `rcl_ret_t`? Shouldn't you compare it to an `RCL_RET_*` define?

rcl/test/rcl/test_events.cpp
```diff
@@ -0,0 +1,546 @@
...
+  }
+
+  // kill the publisher
+  ret = rcl_event_fini(&publisher_event);
+  EXPECT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
+  ret = rcl_publisher_fini(&publisher, this->node_ptr);
+  EXPECT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
+
+  // wait for lease duration to expire
+  std::this_thread::sleep_for(LIVELINESS_LEASE_DURATION_IN_S + milliseconds(500));
```
wjwwood: I would like to see all sleeps gone if possible.

Instead, I think it would be better if `wait_for_msgs_and_events` would check the status during this time (even though we expect it not to change), and even compare it to a timestamp taken before the `fini` to make sure it doesn't expire early. Expiring shortly after the lease duration is ok, but before should be a bug (in my opinion), similar to the contract given by `sleep_for` (i.e. it may sleep >= period, but not less than).

rcl/test/rcl/test_events.cpp
```diff
@@ -0,0 +1,546 @@
...
+};
+
+rcl_ret_t
+wait_for_msgs_and_events(
+  rcl_subscription_t * subscription,
+  rcl_event_t * subscription_event,
+  rcl_event_t * publisher_event,
+  rcl_context_t * context,
+  int64_t period_ms,
+  bool & msg_ready,
```
wjwwood: Can you take `bool *` instead here? The current state leads to misleading usage, e.g.:

```c++
  // wait for events
  bool msg_ready, subscription_event_ready, publisher_event_ready;

  // wait for lease duration to expire
  std::this_thread::sleep_for(DEADLINE_PERIOD_IN_S + milliseconds(500));
  ASSERT_EQ(wait_for_msgs_and_events(&subscription, &subscription_event, &publisher_event,
    context_ptr, 1000, msg_ready, subscription_event_ready, publisher_event_ready), RCL_RET_OK);
//                     ^ I assume a copy is made, or a const & is being used, when passing a value like this
```

Also the Google Style guide for C++ (which our style is based on) does not allow non-const references:

https://google.github.io/styleguide/cppguide.html#Reference_Arguments

rcl/test/rcl/test_events.cpp
wjwwood: Testing for the "happens too early" case might be tricky, so you don't have to do that, but I'd still rather just have it sitting in `rcl_wait()` than sleeping here.

rcl/test/rcl/test_events.cpp
```diff
@@ -0,0 +1,546 @@
...
+    ret = rcl_publish(&publisher, &msg);
+    test_msgs__msg__Primitives__fini(&msg);
+    ASSERT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
+  }
+
+  // wait for events
+  bool msg_ready, subscription_event_ready, publisher_event_ready;
+
+  // wait for lease duration to expire
+  std::this_thread::sleep_for(DEADLINE_PERIOD_IN_S + milliseconds(500));
```
wjwwood: Again, any sleeping in a test needs to be justified in my opinion and avoided entirely if possible. They usually lead to flaky tests due to assuming that the timing will behave in CI every time like it does locally on your machine.

There are cases where it makes sense, but I'd rather be waiting for an event (and knowing or expecting to wait at least some period of time blocking for that event) rather than sleeping.

rcl/test/rcl/test_get_actual_qos.cpp
```diff
@@ -199,24 +201,24 @@ get_parameters()
...
+      rmw_implementation_str == "rmw_opensplice_cpp")
     {
       rmw_qos_profile_t expected_system_default_qos = expected_system_default_qos_profile();
       parameters.push_back({
-        rmw_qos_profile_system_default,
-        expected_system_default_qos,
-        "publisher_system_default_qos"});
+       rmw_qos_profile_system_default,
+       expected_system_default_qos,
+       "publisher_system_default_qos"});
```
wjwwood: Why did you change this to 1 space of indention?

Actually why was this file changed at all?

rcl/test/rcl/test_events.cpp
ross-desmond: In this case, we could modify this test to do 2 wait() function calls, one for message retrieval and one for event condition.
Changing

rcl/test/rcl/test_get_actual_qos.cpp
ross-desmond: rmw_qos_profile_t has changed, and this file manually creates them using item initializer.
See: https://github.com/ros2/rmw/blob/ba0a71707236939c08ece8265a18b19758dd3121/rmw/include/rmw/types.h#L211

Since this file was changed for this reason, I've improved the readability of the if statements as
Clang tidy will recommend "if an equality or inequality check can suffice, that is recommended."

The 1 space indentation may have been an accident, ament_uncrustify --refactor was run on this file, it may have been my code style settings that have changed this.

rcl/test/rcl/test_events.cpp
```diff
@@ -0,0 +1,546 @@
...
+{
+public:
+  void SetUp()
+  {
+    is_opensplice =
+      std::string(rmw_get_implementation_identifier()).find("rmw_opensplice") == 0;
+    is_fastrtps =
+      std::string(rmw_get_implementation_identifier()).find("rmw_fastrtps") == 0;
+
+    is_unsupported = is_fastrtps;
```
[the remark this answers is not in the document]
wjwwood: Alternatively, you could determine this programmatically, e.g. call one of the functions and see if it returns unsupported. That way this could be enabled/disabled automatically even for new rmw's. On the other hand, they should all support this feature as soon as possible, as we already discussed on the design document, I don't think optional features are a good thing to have long term.

rcl/test/rcl/test_events.cpp
wjwwood: That's sort of what I was thinking as well. If you can't do it that's ok, but please have a comment justifying the sleep if you cannot. Thanks!

rcl/test/rcl/test_get_actual_qos.cpp
wjwwood: Ok, let's fix the indentation, it should always be a multiple of 2 for C++.

I'll re-review the file with this new info in mind, thanks.

wjwwood: Ok, the style changes are unrelated to the pr, but ok by me. However, I don't follow what you mean about the initializer list changing. Nothing changed as far as I can tell. The only thing I see in the diff is whitespace changes.

wjwwood:
Ok, I'm done reviewing for now, the comments I had were pretty minor. I'll try to quickly re-review when the feedback has been addressed. Thanks for iterating and for being patient with me :)

rcl/test/rcl/test_events.cpp
ross-desmond: I will change it to a bool* since the google style guide does not allow non-const references.
 
bool & msg_ready implicitly enforces references to NULL, void, invalid objects or functions, and bit fields. Google's Con for this does not seem to outweigh its benefits "References can be confusing, as they have value syntax but pointer semantics." This enforces the intention of this function and reduces the bloating of invalid param checks.

rcl/test/rcl/test_events.cpp
ross-desmond: Yes this should be RCL_RET_ERROR

rcl/include/rcl/event.h
```diff
@@ -0,0 +1,171 @@
...
+#endif
+
+#include "rcl/client.h"
+#include "rcl/macros.h"
+#include "rcl/publisher.h"
+#include "rcl/service.h"
+#include "rcl/subscription.h"
+#include "rcl/visibility_control.h"
+
+
```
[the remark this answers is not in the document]
mm318: Will do.

rcl/include/rcl/event.h
```diff
@@ -0,0 +1,171 @@
...
+/// Internal rcl implementation struct.
+struct rcl_event_impl_t;
+
+/// Structure which encapsulates a ROS QoS event handle.
+typedef struct rcl_event_t
+{
+  struct rcl_event_impl_t * impl;
+} rcl_event_t;
+
+
```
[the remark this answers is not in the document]
mm318: Will do.

rcl/include/rcl/types.h
```diff
@@ -99,6 +99,13 @@ typedef rmw_ret_t rcl_ret_t;
 /// Argument is not a valid log level rule
 #define RCL_RET_INVALID_LOG_LEVEL_RULE 1020
 
+// rcl event specific ret codes in 20XX
+/// Invalid rcl_event_t given return code.
+#define RCL_RET_EVENT_INVALID 2000
+/// Failed to take an event from the event handle
+#define RCL_RET_EVENT_TAKE_FAILED 2001
+
+
```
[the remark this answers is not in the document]
mm318: Will do.

rcl/src/rcl/event.c
```diff
@@ -0,0 +1,184 @@
...
+  const rcl_publisher_event_type_t event_type)
+{
+  rcl_ret_t ret = RCL_RET_OK;
+  // Check publisher and allocator first, so allocator can be used with errors.
+  RCL_CHECK_ARGUMENT_FOR_NULL(publisher, RCL_RET_INVALID_ARGUMENT);
+  rcl_allocator_t * allocator = &publisher->impl->options.allocator;
+  RCL_CHECK_ALLOCATOR_WITH_MSG(allocator, "invalid allocator", return RCL_RET_INVALID_ARGUMENT);
+
+  // Allocate space for the implementation struct.
+  event->impl = (rcl_event_impl_t *) allocator->allocate(
```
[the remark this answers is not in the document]
mm318: Will do. Some of the other functions are already doing `RCL_CHECK_ARGUMENT_FOR_NULL(event, RCL_RET_EVENT_INVALID)`.

rcl/src/rcl/subscription_impl.h
```diff
@@ -0,0 +1,29 @@
...
+// limitations under the License.
+
+#ifndef RCL__SUBSCRIPTION_IMPL_H_
+#define RCL__SUBSCRIPTION_IMPL_H_
+
+#include "rmw/rmw.h"
+
+#include "rcl/subscription.h"
+
+
```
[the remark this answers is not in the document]
mm318: Will do.

rcl/test/rcl/test_events.cpp
```diff
@@ -0,0 +1,546 @@
...
+
+protected:
+  rcl_context_t * context_ptr;
+  rcl_node_t * node_ptr;
+
+  rcl_publisher_t publisher;
+  rcl_event_t publisher_event;
+  rcl_subscription_t subscription;
+  rcl_event_t subscription_event;
+  bool is_unsupported, is_opensplice, is_fastrtps;
```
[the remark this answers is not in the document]
mm318: Will do.

rcl/test/rcl/test_events.cpp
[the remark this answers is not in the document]
mm318: Will add the `TODO`.

Many of these tests are testing against `rcl_publisher_init()`/`rcl_subscription_init()`, which we left untouched since the changes made for the QoS features are already quite a sizeable scope, and `rcl_publisher_init()`/`rcl_subscription_init()` don't seem to have the return unsupported behavior.

I agree that having optional features is not a good thing. It'll be up to FastRTPS to provide the features.

rcl/test/rcl/test_events.cpp
```diff
@@ -0,0 +1,546 @@
...
+
+  // test subscriber/datareader liveliness changed status
+  EXPECT_TRUE(subscription_event_ready);
+  {
+    rmw_liveliness_changed_status_t liveliness_status;
+    ret = rcl_take_event(&subscription_event, &liveliness_status);
+    ASSERT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
+    EXPECT_EQ(liveliness_status.alive_count, 0);
+    int32_t alive_count_changed = 0;
+    if (is_opensplice) {
```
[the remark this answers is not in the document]
mm318: Will add comments.

rcl/test/rcl/test_events.cpp
mm318: Sorry, this is supposed to return a `rcl_ret_t` and it's coming from `rcl_subscription_init()`, so there won't be a `RCL_RET_UNSUPPORTED` return code.

rcl/test/rcl/test_events.cpp
mm318: Will change to `RCL_RET_ERROR`.

rcl/test/rcl/test_events.cpp
mm318: Okay, will do.

rcl/test/rcl/test_get_actual_qos.cpp
mm318: The initializer list changes were originally part of this PR, but ended up being merged in through https://github.com/ros2/rcl/pull/416 instead. Sorry for the confusion.

Anyway, will fix the indentation issue, and will keep the use of `==` as it expresses the intent better.

rcl/src/rcl/event.c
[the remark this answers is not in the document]
wjwwood: Thanks.

rcl/test/rcl/test_events.cpp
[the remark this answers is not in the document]
wjwwood: Ok, sounds good.

rcl/test/rcl/test_events.cpp
wjwwood: Makes sense, thanks for explaining it.

rcl/test/rcl/test_events.cpp
```diff
@@ -0,0 +1,553 @@
...
+    ret = setup_subscriber(deadline, lifespan, lease_duration, liveliness_policy);
+    ASSERT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
+
+    // init subscription event
+    subscription_event = rcl_get_zero_initialized_event();
+    ret = rcl_subscription_event_init(&subscription_event, &subscription, sub_event_type);
+    ASSERT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
+
+    // wait for discovery
+    // total wait time of 10 seconds, if never ready
```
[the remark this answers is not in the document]
wjwwood: I'm fine with what ever value (I just want it to be consistent with the comment), though making it larger is probably ok, since it will avoid flaky behavior more often and still should run quickly most of the time.

rcl/test/rcl/test_events.cpp
mm318: With the change to 2 waits, I am running into some issues with this particular test case (`test_pubsub_deadline_missed`) for Connext (but not OpenSplice). `rcl_wait()` is returning OK (not timing out), but nothing in the wait_set is ready.

rcl/test/rcl/test_events.cpp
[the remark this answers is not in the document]
mm318: I have increased the `max_iterations` to 1000 to match the total wait time of 10 seconds.

rcl/test/rcl/test_events.cpp
wjwwood: Seems like this got added back in the latest commit?

rcl/test/rcl/test_events.cpp
```diff
@@ -269,6 +269,71 @@ wait_for_msgs_and_events(
   return RCL_RET_OK;
 }
 
+// @brief Conditional function for determining when the wait_for_msgs_and_events loop is complete.
```
wjwwood: nitpick: please use `///` and no `brief`. Also we use `\` rather than `@`.

rcl/test/rcl/test_events.cpp
```diff
@@ -370,56 +435,50 @@ TEST_F(CLASSNAME(TestEventFixture, RMW_IMPLEMENTATION), test_pubsub_liveliness_k
...
 
+  if (msg_persist_ready) {
+    EXPECT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
+    EXPECT_EQ(std::string(msg.string_value.data, msg.string_value.size),
+      std::string(test_string));
+  }
+  if (subscription_persist_ready) {
+    EXPECT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
+    EXPECT_EQ(liveliness_status.alive_count, 0);
+    // TODO(ross-desmond): Connext and OpenSplice seem to be counting liveliness differently
```
wjwwood: This is concerning... I'd actual say that opensplice is just broken (or maybe Connext is?). Do you have a reason that they might be disagreeing?

rcl/test/rcl/test_events.cpp
```diff
@@ -269,6 +269,71 @@ wait_for_msgs_and_events(
...
+// @brief Conditional function for determining when the wait_for_msgs_and_events loop is complete.
+/**
+ * Conditional function for determining when the wait_for_msgs_and_events loop is complete.
+ *
+ * @param msg_persist_ready true if a msg has ever been received
+ * @param subscription_persist_readytrue if a subscription has been received
+ * @param publisher_persist_ready true if a pulbisher event has been received
+ * @return true when the desired conditions are met
+ */
+using ConditionalFunction = std::function<
```
wjwwood: Naming is really inconsequential here, but I'd call this a predicate.

rcl/include/rcl/node.h
```diff
@@ -361,6 +361,32 @@ RCL_WARN_UNUSED
...
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | Yes
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] node handle to the node that needs liveliness to be asserted
+ * \return `RCL_RET_OK` if the liveliness assertion was completed successfully, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
```
jacobperron: nit: there is no case where this function returns `RCL_RET_INVALID_ARGUMENT`. This can probably be removed.

jacobperron: Same for `rcl_publisher_assert_liveliness`.

rcl/test/rcl/test_events.cpp
mm318: This has been fixed by basically looping around `rcl_wait()`.

rcl/test/rcl/test_events.cpp
mm318: Fixed.

rcl/test/rcl/test_events.cpp
mm318: Fixed.

rcl/test/rcl/test_events.cpp
mm318: Fixed.

rcl/include/rcl/node.h
mm318: Fixed.

rcl/test/rcl/test_events.cpp
mm318: @ross-desmond, @nburek, and I believe this is a case of "distance vs. displacement". In this test, the publisher comes alive (`alive_count` +1) and is then shut down (`alive_count` -1). Although `alive_count` ends up where it started (at 0), that's two state changes to `alive_count`. So it looks like Connext's `alive_count_change` represents the literal change to `alive_count` while OpenSplice's `alive_count_change` represents the number of state changes to `alive_count`.

mm318: The spec reads that `alive_count_change` is "The change in the alive_count since the last time the listener was called or the status was read". Our interpretation of that statement leads us to believe that Connext's behavior is the correct one.

wjwwood: Sounds like a potential bug in OpenSplice then, or at least it would be nice to know if this is intentional or not. Can one of you summarize this issue on OpenSplice's issue tracker and ask for their opinion on this?

https://github.com/ADLINK-IST/opensplice/issues

/cc @dirk-thomas

wjwwood: For now we can continue with the exception in the test and a TODO, but ideally we'd come back and address this with a definitive comment that they disagree intentionally, or that we adjust for the difference of opinion in our layer, or something like that.

mm318: I have opened an issue on OpenSplice's issue tracker here: https://github.com/ADLINK-IST/opensplice/issues/88.

Thanks.

wjwwood: Thank you!

rcl/test/rcl/test_events.cpp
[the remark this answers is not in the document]
sloretz: @mm318 If you think the OSX connext failure is a timeout issue then I recommend upping this wait time (30 seconds?) before re-running

PR #409, opened 2019-03-30
Rename action state transitions

### LINKED ISSUE #399: Rename action state transitions
#### Feature description

As discussed in the [actions design doc](https://github.com/ros2/design/pull/193/files/1c98b9b7fbbeb5bde7cd36301c914e46016e1c0b#r229870628), we should rename the goal state transitions to use active verbs.

Basically, the following event names:

https://github.com/ros2/rcl/blob/1d8b831bdfe8c700a1934f4d6dc1803d49565865/rcl_action/include/rcl_action/types.h#L101-L103

can be renamed to

```
 GOAL_EVENT_SUCCEED, 
 GOAL_EVENT_ABORT, 
 GOAL_EVENT_CANCEL, 
```

This likely affects some downstream dependencies (e.g. `rclcpp_action` and `rclpy`), although it should not have a big impact since it's just renaming variables/functions.

jacobperron:
Now using active verbs.
Resolves #399.

### DIFF
--- a/rcl_action/test/rcl_action/test_goal_handle.cpp
+++ b/rcl_action/test/rcl_action/test_goal_handle.cpp
@@ -161,7 +161,7 @@ using EventStateActiveCancelableTuple =
   std::tuple<rcl_action_goal_event_t, rcl_action_goal_state_t, bool, bool>;
 using StateTransitionSequence = std::vector<EventStateActiveCancelableTuple>;
 const std::vector<std::string> event_strs = {
-  "EXECUTE", "CANCEL", "SUCCEED", "ABORT", "CANCEL"};
+  "EXECUTE", "REQUEST_CANCEL", "SUCCEED", "ABORT", "CANCEL"};
 
 class TestGoalHandleStateTransitionSequence
   : public ::testing::TestWithParam<StateTransitionSequence>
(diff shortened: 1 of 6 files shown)


jacobperron:
...I just released that there are now two events named `GOAL_EVENT_CANCEL`. Rethinking the change.

rcl_action/test/rcl_action/test_goal_handle.cpp
```diff
@@ -161,7 +161,7 @@ using EventStateActiveCancelableTuple =
   std::tuple<rcl_action_goal_event_t, rcl_action_goal_state_t, bool, bool>;
 using StateTransitionSequence = std::vector<EventStateActiveCancelableTuple>;
 const std::vector<std::string> event_strs = {
-  "EXECUTE", "CANCEL", "SET_SUCCEEDED", "SET_ABORTED", "SET_CANCELED"};
+  "EXECUTE", "CANCEL", "SUCCEED", "ABORT", "CANCEL"};
```
hidmic: @jacobperron shouldn't it be:
```suggestion
  "EXECUTE", "REQUEST_CANCEL", "SUCCEED", "ABORT", "CANCEL"};
```
instead?

jacobperron: Yep, good catch. fixed in 225d805

jacobperron:
Updated based on design discussion (https://github.com/ros2/design/pull/226). @hidmic, please take another look.

Issue #269, opened 2018-07-11
replace uses of reinterpret_cast with static_cast where possible


wjwwood:
See: https://github.com/ros2/rcl/pull/261#discussion_r195932052

jhdcs:
I think I might have replaced a large chunk of the reinterpret_casts. I'm not 100% sure I trust the testing framework to catch any errors, but so far everything seems to be building correctly.

Want me to open a pull request?

wjwwood:
> Want me to open a pull request?

I would default to yes, pull requests can always be declined or refactored, and don't take much to open :)

jhdcs:
Well, I still want to be polite ^_^

sloretz:
Fixed by ros2/rcl#410

PR #410, opened 2019-04-01
Replace reinterperet_cast with static_cast

jhdcs:
I went through the rcl project with an editor and replaced things I felt I could safely get away with (mostly areas that explicitly worked with allocators). I'll have to look at the pimpls later, unless you want to leave them as-is.

Connects to issue #269 

Signed-off-by: Jacob Hassold <jhassold@dcscorp.com>

### DIFF
--- a/rcl/test/rcl/test_time.cpp
+++ b/rcl/test/rcl/test_time.cpp
@@ -222,7 +222,7 @@ TEST(CLASSNAME(rcl_time, RMW_IMPLEMENTATION), default_clock_instanciation) {
   });
   ASSERT_TRUE(rcl_clock_valid(&ros_clock));
 
-  rcl_clock_t * steady_clock = reinterpret_cast<rcl_clock_t *>(
+  auto * steady_clock = static_cast<rcl_clock_t *>(
     allocator.zero_allocate(1, sizeof(rcl_clock_t), allocator.state));
   OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT({
     allocator.deallocate(steady_clock, allocator.state);
@@ -235,7 +235,7 @@ TEST(CLASSNAME(rcl_time, RMW_IMPLEMENTATION), default_clock_instanciation) {
   });
   ASSERT_TRUE(rcl_clock_valid(steady_clock));
 
-  rcl_clock_t * system_clock = reinterpret_cast<rcl_clock_t *>(
+  auto * system_clock = static_cast<rcl_clock_t *>(
     allocator.zero_allocate(1, sizeof(rcl_clock_t), allocator.state));
   OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT({
     allocator.deallocate(system_clock, allocator.state);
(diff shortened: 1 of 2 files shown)


wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6572/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2910/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5388/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6375/)

jhdcs:
There are a few left (27, including comments, according to CLion). But these were the easy ones, and I figured small incremental changes are best when dealing with such things as casting pointers. Especially if the code was working before.

wjwwood:
Sure, we can merge this one (with green CI) and leave the original issue open.

wjwwood:
Redoing the Windows CI as it looks unrelated and there have been passing CI since: [Build Status](https://ci.ros2.org/job/ci_windows/6380/)

jhdcs:
So we have 3 unstables and a failing test. Do your CIs give any reports that can be used to figure out which casts (if any) are causing the issue?

wjwwood:
The error you're speaking about sounds like interferrence from someone else (or yourself) in your building running the `ros1_bridge` on the sams `ROS_DOMAIN_ID` as you. We always run our tests with different `ROS_DOMAIN_ID`'s set to avoid cross talk.

Our CI summarizes the test failures:

- https://ci.ros2.org/job/ci_linux/6572/testReport/

But luckily for you, those two failures are known flakes, so don't worry about them.

Looks like on Windows we got unlucky and our CI started between pull request merges, so here's another:

[Build Status](https://ci.ros2.org/job/ci_windows/6412/)

jhdcs:
So, what we're basically seeing is that, for some reason changing a reinterperet_cast to a static_cast will cause the code to be unable to build on Windows machines?

Isn't that the _opposite_ of what we'd expect?

In any case, since I'm developing on Linux, do you have any suggestions on how to test this? Would building the code inside an emulator work?

jhdcs:
@wjwwood Would I be able to request another CI on this? Or have I used all of mine up?

I've been combing over the code, and I'm not really seeing where it would be causing an issue. I only made changes to two testing files, and this has been building reliably on my system...

rcl/test/rcl/test_time.cpp
```diff
@@ -289,8 +289,8 @@ TEST(CLASSNAME(rcl_time, RMW_IMPLEMENTATION), specific_clock_instantiation) {
 
 TEST(CLASSNAME(rcl_time, RMW_IMPLEMENTATION), rcl_time_difference) {
   rcl_allocator_t allocator = rcl_get_default_allocator();
-  rcl_clock_t * ros_clock =
-    reinterpret_cast<rcl_clock_t *>(allocator.allocate(sizeof(rcl_clock_t), allocator.state));
+  auto ros_clock =
```
allenh1: I think there's a missing `auto *` on this line, but that's not going to cause the CI failures.

jhdcs: I didn't think that auto needed the * since I thought it would infer it's correct type from the static cast. Is that not the case?

allenh1: > I didn't think that auto needed the * since I thought it would infer it's correct type from the static cast

This is indeed the case -- but, as you stated, it is inconsistent with other instances. Honestly, I am not a fan of this part of the `auto` keyword's behavior...

jhdcs:
Am I allowed another CI on this?

sloretz:
CI (testing just `rcl`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6765/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3067/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5558/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6590/)

@jhdcs There's no limit to the number of CI runs. Sometimes it just takes a while to get back to a PR. Please do comment when you'd like another run, it definitely helps. I came to start CI after getting a github notification email when you made a comment :+1:

jhdcs:
And hey, look at that! All passed!

Phew. Casts always make me a bit nervous, lol.

PR #413, opened 2019-04-09
fix missing include for std::runtime_error

dirk-thomas:
Which came up with VS2019.

### DIFF
--- a/rcl/test/rcl/client_fixture.cpp
+++ b/rcl/test/rcl/client_fixture.cpp
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <chrono>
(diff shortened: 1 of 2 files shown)


rcl/test/rcl/client_fixture.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include <chrono>
 #include <string>
+#include <stdexcept>
```
Karsten1987: alpha order

dirk-thomas: Right in one file, wrong in the other :worried: Fixed.

PR #415, opened 2019-04-11
Correct bool reading from yaml files

ivanpauno:
See https://github.com/ros2/launch_ros/issues/12#issuecomment-482275149

Basically, value="of" was comparing true with 'off', and it shouldn't.
Combining strncmp with strnlen is not a good idea. strcmp is enough, it returns only 0 when the strings are equal and have the same length.

Edit:
I have also added an extra checking. If a value is quoted, it's always detected as a string:
```yaml
a: 'true'
b: true
```
Before, both `a` and `b` where recognized like bools. Now it is corrected.

### DIFF
--- a/rcl_yaml_param_parser/src/parser.c
+++ b/rcl_yaml_param_parser/src/parser.c
@@ -118,6 +118,7 @@ static rcl_ret_t replace_ns(
 
 static void * get_value(
   const char * const value,
+  yaml_scalar_style_t style,
   data_types_t * val_type,
   const rcl_allocator_t allocator);
 
@@ -709,6 +710,7 @@ static rcl_ret_t add_val_to_string_arr(
 ///
 static void * get_value(
   const char * const value,
+  yaml_scalar_style_t style,
   data_types_t * val_type,
   const rcl_allocator_t allocator)
 {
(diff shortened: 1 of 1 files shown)


wjwwood:
You should run CI to look for Windows warnings, as it can be picky about these functions.

wjwwood:
(you should always run CI, but in case anyone was temped to call this trivial and just merge it)

rcl_yaml_param_parser/src/parser.c
```diff
@@ -744,17 +744,17 @@ static void * get_value(
...
+    (0 == strcmp(value, "n")) ||
+    (0 == strcmp(value, "no")) ||
+    (0 == strcmp(value, "No")) ||
+    (0 == strcmp(value, "NO")) ||
+    (0 == strcmp(value, "false")) ||
+    (0 == strcmp(value, "False")) ||
+    (0 == strcmp(value, "FALSE")) ||
+    (0 == strcmp(value, "off")) ||
+    (0 == strcmp(value, "Off")) ||
+    (0 == strcmp(value, "OFF")))
```
wjwwood: On an unrelated note, why do we need to determine this for ourselves? Shouldn't libyaml tell us the YAML type of something given a string?

wjwwood: I asked @anup-pem and apparently libyaml doesn't provide this functionality :| So never mind.

ivanpauno: The only problem that I'm still seeing, is the following one:
```yaml
a: 'true'
b: true
```
The first one should be recognized as an string, and the second one as a bool. Now, both are being recognized like a bool.

ivanpauno: I have solved the problem I described above in my last commit.

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6671/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2990/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5479/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6500/)

ivanpauno:
@anup-pem @wjwwood Is this ok for merging?
[Here](https://github.com/yaml/libyaml/blob/690a781ef6af70ce6749d6e2be91743345123998/include/yaml.h#L166-L182) is the documentation about `yaml_scalar_style_e`.

PR #421, opened 2019-04-27
Fix argument passed to logging macros

thomas-moulard:
This commit fixes two problems with the logging
statements:
- %p requires a void * object to be passed
- %d expects an integer. For size_t objects, %zu should be used

Signed-off-by: Thomas Moulard <tmoulard@amazon.com>

### DIFF
--- a/rcl/src/rcl/init.c
+++ b/rcl/src/rcl/init.c
@@ -60,7 +60,7 @@ rcl_init(
 
   RCUTILS_LOG_DEBUG_NAMED(
     ROS_PACKAGE_NAME,
-    "Initializing ROS client library, for context at address: %p", context);
+    "Initializing ROS client library, for context at address: %p", (void *) context);
 
   // test expectation that given context is zero initialized
   if (NULL != context->impl) {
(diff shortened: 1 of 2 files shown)


dirk-thomas:
> %p requires a void * object to be passed

From reading the [docs](https://en.cppreference.com/w/cpp/io/c/fprintf) I would think `%p` works with any kind of pointer:

    > writes an implementation defined character sequence defining a pointer.

thomas-moulard:
If you look at the table, the cell states that the expected type is `void *` though ("Argument type" / "no modifier).

GCC is giving me a warning for those lines after applying the following PR: https://github.com/ros2/rcutils/pull/154

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6829/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3124/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5616/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6654/)

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6830/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3125/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5617/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6655/)

dirk-thomas:
Thanks for the patch..

PR #422, opened 2019-04-29
Add return code to CancelGoal service response

jacobperron:
Connects to https://github.com/ros2/rcl_interfaces/issues/63

### DIFF
--- a/rcl_action/test/rcl_action/test_types.cpp
+++ b/rcl_action/test/rcl_action/test_types.cpp
@@ -67,6 +67,7 @@ TEST(TestActionTypes, test_get_zero_initialized_cancel_response)
   rcl_action_cancel_response_t cancel_response = rcl_action_get_zero_initialized_cancel_response();
   EXPECT_EQ(cancel_response.msg.goals_canceling.size, 0u);
   EXPECT_EQ(cancel_response.msg.goals_canceling.data, nullptr);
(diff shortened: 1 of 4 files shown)


dirk-thomas:
Could you use `ERROR_NONE` in more places which currently use `0`?

jacobperron:
> Could you use ERROR_NONE in more places which currently use 0?

I believe all the places I've left it as zero is because I'm testing if the message is zero initialized (regardless of what the name of the constant is). Does that make sense?

rcl_action/test/rcl_action/test_types.cpp
```diff
@@ -152,6 +155,7 @@ TEST(TestActionTypes, test_init_fini_cancel_response)
   EXPECT_EQ(ret, RCL_RET_OK);
   EXPECT_EQ(cancel_response.msg.goals_canceling.size, num_goals_canceling);
   EXPECT_NE(cancel_response.msg.goals_canceling.data, nullptr);
+  EXPECT_EQ(cancel_response.msg.return_code, 0);
```
jacobperron: I might update this one to use the constant though.

jacobperron: Done 06c8c49

jacobperron:
I've also added logic for setting the return code `GOAL_TERMINATED` in bdfb652

PR #425, opened 2019-04-30
Changes the default 3rd party logger from rcl_logging_noop to rcl_logging_log4cxx

nburek:
When we initially did the logging work we had shipped it using the rcl_logging_noop implementation by default because we were having issues getting rcl_logging_log4cxx fully built on Windows and Mac. This allowed the logging changes to make it in for Crystal, but meant that file based logging still wasn't enable. 

Now that rcl_logging_log4cxx has been building, switching the default logger over will enable logs to start being generated by ROS nodes and will enable advanced logging configuration via a log4cxx config file. 

Without this change, anyone who wants the file based logging would have to rebuild from source with the environment variable set to change from the default logger to explicitly use log4cxx. 

Connects to https://github.com/ros2/rcl_logging/pull/7
Related to https://github.com/ros2/rcl_logging/pull/7

### DIFF
--- a/rcl/cmake/get_default_rcl_logging_implementation.cmake
+++ b/rcl/cmake/get_default_rcl_logging_implementation.cmake
@@ -24,13 +24,13 @@
 macro(get_default_rcl_logging_implementation var)
 
   # if logging implementation already specified or RCL_LOGGING_IMPLEMENTATION environment variable
-  # is set then use that, otherwise default to using rcl_logging_noop
+  # is set then use that, otherwise default to using rcl_logging_log4cxx
   if(NOT "${RCL_LOGGING_IMPLEMENTATION}" STREQUAL "")
     set(_logging_implementation "${RCL_LOGGING_IMPLEMENTATION}")
   elseif(NOT "$ENV{RCL_LOGGING_IMPLEMENTATION}" STREQUAL "")
     set(_logging_implementation "$ENV{RCL_LOGGING_IMPLEMENTATION}")
   else()
-    set(_logging_implementation rcl_logging_noop)
+    set(_logging_implementation rcl_logging_log4cxx)
   endif()
 
   # persist implementation decision in cache


nburek:
I've asked Thomas to kick off a CI build with this change here https://github.com/ros2/rcl_logging/pull/7

thomas-moulard:
CC @juanrh 
Is this gated on landing the ros1_bridge fix? (https://github.com/ros2/ros1_bridge/pull/174)

nburek:
> CC @juanrh
> Is this gated on landing the ros1_bridge fix? ([ros2/ros1_bridge#174](https://github.com/ros2/ros1_bridge/pull/174))

It should not be gated on that. The /rosout feature is already enabled by default, but can be disabled at runtime with a command line parameter. This change is swapping the default 3rd party logger from one that doesn't do anything to one that logs with log4cxx in addition to the stdout and /rosout handlers.

thomas-moulard:
SGTM, thanks for the clarification!

clalancette:
This change looks good to me, and it looks like the CI infrastructure is all setup for it.  I'd like to see green CI with this change before I approve (it looks like the build from https://github.com/ros2/rcl_logging/pull/7 is testing a different branch).

clalancette:
All right, ros2/rcl_logging#7 is merged and this is rebased.  Kicking off a full CI run; if that comes back green, I'll approve and merge:

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6885/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3168/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5656/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6708/)

clalancette:
So CI is still running, and it looks like most of the tests are passing.  However, I'm seeing a lot of spam in the logs in the style of:

```
01:01:01.122 14: log4cxx: setFile(/var/log/ros/11202.log,true) call failed.
01:01:01.122 14: log4cxx: IO Exception : status code = 2
01:01:01.128 14: log4cxx: setFile(/var/log/ros/11202.log,true) call failed.
```

@nburek @thomas-moulard Any ideas on why, and how we can fix that?

thomas-moulard:
@clalancette errno is "no such file or directory" - wild guess, are we allowed to write to `/var/log/ros` in the CI worker?

nburek:
Yes, as Thomas mentioned the most likely cause is not having permission to write to /var/log.

The easiest solution for this will be to change the default log directory

nuclearsandwich:
> Yes, as Thomas mentioned the most likely cause is not having permission to write to /var/log.
> 
> The easiest solution for this will be to change the default log directory

Would you suggest changing the default log location in source or via additional configuration in CI? ROS processes likely won't have permission to write to `/var/log` on development or production hosts without user intervention either which inclines me toward the former.

nburek:
Agreed that the better place to change it is in the source code. I didn't notice this during development since I did all my testing in a docker container as root.

nburek:
I put a quick change up for PR that changes the behavior on Linux to be the same as Windows, which is to log to a local directory.  https://github.com/ros2/rcl_logging/pull/9 

We could also look to change it so that it logs to "~/.ros/log/" by default like ROS1 does, but that's a more involved changed and I don't have the bandwidth to start that until next week.

clalancette:
CI for this and dependent PRs:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6970/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3227/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5717/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6791/)

clalancette:
All right, all CI is green!  This one, https://github.com/ros2/rcutils/pull/155, and https://github.com/ros2/rcl_logging/pull/9 are ready to go, so I'll switch them all to "in review".

nuclearsandwich:
Fresh CI

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6988/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3239/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5732/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6810/)

PR #428, opened 2019-05-01
Rmw preallocate

mjcarroll:
Replaces #383 

This PR is the changes #383, but with the actual test implementation removed, it just contains the relevant API changes and adjustments.

Connects to ros2/rmw#159

### DIFF
--- a/rcl/src/rcl/publisher.c
+++ b/rcl/src/rcl/publisher.c
@@ -38,7 +38,6 @@ typedef struct rcl_publisher_impl_t
   rmw_qos_profile_t actual_qos;
   rcl_context_t * context;
   rmw_publisher_t * rmw_handle;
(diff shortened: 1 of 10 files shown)


rcl/src/rcl/publisher.c
```diff
@@ -74,6 +75,8 @@ rcl_publisher_init(
   RCL_CHECK_ARGUMENT_FOR_NULL(topic_name, RCL_RET_INVALID_ARGUMENT);
   RCUTILS_LOG_DEBUG_NAMED(
     ROS_PACKAGE_NAME, "Initializing publisher for topic name '%s'", topic_name);
+
```
clalancette: Nit; I like adding one newline, two seems excessive :).

PR #430, opened 2019-05-03
Add an allocator to the external logging initialization.

clalancette:

In order to finish the implementation of https://github.com/ros2/rcl_logging/pull/9, we're going to need to allocate some memory.  In order to comply with the way the rest of the system does it, we are going to need to pass an allocator through to `rcl_logging_external_initialize`.  This PR (and the connected one in rcl_logging) stub that out; the implementation that actually uses the allocator will come later, as it will not be an API break.

### DIFF
--- a/rcl/include/rcl/logging_external_interface.h
+++ b/rcl/include/rcl/logging_external_interface.h
@@ -25,12 +25,14 @@
  *   logging library should use to configure itself.
  *   If no config file is provided this will be set to an empty string.
  *   Must be a NULL terminated c string.
- * \param[in] allocator The allocator to use for memory allocation.
+ * \param[in] allocator The allocator to use for memory allocation.  This is
+ *   an rcutils_allocator_t rather than an rcl_allocator_t to ensure that the
+ *   rcl_logging_* packages don't have a circular dependency back to rcl.
  * \todo TODO(clalancette) This API is marked RCL_PUBLIC, but is not built or
- *       exported from librcl.  Instead, these headers should be split into a
- *       separate package which is then depended on by both rcl and the
- *       rcl_logging_* implementations.  The duplicated headers from the
- *       implementations could then be removed.
+ *   exported from librcl.  Instead, these headers should be split into a
+ *   separate package which is then depended on by both rcl and the
+ *   rcl_logging_* implementations.  The duplicated headers from the
+ *   implementations could then be removed.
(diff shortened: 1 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6926/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3194/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5686/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6757/)

rcl/include/rcl/logging_external_interface.h
```diff
@@ -31,7 +31,7 @@
 RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_ret_t
-rcl_logging_external_initialize(const char * config_file);
+rcl_logging_external_initialize(const char * config_file, rcutils_allocator_t allocator);
```
tfoote: For semantics shouldn't this be an rcl_allocator_t? 

That's what's used internallly currently: https://github.com/ros2/rcl/blob/eabe42686688b652be0aa020db2b396c393536cc/rcl/src/rcl/logging.c#L41

Of course it's the same via typedef: https://github.com/ros2/rcl/blob/3b9475bb142bc6226b7441e5926eecb3f3dbf9d9/rcl/include/rcl/allocator.h#L25

wjwwood: We discussed offline about the fact that this needs to be deduplicated, but @clalancette is just going to make a TODO for now.

clalancette: The problem is that these packages shouldn't depend on `rcl`, as it creates a circular dependency.  Also, all of the functions I'm going to eventually call are in `rcutils`, so I think it makes more sense as an `rcutils_allocator_t`.

wjwwood: 👍 That actually is a good idea. Though they are just a `typedef` to one another right now.

clalancette:
Rebuilding just Linux to ensure I don't trip the linters.  If this succeeds, I'll merge both this and https://github.com/ros2/rcl_logging/pull/10 .

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6926/)

PR #432, opened 2019-05-03
Add new RCL_RET_UNSUPPORTED (for QoS)

emersonknapp:
Add a new rcl return type for unsupported features.
Unblocks ros2/rclpy#316

Signed-off-by: Emerson Knapp <eknapp@amazon.com>


mjcarroll:
Ah, this is fantastic!  I need to add this to some of the allocator work that I just added.  Thanks for the patch.

emersonknapp:
Thanks for the approvals, can we merge this? Passed CI on https://github.com/ros2/rmw_fastrtps/pull/276

PR #435, opened 2019-05-09
add explicit dependency on rcl_logging_log4cxx

dirk-thomas:
Makes it possible for the devel / PR jobs to pass, e.g. http://build.ros2.org/view/Ddev/job/Ddev__rcl__ubuntu_bionic_amd64/21/


dirk-thomas:
Similar to https://github.com/ros2/rmw_implementation/blob/7e88ba2afd6b6b0066c4f7b5fd6ed4aaa9d92667/rmw_implementation/package.xml#L14-L21

PR #436, opened 2019-05-14
Revert "Changes the default 3rd party logger from rcl_logging_noop to…

clalancette:
… rcl_logging_log4cxx (#425)"

This reverts commit ac8ee907405748684d1184ca4fd6257313f20a5c.

As discussed at length in https://github.com/ros2/build_cop/issues/189 and https://github.com/ros2/build_cop/issues/188, there are enough problems in the rcl_logging_log4cxx backend that we should switch back to noop for now and then revisit the issue immediately after dashing.

### DIFF
--- a/rcl/cmake/get_default_rcl_logging_implementation.cmake
+++ b/rcl/cmake/get_default_rcl_logging_implementation.cmake
@@ -24,13 +24,13 @@
 macro(get_default_rcl_logging_implementation var)
 
   # if logging implementation already specified or RCL_LOGGING_IMPLEMENTATION environment variable
-  # is set then use that, otherwise default to using rcl_logging_log4cxx
+  # is set then use that, otherwise default to using rcl_logging_noop
   if(NOT "${RCL_LOGGING_IMPLEMENTATION}" STREQUAL "")
     set(_logging_implementation "${RCL_LOGGING_IMPLEMENTATION}")
   elseif(NOT "$ENV{RCL_LOGGING_IMPLEMENTATION}" STREQUAL "")
     set(_logging_implementation "$ENV{RCL_LOGGING_IMPLEMENTATION}")
   else()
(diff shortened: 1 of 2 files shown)


clalancette:
CI running a "Debug" build, as that shows more of the problems:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7066/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3294/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5790/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6882/)

dirk-thomas:
#435 needs to be part of this revert too. And the noop package needs to be dependent on instead.

clalancette:
> #435 needs to be part of this revert too.

Good call, done in 1e86e7f

> And the noop package needs to be dependent on instead.

Sorry, I'm not sure what this bit means.  `rcl` depends on the `rcl_logging_packages` group, which `rcl_logging_noop` provides.  What else do we need to change?

dirk-thomas:
The package dependency of the default plugin needs to be explicit (same reason why it was added for log4cxx) since some parts of the pipeline don't consider the group dependencies.

clalancette:
> I think this is probably the best thing to do for the time. I would also like to see a package introduced to CI that will at least exercise the log4cxx build/run if we intend to switch it back on in the long term.

I think we should look at alternatives to log4cxx.  There are quite a few available, and several of them are much more actively maintained than log4cxx.  I did a prototype of an rcl logging backend using [spdlog](https://github.com/gabime/spdlog), which seemed to work OK in basic testing and is active.  I'm not suggesting we necessarily need to use that one, but we should look at our options.

dirk-thomas:
I will go ahead and merge this before the nightlies are started,..

PR #459, opened 2019-06-12
Add function for getting clients by node

jacobperron:
Connects to https://github.com/ros2/rmw/pull/179

Along with the related PRs, this changes the behavior of `rcl_get_service_names_and_types_by_node` such that it only returns a list of server servers. Previously, it returned a list for both service clients and service servers. The new function `rcl_get_client_names_and_types_by_node` returns a list of service clients.

### DIFF
--- a/rcl/include/rcl/graph.h
+++ b/rcl/include/rcl/graph.h
@@ -135,7 +135,7 @@ rcl_get_subscriber_names_and_types_by_node(
   const char * node_namespace,
   rcl_names_and_types_t * topic_names_and_types);
 
-/// Return a list of service names and types for associated with a node.
+/// Return a list of service names and types associated with a node.
 /**
  * The `node` parameter must point to a valid node.
  *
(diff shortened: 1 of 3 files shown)


rcl/include/rcl/graph.h
```diff
@@ -180,6 +180,51 @@ rcl_get_service_names_and_types_by_node(
...
+ * The `service_names_and_types` parameter must be allocated and zero initialized.
+ * This function allocates memory for the returned list of names and types and so it is the
+ * callers responsibility to pass `service_names_and_types` to rcl_names_and_types_fini()
+ * when it is no longer needed.
+ * Failing to do so will result in leaked memory.
+ *
+ * \see rcl_get_publisher_names_and_types_by_node for details on the `no_demangle` parameter.
+ *
+ * The returned names are not automatically remapped by this function.
+ * Attempting to create clients using names returned by this function may not
```
mjcarroll: Should this be "attempting to create service servers"?

mjcarroll: And vice-versa on the other doc?

jacobperron: Yeah that makes sense. Updated in da774b1

PR #478, opened 2019-08-01
Fix rosout handler visibility

alsora:
The `rcl/logging_rosout.c` file is not including its header `rcl/logging_rosout.h`.

For some reasons it still builds, but this has the side-effect of neglecting the visibility macros explicitly declared in the header file.

This PR partially address https://github.com/ros2/rcl/issues/476

Issue #476, opened 2019-07-29
Undefined reference to rcl_logging functions


alsora:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - source 
- Version or commit hash:
  - master
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
```
#include "rclcpp/rclcpp.hpp"
#include "rcl/logging_rosout.h"
#include <rcl/logging.h>
using namespace std::chrono_literals;

class LoggerNode : public rclcpp::Node
{
public:
  LoggerNode()
  : Node("logger_node")
  {
    //auto allocator = rcl_get_default_allocator();
    //rcl_ret_t status = rcl_logging_rosout_init(&allocator);
    rcutils_logging_set_output_handler(rcl_logging_rosout_output_handler);
   
    auto timer_callback =
      [this]() -> void {
        RCLCPP_INFO(this->get_logger(), "Hello World");
      };
    timer_ = this->create_wall_timer(500ms, timer_callback);
  }

private:
  rclcpp::TimerBase::SharedPtr timer_;
};

int main(int argc, char * argv[])
{
  rclcpp::init(argc, argv);
  rclcpp::spin(std::make_shared<LoggerNode>());
  rclcpp::shutdown();
  return 0;
}

```

#### Expected behavior

I create a node with a custom logging output handler (in this case it should only log to `rosout` topic).

#### Actual behavior

The above code doesn't build as I get a linker error:
```
undefined reference to `rcl_logging_rosout_output_handler'
```

This function is marked as `RCL_PUBLIC`, so it should be accessible.

Note that if I comment the following line in `rcl/CMakeLists.txt` the compilation succeeds.
```
rcl_set_symbol_visibility_hidden(${PROJECT_NAME} LANGUAGE "C")
```

#### Additional notes

On the other hand, the function `rcl_logging_rosout_init` is marked as `RCL_LOCAL`, thus it's hidden and I can't use it in my code.
Assuming that the `rcl_logging_rosout_output_handler` has to be accessible somehow, there should also be a way for checking if it has already been initialized and to initialize it.

This issue is linked to 

https://answers.ros.org/question/329532/ros2-how-to-programmatically-set-rosout-logging-handler/
https://discourse.ros.org/t/set-symbol-visibility-to-hidden-for-rmw-and-rcl-packages/7981/5?u=alsora

wjwwood:
[...]

PR #479, opened 2019-08-01
rosout init and fini marked as RCL_PUBLIC

alsora:
The functions for setting logger handlers are accessible from the `rcl` library, together with `rcl_logging_rosout_output_handler` (a bug in the visibility setting is addressed in https://github.com/ros2/rcl/pull/478).

However, the `rcl_logging_rosout_output_handler` is not usable from outside this library without access to the `init` and `fini` functions.

This PR addresses https://github.com/ros2/rcl/issues/476

### DIFF
--- a/rcl/include/rcl/logging_rosout.h
+++ b/rcl/include/rcl/logging_rosout.h
@@ -45,7 +45,7 @@ extern "C"
  * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
  * \return `RCL_RET_ERROR` if an unspecified error occurs.
  */
-RCL_LOCAL
+RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_ret_t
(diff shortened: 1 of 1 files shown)


nburek:
This makes sense to me. It was originally intend that you could reuse any of the individual logging handlers instead of the default composition of stdout, rosout, and external. We just missed the fact that you'd need the init and fini functions publicly accessible in order to do that. 

The rosout handlers should only be defined in the `rcl` package. Some of the external logger handlers were redefined in the individual logging packages in order to avoid a circular dependency and so that they could be publicly exported as well. That shouldn't impact the rosout handler.

alsora:
@wjwwood any updates on this?

nuclearsandwich:
CI for this together with #478 via an omnibus branch is linked here: https://github.com/ros2/rcl/pull/478#issuecomment-521663707

PR #480, opened 2019-08-02
Improve security error messages 

mikaelarguedas:
Signed-off-by: Mikael Arguedas <mikael.arguedas@gmail.com>

### DIFF
--- a/rcl/src/rcl/security_directory.c
+++ b/rcl/src/rcl/security_directory.c
@@ -247,7 +247,7 @@ char * rcl_get_secure_root(
     // Check node_secure_root is not NULL before checking directory
     if (NULL == node_secure_root) {
       RCL_SET_ERROR_MSG_WITH_FORMAT_STRING(
-        "SECURITY ERROR: unable to find a folder matching the node name '%s' in '%s%s'.\n"
+        "SECURITY ERROR: unable to find a folder matching the node name '%s' in '%s%s'. "
         "Lookup strategy: %s",
         node_name, ros_secure_root_env, node_namespace, lookup_strategy);
     } else {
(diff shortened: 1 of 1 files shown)


rcl/src/rcl/security_directory.c
```diff
@@ -247,12 +247,12 @@ char * rcl_get_secure_root(
     // Check node_secure_root is not NULL before checking directory
     if (NULL == node_secure_root) {
       RCL_SET_ERROR_MSG_WITH_FORMAT_STRING(
-        "SECURITY ERROR: unable to find a folder matching the node name in %s%s."
-        "Lookup strategy: %s",
-        ros_secure_root_env, node_namespace, lookup_strategy);
+        "SECURITY ERROR: unable to find a folder matching the node name '%s' in '%s%s'.\n"
+        "Lookup strategy: '%s'",
```
dirk-thomas: Since the `lookup_strategy` is already separated with the colon we usually don't quote it.

Same below.

rcl/src/rcl/security_directory.c
```diff
@@ -247,12 +247,12 @@ char * rcl_get_secure_root(
     // Check node_secure_root is not NULL before checking directory
     if (NULL == node_secure_root) {
       RCL_SET_ERROR_MSG_WITH_FORMAT_STRING(
-        "SECURITY ERROR: unable to find a folder matching the node name in %s%s."
+        "SECURITY ERROR: unable to find a folder matching the node name '%s' in '%s%s'.\n"
```
dirk-thomas: Sorry for another comment: how about just adding a trailing space instead of a newline? That keeps the lookup strategy on the same line as the error message.

mikaelarguedas: Yeah, I originally also printed the node namespace and name on a different line, hence the newline here.
Now that it's incorporated in the same line, the trailing space is better:+1:

PR #488, opened 2019-08-19
Allow get_node_names to return result in any order

eboasson:
The ``get_node_names`` test fails for Cyclone DDS because the order in which the results are returned by its RMW implementation is different from the order in which they were created. It looks to me like the test is too picky in requiring that order.

Therefore, I propose a change to allow them to be returned in any order. I've taken a rather simple approach that should be ok for 4 nodes but that doesn't scale and that does require some maintenance if the set of node ever changes. That seems to happen rather infrequently and so I don't think that would be a major problem, but there are other ways of going about it, too, obviously.

### DIFF
--- a/rcl/test/rcl/test_get_node_names.cpp
+++ b/rcl/test/rcl/test_get_node_names.cpp
@@ -107,16 +107,25 @@ TEST_F(CLASSNAME(TestGetNodeNames, RMW_IMPLEMENTATION), test_rcl_get_node_names)
     ss << node_names.data[i] << std::endl;
   }
   EXPECT_EQ(size_t(4), node_names.size) << ss.str();
-  EXPECT_STREQ(node1_name, node_names.data[0]);
-  EXPECT_STREQ(node2_name, node_names.data[1]);
-  EXPECT_STREQ(node3_name, node_names.data[2]);
-  EXPECT_STREQ(node4_name, node_names.data[3]);
-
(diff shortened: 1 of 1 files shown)


thomas-moulard:
`EXPECT_THAT` and `UnorderedElementsAre` from `gmock` is probably what you're looking for here instead of reimplementing the logic manually: https://github.com/google/googletest/blob/master/googlemock/docs/cheat_sheet.md

eboasson:
@thomas-moulard, those are pretty nice, thanks for the tip. I don't think they really do the right thing here because the nodes and namespaces are in separate arrays but are considered as pairs (at least that's what it seems to do, and that's what makes the most sense to me).

So I've rewritten it to build up a ``std::set`` of (name, namespace) pairs of expected and discovered nodes, and then simply use std::set equality. Hopefully that'll be an acceptable way, although arguably a table-driven rewrite would be better still as it would avoid the quadruplication creating nodes in the first place.

thomas-moulard:
I see, your implementation LGTM!

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7969/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3993/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6487/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7823/)

PR #496, opened 2019-09-12
Bring back deprecated CLI arguments

hidmic:
Follow-up after https://github.com/ros2/rcl/pull/495#discussion_r322366467. This pull request brings back the CLI as of Dashing, with added deprecation warnings upon use.

### DIFF
--- a/rcl/include/rcl/types.h
+++ b/rcl/include/rcl/types.h
@@ -101,7 +101,7 @@ typedef rmw_ret_t rcl_ret_t;
 /// Argument is not a valid parameter rule
 #define RCL_RET_INVALID_PARAM_RULE 1010
 /// Argument is not a valid log level rule
-#define RCL_RET_INVALID_LOG_LEVEL 1020
+#define RCL_RET_INVALID_LOG_LEVEL_RULE 1020
 
 // rcl event specific ret codes in 20XX
 /// Invalid rcl_event_t given return code.
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -152,6 +152,24 @@ _rcl_parse_log_level(
   rcl_allocator_t allocator,
   int * log_level);
 
+/// Parse an argument that may or may not be a log level rule.
+/**
+ * \param[in] arg the argument to parse
+ * \param[in] allocator an allocator to use
+ * \param[in,out] log_level parsed log level represented by `RCUTILS_LOG_SEVERITY` enum
+ * \return RCL_RET_OK if a valid log level was parsed, or
+ * \return RCL_RET_INVALID_LOG_LEVEL_RULE if the argument is not a valid rule, or
+ * \return RCL_RET_BAD_ALLOC if an allocation failed, or
+ * \return RLC_RET_ERROR if an unspecified error occurred.
+ * \deprecated to be removed in F-Turtle
+ */
+RCL_LOCAL
+rcl_ret_t
+_rcl_parse_log_level_rule(
+  const char * arg,
+  rcl_allocator_t allocator,
+  int * log_level);
+
 /// Parse an argument that may or may not be a log configuration file.
 /**
  * \param[in] arg the argument to parse
@@ -164,6 +182,23 @@ _rcl_parse_log_level(
 RCL_LOCAL
 rcl_ret_t
 _rcl_parse_external_log_config_file(
+  const char * arg,
+  rcl_allocator_t allocator,
+  char ** log_config_file);
+
+/// Parse an argument that may or may not be a log file rule.
+/**
+ * \param[in] arg the argument to parse
+ * \param[in] allocator an allocator to use
+ * \param[in,out] log_config_file parsed log configuration file
+ * \return RCL_RET_OK if a valid log config file was parsed, or
+ * \return RCL_RET_BAD_ALLOC if an allocation failed, or
+ * \return RLC_RET_ERROR if an unspecified error occurred.
+ * \deprecated to be removed in F-Turtle
+ */
+RCL_LOCAL
+rcl_ret_t
+_rcl_parse_external_log_config_file_rule(
   const char * arg,
   rcl_allocator_t allocator,
   char ** log_config_file);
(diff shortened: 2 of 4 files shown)


hidmic:
CI (up to `rcl`):

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/8119/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4110/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6611/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8004/)

rcl/include/rcl/types.h
```diff
@@ -101,7 +101,7 @@ typedef rmw_ret_t rcl_ret_t;
 /// Argument is not a valid parameter rule
 #define RCL_RET_INVALID_PARAM_RULE 1010
 /// Argument is not a valid log level rule
-#define RCL_RET_INVALID_LOG_LEVEL 1020
+#define RCL_RET_INVALID_LOG_LEVEL_RULE 1020
```
jacobperron: Do we need to revert this change too? I thought we essentially just wanted users to be able to provide the double-underscore versions; not sure that we need to maintain this particular API compatibility.
Alternatively, we can not compromise the new API by adding both the new and old definitions.

rcl/src/rcl/arguments.c
```diff
@@ -1444,9 +1653,32 @@ _rcl_parse_log_level(
     return RCL_RET_OK;
   }
   RCL_SET_ERROR_MSG("Argument does not use a valid severity level");
-  return RCL_RET_INVALID_LOG_LEVEL;
+  return RCL_RET_ERROR;
```
jacobperron: I think we should undo this change (see my comment above).

rcl/src/rcl/arguments.c
```diff
@@ -1444,9 +1653,32 @@ _rcl_parse_log_level(
...
 }
 
+rcl_ret_t
+_rcl_parse_log_level_rule(
```
jacobperron: Might be worth adding a TODO or \deprecated marker to remove this function in F-turtle (same for others).

rcl/src/rcl/arguments.c
```diff
@@ -1612,28 +1895,75 @@ _rcl_parse_disabling_flag(
   RCL_CHECK_ARGUMENT_FOR_NULL(suffix, RCL_RET_INVALID_ARGUMENT);
   RCL_CHECK_ARGUMENT_FOR_NULL(disable, RCL_RET_INVALID_ARGUMENT);
 
-  const char * true_prefix = "--enable-";
-  const size_t true_prefix_len = strlen(true_prefix);
+  const size_t enable_prefix_len = strlen(RCL_ENABLE_FLAG_PREFIX);
   if (
-    strncmp(true_prefix, arg, true_prefix_len) == 0 &&
-    strcmp(suffix, arg + true_prefix_len) == 0)
+    strncmp(RCL_ENABLE_FLAG_PREFIX, arg, enable_prefix_len) == 0 &&
```
jacobperron: nitpick: While this change is good, it would have been better as a separate commit (for future reference).

rcl/src/rcl/arguments.c
```diff
@@ -310,13 +404,13 @@ rcl_parse_arguments(
             prev_error_string.str);
         } else {
           RCL_SET_ERROR_MSG_WITH_FORMAT_STRING(
-            "Couldn't parse trailing parameter override rule: '%s'. No rule found.", argv[i]);
+            "Couldn't parse trailing %s flag. No parameter override rule found.", argv[i]);
```
jacobperron: nitpick: I guess the changes to this error message and log messages are unrelated and could be a separate commit (or PR).

rcl/src/rcl/arguments.c
```diff
@@ -1444,9 +1653,32 @@ _rcl_parse_log_level(
...
+  rcutils_ret_t ret = rcutils_logging_severity_level_from_string(
+    arg + strlen(RCL_LOG_LEVEL_ARG_RULE), allocator, log_level);
+  if (RCUTILS_RET_OK == ret) {
+    return RCL_RET_OK;
+  }
+  RCL_SET_ERROR_MSG("Argument does not use a valid severity level");
+  return RCL_RET_INVALID_LOG_LEVEL_RULE;
+}
+
+
```
jacobperron: nit: remove extra vertical space.

rcl/src/rcl/arguments.c
```diff
@@ -1566,6 +1798,7 @@ _rcl_parse_param_rule(
   return ret;
 }
 
+RCL_LOCAL
```
jacobperron: Seems unrelated to the PR title.

rcl/include/rcl/types.h
hidmic: I don't have a strong opinion on this, mainly because these return codes are not being used by the public API -- which raises the question as to why these are exposed in the first place. All `RCL_RET_*_RULE` return codes are used internally only. 

The sole reason why `RCL_RET_INVALID_LOG_LEVEL_RULE` was switched to `RCL_RET_INVALID_LOG_LEVEL` is that the former no longer made sense with the rule-like arguments gone.

Moving forward, we can (1) go back and find another return code for the now deprecated log level rule parsing function, (2) stay as-is, as there's no actual need for a special return code or (3) keep both but outside our public API.

rcl/src/rcl/arguments.c
hidmic: I've added so in each forward declaration, but didn't duplicate it here. Would you do that twice?

rcl/src/rcl/arguments.c
hidmic: There was no need to turn these literals into preprocessor variables until I had to reuse them to provide hints in warnings, thus why they are here. Having said, sure, we can split them out.

rcl/src/rcl/arguments.c
hidmic: Sure.

rcl/include/rcl/types.h
jacobperron: If it's not necessary to be public, then I'm fine leaving it as-is (returning RCL_RET_ERROR).

rcl/src/rcl/arguments.c
jacobperron: It's fine, I missed it.

hidmic:
Alright, @jacobperron @ivanpauno anything else you'd like to suggest? Otherwise I'll go ahead and merge.

hidmic:
Reproducing the one above, CI (up to `rcl`):

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/8132/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4120/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6622/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8015/)

hidmic:
Merging!

rcl/src/rcl/arguments.c
```diff
@@ -492,6 +587,120 @@ rcl_parse_arguments(
         continue;
       }
 
+      // Attempt to parse argument as remap rule in its deprecated form
```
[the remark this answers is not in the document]
hidmic: See https://github.com/ros2/rcl/issues/500.

PR #492, opened 2019-08-29
Add specific return code for non existent node

ivanpauno:
Addresses https://github.com/ros2/ros2cli/pull/322#issuecomment-525364023

PR #499, opened 2019-09-13
Add node name and namespace validation to graph functions

ivanpauno:
Taken from https://github.com/ros2/rcl/pull/492.
I've split the PR to avoid merging unrelated changes together (though I need this change for #492).

### DIFF
--- a/rcl/test/rcl/test_graph.cpp
+++ b/rcl/test/rcl/test_graph.cpp
@@ -274,7 +274,7 @@ TEST_F(
   rcl_allocator_t zero_allocator = static_cast<rcl_allocator_t>(
     rcutils_get_zero_initialized_allocator());
   rcl_node_t zero_node = rcl_get_zero_initialized_node();
-  const char * unknown_node_name = "/test_rcl_get_publisher_names_and_types_by_node";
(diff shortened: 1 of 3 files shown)


rcl/test/rcl/test_graph.cpp
```diff
@@ -274,7 +274,7 @@ TEST_F(
   rcl_allocator_t zero_allocator = static_cast<rcl_allocator_t>(
     rcutils_get_zero_initialized_allocator());
   rcl_node_t zero_node = rcl_get_zero_initialized_node();
-  const char * unknown_node_name = "/test_rcl_get_publisher_names_and_types_by_node";
+  const char * unknown_node_name = "test_rcl_get_publisher_names_and_types_by_node";
```
jacobperron: Is this change necessary? I believe the test is for an *unknown* node, not an invalid node name.
Same for similar changes to below.

ivanpauno: Node names can't start with `/`, namespaces can.
If I left the `/` at the start, I will get an invalid name error, and that's not the idea.

jacobperron: Ok, got it.

PR #504, opened 2019-09-19
Fix rcl_action test_graph

ivanpauno:
Fixes https://github.com/ros2/build_cop/issues/241
Fixes https://github.com/ros2/build_cop/issues/213 (it was extremely flake, but I reproduced it locally)

### DIFF
--- a/rcl_action/test/rcl_action/test_graph.cpp
+++ b/rcl_action/test/rcl_action/test_graph.cpp
@@ -51,6 +51,7 @@ public:
   rcl_node_t zero_node;
   const char * test_graph_node_name = "test_action_graph_node";
   const char * test_graph_old_node_name = "test_action_graph_old_node_name";
+  const char * test_graph_unknown_node_name = "test_action_graph_unknown_node_name";
 
   void SetUp()
   {
@@ -132,17 +133,14 @@ TEST_F(
   rcl_reset_error();
   // Invalid node name
   ret = rcl_action_get_client_names_and_types_by_node(
-    &this->node, &this->allocator, "_test_this_Isnot_a_valid_name", "", &nat);
-  EXPECT_EQ(RCL_RET_ERROR, ret) << rcl_get_error_string().str;
+    &this->node, &this->allocator, "_!test_this_is_not_a_valid_name", "", &nat);
+  EXPECT_EQ(RCL_RET_NODE_INVALID_NAME, ret) << rcl_get_error_string().str;
   rcl_reset_error();
   // Non-existent node
-  // Note, Opensplice successfully reports graph information about finalized nodes
-  if (!is_opensplice) {
-    ret = rcl_action_get_client_names_and_types_by_node(
-      &this->node, &this->allocator, this->test_graph_old_node_name, "", &nat);
-    EXPECT_EQ(RCL_RET_ERROR, ret) << rcl_get_error_string().str;
-    rcl_reset_error();
-  }
+  ret = rcl_action_get_client_names_and_types_by_node(
+    &this->node, &this->allocator, this->test_graph_unknown_node_name, "", &nat);
+  EXPECT_EQ(RCL_RET_NODE_NAME_NON_EXISTENT, ret) << rcl_get_error_string().str;
+  rcl_reset_error();
   // Invalid names and types
   ret = rcl_action_get_client_names_and_types_by_node(
     &this->node, &this->allocator, this->test_graph_node_name, "", nullptr);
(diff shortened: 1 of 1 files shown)


rcl_action/test/rcl_action/test_graph.cpp
```diff
@@ -132,17 +133,14 @@ TEST_F(
   rcl_reset_error();
   // Invalid node name
   ret = rcl_action_get_client_names_and_types_by_node(
-    &this->node, &this->allocator, "_test_this_Isnot_a_valid_name", "", &nat);
-  EXPECT_EQ(RCL_RET_ERROR, ret) << rcl_get_error_string().str;
+    &this->node, &this->allocator, "_!test_this_is_not_a_valid_name", "", &nat);
```
jacobperron: Is the previous value , "_test_this_Isnot_a_valid_name", now valid?

rcl_action/test/rcl_action/test_graph.cpp
```diff
@@ -132,17 +133,14 @@ TEST_F(
...
   // Non-existent node
-  // Note, Opensplice successfully reports graph information about finalized nodes
-  if (!is_opensplice) {
-    ret = rcl_action_get_client_names_and_types_by_node(
-      &this->node, &this->allocator, this->test_graph_old_node_name, "", &nat);
-    EXPECT_EQ(RCL_RET_ERROR, ret) << rcl_get_error_string().str;
-    rcl_reset_error();
-  }
+  ret = rcl_action_get_client_names_and_types_by_node(
+    &this->node, &this->allocator, this->test_graph_unknown_node_name, "", &nat);
```
jacobperron: This is now testing a different case than before. Before It was testing that a finalized node is no longer exists.
I think this change is fine since according the comment what we really want to test for an unknown node. But perhaps the previous test is also valid and worth keeping in addition to the new check? What do you think?

rcl_action/test/rcl_action/test_graph.cpp
ivanpauno: Yes, it was always valid.
Before, the error code for invalid node name and for nonexistent nodes was the same, so the test was passing.
But, AFAIK, caps were always accepted as topic name: https://github.com/ros2/design/pull/241.

jacobperron: Ah, I see. Thanks.

rcl_action/test/rcl_action/test_graph.cpp
```diff
@@ -132,17 +133,14 @@ TEST_F(
...
   // Invalid node name
   ret = rcl_action_get_client_names_and_types_by_node(
-    &this->node, &this->allocator, "_test_this_Isnot_a_valid_name", "", &nat);
-  EXPECT_EQ(RCL_RET_ERROR, ret) << rcl_get_error_string().str;
+    &this->node, &this->allocator, "_!test_this_is_not_a_valid_name", "", &nat);
+  EXPECT_EQ(RCL_RET_NODE_INVALID_NAME, ret) << rcl_get_error_string().str;
   rcl_reset_error();
   // Non-existent node
-  // Note, Opensplice successfully reports graph information about finalized nodes
-  if (!is_opensplice) {
```
jacobperron: If we aren't keeping this particular test, then we can remove the definition of `is_opensplice` near the top of the file.

rcl_action/test/rcl_action/test_graph.cpp
ivanpauno: This is a flaky test, because discovery may be slow after you delete a node, and the other nodes can think it's still there.
We should definitely have system level test for the graph API, that creates/deletes nodes and queries info,

But trying to do that at unit test level is quite flaky. The reason the test was skipped in OpenSplice is'nt that they don't report node finalizing data, but they are just slower.

https://github.com/ros2/build_cop/issues/213 is reporting that this test is flaky.

----

Instead of checking for an old node, I'm checking for one that doesn't exist, which is a similar case.
We're just doing that in `rcl test_graph`, and we don't check what happens with deleted nodes.

jacobperron: I see. The test would have to be restructured so that we wait until we get the expected value or timeout. I'm okay with leaving that for future work (possibly as a  system level test as you proprose).

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8183/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4182/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6665/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8061/)

PR #509, opened 2019-09-25
free valid_transitions for all states

vmayoral:
transition_map->states[i].valid_transitions wasn't being release which was
leaking about 800 bytes per instantiation.

See https://github.com/aliasrobotics/RVD/issues/333. There's a pending quality-related leak in case someone wants to look at it.


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8220/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4193/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6701/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8104/)

clalancette:
@vmayoral The test failures in the Linux CI all seem new with this CI; they didn't happen in the nightlies: https://ci.ros2.org/view/nightly/job/nightly_linux_debug/1325/ .  Thoughts?

vmayoral:
Our dynamic testing is running on top of the latest [ros2 repos](https://github.com/ros2/ros2/blob/master/ros2.repos), I believe that corresponds with your nightly builds though I'm not sure.

This flaw was detected using Google Sanitizers thereby I don't see any reason not to appear in your builds as well. It should IMHO.

As for the bug, I started researching this because it was leaking 1MB+ on initialization. It looked like a potential vulnerability but I don't believe it now since it'd be hard to exploit beyond forcing additional initializations. Memory should be released though. The patch is just a proposal but feel free to change it if you guys would prefer to address this differently.

clalancette:
> This flaw was detected using Google Sanitizers thereby I don't see any reason not to appear in your builds as well. It should IMHO.

I totally agree; this seems like a bug.  And your fix seems reasonable.  But it is causing test failures when I ran it through CI, so I think this needs a closer look.

rcl_lifecycle/src/transition_map.c
```diff
@@ -55,6 +55,12 @@ rcl_lifecycle_transition_map_fini(
 {
   rcl_ret_t fcn_ret = RCL_RET_OK;
 
+  // free valid transitions for all states
+  for (unsigned int i = 0; i < transition_map->states_size; ++i) {
+      if (transition_map->states[i].valid_transitions != NULL) {
+          allocator->deallocate(transition_map->states[i].valid_transitions, allocator->state);
```
jacobperron: I'm not sure if it will resolve the test issues, but we should probably set the address to `NULL` after deallocating:

```c
transition_map->states[i].valid_transitions = NULL;
```

clalancette: Good call.  I did that in 05a6e8a, along with fixing the minor style issue.  I'll run another batch of CI here to see if that made a difference.

clalancette:
New CI: * Linux [Build Status](http://ci.ros2.org/job/ci_linux/8247/)

dirk-thomas:
@vmayoral @clalancette What is the status of this patch? On which side is the ball to follow up on this?

vmayoral:
@dirk-thomas It's up to you guys.

dirk-thomas:
Latest CI: [Build Status](https://ci.ros2.org/job/ci_linux/8705/)

@vmayoral Please rebase the branch against the latest state of the target branch in order to run CI.

vmayoral:
Follows at https://github.com/ros2/rcl/pull/537

PR #186, opened 2017-11-20
Switch rcl_lifecyle to ament_cmake_ros

dhood:
connects to ros2/ros2#306
parallel of https://github.com/ros2/rcl/pull/93

I am interested in this to get access to `ROS_PACKAGE_NAME` for logging (see https://github.com/ros2/ament_cmake_ros/pull/2#issuecomment-345107445)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3571/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/768/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2911/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3664/)

Issue #82, opened 2016-09-24
Reasons intra/interprocess message transport isn't implemented transparent in rcl


firesurfer:
After having a look at the implementation of the rclcpp I was wondering why there are special mechanisms implemented in the rclcpp for interprocess message transport and intraprocess message transport (there's an extra intra_process_manager)?  Wouldn't it be more convenient to have these mechanism transparently implemented in the rcl? By transparent I mean: The client lib on top of the rcl doesn't have to care how the message is transported.

wjwwood:
One reason is historical, we wanted to prototype intra-process communications before `rcl` was in place. At that time `rclcpp` called to `rmw` directly.

Another reason is that the intra-process communication makes heavy use of smart pointer semantics to help manage ownership of data. While it's not impossible to do this in C, it would be a bit more complicated.

Finally, if we move intra-process communications into `rcl` we need to sort out how to handle intra-process plus intra-language communications. Currently `rcl` doesn't know or care what types of objects are being passed through the `publish` and `take` calls, but if it starts holding those to deliver them to peers in the same process then it needs to know that the incoming data type is the right thing to return to the matching subscriber. For example, if you publish from a C++ program and store the C++ message for intra-process delivery but the subscriber is a C or Jave or Python program, then it wouldn't make sense to give it the C++ message, as a copy or the original. So then you need to convert from C++ to C or PyObject or whatever and need to know how to do that. If you have no conversion, then you need to first serialize the C++ object and then deserialize it in the type you want. All of this stuff about knowing when and how to convert between types is also associated with type masquerading.

wjwwood:
Also, the intra-process communications have to deal with multi-threading concerns, and up till now we've managed to avoid using threads or thread synchronization primitives in `rcl`, preferring to leave those things up to the client library implementations since each language has it's own mechanisms for dealing with that.

gbiggs:
It does sound like the use case for multi-language intra-process communication is too small to justify the complexity it would introduce.

clalancette:
@wjwwood @firesurfer Given this (old) discussion, should we close this ticket out?

firesurfer:
From my point of view it is still up to date because I still think it would be very useful for other language bindings, but regarding the extra work this would cause I think it's okey to close this ticket.

PR #72, opened 2016-08-22
Changed RCL_RET_SUBSCRIPTION_TAKE_FAILED to 401

firesurfer:
Changed RCL_RET_SUBSCRIPTION_TAKE_FAILED to 401 instead of 501. Seems to be a copy&paste error: https://github.com/ros2/rcl/issues/71

PR #511, opened 2019-10-04
Implement functions to get publisher and subcription informations like QoS policies from topic name

jaisontj:
**NOTE: DO NOT MERGE until [rmw #186](https://github.com/ros2/rmw/pull/186) and [rmw_implementation #72](https://github.com/ros2/rmw_implementation/pull/72) are merged.**

This PR makes the necessary changes to implement [this](https://github.com/ros2/rmw/issues/151) feature request. The `RCL` layer needs to expose functions to retrieve information about publishers and subscriptions to a topic. 

Summary of changes:
- Added and implemented functions `rcl_get_publishers_info_by_topic` and `rcl_get_subscriptions_info_by_topic`
- Wrote tests for the in-variants to these functions.

Related to issues in aws-roadmap [#82](https://github.com/ros-security/aws-roadmap/issues/82)


jaisontj:
@ivanpauno

rcl/test/rcl/test_graph.cpp
```diff
@@ -63,6 +63,9 @@ class CLASSNAME (TestGraphFixture, RMW_IMPLEMENTATION) : public ::testing::Test
   rcl_wait_set_t * wait_set_ptr;
   const char * test_graph_node_name = "test_graph_node";
 
+  rmw_topic_info_array_t * topic_info_array;
+  const char * topic_name = "valid_topic_name";
```
zmichaels11: nit: 
you probably want:
```c
const char * const topic_name = "valid_topic_name";
```
since that's a "constant pointer to constant chars"

rcl/test/rcl/test_graph.cpp
```diff
@@ -101,6 +104,12 @@ class CLASSNAME (TestGraphFixture, RMW_IMPLEMENTATION) : public ::testing::Test
     ret = rcl_wait_set_init(
       this->wait_set_ptr, 0, 1, 0, 0, 0, 0, this->context_ptr, rcl_get_default_allocator());
     ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+
+    rmw_topic_info_array_t valid_topic_info_array = {
```
zmichaels11: Any reason you didn't do:
```c
valid_topic_info_array = {0};
```
since that will init all values in the struct to 0.

rcl/test/rcl/test_graph.cpp
```diff
@@ -101,6 +104,12 @@ class CLASSNAME (TestGraphFixture, RMW_IMPLEMENTATION) : public ::testing::Test
...
+      0,     /*count*/
+      nullptr     /*info_array*/
+    };
+    this->topic_info_array = &valid_topic_info_array;
```
zmichaels11: you can't do this.
`valid_topic_info_array` is stack allocated and you're leaking a pointer of it to somewhere else.

ideally you'd do 
```c++
this->topic_info_array = std::make_unique<rmw_topic_info_array_t>();
```
However this requires `topic_info_array` to be a `unique_ptr`.
If you **must** use a raw pointer here, use `new`, however the class should probably be changed to use a `unique_ptr` some time in the future.

rcl/test/rcl/test_graph.cpp
```diff
@@ -1321,3 +1330,172 @@ TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_rcl_service_server_
...
+
+/*
+ * This does not test content of the response.
+ * It only tests if the return code is the one expected.
+ */
+TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION),
+  test_rcl_get_publishers_info_by_topic_null_node)
+{
+  rcl_allocator_t allocator = rcl_get_default_allocator();
+  auto ret = rcl_get_publishers_info_by_topic(nullptr,
```
zmichaels11: nit: `const auto ret = ...`

`const` is not really needed, however it indicates to reviewers that it is not reassigned later inside the function.

rcl/test/rcl/test_graph.cpp
```diff
@@ -1321,3 +1330,172 @@ TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_rcl_service_server_
...
+}
+
+/*
+ * This does not test content of the response.
+ * It only tests if the return code is the one expected.
+ */
+TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION),
+  test_rcl_get_subscriptions_info_by_topic_null_node)
+{
+  rcl_allocator_t allocator = rcl_get_default_allocator();
```
zmichaels11:
```c++
auto allocator = rcl_get_default_allocator();
const auto ret = ...
```

same as above; `const` is not really needed on `ret`, however it indicates to the reviewer that it is not reassigned later inside the function.
`allocator` cannot be `const` unless it and its fields can be accessed as `const`.

rcl/test/rcl/test_graph.cpp
```diff
@@ -1321,3 +1330,172 @@ TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_rcl_service_server_
...
+/*
+ * This does not test content of the response.
+ * It only tests if the return code is the one expected.
+ */
+TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION),
+  test_rcl_get_publishers_info_by_topic_invalid_participants)
+{
+  // this participant is invalid as the pointer "participants" inside is expected to be null.
+  this->topic_info_array->info_array =
+    reinterpret_cast<rmw_topic_info_t *>(malloc(sizeof(rmw_topic_info_t *)));
```
zmichaels11: is this correct?
Its allocating a pointer of size (rmw_topic_info_t pointer size).
Did you mean:
```c++
reinterpret_cast<rmw_topic_info_t *>(malloc(sizeof(rmw_topic_info_t)));
```

This is a personal nit, but I'd use `calloc` for tests since it zeroes out the allocated memory where as `malloc` is allowed to allocate chunks of garbage. 

I'd say the performance cost is worth consistency, especially in tests.

rcl/test/rcl/test_graph.cpp
jaisontj: I learnt that this was possible from your previous review. I had implemented this before that review and kinda missed this one, gah! Thanks for this though!

rcl/test/rcl/test_info_by_topic.cpp
```diff
@@ -0,0 +1,275 @@
...
+}
+
+/*
+ * This does not test content of the response.
+ * It only tests if the return code is the one expected.
+ */
+TEST_F(CLASSNAME(TestInfoByTopicFixture, RMW_IMPLEMENTATION),
+  test_rcl_get_publishers_info_by_topic_invalid_participants)
+{
+  // this participant is invalid as the pointer "participants" inside is expected to be null.
```
ivanpauno: nit: update the comment

rcl/test/rcl/test_info_by_topic.cpp
```diff
@@ -0,0 +1,275 @@
...
+  temp_info_array->info_array =
+    reinterpret_cast<rmw_topic_info_t *>(calloc(1, sizeof(rmw_topic_info_t)));
+  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT({
+    free(temp_info_array->info_array);
+  });
+  rcl_allocator_t allocator = rcl_get_default_allocator();
+  const auto ret = rcl_get_subscriptions_info_by_topic(this->node_ptr,
+      &allocator, this->topic_name, false, temp_info_array);
+  EXPECT_EQ(RCL_RET_INVALID_ARGUMENT, ret);
+}
```
ivanpauno: It would be good to add some actual testing of the response

rcl/test/rcl/test_info_by_topic.cpp
```diff
@@ -0,0 +1,378 @@
...
+  subscription_options.qos = default_qos_profile;
+  ret = rcl_subscription_init(
+    &subscription,
+    this->node_ptr,
+    ts,
+    this->topic_name,
+    &subscription_options);
+  ASSERT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
+  std::string fqdn = this->topic_name;
+  fqdn = "/" + fqdn;
```
hidmic: @jaisontj nit: can be collapsed into the line above and made `const`

rcl/test/rcl/test_info_by_topic.cpp
```diff
@@ -0,0 +1,378 @@
...
+
+    ret = rcl_init(0, nullptr, &init_options, this->context_ptr);
+    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+    this->node_ptr = new rcl_node_t;
+    *this->node_ptr = rcl_get_zero_initialized_node();
+    const char * name = "test_graph_node";
+    ret = rcl_node_init(this->node_ptr, name, "", this->context_ptr, &node_options);
+    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+
+    this->topic_info_array = std::make_unique<rmw_topic_info_array_t>();
```
hidmic: @jaisontj again, why does it have to heap allocated? Also, if used, `std::unique_ptr` should have an associated `Deleter` to properly finalize the allocated `rmw_topic_info_array_t`.

rcl/test/rcl/test_info_by_topic.cpp
```diff
@@ -0,0 +1,378 @@
...
+    this->old_context_ptr = new rcl_context_t;
+    *this->old_context_ptr = rcl_get_zero_initialized_context();
+    ret = rcl_init(0, nullptr, &init_options, this->old_context_ptr);
+    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+    this->old_node_ptr = new rcl_node_t;
+    *this->old_node_ptr = rcl_get_zero_initialized_node();
+    const char * old_name = "old_node_name";
+    rcl_node_options_t node_options = rcl_node_get_default_options();
+    ret = rcl_node_init(this->old_node_ptr, old_name, "", this->old_context_ptr, &node_options);
+    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
```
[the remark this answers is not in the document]
mm318: Isn't the code already doing that? Or do you mean `rcl_node_fini()` should always be called?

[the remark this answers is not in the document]
mm318: I'll add `rcl_shutdown(&this->old_context)` to `TearDown()`.

rcl/test/rcl/test_info_by_topic.cpp
```diff
@@ -0,0 +1,378 @@
...
+
+    this->context_ptr = new rcl_context_t;
+    *this->context_ptr = rcl_get_zero_initialized_context();
+
+    ret = rcl_init(0, nullptr, &init_options, this->context_ptr);
+    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+    this->node_ptr = new rcl_node_t;
+    *this->node_ptr = rcl_get_zero_initialized_node();
+    const char * name = "test_graph_node";
+    ret = rcl_node_init(this->node_ptr, name, "", this->context_ptr, &node_options);
```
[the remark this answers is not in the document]
mm318: `rcl_shutdown(&this->context)` will be called in `TearDown()` if `rcl_node_init()` fails in `SetUp()`.

rcl/test/rcl/test_info_by_topic.cpp
```diff
@@ -0,0 +1,378 @@
...
+ * It only tests if the return code is the one expected.
+ */
+TEST_F(CLASSNAME(TestInfoByTopicFixture, RMW_IMPLEMENTATION),
+  test_rcl_get_subscriptions_info_by_topic_invalid_participants)
+{
+  // temp_info_array is invalid because it is expected to be zero initialized
+  // and the info_array variable inside it is expected to be null.
+  const auto & temp_info_array = this->topic_info_array.get();
+  temp_info_array->info_array =
+    reinterpret_cast<rmw_topic_info_t *>(calloc(1, sizeof(rmw_topic_info_t)));
```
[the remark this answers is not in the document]
mm318: I think it's for guaranteeing that all bits are initialized to 0.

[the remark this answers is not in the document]
mm318: Though I do think this should be a `static_cast` instead of a `reinterpret_cast`.

rcl/include/rcl/graph.h
```diff
@@ -524,6 +525,120 @@ rcl_count_subscribers(
...
+ *
+ * It is the responsibility of the caller to ensure that `publishers_info` parameter points
+ * to a valid struct of type rmw_topic_info_array_t. The `count` field inside the struct
+ * must be set to 0 and the `info_array` field inside the struct must be set to null.
+ * @see rmw_get_zero_initialized_topic_info_array
+ *
+ * The `allocator` will be used to allocate memory to the `info_array` member
+ * inside of `publishers_info`.
+ * Moreover, every const char * member inside of
+ * rmw_topic_info_t will be allocated memory and assigned a copied value.
```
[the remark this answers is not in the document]
mm318: Done.

rcl/test/rcl/test_info_by_topic.cpp
```diff
@@ -0,0 +1,378 @@
...
+# define CLASSNAME(NAME, SUFFIX) NAME
+#endif
+
+class CLASSNAME (TestInfoByTopicFixture, RMW_IMPLEMENTATION) : public ::testing::Test
+{
+public:
+  rcl_context_t * old_context_ptr;
+  rcl_context_t * context_ptr;
+  rcl_node_t * old_node_ptr;
+  rcl_node_t * node_ptr;
```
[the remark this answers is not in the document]
mm318: Made these member into values, instead of pointers.

rcl/test/rcl/test_info_by_topic.cpp
mm318: Done.

rcl/test/rcl/test_info_by_topic.cpp
mm318: Made this member into a value instead of a `unique_ptr`.

rcl/include/rcl/graph.h
```diff
@@ -524,6 +525,120 @@ rcl_count_subscribers(
...
+/**
+ * The `node` parameter must point to a valid node.
+ *
+ * The `topic_name` parameter must not be `NULL`.
+ *
+ * The `no_mangle` parameter determines if the provided topic_name should be
+ * expanded to its fully qualified name.
+ *
+ * It is the responsibility of the caller to ensure that `publishers_info` parameter points
+ * to a valid struct of type rmw_topic_info_array_t. The `count` field inside the struct
```
hidmic: @mm318 nit: one sentence per line

rcl/include/rcl/graph.h
```diff
@@ -524,6 +525,120 @@ rcl_count_subscribers(
...
+/**
+ * The `node` parameter must point to a valid node.
+ *
+ * The `topic_name` parameter must not be `NULL`.
+ *
+ * The `no_mangle` parameter determines if the provided topic_name should be
+ * expanded to its fully qualified name.
+ *
+ * It is the responsibility of the caller to ensure that `subscriptions_info` parameter points
+ * to a valid struct of type rmw_topic_info_array_t. The `count` field inside the struct
```
hidmic: @mm318 nit: one sentence per line

rcl/include/rcl/graph.h
```diff
@@ -524,6 +525,120 @@ rcl_count_subscribers(
...
+ * Uses Atomics       | No
+ * Lock-Free          | Maybe [1]
+ * <i>[1] implementation may need to protect the data structure with a lock</i>
+ *
+ * \param[in] node the handle to the node being used to query the ROS graph
+ * \param[in] allocator allocator to be used when allocating space for
+ *            the array inside publishers_info
+ * \param[in] topic_name the name of the topic in question
+ * \param[in] no_mangle if true, the topic_name will be expanded to its fully qualified name.
+ * \param[out] subscriptions_info a struct representing a list of subscriptions information.
```
hidmic: @mm318 here and elsewhere, sometimes there's a trailing period, sometimes there isn't. Consider normalizing punctuation.

rcl/src/rcl/graph.c
```diff
@@ -375,6 +377,81 @@ rcl_count_subscribers(
...
+  rmw_ret = get_topic_info(
+    rcl_node_get_rmw_handle(node),
+    allocator,
+    topic_name,
+    no_mangle,
+    info_array);
+  if (rmw_ret != RMW_RET_OK) {
+    error_string = rmw_get_error_string();
+    rmw_reset_error();
+    RCL_SET_ERROR_MSG(error_string.str);
```
hidmic: @mm318 I think you don't have to re-set the error message, though it makes me wonder why `rcl` and `rmw` share error code and message storage through `rcutils`. @wjwwood @jacobperron thoughts?

rcl/test/rcl/test_info_by_topic.cpp
```diff
@@ -0,0 +1,370 @@
...
+/*
+ * This does not test content of the response.
+ * It only tests if the return code is the one expected.
+ */
+TEST_F(CLASSNAME(TestInfoByTopicFixture, RMW_IMPLEMENTATION),
+  test_rcl_get_publishers_info_by_topic_invalid_participants)
+{
+  // temp_info_array is invalid because it is expected to be zero initialized
+  // and the info_array variable inside it is expected to be null.
+  const auto & temp_info_array = &this->topic_info_array;
```
hidmic: @mm318 a const ref to a pointer? Consider simply using `&this->topic_info-array` where needed.

rcl/test/rcl/test_info_by_topic.cpp
```diff
@@ -0,0 +1,370 @@
...
+/*
+ * This does not test content of the response.
+ * It only tests if the return code is the one expected.
+ */
+TEST_F(CLASSNAME(TestInfoByTopicFixture, RMW_IMPLEMENTATION),
+  test_rcl_get_subscriptions_info_by_topic_invalid_participants)
+{
+  // temp_info_array is invalid because it is expected to be zero initialized
+  // and the info_array variable inside it is expected to be null.
+  const auto & temp_info_array = &this->topic_info_array;
```
hidmic: @mm318 same as above.

rcl/test/rcl/test_info_by_topic.cpp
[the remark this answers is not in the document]
hidmic: Thanks. BTW `rcl_context_is_valid` may help to know whether to shutdown or not. Either that or a comment explaining why we do not assert on its return value in `TearDown()` would be helpful for posterity.

rcl/test/rcl/test_info_by_topic.cpp
```diff
@@ -0,0 +1,370 @@
...
+ * It only tests if the return code is the one expected.
+ */
+TEST_F(CLASSNAME(TestInfoByTopicFixture, RMW_IMPLEMENTATION),
+  test_rcl_get_publishers_info_by_topic_invalid_participants)
+{
+  // temp_info_array is invalid because it is expected to be zero initialized
+  // and the info_array variable inside it is expected to be null.
+  const auto & temp_info_array = &this->topic_info_array;
+  temp_info_array->info_array =
+    static_cast<rmw_topic_info_t *>(calloc(1, sizeof(rmw_topic_info_t)));
```
hidmic: @mm318 I think `new rwm_topic_info_t()` may achieve what you want (see Construction section on [`new` expression documentation](https://en.cppreference.com/w/cpp/language/new)).

rcl/src/rcl/graph.c
wjwwood: You don't have to do it right now, but I would recommend doing it, in order to future proof yourself in the event that we change it so that rmw errors differ from rcl in the future.

rcl/include/rcl/graph.h
mm318: Done.

rcl/include/rcl/graph.h
mm318: Done.

rcl/include/rcl/graph.h
mm318: Removed trailing period everywhere.

rcl/src/rcl/graph.c
mm318: I'll leave it there then.

rcl/test/rcl/test_info_by_topic.cpp
mm318: Oops, this is an artifact of a previous global replace of a `topic_info_array` `unique_ptr` to a `struct`.

It is now fixed.

rcl/test/rcl/test_info_by_topic.cpp
mm318: Fixed.

rcl/test/rcl/test_info_by_topic.cpp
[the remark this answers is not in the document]
mm318: I have added the use of `rcl_context_is_valid()` and a short comment.

rcl/test/rcl/test_info_by_topic.cpp
mm318: Oops, forgot about C++03 features :)

I have changed the use of `calloc()` to using `new()`.

rcl/include/rcl/graph.h
```diff
@@ -524,6 +525,122 @@ rcl_count_subscribers(
   const char * topic_name,
   size_t * count);
 
+/// Returns a list of all publishers to a topic.
+/// Each element in the list will contain the node name, node namespace, topic type,
+/// gid and the qos profile of the publisher.
```
wjwwood: nitpick: keep the brief to one sentence, i.e. move these last two lines into the doc body and don't use the `///` for them.

rcl/include/rcl/graph.h
```diff
@@ -524,6 +525,122 @@ rcl_count_subscribers(
...
+/**
+ * The `node` parameter must point to a valid node.
+ *
+ * The `topic_name` parameter must not be `NULL`.
+ *
+ * The `no_mangle` parameter determines if the provided topic_name should be
+ * expanded to its fully qualified name.
```
wjwwood: Is this accurate? For me the mangle has to do with the "ROS topic name" versus the underlying "middleware topic name" which may or may not be the same. It doesn't imply anything about relative versus absolute (or fully qualified) names.

rcl/include/rcl/graph.h
```diff
@@ -524,6 +525,122 @@ rcl_count_subscribers(
...
+ *
+ * It is the responsibility of the caller to ensure that `publishers_info` parameter points
+ * to a valid struct of type rmw_topic_info_array_t.
+ * The `count` field inside the struct must be set to 0 and the `info_array` field inside
+ * the struct must be set to null.
+ * @see rmw_get_zero_initialized_topic_info_array
```
wjwwood: Please use `\` everywhere.

rcl/include/rcl/graph.h
```diff
@@ -524,6 +525,122 @@ rcl_count_subscribers(
...
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_get_subscriptions_info_by_topic(
+  const rcl_node_t * node,
+  rcutils_allocator_t * allocator,
+  const char * topic_name,
+  bool no_mangle,
+  rmw_topic_info_array_t * subscriptions_info);
+
+
```
wjwwood: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

rcl/src/rcl/graph.c
```diff
@@ -375,6 +377,81 @@ rcl_count_subscribers(
   return rcl_convert_rmw_ret_to_rcl_ret(rmw_ret);
 }
 
+
```
wjwwood: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

rcl/include/rcl/graph.h
```diff
@@ -524,6 +525,122 @@ rcl_count_subscribers(
...
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_get_publishers_info_by_topic(
+  const rcl_node_t * node,
+  rcutils_allocator_t * allocator,
+  const char * topic_name,
+  bool no_mangle,
+  rmw_topic_info_array_t * publishers_info);
+
+
```
wjwwood: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

rcl/src/rcl/graph.c
```diff
@@ -375,6 +377,81 @@ rcl_count_subscribers(
...
+rcl_ret_t
+rcl_get_publishers_info_by_topic(
+  const rcl_node_t * node,
+  rcutils_allocator_t * allocator,
+  const char * topic_name,
+  bool no_mangle,
+  rmw_topic_info_array_t * publishers_info)
+{
+  return __rcl_get_info_by_topic(node, allocator, topic_name, no_mangle, publishers_info,
+           rmw_get_publishers_info_by_topic);
```
wjwwood: nitpick: weird indention

rcl/include/rcl/graph.h
ivanpauno: Yes, I think the wording of the comment is incorrect.

Should the topic name expansion be done in `rcl` or in `rclcpp` and `rclpy`?
See my comment [here](https://github.com/ros2/rclcpp/pull/960#discussion_r364215929).
I also don't find consistency on how this is done in different functions, e.g.:
- `rcl_publisher_init` [does the topic name expansion](https://github.com/ros2/rcl/blob/024c3ee50d9b79a85665efe1c191cc1225026519/rcl/src/rcl/publisher.c#L74-L156).
- `rcl_count_publishers` [does not the expansion]( https://github.com/ros2/rcl/blob/024c3ee50d9b79a85665efe1c191cc1225026519/rcl/src/rcl/graph.c#L374).

In general, the expansion is done in `rcl`, I think that the count publishers/subscriptions functions are the lonely exceptions. We should also create a function in `rcl` that does the hole expand and remap thing, to avoid repeating ourselves (and repeating that in [upper layers](https://github.com/ros2/rclcpp/blob/33a755c535d654c97401eb6d199580368c19eb40/rclcpp/src/rclcpp/expand_topic_or_service_name.cpp)).

wjwwood: I think that it may depend on how it is used. For `rcl_count_publishers()` it may be possible that the caller passes an already mangle topic name? I dunno without digging in some. All I'm saying is that it may intentionally be that way.

rcl/include/rcl/graph.h
mm318: Done.

rcl/include/rcl/graph.h
mm318: Updated the docstring.

rcl/include/rcl/graph.h
mm318: Done.

rcl/include/rcl/graph.h
mm318: Removed an empty line.

rcl/src/rcl/graph.c
mm318: Removed an empty line.

rcl/include/rcl/graph.h
mm318: Removed an empty line.

rcl/src/rcl/graph.c
mm318: Fixed.

rcl/include/rcl/graph.h
ivanpauno: @mm318 I don't see the comment updated.

> I think that it may depend on how it is used. For rcl_count_publishers() it may be possible that the caller passes an already mangle topic name? I dunno without digging in some. All I'm saying is that it may intentionally be that way.

Yeah, maybe doing remapping and expansion is not compatible with passing an already mangled name.
But in the case of `no_mangle=false`, it makes sense to do the `remapping/expansion` in rcl, to avoid repeating code like [this](https://github.com/ros2/rclcpp/blob/4a5eed968ca6fa4979214db8ac945fb163919326/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp#L230) (which is probably repeated in `rclpy` too).
And also, it's strange that the `rcl` function consider the `no_mangle` case and the `rclcpp` doesn't.

Anyways, we can continue the discussion after merging this, and correct `rcl_count_*` and these new functions both together. For now, just keep consistensy with `rcl_count_*`.

The comment may be like:
The `no_mangle` parameter determines if the provided topic_name should be mangled with the ROS conventions (`false`), or if it's a valid topic name for the middleware (useful when combining ROS with native middleware (e.g. DDS) apps) (`true`).

Also, add a comment in the docstring stating that the expected topic name should always be fully qualified (regardless of `no_mangle` param).

mm318: Oops, I updated the other instance of this comment, but forgot to update this one. Thanks, I'll update both instances of the comment with your suggestion.

PR #513, opened 2019-10-04
add mechanism to pass rmw impl specific payloads during pub/sub creation

wjwwood:
Connects to ros2/rclcpp#882


rcl/include/rcl/subscription.h
```diff
@@ -40,11 +40,11 @@ typedef struct rcl_subscription_options_t
 {
   /// Middleware quality of service settings for the subscription.
   rmw_qos_profile_t qos;
-  /// If true, messages published from within the same node are ignored.
-  bool ignore_local_publications;
```
ivanpauno: Why is this being deleted? Consider doing this in a separated PR if it's not related.

ivanpauno: Never mind. I understood after checking https://github.com/ros2/rmw/pull/187/ and seeing again the `rclcpp` PR.

PR #515, opened 2019-10-10
Switch to one Participant per Context

ivanpauno:
API changes needed after https://github.com/ros2/rmw_fastrtps/pull/312.


rcl/test/CMakeLists.txt
```diff
@@ -98,6 +98,7 @@ function(test_target_function)
     APPEND_LIBRARY_DIRS ${extra_lib_dirs}
     LIBRARIES ${PROJECT_NAME}
     AMENT_DEPENDENCIES ${rmw_implementation} "osrf_testing_tools_cpp" "test_msgs"
+    TIMEOUT 120
```
hidmic: @ivanpauno what happened here?

rcl/include/rcl/localhost.h
```diff
@@ -22,18 +22,24 @@ extern "C"
...
 
 extern const char * const RCL_LOCALHOST_ENV_VAR;
 
 /// Determine if the user wants to communicate using loopback only.
 /**
- * Checks if localhost should be used for network communication checking ROS_LOCALHOST_ONLY env
- * variable
- * \returns true if ROS_LOCALHOST_ONLY is set and is 1, false otherwise.
+ * Checks if localhost should be used for network communication.
+ * If `localhost_only` is RCL_LOCALHOST_ONLY_DEFAULT, ROS_LOCALHOST_ONLY env variable is used.
```
fujitatomoya:
```suggestion
 * If `localhost_only` is RMW_LOCALHOST_ONLY_DEFAULT, ROS_LOCALHOST_ONLY env variable is used.
```

rcl/test/CMakeLists.txt
ivanpauno: I saw it was flaky because of timeout, and I aummented it.

ivanpauno:
@kyrofa @ruffsl @mikaelarguedas In commit https://github.com/ros2/rcl/pull/515/commits/c2710820b6c6ca3e3d99d7b11c7d52bf3c78d974, I modified how the security directory lookup work when a participant is mapped to a node. I would like to get feedback on the approach.
@ruffsl any news about the changes in `sros2`?

I haven't extended the context name API in uper layers yet, though that should be strightforward.

I left the `security_options` argument in `rmw_create_node`, just to made review simpler, but it'll be ignored except a `Participant` is mapped to a `Node`.
I think that in a future iteration, we can consider removing it.
For rmw implementations that map one participant to a node, the security path stored in the [context security options](https://github.com/ros2/rmw/blob/8254c7f9166bdc354309143d187428536989ff3a/rmw/include/rmw/init_options.h#L54) can be used as a root and then do the lookup based in node name and namespace (though that would imply moving the lookup code out of rcl).

Another thing, I'm ignoring the `ROS_SECURITY_NODE_DIRECTORY` environment variable, because it doesn't make much sense when a `Participant` is mapped to a `Context`. I'm not sure if somebody was using it before, if not I think we should consider to depecrate it.

----

There's not much changes in the [rmw_fastrtps PR] about security, I just added a function that indicates if using the node name or the context name.
In the [rmw PR](https://github.com/ros2/rmw/pull/189), I've renamed `rmw_node_security_options_t` to `rmw_security_options_t` and added a few new utility functions.

rcl/src/rcl/init.c
```diff
@@ -144,6 +148,60 @@ rcl_init(
...
+  bool use_security = false;
+  ret = rcl_use_security(&use_security);
+  if (RCL_RET_OK != ret) {
+    fail_ret = ret;
+    goto fail;
+  }
+  RCUTILS_LOG_DEBUG_NAMED(
+    ROS_PACKAGE_NAME, "Using security: %s", use_security ? "true" : "false");
+
+  if (!rmw_use_node_name_in_security_directory_lookup()) {
```
kyrofa: So if the RMW sets this to be true, this chunk of code doesn't do anything and the old behavior in rcl_node_init() takes effect, right? It makes for what looks like quite a bit of duplication between these two functions-- is there a reason they can't be closer to each other? Or at least could we refactor this process into a single function call that does the right thing depending on RMW support?

rcl/include/rcl/security.h
```diff
@@ -0,0 +1,104 @@
...
+ * The returned security directory is associated with the node or context depending on the
+ * rmw implementation.
+ *
+ * E.g. for a node/context named "c" in namespace "/a/b", the secure root path will
+ * be "a/b/c", where the delimiter "/" is native for target file system (e.g. "\\" for _WIN32).
+ * If no exact match is found for the node/context name, a best match would be used instead
+ * (by performing longest-prefix matching).
+ *
+ * However, this expansion can be overridden by setting the secure node directory environment
+ * variable (only in rmw implementations that associate a security directory with a node),
```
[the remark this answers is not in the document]
ivanpauno: I think that it also didn't make much sense for nodes.
I mean, you could pass a different `ROS_SECURITY_DIRECTORY` to each process, and that would work too (the files would need to be in `$ROS_SECURITY_DIRECTORY/node_namespace/node_name` instead of `ROS_SECURITY_NODE_DIRECTORY`).
If the intention was to specify the same security directory for many nodes in the same process, it wouldn't make much sense to do the same for context, as you can just use the same context name (node names are supposed to be unique).

In general, I try to avoid features that can cause confusion.
Using `ROS_SECURITY_NODE_DIRECTORY` when you have one `Participant` per `Context` would be confusing. A name for the environment variable like `ROS_SECURITY_CONTEXT_DIRECTORY` would make more sense, but I don't see the value of such an option.

rcl/src/rcl/init.c
ivanpauno: Yes, I think I can avoid most of the duplication.

rcl/include/rcl/security.h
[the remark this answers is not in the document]
kyrofa: Yeah I agree, just makes me nervous to remove stuff I don't think makes sense, as I may not fully understand the problem the feature is trying to solve. Maybe @ruffsl or @mikaelarguedas will have some insight to share, here.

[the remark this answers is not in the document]
kyrofa: Yeah I agree, just makes me nervous to remove stuff I don't think makes sense, as I may not fully understand the problem the feature is trying to solve. Maybe @ruffsl or @mikaelarguedas will have some insight to share, here.

rcl/src/rcl/init.c
mikaelarguedas: Could this be deduplicated? looks like a lot of copy-n-paste happened here

rcl/src/rcl/node.c
```diff
@@ -258,79 +256,57 @@ rcl_node_init(
...
-      if (number == ULONG_MAX) {
-        RCL_SET_ERROR_MSG("failed to interpret ROS_DOMAIN_ID as integral number");
-        goto fail;
-      }
-      domain_id = (size_t)number;
-    }
-  } else {
-    domain_id = node->impl->options.domain_id;
+  domain_id = node->impl->options.domain_id;
+  if (RMW_RET_OK != rcl_domain_id(&domain_id)) {
```
mikaelarguedas: RCL_RET_OK ?

rcl/src/rcl/security.c
```diff
@@ -0,0 +1,310 @@
...
+  return RCL_RET_OK;
+}
+
+/**
+ * A security lookup function takes in the node's name, namespace, a security root directory and an allocator;
+ *  It returns the relevant information required to load the security credentials,
+ *   which is currently a path to a directory on the filesystem containing DDS Security permission files.
+ */
+typedef char * (* security_lookup_fn_t) (
+  const char * node_name,
```
mikaelarguedas: If we were to replace node everywhere, all this file could be updated as well.
Though it's unclear what word to use to capture "either context or node". I mean "participant" but it's not a ROS concept per say

rcl/src/rcl/security.c
```diff
@@ -0,0 +1,310 @@
...
+/**
+ * By using a prefix match, a node named e.g. "my_node_123" will be able to load and use the
+ * directory "my_node" if no better match exists.
+ * \param[in] base_dir
+ * \param[in] node_name
+ * \param[out] matched_name must be a valid memory address allocated with at least
+ * _TINYDIR_FILENAME_MAX characters.
+ * \return true if a match was found
+ */
+static bool get_best_matching_directory(
```
mikaelarguedas: Is this file a copy of `security_directory.c` ?
Is there a reason to duplicate the files?

If no need to keep the old file, could we just rename it and preserve history instead

rcl/include/rcl/security.h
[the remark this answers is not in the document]
mikaelarguedas: The main use I have for this is when I want to apply a single set of artifacts/policies to an entire subsystem.
e.g. "let's run this entire system with encryption and no access control and see how it performs"

Without this option we'd need to set a different environment with an appropriate env variable for each and every process that your launchfile (and included launchfiles) will start (that means a lot of launchfile editing)

Another example is to have an "admin" set of artifacts that can publish/subscribe to everything. You can point all your debugging tools to this single directory and not have to remap all your debug nodes / contexts to have the same name or prefix to fit the matching criteria.

---

So overall, even if it's not meant for use by default, I'm in favor of keeping such a feature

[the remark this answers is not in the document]
kyrofa: > You can point all your debugging tools to this single directory and not have to remap all your debug nodes / contexts to have the same name or prefix to fit the matching criteria.

I think this use-case might be better solved by having the debugging tools setup to use a common context name that can get a set of predefined, wide-open profiles (this can be disabled for production, of course). @SidFaber has done a lot of thinking about that.

rcl/src/rcl/init.c
ivanpauno: Done in https://github.com/ros2/rcl/pull/515/commits/b6ed0b46706415ea063b697da30f75cb8c9749fa.

rcl/include/rcl/security.h
```diff
@@ -0,0 +1,104 @@
...
+ * allowing users to explicitly specify the exact secure root directory to be utilized.
+ * Such an override is useful for where the FQN of a node is non-deterministic before runtime,
+ * or when testing and using additional tools that may not otherwise be easily provisioned.
+ *
+ * \param[in] name validated node name (a single token)
```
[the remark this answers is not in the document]
ivanpauno: Removed, except from the part that's referring to the above discussion: https://github.com/ros2/rcl/pull/515#discussion_r372513435.
I will update when the discussion is solved.

rcl/include/rcl/security.h
```diff
@@ -0,0 +1,104 @@
...
+rcl_get_enforcement_policy(rmw_security_enforcement_policy_t * policy);
+
+/// Return the secure root given a name and namespace.
+/**
+ * The returned security directory is associated with the node or context depending on the
+ * rmw implementation.
+ *
+ * E.g. for a node/context named "c" in namespace "/a/b", the secure root path will
+ * be "a/b/c", where the delimiter "/" is native for target file system (e.g. "\\" for _WIN32).
+ * If no exact match is found for the node/context name, a best match would be used instead
```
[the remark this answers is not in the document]
ivanpauno: Done.

rcl/include/rcl/security.h
```diff
@@ -0,0 +1,104 @@
...
+/// Check if security has to be used, according to the environment.
+/**
+ * \param use_security[out] Must not be NULL.
+ * \returns RCL_RET_INVALID_ARGUMENT if an argument is not valid, or
+ * \returns RCL_RET_ERROR if an unexpected error happened, or
+ * \returns RCL_RET_OK.
+ */
+RCL_PUBLIC
+rcl_ret_t
+rcl_use_security(bool * use_security);
```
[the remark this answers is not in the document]
ivanpauno: Updated to `rcl_security_enabled`.

rcl/src/rcl/init.c
```diff
@@ -141,6 +145,60 @@ rcl_init(
...
+  }
+
+  // Get actual localhost_only value based on environment variable, if needed.
+  ret = rcl_localhost_only(&context->impl->init_options.impl->rmw_init_options.localhost_only);
+  if (RCL_RET_OK != ret) {
+    fail_ret = ret;
+    goto fail;
+  }
+
+  // TODO(ivanpauno): Uncomment this when new path discovery magic is ready.
```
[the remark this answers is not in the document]
ivanpauno: Ups, forgot to delete that.
It was referring to the directory lookup, I borrowed the comment from here https://github.com/ros2/rcl/blob/7537f8304493e9b9568748b5601f5c1b7fa463f0/rcl/src/rcl/node.c#L319.

rcl/src/rcl/init.c
ivanpauno: Deduplicated code in https://github.com/ros2/rcl/pull/515/commits/b6ed0b46706415ea063b697da30f75cb8c9749fa.

rcl/src/rcl/node.c
ivanpauno: Thanks! Corrected in https://github.com/ros2/rcl/pull/515/commits/0317e66b33fd479dddfc38a8297a31d8801957a3.

rcl/src/rcl/security.c
ivanpauno: Oh, I thought I renamed (I actually renamed in the `CMakeLists.txt`, and `security_directory` is not compiled more).
Don't remember what I did, I will try to clean the history.

rcl/include/rcl/security.h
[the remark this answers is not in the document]
ivanpauno: > Without this option we'd need to set a different environment with an appropriate env variable for each and every process that your launchfile (and included launchfiles) will start (that means a lot of launchfile editing)

Maybe we could rename the variable to `ROS_SECURITY_DIRECTORY_OVERRIDE`, and avoid using `NODE` on it. We should keep backwards compatibility in rmw implementations that can actually set a different security directory for each point, and still accept the old name.

There's also another way to solve this. 99% of the processes, will have just one `Context` per process and use the default name on it. In that case, all launched processes will use the same security artifacts/policies. The only exception when you wouldn't use the default name, is when you need to create more than one (bridge between different domains, and other super stranges use cases). In those cases, you actually need to use a different context name, if not you won't be able to remap them separatly. For those exceptions, if you want to use the same security artifacts, you will have to remap both names againt to the "default context name".

ivanpauno:
I think I addressed all the comments, except https://github.com/ros2/rcl/pull/515#discussion_r372513435 where we need to reach concensous.
I've also to clean history, as commented in https://github.com/ros2/rcl/pull/515#discussion_r372635815.

rcl/src/rcl/security.c
ivanpauno: I've cleaned the history now.

rcl/src/rcl/security.c
ivanpauno: Avoided to mention node as far as possible.

rcl/src/rcl/security.c
```diff
@@ -0,0 +1,353 @@
...
+);
+
+char * prefix_match_lookup(
+  const char * name,
+  const char * namespace_,
+  const char * ros_secure_root_env,
+  const rcl_allocator_t * allocator
+);
+
+security_lookup_fn_t g_security_lookup_fns[] = {
```
hidmic: @ivanpauno can be made `const`?

rcl/src/rcl/security.c
```diff
@@ -0,0 +1,353 @@
...
+};
+
+typedef enum ros_security_lookup_type_e
+{
+  ROS_SECURITY_LOOKUP_NODE_OVERRIDE = 0,
+  ROS_SECURITY_LOOKUP_MATCH_EXACT = 1,
+  ROS_SECURITY_LOOKUP_MATCH_PREFIX = 2,
+} ros_security_lookup_type_t;
+
+char * g_security_lookup_type_strings[] = {
```
hidmic: @ivanpauno can be made `const`?

rcl/include/rcl/security.h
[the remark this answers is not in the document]
ivanpauno: @kyrofa @mikaelarguedas Friendly ping.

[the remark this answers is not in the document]
kyrofa: Sorry for the delay @ivanpauno, we've been chatting about exactly this in the WG. I promise we're not ignoring you!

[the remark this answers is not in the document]
kyrofa: The general feeling is that this variable does indeed serve a purpose. I suggest keeping it, although I'm in favor of @ivanpauno's suggested rename.

rcl/include/rcl/security.h
```diff
@@ -0,0 +1,139 @@
...
+ * true.
+ *
+ * \param use_security[out] Must not be NULL.
+ * \returns RCL_RET_INVALID_ARGUMENT if an argument is not valid, or
+ * \returns RCL_RET_ERROR if an unexpected error happened, or
+ * \returns RCL_RET_OK.
+ */
+RCL_PUBLIC
+rcl_ret_t
+rcl_security_enabled(bool * use_security);
```
jacobperron: Why not just return a `bool`?

rcl/src/rcl/security.c
```diff
@@ -0,0 +1,356 @@
...
+    ROS_PACKAGE_NAME, "Using security: %s", use_security ? "true" : "false");
+
+  if (!rmw_use_node_name_in_security_directory_lookup()) {
+    ret = rcl_get_enforcement_policy(&security_options->enforce_security);
+    if (RCL_RET_OK != ret) {
+      return ret;
+    }
+
+    if (!use_security) {
+      security_options->enforce_security = RMW_SECURITY_ENFORCEMENT_PERMISSIVE;
```
jacobperron: We can return early here and avoid the `else`. Also, I think we can move this check to before L54, right?

rcl/src/rcl/security.c
```diff
@@ -0,0 +1,356 @@
...
+  return RCL_RET_OK;
+}
+
+rcl_ret_t
+rcl_security_enabled(bool * use_security)
+{
+  const char * ros_security_enable = NULL;
+  const char * get_env_error_str = NULL;
+
+  if (!use_security) {
```
jacobperron: We have a macro for this: 

`RCL_CHECK_ARGUMENT_FOR_NULL(use_security, RCL_RET_INVALID_ARGUMENT)`

rcl/src/rcl/security.c
```diff
@@ -0,0 +1,356 @@
...
+
+rcl_ret_t
+rcl_get_enforcement_policy(rmw_security_enforcement_policy_t * policy)
+{
+  const char * ros_enforce_security = NULL;
+  const char * get_env_error_str = NULL;
+
+  if (!policy) {
+    return RCL_RET_INVALID_ARGUMENT;
+  }
```
jacobperron: `RCL_CHECK_ARGUMENT_FOR_NULL(policy, RCL_RET_INVALID_ARGUMENT)`

rcl/src/rcl/security.c
```diff
@@ -0,0 +1,356 @@
...
+
+  *policy = (0 == strcmp(ros_enforce_security, "Enforce")) ?
+    RMW_SECURITY_ENFORCEMENT_ENFORCE : RMW_SECURITY_ENFORCEMENT_PERMISSIVE;
+  return RCL_RET_OK;
+}
+
+/**
+ * A security lookup function takes in the node/context's name, namespace, a security root directory and an allocator;
+ *  It returns the relevant information required to load the security credentials,
+ *   which is currently a path to a directory on the filesystem containing DDS Security permission files.
```
jacobperron: nit: The indentation looks wrong here.

rcl/src/rcl/security.c
```diff
@@ -0,0 +1,356 @@
...
+  const rcl_allocator_t * allocator
+);
+
+const security_lookup_fn_t g_security_lookup_fns[] = {
+  NULL,
+  exact_match_lookup,
+  prefix_match_lookup,
+};
+
+typedef enum ros_security_lookup_type_e
```
jacobperron: Is there a reason to use the suffix `_e` instead of `_t` like the other enums in `rcl`?

rcl/src/rcl/security.c
```diff
@@ -0,0 +1,356 @@
...
+  ROS_SECURITY_LOOKUP_MATCH_PREFIX = 2,
+} ros_security_lookup_type_t;
+
+const char * const g_security_lookup_type_strings[] = {
+  "NODE_OVERRIDE",
+  "MATCH_EXACT",
+  "MATCH_PREFIX"
+};
+
+/// Return the directory whose name most closely matches name (longest-prefix match),
```
jacobperron:
```suggestion
/// Return the directory that most closely matches a given name (longest-prefix match),
```

rcl/test/rcl/test_node.cpp
```diff
@@ -120,14 +120,11 @@ TEST_F(CLASSNAME(TestNodeFixture, RMW_IMPLEMENTATION), test_rcl_node_accessors)
...
     rcl_ret_t ret = rcl_node_fini(&invalid_node);
+    EXPECT_EQ(RCL_RET_OK, rcl_context_fini(&invalid_context)) << rcl_get_error_string().str;
     EXPECT_EQ(RCL_RET_OK, ret);
   });
   ret = rcl_shutdown(&invalid_context);  // Shutdown to invalidate the node.
   ASSERT_EQ(RCL_RET_OK, ret);
-  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
-  {
-    EXPECT_EQ(RCL_RET_OK, rcl_context_fini(&invalid_context)) << rcl_get_error_string().str;
-  });
```
jacobperron: nit: This change seems unrelated to the PR

rcl/include/rcl/security.h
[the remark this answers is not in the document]
ruffsl: > you could pass a different ROS_SECURITY_DIRECTORY to each process

Having to specify an entirely different keystore could be tedious in the sense that one would have to take care that the two keystores shared the correct common certificate authorities and governances. I guess that isn't too difficult do one off, would currently consist of copying over the root level keystore files; though keeping the two keystores in sync would be fragile.

> I'm not sure if somebody was using it before

Here are some links to its purpose and origin:

* https://github.com/ros2/sros2/issues/69
* https://github.com/ros2/rcl/pull/300
* a82793ce3c70a9dc80fada177d01ac9181edfea8

> If the intention was to specify the same security directory for many nodes in the same process, it wouldn't make much sense to do the same for context, as you can just use the same context name
>
> ---
> We should keep backwards compatibility in rmw implementations that can actually set a different security directory for each point, and still accept the old name.

Hmm, I think I may have been under the impression that mapping multiple nodes per process to one context would remain optional, thus still warenting the use of `ROS_SECURITY_NODE_DIRECTORY`. However, reading further into this PR, it seems that the old style of mapping has been removed entirely (which I'm ok with). This may make it tricky to measure the minimal span of permissions required for each node that are never launched as their own process, as permissions brought in by other nodes could implicitly fill in the gaps of other node policies, akin to underspecified dependencies in package.xml manifest. But perhaps this just necessitates better measurement tooling of requested graph resources in SROS.

> There's also another way to solve this. 99% of the processes, will have just one Context per process and use the default name on it.

To be clear, while the norm may be for most processes to consist of only one context, different processes should not necessarily share a `default` context directory. Unless of course the user is unconcerned about access control, or the granularity of concern is that of a robot, and not a particular process in the robot. From a security perspective, our goal is to minimize the the attack surface or privilege of each process in the graph to what is only essential for it's application.

> For those exceptions, if you want to use the same security artifacts, you will have to remap both names againt to the "default context name".

Could you clarify by what you mean here? When you have a bridging scenario of one process, consisting of two separate contexts, say for bridging over two different DDS domains? If the two DDS domains where mutually exclusive, e.g. the ranges of domains expected to apply different permissions didn't overlap, then one could just use one context (one set of DDS security artifacts) as the DDS policy schema has it's primary pivot of allowed/denied permissions on the domain range.

https://github.com/ros2/sros2/blob/fc777a2b6957cc7481341bd9f7f49e7538035e9b/sros2/test/policies/permissions.xml#L10-L12

Going back to @mikaelarguedas https://github.com/ros2/rcl/pull/515#discussion_r372643420 , I think as long as users have a method for manually overriding what context directory gets loaded via the environment, for prototyping, deguing, and auditing, then that would serve the same purpose.

rcl/include/rcl/security.h
```diff
@@ -0,0 +1,139 @@
...
+#ifndef ROS_SECURITY_ROOT_DIRECTORY_VAR_NAME
+  #define ROS_SECURITY_ROOT_DIRECTORY_VAR_NAME "ROS_SECURITY_ROOT_DIRECTORY"
+#endif
+
+#ifndef ROS_SECURITY_LOOKUP_TYPE_VAR_NAME
+  #define ROS_SECURITY_LOOKUP_TYPE_VAR_NAME "ROS_SECURITY_LOOKUP_TYPE"
+#endif
+
+#ifndef ROS_SECURITY_STRATEGY_VAR_NAME
+#define ROS_SECURITY_STRATEGY_VAR_NAME "ROS_SECURITY_STRATEGY"
```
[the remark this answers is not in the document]
ivanpauno: I modified the style in this file too:
```c
#ifndef ...
# define ...
#endif
```
which we use everywhere else.

rcl/include/rcl/security.h
```diff
@@ -0,0 +1,139 @@
...
+
+/// Init the security options from the values of the environment variables and passed names.
+/**
+ * Inits the passed security options based on the environment.
+ * For more details:
+ *  \sa rcl_security_enabled
+ *  \sa rcl_get_enforcement_policy
+ *  \sa rcl_get_secure_root
+ *
+ * \param name[in] name used to find the securiy root path.
```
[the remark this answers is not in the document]
ivanpauno: Good catch, thanks!

rcl/include/rcl/security.h
ivanpauno: To differentiate an error condition from security being enbled or not.

rcl/src/rcl/security.c
ivanpauno: Checking `rmw_use_node_name_in_security_directory_lookup` in this function was an error.
I will add a test case to cover that.

rcl/src/rcl/security.c
ivanpauno: This comes from the previous `security_directory.h`. I can fix it if you want, but it's unrelated.

P.S.: I don't know why github just shows a deleted file and a new file in the diff.
  If you check the diff of each commit, you would be able to understand better the history (as there is one commit just renaming files). For the same reason, I will avoid squash and merge, and left a separated commit renaming the files.

rcl/test/rcl/test_node.cpp
ivanpauno: Without that change, `invalid_context` may be finalized before `invalid_node`.
The documentation says that doing so is UB.
Before, doing that in the wrong order didn't create a problem, now it segfaults.

rcl/src/rcl/security.c
jacobperron: I missed that the file was renamed. We can leave it as-is since the change is unrelated.

rcl/test/rcl/test_node.cpp
jacobperron: Thanks.

rcl/include/rcl/security.h
jacobperron: :ok_hand:

rcl/src/rcl/security.c
jacobperron: This change is unrelated.

rcl/src/rcl/domain_id.c
```diff
@@ -0,0 +1,56 @@
...
+rcl_domain_id(size_t * domain_id)
+{
+  const char * ros_domain_id;
+  const char * get_env_error_str;
+
+  if (!domain_id) {
+    return RCL_RET_INVALID_ARGUMENT;
+  }
+
+  if (RCL_DEFAULT_DOMAIN_ID != *domain_id) {
```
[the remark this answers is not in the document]
ivanpauno: Solved.

rcl/src/rcl/localhost.c
```diff
@@ -12,19 +12,39 @@
...
   const char * ros_local_host_env_val = NULL;
-  return rcutils_get_env(RCL_LOCALHOST_ENV_VAR, &ros_local_host_env_val) == NULL &&
-         ros_local_host_env_val != NULL && strcmp(ros_local_host_env_val, "1") == 0;
+  const char * get_env_error_str = NULL;
+
+  if (!localhost_only) {
+    return RCL_RET_INVALID_ARGUMENT;
+  }
+
+  if (RMW_LOCALHOST_ONLY_DEFAULT == *localhost_only) {
```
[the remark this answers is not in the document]
ivanpauno: Solved.

rcl/include/rcl/security.h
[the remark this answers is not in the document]
ivanpauno: > Having to specify an entirely different keystore could be tedious in the sense that one would have to take care that the two keystores shared the correct common certificate authorities and governances. I guess that isn't too difficult do one off, would currently consist of copying over the root level keystore files; though keeping the two keystores in sync would be fragile.

:+1: 

> However, reading further into this PR, it seems that the old style of mapping has been removed entirely (which I'm ok with).

Yes, the implementations will be refactored to map a `Participant` to a `Context`.

> Could you clarify by what you mean here? When you have a bridging scenario of one process, consisting of two separate contexts, say for bridging over two different DDS domains? If the two DDS domains where mutually exclusive, e.g. the ranges of domains expected to apply different permissions didn't overlap, then one could just use one context (one set of DDS security artifacts) as the DDS policy schema has it's primary pivot of allowed/denied permissions on the domain range.

Yes, I was refering to bridging two domains. In that case, you need two Contexts. as you need one participant for each domain. If they can share the security artifacts, both of them could use the same context name.

> Going back to @mikaelarguedas #515 (comment) , I think as long as users have a method for manually overriding what context directory gets loaded via the environment, for prototyping, deguing, and auditing, then that would serve the same purpose.

I will add `ROS_SECURITY_DIRECTORY_OVERRIDE`, which is a clearer name.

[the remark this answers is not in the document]
ivanpauno: Solved in https://github.com/ros2/rcl/pull/515/commits/d284a2e92ac07c3afc362b802d8b2851e302fbec.

rcl/src/rcl/security.c
ivanpauno: See https://github.com/ros2/rcl/pull/515/commits/7111a4598dbd9c37a34cb2fda0fcfa804ee2201f.

rcl/src/rcl/security.c
```diff
@@ -0,0 +1,351 @@
...
+  return secure_root;
+}
+
+char * rcl_get_secure_root(
+  const char * name,
+  const char * namespace_,
+  const rcl_allocator_t * allocator)
+{
+  bool ros_secure_directory_override = true;
+  bool use_node_name_in_lookup = rmw_use_node_name_in_security_directory_lookup();
```
kyrofa: It appears that this is not used, is that true?

ivanpauno: Fixed in https://github.com/ros2/rcl/pull/515/commits/533ea2bbd7a56468cedca1323de9f5f176218920.

rcl/include/rcl/security.h
```diff
@@ -0,0 +1,138 @@
...
+
+#ifndef ROS_SECURITY_STRATEGY_VAR_NAME
+# define ROS_SECURITY_STRATEGY_VAR_NAME "ROS_SECURITY_STRATEGY"
+#endif
+
+#ifndef ROS_SECURITY_ENABLE_VAR_NAME
+# define ROS_SECURITY_ENABLE_VAR_NAME "ROS_SECURITY_ENABLE"
+#endif
+
+/// Init the security options from the values of the environment variables and passed names.
```
hidmic: @ivanpauno nit: consider 
```suggestion
/// Initializes security options from values in the environment and given names.
```

rcl/include/rcl/security.h
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * Inits the passed security options based on the environment.
```
hidmic: @ivanpauno same about using `Init` as a verb.

rcl/src/rcl/init.c
```diff
@@ -155,7 +193,6 @@ rcl_init(
   }
 
   TRACEPOINT(rcl_init, (const void *)context);
-
```
hidmic: @ivanpauno spurious change?

rcl/src/rcl/domain_id.c
```diff
@@ -0,0 +1,52 @@
...
+#include "rcl/error_handling.h"
+#include "rcl/types.h"
+
+const char * const RCL_DOMAIN_ID_ENV_VAR = "ROS_DOMAIN_ID";
+
+rcl_ret_t
+rcl_get_default_domain_id(size_t * domain_id)
+{
+  const char * ros_domain_id;
+  const char * get_env_error_str;
```
hidmic: @ivanpauno consider zero initializing these too.

rcl/src/rcl/security.c
```diff
@@ -0,0 +1,350 @@
...
+    return ret;
+  }
+
+  if (!use_security) {
+    security_options->enforce_security = RMW_SECURITY_ENFORCEMENT_PERMISSIVE;
+    return RMW_RET_OK;
+  }
+
+  RCUTILS_LOG_DEBUG_NAMED(
+    ROS_PACKAGE_NAME, "Using security: %s", use_security ? "true" : "false");
```
hidmic: @ivanpauno this should be moved up, `use_security` will never be `false` at this point.

rcl/src/rcl/security.c
```diff
@@ -0,0 +1,350 @@
...
+
+  get_env_error_str = rcutils_get_env(ROS_SECURITY_ENABLE_VAR_NAME, &ros_security_enable);
+  if (NULL != get_env_error_str) {
+    RCL_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "Error getting env var '" RCUTILS_STRINGIFY(ROS_SECURITY_ENABLE_VAR_NAME) "': %s\n",
+      get_env_error_str);
+    return RCL_RET_ERROR;
+  }
+
+  *use_security = (0 == strcmp(ros_security_enable, "true"));
```
hidmic: @ivanpauno meta: uh, this environment variable expects `true` as a boolean literal, while others above expect `1`. We should probably standardize that in a follow-up PR.

rcl/src/rcl/security.c
```diff
@@ -0,0 +1,350 @@
...
+    }
+    if (file.is_dir) {
+      size_t matched_name_length = strnlen(file.name, sizeof(file.name) - 1);
+      if (0 ==
+        strncmp(
+          file.name, name, matched_name_length) &&
+        matched_name_length > max_match_length)
+      {
+        max_match_length = matched_name_length;
+        memcpy(matched_name, file.name, max_match_length);
```
hidmic: @ivanpauno same as in other PRs, use `strncpy` or `rcutils_format_string` or `rcutils_strdup` instead of plain old `memcpy`

rcl/src/rcl/security.c
```diff
@@ -0,0 +1,350 @@
...
+  if (1 == strlen(namespace_)) {
+    secure_root = rcutils_join_path(ros_secure_root_env, name, *allocator);
+  } else {
+    char * fqn = NULL;
+    char * root_path = NULL;
+    // Combine namespace with name
+    // TODO(ros2team): remove the hard-coded value of the root namespace
+    fqn = rcutils_format_string(*allocator, "%s%s%s", namespace_, "/", name);
+    // Get native path, ignore the leading forward slash
+    // TODO(ros2team): remove the hard-coded length, use the length of the root namespace instead
```
hidmic: @ivanpauno meta: these TODOs seem to have been solved.

rcl/include/rcl/security.h
ivanpauno: I will use imperative mood instead of indicative mood, but otherwise follow your suggestion.

rcl/src/rcl/security.c
ivanpauno: Good catch, thanks!

rcl/src/rcl/security.c
ivanpauno: It was already working like that. See that I first just moved code https://github.com/ros2/rcl/pull/515/commits/4aa706cf34cc69d2c86920c161c4aa964fb706a7.

I agree we should standarize. We should probably have a function in `rcutils` that accepts either "1" "true" "on" (not case sensitive).

rcl/src/rcl/security.c
ivanpauno: I didn't modify, I just moved it.
In this case, `rcutils_strdup` doesn't apply.
I will replace `memcpy` with `strncpy`, which is more appropiate here.

rcl/src/rcl/security.c
ivanpauno: I still see a "/" and a `+ 1` there, so I don't think they were solved.

rcl/include/rcl/security.h
ivanpauno: Done in https://github.com/ros2/rcl/pull/515/commits/a905ae9dc01d24d4b09f0999f538b61a0e83315e.

rcl/include/rcl/security.h
ivanpauno: Done https://github.com/ros2/rcl/pull/515/commits/a905ae9dc01d24d4b09f0999f538b61a0e83315e.

rcl/src/rcl/init.c
ivanpauno: Solved in https://github.com/ros2/rcl/pull/515/commits/a905ae9dc01d24d4b09f0999f538b61a0e83315e.

rcl/src/rcl/domain_id.c
ivanpauno: Done, see https://github.com/ros2/rcl/pull/515/commits/a905ae9dc01d24d4b09f0999f538b61a0e83315e.

rcl/src/rcl/security.c
ivanpauno: See https://github.com/ros2/rcl/pull/515/commits/a905ae9dc01d24d4b09f0999f538b61a0e83315e.

rcl/src/rcl/security.c
ivanpauno: See https://github.com/ros2/rcl/pull/515/commits/a905ae9dc01d24d4b09f0999f538b61a0e83315e.

rcl/src/rcl/security.c
```diff
@@ -0,0 +1,189 @@
...
+
+  ret = rcl_get_enforcement_policy(&security_options->enforce_security);
+  if (RCL_RET_OK != ret) {
+    return ret;
+  }
+
+  // File discovery magic here
+  char * secure_root = rcl_get_secure_root(
+    name,
+    allocator);
```
hidmic: @ivanpauno nit: is there a need to wrap these arguments?

rcl/src/rcl/security.c
hidmic: Oh, I see. Nevermind.

rcl/src/rcl/security.c
```diff
@@ -0,0 +1,189 @@
...
+    // check root directory if override directory environment variable is empty
+    if (rcutils_get_env(ROS_SECURITY_ROOT_DIRECTORY_VAR_NAME, &env_buf)) {
+      return NULL;
+    }
+    if (!env_buf) {
+      return NULL;
+    }
+    if (0 == strcmp("", env_buf)) {
+      return NULL;  // environment variable was empty
+    } else {
```
hidmic: @ivanpauno nit: on need for an `else` clause

rcl/src/rcl/validate_security_context_name.c
```diff
@@ -0,0 +1,148 @@
...
+rcl_ret_t
+rcl_validate_security_context_name(
+  const char * security_context,
+  int * validation_result,
+  size_t * invalid_index)
+{
+  if (!security_context) {
+    return RCL_RET_INVALID_ARGUMENT;
+  }
+  return rmw_validate_namespace_with_size(
```
hidmic: @ivanpauno shouldn't this be `rcl_validate_security_context_name_with_size`?

rcl/src/rcl/validate_security_context_name.c
```diff
@@ -0,0 +1,148 @@
...
+          // explicitly not taking return value which is number of bytes written
+          int ret = rcutils_snprintf(
+            default_err_msg, sizeof(default_err_msg),
+            "rcl_validate_security_context_name_with_size(): "
+            "unknown rmw_validate_namespace_with_size() result '%d'",
+            t_validation_result);
+          if (ret < 0) {
+            RCL_SET_ERROR_MSG(
+              "rcl_validate_security_context_name_with_size(): "
+              "rcutils_snprintf() failed");
```
hidmic: @ivanpauno it would be good to leave a trace of the original error, e.g. 
```suggestion
              "rcutils_snprintf() failed while reporting an unknown validation result");
```

rcl/src/rcl/validate_security_context_name.c
```diff
@@ -0,0 +1,148 @@
...
+  size_t * invalid_index)
+{
+  if (!security_context) {
+    return RCL_RET_INVALID_ARGUMENT;
+  }
+  if (!validation_result) {
+    return RCL_RET_INVALID_ARGUMENT;
+  }
+
+  int t_validation_result;
```
hidmic: @ivanpauno what does the `t_` stand for? Why not: 
```suggestion
  int rmw_validation_result;
```
?

Same elsewhere.

rcl/src/rcl/validate_security_context_name.c
```diff
@@ -0,0 +1,148 @@
...
+    }
+    if (invalid_index) {
+      *invalid_index = t_invalid_index;
+    }
+    return RCL_RET_OK;
+  }
+
+  // security_context might be longer that namespace length, check false positives and correct
+  if (t_validation_result == RMW_NAMESPACE_INVALID_TOO_LONG &&
+    security_context_length <= RCL_CONTEXT_NAME_MAX_LENGTH)
```
hidmic: @ivanpauno mind to explain? Also, I have the impression some result is wrong, as this `if` clause is redundant as-is once you get to this point.

rcl/test/rcl/test_security.cpp
```diff
@@ -0,0 +1,221 @@
...
+#endif
+
+char g_envstring[512] = {0};
+
+static int putenv_wrapper(const char * env_var)
+{
+#ifdef _WIN32
+  return _putenv(env_var);
+#else
+  return putenv(reinterpret_cast<char *>(const_cast<char *>(env_var)));
```
hidmic: Hmm, why an (apparently redundant) `reinterpret_cast` here?

rcl/test/rcl/test_security.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+    // Always make sure the variable we set is unset at the beginning of a test
+    unsetenv_wrapper(ROS_SECURITY_ROOT_DIRECTORY_VAR_NAME);
+    unsetenv_wrapper(ROS_SECURITY_DIRECTORY_OVERRIDE);
+    unsetenv_wrapper(ROS_SECURITY_STRATEGY_VAR_NAME);
+    unsetenv_wrapper(ROS_SECURITY_ENABLE_VAR_NAME);
+    allocator = rcl_get_default_allocator();
+    root_path = nullptr;
+    secure_root = nullptr;
+    base_lookup_dir_fqn = nullptr;
```
hidmic: @ivanpauno nit: any reason not to directly initialize these member fields?

rcl/test/rcl/test_security.cpp
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  void TearDown() final
+  {
+    OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+    {
+      allocator.deallocate(root_path, allocator.state);
+    });
```
hidmic: @ivanpauno why do you need `OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT` here? `rcl`'s default allocator will never throw.

rcl/test/rcl/test_security.cpp
```diff
@@ -0,0 +1,221 @@
...
+TEST_F(TestGetSecureRoot, successScenarios_local_exactMatch) {
+  putenv_wrapper(
+    ROS_SECURITY_ROOT_DIRECTORY_VAR_NAME "="
+    TEST_RESOURCES_DIRECTORY TEST_SECURITY_DIRECTORY_RESOURCES_DIR_NAME);
+
+  secure_root = rcl_get_secure_root(TEST_SECURITY_CONTEXT_ABSOLUTE, &allocator);
+  std::string secure_root_str(secure_root);
+  ASSERT_STREQ(
+    TEST_SECURITY_CONTEXT,
+    secure_root_str.substr(secure_root_str.size() - (sizeof(TEST_SECURITY_CONTEXT) - 1)).c_str());
```
hidmic: @ivanpauno I think you meant `strlen`.

rcl/test/rcl/test_security.cpp
hidmic: I personally think it'd make tests clearer if these deallocations were right next to the allocation spot. I know that's duplicating, but it makes them easier to read.

rcl/test/rcl/test_security.cpp
```diff
@@ -0,0 +1,221 @@
...
+  putenv_wrapper(ROS_SECURITY_ROOT_DIRECTORY_VAR_NAME "=" TEST_RESOURCES_DIRECTORY);
+  ASSERT_STREQ(root_path, TEST_RESOURCES_DIRECTORY);
+}
+
+TEST_F(TestGetSecureRoot, nodeSecurityDirectoryOverride_invalidDirectory) {
+  /* The override provided should exist. Providing correct node/namespace/root dir won't help
+   * if the node override is invalid. */
+  putenv_wrapper(
+    ROS_SECURITY_DIRECTORY_OVERRIDE
+    "=TheresN_oWayThi_sDirectory_Exists_hence_this_would_fail");
```
hidmic: @ivanpauno nit: 
```suggestion
    "=TheresN_oWayThi_sDirectory_Exists_hence_this_should_fail");
```

rcl/test/rcl/test_security.cpp
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  rcl_allocator_t allocator;
+  char * root_path;
+  char * secure_root;
+  char * base_lookup_dir_fqn;
+};
+
+TEST_F(TestGetSecureRoot, failureScenarios) {
+  ASSERT_EQ(
```
hidmic: @ivanpauno here and elsewhere, do we have to `assert` or would an `expect` perform equally well? Particularly in those tests that currently have a single assertion.

rcl/test/rcl/test_security.cpp
```diff
@@ -0,0 +1,221 @@
...
+  ASSERT_STREQ(root_path, TEST_RESOURCES_DIRECTORY);
+}
+
+TEST_F(
+  TestGetSecureRoot,
+  nodeSecurityDirectoryOverride_validDirectory_overrideRootDirectoryAttempt) {
+  /* Setting root dir has no effect */
+  putenv_wrapper(ROS_SECURITY_DIRECTORY_OVERRIDE "=" TEST_RESOURCES_DIRECTORY);
+  root_path = rcl_get_secure_root(
+    "name shouldn't matter", &allocator);
```
hidmic: @ivanpauno nit: do we need to wrap these arguments?

rcl/src/rcl/validate_security_context_name.c
ivanpauno: temporal? to differentiate from the parameter being passed.

rcl/test/rcl/test_security.cpp
ivanpauno: good question, it's preexistence but I will update.

rcl/test/rcl/test_security.cpp
ivanpauno: Preexistent. Some tests set them to different things.

rcl/test/rcl/test_security.cpp
ivanpauno: preexistent

rcl/test/rcl/test_security.cpp
ivanpauno: :sweat_smile:

rcl/src/rcl/init.c
```diff
@@ -144,6 +153,72 @@ rcl_init(
...
+    &validation_result,
+    &invalid_index);
+  if (RCL_RET_OK != ret) {
+    RCL_SET_ERROR_MSG("rcl_validate_security_context_name() failed");
+    fail_ret = ret;
+    goto fail;
+  }
+  if (RCL_CONTEXT_NAME_VALID != validation_result) {
+    RCL_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "rcl_validate_security_context_name result is not valid: '%s'. Invalid index: %zu",
```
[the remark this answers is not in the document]
ivanpauno: See https://github.com/ros2/rcl/pull/515/commits/e3d5ebdf6094274f7f74d752ecce3fbadec84150.

rcl/src/rcl/security.c
ivanpauno: See https://github.com/ros2/rcl/pull/515/commits/e3d5ebdf6094274f7f74d752ecce3fbadec84150.

rcl/src/rcl/security.c
ivanpauno: See https://github.com/ros2/rcl/pull/515/commits/e3d5ebdf6094274f7f74d752ecce3fbadec84150.

rcl/src/rcl/validate_security_context_name.c
ivanpauno: See https://github.com/ros2/rcl/pull/515/commits/e3d5ebdf6094274f7f74d752ecce3fbadec84150.

rcl/src/rcl/validate_security_context_name.c
ivanpauno: See https://github.com/ros2/rcl/pull/515/commits/e3d5ebdf6094274f7f74d752ecce3fbadec84150.

rcl/test/rcl/test_security.cpp
ivanpauno: See https://github.com/ros2/rcl/pull/515/commits/e3d5ebdf6094274f7f74d752ecce3fbadec84150.

rcl/test/rcl/test_security.cpp
ivanpauno: See https://github.com/ros2/rcl/pull/515/commits/e3d5ebdf6094274f7f74d752ecce3fbadec84150.

rcl/test/rcl/test_security.cpp
ivanpauno: See https://github.com/ros2/rcl/pull/515/commits/e3d5ebdf6094274f7f74d752ecce3fbadec84150.

rcl/test/rcl/test_security.cpp
ivanpauno: See https://github.com/ros2/rcl/pull/515/commits/e3d5ebdf6094274f7f74d752ecce3fbadec84150.

rcl/test/rcl/test_security.cpp
ivanpauno: See https://github.com/ros2/rcl/pull/515/commits/e3d5ebdf6094274f7f74d752ecce3fbadec84150.

rcl/src/rcl/validate_security_context_name.c
ivanpauno: The security context name can be longer that a namespace.
I'm reusing the namespace check, and in the case the error was that the name was too long, I recheck if that's actually the case. If not, I correct the validation result to valid and return `RCL_RET_OK`.

Note: My logic is valid because the length is the last thing being checked:
https://github.com/ros2/rmw/blob/f76511333709a6f5d1e4543129ec73f2cd2b4f06/rmw/src/validate_full_topic_name.c#L117
https://github.com/ros2/rmw/blob/f76511333709a6f5d1e4543129ec73f2cd2b4f06/rmw/src/validate_namespace.c#L112

rcl/include/rcl/graph.h
```diff
@@ -440,6 +441,39 @@ rcl_get_node_names(
   rcutils_string_array_t * node_names,
   rcutils_string_array_t * node_namespaces);
 
+/// Return a list of available nodes in the ROS graph, including their security context names.
+/**
+ * \sa Works like rcl_get_node_names, but it also include the security context name the node is
```
hidmic: @ivanpauno `\sa` expects references as arguments.

Maybe you just want something like:
```suggestion
 * \sa An \ref rcl_get_node_names equivalent, but including in its output the security context name the node is using.
```
?

rcl/include/rcl/graph.h
```diff
@@ -440,6 +441,39 @@ rcl_get_node_names(
...
+ * using as an output.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Maybe [1]
+ * <i>[1] implementation may need to protect the data structure with a lock</i>
```
hidmic: @ivanpauno what implementations? This may need some qualification.

rcl/include/rcl/security.h
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * The value of the environment variable `ROS_SECURITY_ROOT_DIRECTORY` is used as a root.
+ * The specific directory to be used, is found from that root using the `name` passed.
+ * E.g. for a context named "/a/b/c" and root "/r", the secure root path will be
+ * "/r/a/b/c", where the delimiter "/" is native for target file system (e.g. "\\" for _WIN32).
+ *
+ * However, this expansion can be overridden by setting the secure directory override environment
+ * (`ROS_SECURITY_DIRECTORY_OVERRIDE`) variable, allowing users to explicitly specify the exact secure
+ * root directory to be utilized.
+ * Such an override is useful for where the security context is non-deterministic before runtime,
```
hidmic: @ivanpauno nit: 
```suggestion
 * Such an override is useful for applications where the security context is non-deterministic before runtime,
```

hidmic: Also, what do you mean by `non-deterministic before runtime`? Unknown?

rcl/include/rcl/validate_security_context_name.h
```diff
@@ -0,0 +1,91 @@
...
+  RMW_NAMESPACE_INVALID_CONTAINS_REPEATED_FORWARD_SLASH
+#define RCL_SECURITY_CONTEXT_NAME_INVALID_NAME_TOKEN_STARTS_WITH_NUMBER \
+  RMW_NAMESPACE_INVALID_NAME_TOKEN_STARTS_WITH_NUMBER
+#define RCL_SECURITY_CONTEXT_NAME_INVALID_TOO_LONG RMW_NAMESPACE_INVALID_TOO_LONG
+
+#define RCL_SECURITY_CONTEXT_NAME_MAX_LENGTH RMW_NODE_NAME_MAX_NAME_LENGTH
+
+/// Determine if a given security context name is valid.
+/**
+ * /sa The same rules as rmw_validate_namespace are used.
```
hidmic: @ivanpauno same comment as elsewhere about using `\sa` with arguments other than references. Also, that slash is backwards :sweat_smile:

rcl/include/rcl/validate_security_context_name.h
```diff
@@ -0,0 +1,91 @@
...
+ * The only difference is the maximum length, which can be 255 characters.
```
hidmic: @ivanpauno nit: 
```suggestion
 * The only difference is in the maximum allowed length, which in this case can be up to 255 characters.
```

rcl/include/rcl/validate_security_context_name.h
```diff
@@ -0,0 +1,91 @@
...
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_validate_security_context_name(
+  const char * security_context,
+  int * validation_result,
+  size_t * invalid_index);
+
+/// Deterimine if a given security context name is valid.
+/**
+ * This is an overload with an extra parameter for the length of security_context.
```
hidmic: @ivanpauno an overload in C? Of what?

rcl/src/rcl/arguments.c
```diff
@@ -1098,6 +1147,7 @@ rcl_arguments_fini(
       args->impl->num_param_files_args = 0;
       args->impl->parameter_files = NULL;
     }
+    args->impl->allocator.deallocate(args->impl->security_context, args->impl->allocator.state);
```
hidmic: @ivanpauno `rcl_arguments_copy` is lacking code to handle this new member field.

rcl/src/rcl/init.c
```diff
@@ -144,6 +153,72 @@ rcl_init(
...
+    &validation_result,
+    &invalid_index);
+  if (RCL_RET_OK != ret) {
+    RCL_SET_ERROR_MSG("rcl_validate_security_context_name() failed");
+    fail_ret = ret;
+    goto fail;
+  }
+  if (RCL_SECURITY_CONTEXT_NAME_VALID != validation_result) {
+    RCL_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "security context name is not valid: '%s'. Invalid index: %zu",
```
hidmic: @ivanpauno nit^N: capitalize, as in 
```suggestion
      "Security context name is not valid: '%s'. Invalid index: %zu",
```

rcl/src/rcl/validate_security_context_name.c
hidmic: Consider using `tmp`, it's somewhat more common.

rcl/src/rcl/validate_security_context_name.c
```diff
@@ -0,0 +1,148 @@
...
+    return RCL_RET_OK;
+  }
+
+  // security_context might be longer that namespace length, check false positives and correct
+  if (t_validation_result == RMW_NAMESPACE_INVALID_TOO_LONG &&
+    security_context_length <= RCL_SECURITY_CONTEXT_NAME_MAX_LENGTH)
+  {
+    *validation_result = RCL_SECURITY_CONTEXT_NAME_VALID;
+    return RCL_RET_OK;
+  }
```
hidmic: @ivanpauno Hmm I though I had brought up this. What if it is too long and it exceeds `RCL_SECURITY_CONTEXT_NAME_MAX_LENGTH`?

rcl/test/rcl/test_get_node_names.cpp
```diff
@@ -162,3 +163,134 @@ TEST_F(CLASSNAME(TestGetNodeNames, RMW_IMPLEMENTATION), test_rcl_get_node_names)
...
+    std::make_tuple(
+      std::string(node4_name),
+      std::string(node4_namespace),
+      std::string(security_context_name)));
+
+  rcl_node_t node5 = rcl_get_zero_initialized_node();
+  const char * node5_name = "node1";
+  const char * node5_namespace = "/";
+  rcl_node_options_t node5_options = rcl_node_get_default_options();
+  ret = rcl_node_init(&node5, node5_name, node5_namespace, &context, &node5_options);
```
hidmic: @ivanpauno hmm, is this checking that the feature works with duplicated FQN/security context tuples?

rcl/include/rcl/graph.h
ivanpauno: Yeah, I was trying to do that. Thanks for the suggestion!

rcl/include/rcl/graph.h
ivanpauno: Preexisting, see https://github.com/ros2/rcl/blob/f242967bf024804b3ed1e8b00ebf3b3b3c9541cd/rcl/include/rcl/graph.h#L456 and documentation of the other functions too.

IMO, the comment is clear.
It means that it might lock or not depending of the rmw implementation, and that the user shouldn't make any assumption.

rcl/include/rcl/security.h
ivanpauno: Again, this was taken from the old `security_directory.h`.
The original comment isn't clear to me.

rcl/include/rcl/validate_security_context_name.h
ivanpauno: It's kind of metaphorical :joy:.
I copied the comment from `rmw_validate_namespace.h`, I can rephrase it if you want.

rcl/include/rcl/graph.h
hidmic: I'd be explicit and `RMW`, but fine with it for now.

rcl/src/rcl/validate_security_context_name.c
ivanpauno: See https://github.com/ros2/rcl/pull/515#discussion_r395865974.

rcl/test/rcl/test_get_node_names.cpp
ivanpauno: yes

rcl/include/rcl/graph.h
ivanpauno: See https://github.com/ros2/rcl/pull/515/commits/44a15be05484088c78a4e7cde08a021356757160

rcl/include/rcl/validate_security_context_name.h
ivanpauno: Done in https://github.com/ros2/rcl/pull/515/commits/44a15be05484088c78a4e7cde08a021356757160.

rcl/include/rcl/validate_security_context_name.h
ivanpauno: See https://github.com/ros2/rcl/pull/515/commits/44a15be05484088c78a4e7cde08a021356757160.

rcl/src/rcl/arguments.c
ivanpauno: Solved in https://github.com/ros2/rcl/pull/515/commits/44a15be05484088c78a4e7cde08a021356757160.

rcl/src/rcl/init.c
ivanpauno: See https://github.com/ros2/rcl/pull/515/commits/44a15be05484088c78a4e7cde08a021356757160.

rcl/src/rcl/validate_security_context_name.c
ivanpauno: See https://github.com/ros2/rcl/pull/515/commits/44a15be05484088c78a4e7cde08a021356757160.

rcl/include/rcl/graph.h
ivanpauno: AFAIK, every of them are locking now (at least, the three default implementations).
It's pretty hard to have a complete list, considering that there are also external rmw implementations.

hidmic: Oh, for sure, I wouldn't mention every existing RMW implementation, I'd just qualify the noun as in:
```suggestion
 * <i>[1] RMW implementation in use may need to protect the data structure with a lock</i>
```

rcl/include/rcl/validate_security_context_name.h
hidmic: Fine with keeping the overload concept, but do add a reference to the function that is being "overloaded".

rcl/src/rcl/validate_security_context_name.c
hidmic: I may be crazy, but, let's say `tmp_validation_result` equals `RMW_NAMESPACE_INVALID_TOO_LONG` and `security_context_length` exceeds `RCL_SECURITY_CONTEXT_NAME_MAX_LENGTH`. The switch statement [will be skipped](https://github.com/ros2/rcl/pull/515/files#diff-6d1e34c077896ab6d3f8e1a936573cccR65), the  if statement immediately below [will be skipped](https://github.com/ros2/rcl/pull/515/files#diff-6d1e34c077896ab6d3f8e1a936573cccR112) as well, and the implementation will conclude that [the name is valid](https://github.com/ros2/rcl/pull/515/files#diff-6d1e34c077896ab6d3f8e1a936573cccR120).

rcl/include/rcl/security.h
ivanpauno: Used your suggestion in https://github.com/ros2/rcl/pull/515/commits/35ec36eb170bb5801c17f03637e2a9b97f13ded7.

rcl/include/rcl/graph.h
ivanpauno: Done in https://github.com/ros2/rcl/pull/515/commits/35ec36eb170bb5801c17f03637e2a9b97f13ded7.

rcl/include/rcl/validate_security_context_name.h
ivanpauno: See https://github.com/ros2/rcl/pull/515/commits/35ec36eb170bb5801c17f03637e2a9b97f13ded7.

rcl/src/rcl/validate_security_context_name.c
ivanpauno: You're right, I failed to get your point the first time.
See https://github.com/ros2/rcl/pull/515/commits/675e4d566319dc784cd514fab12bb2858f34a1f7.

rcl/include/rcl/validate_security_context_name.h
```diff
@@ -44,8 +44,8 @@ extern "C"
 
 /// Determine if a given security context name is valid.
 /**
- * /sa The same rules as rmw_validate_namespace are used.
- * The only difference is the maximum length, which can be 255 characters.
+ * \sa The same rules as \ref rmw_validate_namespace are used.
```
hidmic: @ivanpauno I think you want to drop `\sa` here.

rcl/src/rcl/arguments.c
```diff
@@ -1100,6 +1100,15 @@ rcl_arguments_copy(
       }
     }
   }
+  char * security_context_copy = rcutils_strdup(args->impl->security_context, allocator);
+  if (args->impl->security_context && !security_context_copy) {
+    if (RCL_RET_OK != rcl_arguments_fini(args_out)) {
+      RCL_SET_ERROR_MSG("Error while finalizing arguments due to another error");
+    }
+    RCL_SET_ERROR_MSG("Error while copying security context argument");
```
hidmic: @ivanpauno won't you get a complain about the error message being overwritten if `rcl_arguments_fini` fails?

rcl/include/rcl/validate_security_context_name.h
```diff
@@ -64,10 +64,10 @@ rcl_validate_security_context_name(
 
 /// Deterimine if a given security context name is valid.
 /**
- * This is an overload with an extra parameter for the length of security_context.
- * \param[in] security_context The number of characters in security_context.
+ * \sa This is an overload of \ref rcl_validate_security_context_name with an extra parameter
```
hidmic: @ivanpauno same, I think you want to drop `\sa` here.

rcl/include/rcl/validate_security_context_name.h
ivanpauno: See https://github.com/ros2/rcl/pull/515/commits/77aad14964b484c6091bebe0cec28169546303a9.

rcl/src/rcl/arguments.c
ivanpauno: Done in https://github.com/ros2/rcl/pull/515/commits/77aad14964b484c6091bebe0cec28169546303a9.

rcl/include/rcl/validate_security_context_name.h
ivanpauno: Solved in https://github.com/ros2/rcl/pull/515/commits/77aad14964b484c6091bebe0cec28169546303a9.

PR #516, opened 2019-10-10
Remove vestigial references to rcl_ok()

jacobperron:
Minor thing I noticed while doing something else.

### DIFF
--- a/rcl/test/rcl/test_init.cpp
+++ b/rcl/test/rcl/test_init.cpp
@@ -98,9 +98,9 @@ private:
   FakeTestArgv(const FakeTestArgv &) = delete;
 };
 
-/* Tests the rcl_init(), rcl_ok(), and rcl_shutdown() functions.
+/* Tests the rcl_init() and rcl_shutdown() functions.
  */
-TEST_F(CLASSNAME(TestRCLFixture, RMW_IMPLEMENTATION), test_rcl_init_and_ok_and_shutdown) {
+TEST_F(CLASSNAME(TestRCLFixture, RMW_IMPLEMENTATION), test_rcl_init_and_shutdown) {
   rcl_ret_t ret;
   rcl_init_options_t init_options = rcl_get_zero_initialized_init_options();
   ret = rcl_init_options_init(&init_options, rcl_get_default_allocator());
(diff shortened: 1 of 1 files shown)


rcl/test/rcl/test_init.cpp
```diff
@@ -98,9 +98,9 @@ struct FakeTestArgv
   FakeTestArgv(const FakeTestArgv &) = delete;
 };
 
-/* Tests the rcl_init(), rcl_ok(), and rcl_shutdown() functions.
+/* Tests the rcl_init() and rcl_shutdown() functions.
  */
-TEST_F(CLASSNAME(TestRCLFixture, RMW_IMPLEMENTATION), test_rcl_init_and_ok_and_shutdown) {
+TEST_F(CLASSNAME(TestRCLFixture, RMW_IMPLEMENTATION), test_rcl_init_and_shutdown) {
```
ivanpauno: This is also testing `rcl_context_is_valid` which replaced `rcl_ok`, though I don't have a better name for the test case.

rcl/test/rcl/test_init.cpp
```diff
@@ -206,9 +206,9 @@ TEST_F(CLASSNAME(TestRCLFixture, RMW_IMPLEMENTATION), test_rcl_init_and_ok_and_s
   });
 }
 
-/* Tests the rcl_get_instance_id() and rcl_ok() functions.
+/* Tests the rcl_get_instance_id() function.
  */
-TEST_F(CLASSNAME(TestRCLFixture, RMW_IMPLEMENTATION), test_rcl_get_instance_id_and_ok) {
+TEST_F(CLASSNAME(TestRCLFixture, RMW_IMPLEMENTATION), test_rcl_get_instance_id) {
```
ivanpauno: The same applies here

rcl/test/rcl/test_init.cpp
jacobperron: Yeah, I'm not sure about the name either. While this test happens to be checking `rcl_context_is_valid`, IMO, any unit tests specifically targeting that function should go in `test_context.c` anyways, so I'll leave this as-is.

PR #520, opened 2019-10-16
Add localhost option to node creation

BMarchi:
Add the boolean that determines if the user provided the env var ROS_LOCALHOST_ONLY as 1 to restrict network traffic. Connected to this [issue](https://github.com/ros2/ros2/issues/798)

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8441/)
- Arch [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4343/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/8352/)
- OSX [Build Status](https://ci.ros2.org/job/ci_osx/6866/)

### DIFF
--- a/rcl/include/rcl/localhost.h
+++ b/rcl/include/rcl/localhost.h
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef RCL__LOCALHOST_H
-#define RCL__LOCALHOST_H
+#ifndef RCL__LOCALHOST_H_
+#define RCL__LOCALHOST_H_
 
 #ifdef __cplusplus
 extern "C"
@@ -23,7 +23,7 @@ extern "C"
 #include "rcl/types.h"
 #include "rcl/visibility_control.h"
 
-extern const char * RCL_LOCALHOST_ENV_VAR;
+extern const char * const RCL_LOCALHOST_ENV_VAR;
 
 /// Determine if the user wants to communicate using loopback only.
 /**
@@ -39,4 +39,4 @@ rcl_localhost_only();
 }
 #endif
 
-#endif // RCL__LOCALHOST_H
+#endif  // RCL__LOCALHOST_H_
--- a/rcl/include/rcl/localhost.h
+++ b/rcl/include/rcl/localhost.h
@@ -1,0 +1,42 @@
+// Copyright 2019 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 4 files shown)


dirk-thomas:
See my other comment about where to place this function: https://github.com/ros2/rmw/pull/190#issuecomment-542757995

rcl/include/rcl/localhost.h
```diff
@@ -0,0 +1,42 @@
...
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#include "rcl/types.h"
+#include "rcl/visibility_control.h"
+
+extern const char * RCL_LOCALHOST_ENV_VAR;
```
hidmic: consider:
```suggestion
extern const char * const RCL_LOCALHOST_ENV_VAR;
```

rcl/src/rcl/localhost.c
```diff
@@ -0,0 +1,30 @@
...
+// limitations under the License.
+
+#include <rcl/localhost.h>
+
+#include "rcutils/get_env.h"
+
+#include <stdlib.h>
+#include <string.h>
+
+const char * RCL_LOCALHOST_ENV_VAR = "ROS_LOCALHOST_ONLY";
```
hidmic: likewise, consider:
```suggestion
const char * const RCL_LOCALHOST_ENV_VAR = "ROS_LOCALHOST_ONLY";
```

rcl/include/rcl/localhost.h
fujitatomoya: does it have to be extern here? any plan to use this value in other files as globally in the future? i do not see that is coming since there is a API provided to check if localhost is enabled or not on this.

rcl/src/rcl/localhost.c
fujitatomoya: and also could be static value, since the scope is only localhost.c file ?

rcl/include/rcl/localhost.h
BMarchi: Right now is not used by anyone outside the localhost.c file. I made it extern to avoid copies using static and to fulfill [dirk's suggestion](https://github.com/ros2/rmw/pull/190#discussion_r334059644).
@dirk-thomas , any thoughts about keeping this variable in this context only?

dirk-thomas: There is no point in hiding the definition of the name of the env var. If we do that people which want to use that will just copy-n-paste the name of the env var which is strictly worse. So keeping the name public is desired.

dirk-thomas: (If we would do the same for `ROS_DOMAIN_ID` it would actually avoid it being define in three different source files.)

Issue #244, opened 2018-05-29
[YAML parser] Configurable maximum values during parsing 


mikaelarguedas:
This set of maximum values should be reevaluated: https://github.com/ros2/rcl/blob/5edf27b032cdaa324349bbdc02a9772eb4d9443b/rcl_yaml_param_parser/src/parser.c#L65-L71

kekstee:
I was just about to open an issue about this set of limits.

More specifically I easily hit the 128 byte string length on resource file paths which will look something like:
"sharepath(absolute) + some subfolder + filename"

Maybe the length limit for keys and values should be separated?
Is this limit on strings enforced anywhere else?

jacobperron:
We should making the maximum values for string size, number of nodes, and number of parameters configurable. Perhaps it makes sense to default all of them to "no limit" and users can configure a hard limit to avoid dynamic memory allocation.

jdlangs:
@jacobperron could you provide some thoughts on what you consider configurable? Can we get away with compile-time definitions or is something more complex needed?

jacobperron:
Compile-time definitions might be okay, but then I guess one would have to compile `rcl_yaml_param_parser` from source to change a value. This is inconvenient if you've installed ROS 2 binaries. It might be nice to have more flexibility and configure these parser parameters at runtime. For example, we could bootstrap parameter parsing by first parsing an optional YAML file containing the parser parameters.

It would be good to get other thoughts on the matter.

clalancette:
Before we go too much further, I took a quick look at the code to make sure I could say something intelligent.  If you look at the places that `MAX_STRING_SIZE` is used ([here](https://github.com/ros2/rcl/blob/master/rcl_yaml_param_parser/src/parser.c#L210), [here](https://github.com/ros2/rcl/blob/master/rcl_yaml_param_parser/src/parser.c#L1119), [here](https://github.com/ros2/rcl/blob/master/rcl_yaml_param_parser/src/parser.c#L1318), and [here](https://github.com/ros2/rcl/blob/master/rcl_yaml_param_parser/src/parser.c#L1429)), it doesn't actually prevent allocations from happening.  It really just bounds the size that they could potentially be.

[...]

PR #521, opened 2019-10-17
Add compile flag to disable string parameter size limit

jdlangs:
This is intended to be a minimal change for potential inclusion for the Eloquent release with follow-on work to come for making an easier interface to recover this behavior than rebuilding from source. See #244 for additional discussion on this topic.

If a change like this that makes string parameters unbounded by default is considered a breaking change that can't be done mid-distribution, I'd really appreciate a quick evaluation if we can consider including it for this release. I think of this as fixing a notable regression from ROS1.

Thanks,
Josh

### DIFF
--- a/rcl_yaml_param_parser/CMakeLists.txt
+++ b/rcl_yaml_param_parser/CMakeLists.txt
@@ -27,6 +27,10 @@ add_library(
   ${rcl_yaml_parser_sources})
 ament_target_dependencies(${PROJECT_NAME} "yaml" "rcutils")
 
+# This compile definition used to enable bounding of string parameters
+# see MAX_STRING_SIZE definition in src/parser.c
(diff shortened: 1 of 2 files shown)


clalancette:
I'd much rather just either increase the limit or remove it entirely.  If we don't hear back from the original author of the PR soon, I'll look at doing one of those instead.

jdlangs:
That works too. In case it does turn out that still having a limit makes sense, how would the size be decided?

clalancette:
>  In case it does turn out that still having a limit makes sense, how would the size be decided?

I was thinking of 8MB.  It's not too much memory on modern machines, it's larger than any URDF (for instance) than I've ever seen, and it matches the default stack limit on Linux, so it has nice symmetry.  But I could be convinced of other numbers too.

clalancette:
See https://github.com/ros2/rcl/pull/524 , where we remove the string limit completely.

PR #529, opened 2019-10-23
Specify test working directory

rotu:
Avert breakage from https://github.com/ament/ament_cmake/pull/206
Signed-off-by: Dan Rose <dan@digilabs.io>


dirk-thomas:
Thanks.

Issue #481, opened 2019-08-07
Support parameter descriptions in yaml parser


bpwilcox:
----
## Feature request
Add support to parse parameter descriptions in yaml file.

#### Feature description

Given that parameters must now be declared and have optional ParameterDescriptors, it would be an enhancement to support parsing of a parameter description from the yaml file format. This could be convenient for users to not have to specify descriptions in code or to change their parameter settings without recompiling, but another example use case would be to control read-only or range validation for nodes with `NodeOptions.automatically_declare_from_overrides` set as true (such as a "parameter blackboard" ). Since only the default descriptors are currently declared for parameters passed via yaml, any other node can change the value to anything, which may not be desired

This feature would allow users to specify optional descriptions in the yaml file, aligning with the fields in `rcl_interfaces/ParmeterDescriptor.msg`.

An example format could be:
``` yaml
node:
  ros__parameters:
    param1: value
      read_only: true
      description: "this value is read-only"
   param2: value
      from_value: min_value
      to_value: max_value
```

#### Implementation considerations

- Requires matching changes in rclcpp to create ParameterDescriptor objects from parsed yaml and pass to Node parameter overrides

hidmic:
@bpwilcox could you further expand on a use case for this feature? A node's implementation usually relies on the existence of some parameters, in which case I think they should explicitly declare them and not rely on an external YAML file to do so -- particularly for client libraries in languages that are statically typed and will likely fail ungracefully if things are not the way they should.

hidmic:
Hmm, https://github.com/ros2/rclcpp/issues/807 suggests a parameter blackboard i.e. not an application-specific node but a generic parameter container. Is that what you had in mind?

bpwilcox:
@hidmic The parameter blackboard is a use-case that I believe well deserves this feature, but I believe even application-specific nodes could benefit.

>  A node's implementation usually relies on the existence of some parameters, in which case I think they should explicitly declare them and not rely on an external YAML file to do so
>
I believe this would be the case as-is, nodes relying on certain parameters may explicitly declare them, but already the YAML provides an opportunity to overwrite these parameters. As it pertains to parameter descriptors, you can overwrite (or declare if not already provided) key descriptions for paramters as they fit the system. Some example use-cases:
- set certain values to read-only for a particular run of a node
- adapt integer/floating-point ranges to match a variety of different robot/context configurations
  - if using a parameter tuning plugin, this helps define allowable ranges to expose
[...]

PR #533, opened 2019-10-25
Parse parameter descriptors in yaml

bpwilcox:
This PR addresses the feature request in #481, namely parsing parameter descriptor keys in the yaml file. This feature will enable parameter descriptors to be added into nodes at run-time. 

Overview of additions are:
- New struct types to contain parameter descriptor information (in alignment with the `ParameterDescriptor.msg`)
  - `rcl_param_descriptor_t`, `rcl_node_params_descriptor_t`
  - descriptor keys are compared for assignment into struct
- Added map level for `MAP_PARAM_DESCRIPTORS_LVL` to parse descriptors
  - new map level is triggered when the `'parameter__descriptors'` key is parsed by the yaml event
  - some added API to handle descriptor keys/values
- Independent allocation/access to parameters and parameter descriptors in the `params_t` struct
  - each node will have its list of parameters as well as descriptors
  - parameter descriptors may be added in yaml file without corresponding parameter or mixed

The proposed syntax for this PR is:
```
  node1:
    ros__parameters:
      foo: 42
      bar: 25.0
      param1: 10
      parameter__descriptors:
        param1:
          type: 2
          min_value: 5
          max_value: 100
          step: 5
          read_only: false
          description: "int parameter"
          additional_constraints: "only multiples of 5"
        param2:
          min_value: 1.0
          max_value: 10.0
          description: "double parameter"
```

I've made accompanying changes in rclcpp to extract the parameter descriptors struct into a `ParameterDescriptor.msg` which will override when calling `declare_parameter` (or if the overrides are automatically declared). I will submit that PR to the rclcpp repo soon, though this is a pre-requisite.


rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
```diff
@@ -74,12 +74,39 @@ typedef struct rcl_node_params_s
...
+  uint8_t * type;
+  char * description;
+  char * additional_constraints;
+  double * min_value_double;
+  double * max_value_double;
+  double * step_double;
+  int64_t * min_value_int;
+  int64_t * max_value_int;
+  int64_t * step_int;
+} rcl_param_descriptor_t;
```
hidmic: @wjwwood meta: not a comment for you @bpwilcox, but replicating `rcl_interfaces/msg/ParameterDescriptor` structure is not ideal. The same goes with `rcl_interfaces/msg/Parameter`, so I wonder if we should be re-using the message C representation instead (like we partially do in [`rcl_action`](https://github.com/ros2/rcl/blob/9034c406c2d5bc4ec291f8e1ad4303cde0184cbc/rcl_action/include/rcl_action/types.h#L65-L80) already). Thoughts?

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
```diff
@@ -74,12 +74,39 @@ typedef struct rcl_node_params_s
...
+
+/// \typedef rcl_node_params_descriptors_t
+/// \brief node_params_descriptors_t stores all the parameter descriptors of a single node
+typedef struct rcl_node_params_descriptors_s
+{
+  char ** parameter_names;  ///< Array of parameter names (keys)
+  rcl_param_descriptor_t * parameter_descriptors;  ///< Array of coressponding parameter descriptors
```
hidmic: @bpwilcox typo: 
```suggestion
  rcl_param_descriptor_t * parameter_descriptors;  ///< Array of corresponding parameter descriptors
```

rcl_yaml_param_parser/src/parser.c
```diff
@@ -63,9 +65,12 @@ typedef struct namespace_tracker_s
   uint32_t num_node_ns;
   char * parameter_ns;
   uint32_t num_parameter_ns;
+  char * descriptor_key_ns;
+  uint32_t num_descriptor_key_ns;
 } namespace_tracker_t;
 
 #define PARAMS_KEY "ros__parameters"
+#define PARAMS_DESCRIPTORS_KEY "parameter__descriptors"
```
hidmic: @bpwilcox meta: why not `ros__parameter_descriptors` for consistency?

rcl_yaml_param_parser/src/parser.c
```diff
@@ -826,6 +1136,77 @@ void rcl_yaml_node_struct_print(
       }
     }
   }
+  printf("\n Node Name\t\t\t\tParameter Descriptors\n");
```
hidmic: @bpwilcox nit: I guess it's fine to begin with, as it's precludes the need to match parameter values with parameter descriptors, but it'd be nice to collapse both printed tables into one.

rcl_yaml_param_parser/src/parser.c
```diff
@@ -826,6 +1136,77 @@ void rcl_yaml_node_struct_print(
...
+      printf("%s\n", node_name);
+    }
+
+    if (NULL != params_st->descriptors) {
+      rcl_node_params_descriptors_t * node_descriptors_st = &(params_st->descriptors[node_idx]);
+      for (size_t parameter_idx = 0U; parameter_idx < node_descriptors_st->num_params;
+        parameter_idx++)
+      {
+        if (
+          (NULL != node_descriptors_st->parameter_names))
```
hidmic: @bpwilcox nit: can't this line be collapsed with the one above?

rcl_yaml_param_parser/src/parser.c
```diff
@@ -1278,6 +1659,201 @@ static rcutils_ret_t parse_value(
...
 
+static rcutils_ret_t parse_descriptor(
+  namespace_tracker_t * ns_tracker,
+  const yaml_event_t event,
+  const bool is_seq,
+  const size_t node_idx,
+  const size_t parameter_idx,
+  rcl_params_t * params_st)
+{
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(params_st, RCUTILS_RET_INVALID_ARGUMENT);
```
hidmic: @bpwilcox consider adding a couple `assert`s to ensure both `*_idx` arguments are within bounds.

rcl_yaml_param_parser/src/parser.c
```diff
@@ -1278,6 +1659,201 @@ static rcutils_ret_t parse_value(
...
+    RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "Error parsing value %s at line %d", value, line_num);
+    return RCUTILS_RET_ERROR;
+  }
+
+  if (NULL == ns_tracker->parameter_ns) {
+    RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "Parameter assignment at line %d unallowed in parameter__descriptors", line_num);
+    return RCUTILS_RET_ERROR;
+  } else {
```
hidmic: @bpwilcox no need for an `else` clause, there's a return in the `if` clause.

rcl_yaml_param_parser/src/parser.c
```diff
@@ -1278,6 +1659,201 @@ static rcutils_ret_t parse_value(
...
+    params_st->descriptors[node_idx].parameter_names[parameter_idx] =
+      rcutils_strdup(ns_tracker->parameter_ns, allocator);
+  }
+  if (NULL == param_descriptor->name) {
+    param_descriptor->name =
+      rcutils_strdup(params_st->descriptors[node_idx].parameter_names[parameter_idx], allocator);
+    rcl_node_params_descriptors_t * node_descriptors_st = &(params_st->descriptors[node_idx]);
+    node_descriptors_st->num_params++;
+  }
+  if (true == is_seq) {
```
hidmic: @bpwilcox I think this should be way above.

hidmic: Also, why not simply `if(is_seq)`?

rcl_yaml_param_parser/src/parser.c
```diff
@@ -1422,6 +2003,83 @@ static rcutils_ret_t parse_key(
...
+        // Guard against adding more than the maximum allowed parameters
+        if (params_st->descriptors[*node_idx].num_params >= MAX_NUM_PARAMS_PER_NODE) {
+          RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(
+            "Exceeded maximum allowed number of parameters for a node (%d)",
+            MAX_NUM_PARAMS_PER_NODE);
+          ret = RCUTILS_RET_ERROR;
+          break;
+        }
+
+        /// Add a parameter name into the node parameters
```
hidmic: @bpwilcox `node parameter descriptors` maybe?

rcl_yaml_param_parser/src/parser.c
```diff
@@ -1278,6 +1659,201 @@ static rcutils_ret_t parse_value(
...
+      RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(
+        "Value type 'string' expected at line %d for "
+        "parameter__descriptors key: additional_constraints",
+        line_num);
+      return RCUTILS_RET_ERROR;
+    }
+  } else if (0 == strncmp("type", ns_tracker->descriptor_key_ns, strlen("type"))) {
+    if (val_type == DATA_TYPE_INT64) {
+      param_descriptor->type = (uint8_t *)ret_val;
+      if (NULL == param_descriptor->type) {
```
hidmic: @bpwilcox here and everywhere else within this function, why do you check for `ret_val` nullity again? And why does the error message set reads as a memory allocation error?? Bad copy-paste?

rcl_yaml_param_parser/src/parser.c
```diff
@@ -38,6 +38,7 @@ typedef enum yaml_map_lvl_e
   MAP_UNINIT_LVL = 0U,
   MAP_NODE_NAME_LVL = 1U,
   MAP_PARAMS_LVL = 2U,
+  MAP_PARAMS_DESCRIPTORS_LVL = 3U
```
hidmic: @bpwilcox hmm, putting parameter descriptors within the `ros__parameters` and relying on users not using the `parameter__descriptors` key for their own purposes under the assumption that it's unlikely enough is not ideal. Consider either making this map a sibling of `ros__parameters` or supporting further nesting e.g.:

```yaml
my_node:
    ros__parameters:
        param1: 1.0 
        param2:
            value: foo
            descriptor:
                read_only: true
```

rcl_yaml_param_parser/src/parser.c
```diff
@@ -1278,6 +1659,201 @@ static rcutils_ret_t parse_value(
...
+  void * ret_val = get_value(value, style, &val_type, allocator);
+  if (NULL == ret_val) {
+    RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "Error parsing value %s at line %d", value, line_num);
+    return RCUTILS_RET_ERROR;
+  }
+
+  if (NULL == ns_tracker->parameter_ns) {
+    RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "Parameter assignment at line %d unallowed in parameter__descriptors", line_num);
```
hidmic: @bpwilcox here and elsewhere, isn't it the point of having a `PARAMS_DESCRIPTORS_KEY` macro to not use the string literal everywhere?

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
```diff
@@ -74,12 +74,39 @@ typedef struct rcl_node_params_s
...
+  int64_t * min_value_int;
+  int64_t * max_value_int;
+  int64_t * step_int;
+} rcl_param_descriptor_t;
+
+/// \typedef rcl_node_params_descriptors_t
+/// \brief node_params_descriptors_t stores all the parameter descriptors of a single node
+typedef struct rcl_node_params_descriptors_s
+{
+  char ** parameter_names;  ///< Array of parameter names (keys)
```
hidmic: @bpwilcox I don't follow why there's an array of parameter names here and a name on each descriptor.

rcl_yaml_param_parser/src/parser.c
bpwilcox: Given it is already scoped under ros__parameters level it seemed redundant, but I'm not opposed.

This does bring up my original approach to have the `parameter__descriptors` be at the same level as `ros__parameters` (e.g. no longer scoped), so the two would be uncoupled. It seemed like much more work at the time with not too significant benefits, but I'm open to discussing it.

rcl_yaml_param_parser/src/parser.c
bpwilcox: Right, I didn't want to have to match values and descriptors.

rcl_yaml_param_parser/src/parser.c
bpwilcox: You're referring to why I am using the `ns_tracker->descriptor_key_ns` in this function?

rcl_yaml_param_parser/src/parser.c
bpwilcox: I'm not checking `ret_val` for nullity directly here, that is already done once when it is first assigned. Here I am following the same pattern as in the `parse_value`, where it checks the allocated descriptor struct field after being cast from `ret_val`. Am I misunderstanding the original code on that pattern?

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
bpwilcox: Parallelism? I think I had intended to parallel the structure of `rcl_node_params` and `rcl_node_params_descriptors`. It made the code more convenient to write, I believe, though I suppose we could remove one or the other.

rcl_yaml_param_parser/src/parser.c
bpwilcox: I would think the further nesting could be more confusing to the user, just my opinion (it seems to conflate the yaml syntax rules for when it is a parameter vs a descriptor). I mentioned this in a another comment, but yes I can see `parameter__descriptors` as a sibling to `ros__parameters`,but as I did attempt that first, I think it is a bit more messy to write the code for doing so than simply using the key to trigger a new map level. If per your suggestion, we change the key to `ros__parameter_descriptors`, do you think it is unique enough for that assumption?

rcl_yaml_param_parser/src/parser.c
hidmic: Hmm I'm not following you, I don't see such pattern. But even if there was, you're not allocating memory here when you cast (as the log message below suggests). It's a C-like cast, it won't result in NULL if it isn't NULL already. Or I completely missed the point of this code.

rcl_yaml_param_parser/src/parser.c
hidmic: I'm wary about the word `enough` there. I'd rather just preclude name collisions. But maybe others don't think alike. @ivanpauno @jacobperron @wjwwood thoughts?

rcl_yaml_param_parser/src/parser.c
hidmic: > You're referring to why I am using the `ns_tracker->descriptor_key_ns` in this function?

Not exactly, that I understand. What I meant to say is, let's try use `PARAMS_DESCRIPTORS_KEY` instead of `parameter_descriptors` where possible. For this particular line:

```c
RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(
      "Parameter assignment at line %d unallowed in " PARAMS_DESCRIPTORS_KEY, line_num);
```

Sorry for not being clear enough.

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
hidmic: Hmm I'm not entirely convinced it's worth the duplication. You can get away with just an `rcl_param_descriptor_t`  array.

rcl_yaml_param_parser/src/parser.c
ivanpauno: I don't see a reason to nest `parameter_descriptors` under `ros__parameters`, and I would rather implement it as a sibling of `ros__parameters`, as @hidmic suggested. `ros__parameter_descriptors` is a good name for that sibling.
I also agree that the alternative of further nesting is confusing.

jacobperron: I prefer the extra nesting since it keeps the descriptors physically close to the parameters they are describing. As a sibling to `ros__parameters`, I think it requires more mental effort. My two cents.

ivanpauno: Well, actually that option isn't bad either. But both the value and the descriptor should be optional (and at least one has to be specified).

wjwwood: I don't have a strong opinion. I see merit in both keeping it separate and having descriptors near the overrides.

bpwilcox: In regards to the extra nesting, I am unsure what would be a preferable syntax. The one @hidmic  presented as an example implicates that there are two different ways to define an override, the original way, and the nested version, where the descriptor is attached and there's 2 new key words, `value` and `descriptor`. In my opinion, I think it obfuscates the parameter namespacing syntax by adding even more key words to check.  For that reason, I'm still in favor of a separate map level (either sibling or scoped under `ros__parameters`) for the `ros__parameter_descriptors` where the rules for parsing are cleanly separated. In addition, I wouldn't want to require that both a parameter description and override have to exist, but a user can set either.

rcl_yaml_param_parser/src/parser.c
bpwilcox: I understand now, thanks for clarifying

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
bpwilcox: Fair point

rcl_yaml_param_parser/src/parser.c
bpwilcox: I realized my mistake, you're correct. The check was required when allocating the arrays in `parse_value`, which I am not doing here.

rcl_yaml_param_parser/src/parser.c
hidmic: > The one @hidmic presented as an example implies that there are two different ways to define an override, the original way, and the nested version,

That's correct.

> In my opinion, I think it obfuscates the parameter namespacing syntax by adding even more key words to check.

That sounds a bit like an implementation detail i.e. something a user wouldn't (have to) care about.

In any case, I don't have a strong opinion in favor or against a sibling descriptor mapping or nested descriptors (and neither do others, or at least there's no general consensus), but I do think it should be one of those two collision free approaches and not the current one.

hidmic:
@bpwilcox are you planning to circle back on this?

bpwilcox:
@hidmic Yes, I'd started some local changes before the new year but had been preoccupied since. Most notably, as suggested I'm intending to move the ros__parameters and ros__parameter_descriptors to be at sibling levels. With regards to removing parameter names from the descriptor type, I'm not sure if that is worthwhile as it breaks some of the intended code similarity I'd added intentionally before. I'll try to get an update in next week or so.

hidmic:
@bpwilcox do you still plan on submitting changes? Or shall I close this patch?

bpwilcox:
> @bpwilcox do you still plan on submitting changes? Or shall I close this patch?

Yes, I believe I should have some more cycles to work on this in the near future.

bpwilcox:
@hidmic I've re-done this work and addressed feedback on a [separate branch](https://github.com/bpwilcox/rcl/tree/param_descriptors_feedback_revised). It was easier and cleaner than trying to resolve all of the merge conflicts considering the many changes since I've last touched this PR. Would it be better to close this PR and open a new one, or to force push/reset that branch onto this one?

bpwilcox:
@mjeronimo I reset this branch with commits from my revised branch. This way the previous PR comments can be referenced. Maybe we can get some re-reviews of this work while I push ahead in this next month on the rclcpp layer in https://github.com/ros2/rclcpp/pull/909

rcl_yaml_param_parser/src/parse.c
```diff
@@ -786,13 +786,6 @@ rcutils_ret_t parse_key(
       {
         /// Till we get PARAMS_KEY or PARAMS_DESCRIPTORS_KEY, keep adding to node namespace
         if (0 == strncmp(PARAMS_KEY, value, strlen(PARAMS_KEY))) {
-          if (0U == ns_tracker->num_node_ns) {
```
bpwilcox: In the current unit tests, it was only testing the case of a node with a namespace. With no namespace, this error occurs when a yaml file includes both parameters (PARAMS_KEY) and descriptors (PARAM_DESCRIPTORS_KEY) under the same scope.

For example, this yaml would cause the error:
```
parameters_node:
  ros__parameters:
    param1: 30
    param3: "hello world"
  ros__parameter_descriptors:
    param1:
      type: 2
      min_value: 5
      max_value: 100
      step: 5
      read_only: false
      description: "int parameter"
      additional_constraints: "only multiples of 5"
    param2:
      min_value: 1.0
      max_value: 10.0
      description: "double parameter"
```

If these lines are removed, then this yaml becomes acceptable and the unit tests still pass. There will still be an error if trying to put a key without a node name (which is expected), but it is not caught here. I'm still trying to determine what should be the proper solution if anyone has some thoughts to add.

rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
```diff
@@ -119,6 +119,17 @@ rcl_variant_t * rcl_yaml_node_struct_get(
   const char * param_name,
   rcl_params_t * params_st);
 
+/// \brief Get the descriptor struct for a given parameter
+/// \param[in] node_name is the name of the node to which the parameter belongs
+/// \param[in] param_name is the name of the parameter whose value is to be retrieved
```
hidmic, line 124: @bpwilcox this needs an update: 
```suggestion
/// \param[in] param_name is the name of the parameter whose descriptor is to be retrieved
```

rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
```diff
@@ -119,6 +119,17 @@ rcl_variant_t * rcl_yaml_node_struct_get(
...
+/// \param[inout] params_st points to the populated (or to be populated) parameter struct
+/// \return parameter descriptor struct on success and NULL on failure
```
hidmic, line 126: @bpwilcox what if no descriptor was specified? what will this function do?

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
hidmic: This comment still applies i.e. no need to have the parameter name in three (3) places.

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
```diff
@@ -95,6 +95,33 @@ typedef struct rcl_node_params_s
...
+  rcl_param_descriptor_t * parameter_descriptors;  ///< Array of corresponding parameter descriptors
+  size_t num_params;  ///< Number of parameters in the node
```
hidmic, line 121: @bpwilcox did you mean `num_descriptors`?

rcl_yaml_param_parser/src/impl/yaml_descriptor.h
```diff
@@ -0,0 +1,51 @@
...
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+///
+/// Finalize an rcl_param_descriptor_t
+///
+RCL_YAML_PARAM_PARSER_PUBLIC
+void rcl_yaml_descriptor_fini(
```
hidmic, line 33: @bpwilcox did you mean `rcl_param_descriptor_fini`? Same elsewhere. YAML is implied.

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
```diff
@@ -95,6 +95,33 @@ typedef struct rcl_node_params_s
...
+  size_t capacity_descriptors;  ///< Capacity of parameters in the node
+} rcl_node_params_descriptors_t;
```
hidmic, line 123: @bpwilcox wouldn't adding an `rcl_param_descriptor_t ** parameter_descriptors;` field to the `rcl_node_params_s` struct simplify the code a bit? Using NULL to indicate when a descriptor is not available, precluding parameter descriptors without an associated parameter value.

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
hidmic: In a related note, I understand pointers are used here to have indication of absence. At the same time, ROS 2 interfaces have no optional fields, so something, somewhere, will choose defaults. If we had all of that here, we could get rid of many heap allocations and simplify the code.

hidmic (commented):
First pass, mostly high-level comments. 

Thanks for picking this up again @bpwilcox !

hidmic:
@bpwilcox friendly ping

rcl_yaml_param_parser/src/impl/yaml_descriptor.h
bpwilcox: This parallels the function naming in `yaml_variant.h` so I just kept it. If we change the name to your suggestion, I'd recommend we do the same for the variant version. I don't know if this is a big concern, however.

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
bpwilcox: I was initially going to put the together as you recommend, however I do not think we would want the parameter descriptors to require association with a parameter value. In my view, it should be perfectly acceptable to have a yaml file with descriptors only and declared parameters in code may comply with these descriptions.

rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
bpwilcox: If the parameter name or node name does not exist for the descriptor, then it will still return a descriptor (not NULL). I found that this is the same behavior as what happens with `rcl_yaml_node_struct_get`.

rcl_yaml_param_parser/include/rcl_yaml_param_parser/types.h
bpwilcox: I removed the name field in `rcl_param_descriptor_t` instead since it made more sense following the parallel logic of the code with the parameter values.

bpwilcox:
Hey @hidmic Could you take another pass through this review since my latest commits?

rcl_yaml_param_parser/test/test_node_params_descriptors.cpp
```diff
@@ -0,0 +1,84 @@
...
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include "../src/impl/node_params_descriptors.h"
+#include "rcutils/allocator.h"
+
+TEST(TestNodeParamsDescriptors, init_fini) {
+  rcutils_allocator_t allocator = rcutils_get_default_allocator();
+  rcl_node_params_descriptors_t node_descriptors = {NULL, NULL, 0u, 128u};
```
hidmic, line 22: @bpwilcox here and elsewhere, why are we not zero initializing `rcl_node_params_descriptors_t` structs?

rcl_yaml_param_parser/test/test_parse_yaml.cpp
```diff
@@ -661,6 +661,203 @@ TEST(test_file_parser, wildcards_partial) {
...
+  ASSERT_TRUE(NULL != copy_of_params_hdl) << rcutils_get_error_string().str;
+  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+  {
+    rcl_yaml_node_struct_fini(copy_of_params_hdl);
+  });
+
+  rcl_params_t * params_hdl_set[] = {params_hdl, copy_of_params_hdl};
+  for (rcl_params_t * params : params_hdl_set) {
+    rcl_param_descriptor_t * param_descriptor = NULL;
+    param_descriptor = rcl_yaml_node_struct_get_descriptor("node_ns/node1", "param1", params);
```
hidmic, line 710: @bpwilcox nit: you may collapse this two lines into one.

rcl_yaml_param_parser/test/test_yaml_descriptor.cpp
```diff
@@ -0,0 +1,111 @@
...
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include "osrf_testing_tools_cpp/scope_exit.hpp"
+#include "../src/impl/yaml_descriptor.h"
+#include "rcutils/allocator.h"
+#include "rcutils/strdup.h"
+
+#define TEST_DESCRIPTOR_COPY(field, tmp_field) \
```
hidmic, line 22: @bpwilcox nit^2: consider `ASSERT_PARAM_DESCRIPTOR_MEMBER_CAN_BE_COPIED()` for a somewhat more accurate macro name

rcl_yaml_param_parser/test/test_yaml_descriptor.cpp
```diff
@@ -0,0 +1,111 @@
...
+  do { \
+    SCOPED_TRACE("TEST_DESCRIPTOR_COPY " #field); \
+    rcl_param_descriptor_t src_descriptor{}; \
+    rcl_param_descriptor_t dest_descriptor{}; \
+    rcutils_allocator_t allocator = rcutils_get_default_allocator(); \
+    src_descriptor.field = &tmp_field; \
```
hidmic, line 28: @bpwilcox nit: you may also do:
```suggestion
    auto tmp_field = field_value;
    src_descriptor.field = &tmp_field; \
```
to not have to add a temporary lvalue for each literal rvalue you want to use for testing.

rcl_yaml_param_parser/test/test_yaml_descriptor.cpp
```diff
@@ -0,0 +1,111 @@
...
+  TEST_DESCRIPTOR_COPY(step_int, tmp_step_int);
+}
+
+TEST(TestYamlDescriptor, copy_string_fields) {
+  // String version is slightly different and can't use the above macro
+  rcl_param_descriptor_t src_descriptor{};
+  rcl_param_descriptor_t dest_descriptor{};
+  rcutils_allocator_t allocator = rcutils_get_default_allocator();
+
+  char * tmp_description = rcutils_strdup("param description", allocator);
```
hidmic, line 87: @bpwilcox nit: following above's suggestion, you may be able to merge this test into the one before by directly passing `rcutils_strdup()` and setting up a scope exit block to finalize `src_descriptor`.

rcl_yaml_param_parser/src/parse.c
```diff
@@ -616,60 +795,151 @@ rcutils_ret_t parse_key(
...
+              break;
+            }
+
+            ret = rem_name_from_ns(ns_tracker, NS_TYPE_NODE, allocator);
+            if (RCUTILS_RET_OK != ret) {
+              RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING(
+                "Internal error adding node namespace at line %d", line_num);
+              break;
+            }
+          }
```
hidmic, line 858: @bpwilcox nit: can we reduce code duplication? `map_level` bumps seem to be the only difference between the last two clauses.

rcl_yaml_param_parser/src/parse.c
```diff
@@ -801,13 +1083,19 @@ rcutils_ret_t parse_file_events(
               yaml_event_delete(&event);
               return RCUTILS_RET_ERROR;
             }
-            if (0U == params_st->params[node_idx].num_params) {
+            if (0U == params_st->params[node_idx].num_params && map_level != 3U) {
```
hidmic, line 1086: @bpwilcox meta: since `ros__parameters` and `ros__parameter_descriptors` are siblings, perhaps we should have another flag to signal which one we're processing as opposed to repurposing `map_level`

hidmic (commented):
Alright, another pass @bpwilcox. Sorry for the delay. Too much to do, and this PR's a lengthy one.

Shubham2S:
Hi, I'm looking forward to use parameter descriptors in the YAML file. What is the status of this PR?

fujitatomoya:
@bpwilcox are you planning to work on this?

Mat198:
What is the status of this PR? Any updates?

fujitatomoya:
it's been 3 years, i think nobody is working on this anymore.

Mat198:
If somebody google this in the future I suggest to try this:

https://github.com/PickNikRobotics/generate_parameter_library

Issue #539, opened 2019-11-28
Segmentation fault in void __cleanup_context(rcl_context_t * context)


BorjaOuterelo:
## Bug report

**Required Info:**

- Operating System: 
  - Ubuntu 18.04.
- Installation type:
  - Binaries.
- Version or commit hash:
  - Dashing, Eloquent, Master.
- DDS implementation:
  - Custom. N/A.
- Client library (if applicable):
  - N/A.

#### Steps to reproduce issue

``` C
  rcl_ret_t rv;

  rcl_init_options_t options = rcl_get_zero_initialized_init_options();
  rv = rcl_init_options_init(&options, rcl_get_default_allocator());
  if (RCL_RET_OK != rv) {
    printf("rcl init options error: %s\n", rcl_get_error_string().str);
    return 1;
  }

  rcl_context_t context = rcl_get_zero_initialized_context();
  rv = rcl_init(argc, argv, &options, &context);
```
Above code in combination with a 'rmw_init` implementation which returns something that is not RMW_RET_OK,

#### Expected behaviour

It should fail gently setting errors accordingly.

#### Actual behavior

Segmentation fault in: https://github.com/ros2/rcl/blob/fbe299af7f96d2d530d55e633ab82a4514d045db/rcl/src/rcl/context.c#L155

#### Additional information

The dump is generated due to allocator being 0.

Combination of 
https://github.com/ros2/rcl/blob/9034c406c2d5bc4ec291f8e1ad4303cde0184cbc/rcl/src/rcl/init.c#L151

but it is setting the allocators in a different line.

https://github.com/ros2/rcl/blob/9034c406c2d5bc4ec291f8e1ad4303cde0184cbc/rcl/src/rcl/init.c#L155

PR #540, opened 2019-11-28
[dashing] Set allocator before goto fail.

BorjaOuterelo:
Fix https://github.com/ros2/rcl/issues/539

It should be added also in master and Eloquent

### DIFF
--- a/rcl/src/rcl/init.c
+++ b/rcl/src/rcl/init.c
@@ -82,6 +82,9 @@ rcl_init(
   // Zero initialize rmw context first so its validity can by checked in cleanup.
   context->impl->rmw_context = rmw_get_zero_initialized_context();
 
+  // Store the allocator.
+  context->impl->allocator = allocator;
+
   // Copy the options into the context for future reference.
   rcl_ret_t ret = rcl_init_options_copy(options, &(context->impl->init_options));
   if (RCL_RET_OK != ret) {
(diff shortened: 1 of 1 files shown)


rcl/src/rcl/init.c
```diff
@@ -82,6 +82,9 @@ rcl_init(
   // Zero initialize rmw context first so its validity can by checked in cleanup.
   context->impl->rmw_context = rmw_get_zero_initialized_context();
 
+  // Store the allocator.
+  context->impl->allocator = allocator;
```
jacobperron: The allocator is also stored below (L154). I think we should remove that redundant logic.

BorjaOuterelo: Right, I already fixed it.

jacobperron:
@BorjaOuterelo I just noticed that this PR is targeting `dashing`. It would be better to target `master` and then we can backport the changes for Eloquent and Dashing.

jacobperron:
I've ported the change to the `master` branch in https://github.com/ros2/rcl/pull/546

BorjaOuterelo:
Shall I be worried for the CI fail?
http://build.ros2.org/job/Epr__rcl__ubuntu_bionic_amd64/77/ It seems more like an infrastructure issue.

jacobperron:
> Shall I be worried for the CI fail?

No. The failing job is for Eloquent (it was accidentally triggered when I temporary changed the target of this PR).

jacobperron:
I'm waiting for #546 to be merged before this one.

PR #541, opened 2019-11-28
should also check if action status publisher is ready.

fujitatomoya:
current action client code just checks if feedback publisher is ready on action server side. it should also check if status publisher is ready on action server side.


fujitatomoya:
@ivanpauno 

this is an easy fix, could you take a look at it?

ivanpauno:
Merging! Thanks for the fix @fujitatomoya !

PR #544, opened 2019-12-03
Feature/remove security directory

BorjaOuterelo:
Fix https://github.com/ros2/rcl/issues/545


jacobperron:
Please fill out the description explaining the change (or at the very least referencing the related issue) for context.

---

I believe there was a past discussion about relocating/refactoring the security directory implementation, but I'm unable to find it at the moment. In any case, I don't think that moving the functionality to `rcutils` is the right thing to do simply because `rcutils` is a place for ROS-agnostic helpers for the ROS code base. Instead, perhaps creating a new package to contain the core security functions makes sense (e.g. `rcl_security`). I'd have to think about it some more.

clalancette:
There's been no response to this in almost a year, so I'm going to go ahead and close it out.  If you'd like to continue with this, please feel free to address @jacobperron 's comments and reopen.

PR #548, opened 2019-12-04
Revert "should also check if action status publisher is ready. (#541)"

ivanpauno:
This reverts commit 80340c8b5ae117bbb36e18485d3f078e56926d3b.

I still think that #541 is reasonable, but I have tested locally that all the [failures](https://ci.ros2.org/view/nightly/job/nightly_linux_release/1383/#showFailuresLink) in `rclpy` were caused by this.

Full CI to double check that reverting actually solves the problem:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8797/)

### DIFF
--- a/rcl_action/src/rcl_action/action_client.c
+++ b/rcl_action/src/rcl_action/action_client.c
@@ -284,13 +284,6 @@ rcl_action_server_is_available(
   }
   *is_available = *is_available && (number_of_publishers != 0);
 
-  ret = rcl_subscription_get_publisher_count(
-    &(client->impl->status_subscription), &number_of_publishers);
-  if (RCL_RET_OK != ret) {
-    return ret;  // error is already set
-  }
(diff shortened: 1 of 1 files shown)


ivanpauno:
@jacobperron Could you double check if #541 was ok or not?
In case the answer is yes, I would rather look for the real cause of the bug than reverting #541 (though, I haven't done that yet).

jacobperron:
@ivanpauno The additional check seems reasonable to me too. The only thing I can think of is that the status publisher uses [different QoS settings](https://github.com/ros2/rcl/blob/3a5c3a34191bd95491cb72c7458b9409452a28f0/rcl_action/include/rcl_action/default_qos.h#L26-L37) than the feedback publisher. Maybe it's a bug in `rcl_subscription_get_publisher_count`? Just a wild guess.

ivanpauno:
Closing in favor of https://github.com/ros2/rclpy/pull/471.

PR #556, opened 2020-01-17
Fix behavior when provided a malformed parameters file

### LINKED ISSUE #553: SIGABRT parsing yaml config file (double free detected)
## Bug report
rclcpp::init(...) causing SIGABRT while parsing config yaml that is well formed yaml but not valid ROS2 parameters format. For example, if yaml has only a single underscore in ros__parameters.

- Operating System: Ubuntu 19.04 on x64 and also reproduced on RPI4 using Ubuntu 19
- Installation type: source
- Version or commit hash: eloquent branch
- DDS implementation: default
- Client library (if applicable): rclcpp

#### Steps to reproduce issue
remove an underscore from ros__parameters. Or have a yaml without node-name, such as just:
`joints: 17`

#### Expected behavior
Should produce an error saying the yaml file is malformed, or at least not crash but node would not receive parameters.

#### Actual behavior
Program produces the following output:
```
free(): double free detected in tcache 2
Aborted (core dumped)

```

#### Additional information
Using `--cmake-args -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS="-fsanitize=address"` produces this output:
```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==24262==ERROR: AddressSanitizer: SEGV on unknown address 0x00002dfffff1 (pc 0x7f3437322962 bp 0x00002dfffff1 sp 0x7ffecf4d24c0 T0)
==24262==The signal is caused by a WRITE memory access.
    #0 0x7f3437322961  (/usr/lib/x86_64-linux-gnu/libasan.so.5+0x28961)
    #1 0x7f3437406014 in __interceptor_free (/usr/lib/x86_64-linux-gnu/libasan.so.5+0x10c014)
    #2 0x7f3436d0a6de in __default_deallocate (/opt/ros2/eloquent/install/rcutils/lib/librcutils.so+0x46de)
    #3 0x7f34367308e8 in rcl_yaml_node_struct_fini (/opt/ros2/eloquent/install/rcl_yaml_param_parser/lib/librcl_yaml_param_parser.so+0x28e8)
    #4 0x7f343676ab8e in rcl_arguments_fini (/opt/ros2/eloquent/install/rcl/lib/librcl.so+0xfb8e)
    #5 0x7f3436769ab0 in rcl_parse_arguments (/opt/ros2/eloquent/install/rcl/lib/librcl.so+0xeab0)
    #6 0x7f3436772f19 in rcl_init (/opt/ros2/eloquent/install/rcl/lib/librcl.so+0x17f19)
    #7 0x7f34370f5d69 in rclcpp::Context::init(int, char const* const*, rclcpp::InitOptions const&) (/opt/ros2/eloquent/install/rclcpp/lib/librclcpp.so+0x3d9d69)
    #8 0x7f3437212fa7 in rclcpp::init(int, char const* const*, rclcpp::InitOptions const&) (/opt/ros2/eloquent/install/rclcpp/lib/librclcpp.so+0x4f6fa7)
    #9 0x55c93cebce1a in main /home/guru/src/humanoid/ros2/humanoid/src/lss_joint_publisher/src/lss_joint_states.cpp:96
    #10 0x7f3436943b6a in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x26b6a)
    #11 0x55c93ceba619 in _start (/home/guru/src/humanoid/ros2/humanoid/build/lss_joint_publisher/lss_joint_states+0xf619)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (/usr/lib/x86_64-linux-gnu/libasan.so.5+0x28961)

### PR DESCRIPTION
**Pull request by jacobperron**
Fixes #553. 

* Don't finalize parameters struct in rcl_parse_yaml_file function
The function assumes that the structure has already been initialized, so it should be the callers
responsibility to finalize it. Otherwise, this may lead to a double free as reported in #553.
* Return an error code if there's a malformed parameters file
This restores behavior that was lost when the '__params:=' syntax was deprecated in #495.
* An extra guard was also added when finalizing the parameters struct.

### DIFF
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -625,8 +625,8 @@ rcl_parse_arguments(
           args_impl->num_param_files_args);
         continue;
       } else if (RCL_RET_ERROR == ret) {
-        // If we return RCL_RET_ERROR then the argument contained the prefix '__params:=',
-        // but the parameter file may be malformed or does not exist.
+        // If _rcl_parse_param_file_rule() returned RCL_RET_ERROR then the argument contained the
+        // '__params:=' prefix, but parsing the parameter file failed.
         goto fail;
       }
       RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME,
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -613,12 +613,17 @@ rcl_parse_arguments(
       ret = _rcl_parse_param_file_rule(
         argv[i], allocator, args_impl->parameter_overrides,
         &args_impl->parameter_files[args_impl->num_param_files_args]);
-      if (RCL_RET_OK == ret) {
-        ++(args_impl->num_param_files_args);
+
+      // Deprecation warning regardless if there is an error parsing the file
+      if (RCL_RET_INVALID_PARAM_RULE != ret) {
         RCUTILS_LOG_WARN_NAMED(ROS_PACKAGE_NAME,
           "Found parameter file rule '%s'. This syntax is deprecated. Use '%s %s %s' instead.",
           argv[i], RCL_ROS_ARGS_FLAG, RCL_PARAM_FILE_FLAG,
-          args_impl->parameter_files[args_impl->num_param_files_args - 1]);
+          args_impl->parameter_files[args_impl->num_param_files_args]);
+      }
+
+      if (RCL_RET_OK == ret) {
+        ++(args_impl->num_param_files_args);
         RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME,
           "params rule : %s\n total num param rules %d",
           args_impl->parameter_files[args_impl->num_param_files_args - 1],
@@ -627,6 +632,12 @@ rcl_parse_arguments(
       } else if (RCL_RET_ERROR == ret) {
         // If _rcl_parse_param_file_rule() returned RCL_RET_ERROR then the argument contained the
         // '__params:=' prefix, but parsing the parameter file failed.
+        rcl_error_string_t prev_error_string = rcl_get_error_string();
+        rcl_reset_error();
+        RCL_SET_ERROR_MSG_WITH_FORMAT_STRING(
+          "Couldn't parse params file: '%s'. Error: %s",
+          args_impl->parameter_files[args_impl->num_param_files_args],
+          prev_error_string.str);
         goto fail;
       }
       RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME,
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -610,11 +610,10 @@ rcl_parse_arguments(
 
       // Attempt to parse argument as parameter file rule
       args_impl->parameter_files[args_impl->num_param_files_args] = NULL;
-      if (
-        RCL_RET_OK == _rcl_parse_param_file_rule(
-          argv[i], allocator, args_impl->parameter_overrides,
-          &args_impl->parameter_files[args_impl->num_param_files_args]))
-      {
+      ret = _rcl_parse_param_file_rule(
+        argv[i], allocator, args_impl->parameter_overrides,
+        &args_impl->parameter_files[args_impl->num_param_files_args]);
+      if (RCL_RET_OK == ret) {
         ++(args_impl->num_param_files_args);
         RCUTILS_LOG_WARN_NAMED(ROS_PACKAGE_NAME,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by jacobperron** [comment]
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9079/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4826/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7430/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9015/)

File: rcl/src/rcl/arguments.c [hunk 1]
```diff
@@ -625,6 +624,10 @@ rcl_parse_arguments(
           args_impl->parameter_files[args_impl->num_param_files_args - 1],
           args_impl->num_param_files_args);
         continue;
+      } else if (RCL_RET_ERROR == ret) {
+        // If we return RCL_RET_ERROR then the argument contained the prefix '__params:=',
+        // but the parameter file may be malformed or does not exist.
```
clalancette: I'd suggest something like:

```
// If _rcl_parse_param_file_rule() returned RCL_RET_ERROR, then the argument contained the '__params:=' prefix,
// but parsing of the parameter file failed.  Get out early here.
```

jacobperron: Refactored: 0937963

jacobperron:
@clalancette I've added a test that fails without this fix. PTAL. 98433bc

rcl/test/rcl/test_arguments.cpp
```diff
@@ -897,6 +897,26 @@ TEST_F(CLASSNAME(TestArgumentsFixture, RMW_IMPLEMENTATION), test_deprecated_para
...
+  const char * argv[] = {
+    "process_name", parameter_rule.c_str()
+  };
+  int argc = sizeof(argv) / sizeof(const char *);
+  rcl_ret_t ret;
+
+  rcl_allocator_t alloc = rcl_get_default_allocator();
+  rcl_arguments_t parsed_args = rcl_get_zero_initialized_arguments();
+
+  ret = rcl_parse_arguments(argc, argv, alloc, &parsed_args);
```
hidmic: @jacobperron nit: why not declaring `ret` on first use?

rcl/src/rcl/arguments.c
```diff
@@ -625,6 +624,10 @@ rcl_parse_arguments(
           args_impl->parameter_files[args_impl->num_param_files_args - 1],
           args_impl->num_param_files_args);
         continue;
+      } else if (RCL_RET_ERROR == ret) {
+        // If _rcl_parse_param_file_rule() returned RCL_RET_ERROR then the argument contained the
+        // '__params:=' prefix, but parsing the parameter file failed.
```
hidmic: @jacobperron would it make sense to provide a hint to the user?

rcl/test/rcl/test_arguments.cpp
jacobperron: The boilerplate of this test was copied from the previous test. I'll leave it since it's the same in all other tests, but it could be refactored if someone wants to do it.

rcl/src/rcl/arguments.c
jacobperron: Good catch. I've copied the error message when when we use the non-deprecated form: f308b49

Also, warning about deprecated usage if there is an error parsing.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9156/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4874/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7477/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9080/)

rcl/test/rcl/test_arguments.cpp
hidmic: Fair enough.

clalancette:
@jacobperron I may be misunderstanding this, but I only see this commit on the Eloquent branch, not the master branch.  If that is the case, we need to port it over to the master branch as well.

jacobperron:
This commit was only intended for the Eloquent branch. The bug was introduced when the old ROS CLI arguments were deprecated (see https://github.com/ros2/rcl/issues/553#issuecomment-575400745). Since the deprecated syntax has since been removed on master, there is nothing to fix.

clalancette:
> This commit was only intended for the Eloquent branch. The bug was introduced when the old ROS CLI arguments were deprecated (see [#553 (comment)](https://github.com/ros2/rcl/issues/553#issuecomment-575400745)). Since the deprecated syntax has since been removed on master, there is nothing to fix.

Ah!  I forgot about that, I was just noticing commits only on this branch.  Thanks for the reminder.

PR #557, opened 2020-01-18
Fix some doc strings regarding no_mangle input parameter

mm318:
This mistake was found while reviewing https://github.com/ros2/rclcpp/pull/960 and https://github.com/ros2/rclpy/pull/454.

The `no_mangle` input parameter is not related to whether or not the inputted topic name is fully qualified, but rather related to whether it is a raw middleware name or a conventional ROS name.


ivanpauno:
Thanks for the PR!

Issue #551, opened 2019-12-09
`rcl_configure_logging` and `rcl_logging_fini` are called multiple times when creating multiple contexts


ivanpauno:
#### Feature description

Currently, if many contexts are created, [rcl_logging_configure](https://github.com/ros2/rcl/blob/024c3ee50d9b79a85665efe1c191cc1225026519/rcl/src/rcl/init.c#L125) is called many times, overriding preexisting configurations silently.
Same happens with [rcl_logging_fini](https://github.com/ros2/rcl/blob/024c3ee50d9b79a85665efe1c191cc1225026519/rcl/src/rcl/init.c#L188). That destroys the logger when the first context is destructed.

The logger should be only initialized once, and initialization of future contexts should be ignored logging a warning (or handling the error somehow).

#### Implementation considerations

To avoid destructing the logger at the first call, an initialization count can be implemented.
The logger should be destroyed after `rcl_logging_destroy` was called the same times that init.

#### Note

Check if something similar may be happening with other "global" configurations.

fujitatomoya:
@ivanpauno @wjwwood 

let me take care of this, will get back to you.

Best

fujitatomoya:
@ivanpauno 

my thoughts are the followings,

- logging confiration is based on global arguments, so should be initalized and finalized once in the process space.
- rcl_logging_configure will be changed into rcl_logging_init to keep the name consistency for rcl_logging_fini.
- Thread-Safe is not supported, just to get along with current implementation.
- reference counter is provided to keep the consistency, how many times initialized and finalized.
- reference counter and implementation is declared in static scope in logging.

just let me know if i got anything missing and wrong.

Best,

ivanpauno:
> let me take care of this, will get back to you.

Sounds good, thanks!

> logging confiration is based on global arguments, so should be initalized and finalized once in the process space.

Yeap, I agree. Alternativly, we could have some context dependent logging configs, but I don't think that's a good idea (and definetly, much more work).

> rcl_logging_configure will be changed into rcl_logging_init to keep the name consistency for rcl_logging_fini.

That sounds reasonable, though it's not required. You can skip that in the first iteration, in order to minimize changes in upstream packages.

> Thread-Safe is not supported, just to get along with current implementation.

I would ensure thread safety in logging init-fini functions, to avoid strange runtime errors.

> reference counter is provided to keep the consistency, how many times initialized and finalized.
> reference counter and implementation is declared in static scope in logging.

[...]

PR #560, opened 2020-01-23
Ensure no redundant rcl_logging initialization and finalization.

fujitatomoya:
this is fix for https://github.com/ros2/rcl/issues/551.

the following, i would like to hear from others.

- Does it have to introduce thread safe? once mutex is introduced, there would be pain for maintenance and possibilities to increase deadlock. (which we do not really like...)
- it seems to be probable we can avoid multiple call for rcl_logging init/init without thread safe. and even if that happens, there would be no such problem for user application.

### DIFF
--- a/rcl/src/rcl/logging.c
+++ b/rcl/src/rcl/logging.c
@@ -71,7 +71,9 @@ rcl_logging_init(const rcl_arguments_t * global_args, const rcl_allocator_t * al
 {
   RCL_CHECK_ARGUMENT_FOR_NULL(global_args, RCL_RET_INVALID_ARGUMENT);
   RCL_CHECK_ARGUMENT_FOR_NULL(allocator, RCL_RET_INVALID_ARGUMENT);
-
+  if (rcutils_atomic_fetch_add_uint64_t(&__rcl_logging_init_count, 1)) {
+    return RCL_RET_OK;
+  }
   RCUTILS_LOGGING_AUTOINIT
     g_logging_allocator = *allocator;
   int default_level = global_args->impl->log_level;
@@ -81,7 +83,6 @@ rcl_logging_init(const rcl_arguments_t * global_args, const rcl_allocator_t * al
   g_rcl_logging_ext_lib_enabled = !global_args->impl->log_ext_lib_disabled;
   rcl_ret_t status = RCL_RET_OK;
   g_rcl_logging_num_out_handlers = 0;
-  (void) rcutils_atomic_fetch_add_uint64_t(&__rcl_logging_init_count, 1);
 
   if (default_level >= 0) {
     rcutils_logging_set_default_logger_level(default_level);
@@ -112,10 +113,14 @@ rcl_logging_init(const rcl_arguments_t * global_args, const rcl_allocator_t * al
 rcl_ret_t rcl_logging_fini()
 {
   rcl_ret_t status = RCL_RET_OK;
-  uint64_t current_count = rcutils_atomic_fetch_add_uint64_t(&__rcl_logging_init_count, -1);
 
-  if (current_count != 1) {
-    return status;
+  if (rcutils_atomic_load_uint64_t(&__rcl_logging_init_count) > 0) {
+    uint64_t current_count = rcutils_atomic_fetch_add_uint64_t(&__rcl_logging_init_count, -1);
+    if (current_count != 1) {
+      return RCL_RET_OK;
+    }
+  } else {
+    return RCL_RET_OK;
   }
 
   rcutils_logging_set_output_handler(rcutils_logging_console_output_handler);
--- a/rcl/src/rcl/logging.c
+++ b/rcl/src/rcl/logging.c
@@ -21,7 +21,6 @@ extern "C"
 #include <inttypes.h>
 #include <stdint.h>
 #include <string.h>
-#include <pthread.h>
 
 #include "./arguments_impl.h"
 #include "rcl/allocator.h"
@@ -32,7 +31,7 @@ extern "C"
 #include "rcl/macros.h"
 #include "rcutils/logging.h"
 #include "rcutils/time.h"
-#include "rcutils/logging_macros.h"
+#include "rcutils/stdatomic_helper.h"
 
 #define RCL_LOGGING_MAX_OUTPUT_FUNCS (4)
 
@@ -45,8 +44,7 @@ static bool g_rcl_logging_stdout_enabled = false;
 static bool g_rcl_logging_rosout_enabled = false;
 static bool g_rcl_logging_ext_lib_enabled = false;
 
-static pthread_mutex_t g_rcl_logging_mutex = PTHREAD_MUTEX_INITIALIZER;
-static bool g_rcl_logging_initialized = false;
+static atomic_uint_least64_t __rcl_logging_init_count = ATOMIC_VAR_INIT(0);
 
 /**
  * An output function that sends to multiple output appenders.
@@ -71,12 +69,6 @@ rcl_logging_ext_lib_output_handler(
 rcl_ret_t
 rcl_logging_init(const rcl_arguments_t * global_args, const rcl_allocator_t * allocator)
 {
-  pthread_mutex_lock(&g_rcl_logging_mutex);
-  if (g_rcl_logging_initialized) {
-    RCUTILS_LOG_WARN_NAMED("rcl.logging", "rcl_logging_init is already done.");
-    pthread_mutex_unlock(&g_rcl_logging_mutex);
-    return RCL_RET_ALREADY_INIT;
-  }
   RCL_CHECK_ARGUMENT_FOR_NULL(global_args, RCL_RET_INVALID_ARGUMENT);
   RCL_CHECK_ARGUMENT_FOR_NULL(allocator, RCL_RET_INVALID_ARGUMENT);
 
@@ -89,6 +81,7 @@ rcl_logging_init(const rcl_arguments_t * global_args, const rcl_allocator_t * al
   g_rcl_logging_ext_lib_enabled = !global_args->impl->log_ext_lib_disabled;
   rcl_ret_t status = RCL_RET_OK;
   g_rcl_logging_num_out_handlers = 0;
+  (void) rcutils_atomic_fetch_add_uint64_t(&__rcl_logging_init_count, 1);
 
   if (default_level >= 0) {
     rcutils_logging_set_default_logger_level(default_level);
@@ -113,24 +106,17 @@ rcl_logging_init(const rcl_arguments_t * global_args, const rcl_allocator_t * al
     }
   }
   rcutils_logging_set_output_handler(rcl_logging_multiple_output_handler);
-
-  if (RCL_RET_OK == status) {
-    g_rcl_logging_initialized = true;
-  }
-  pthread_mutex_unlock(&g_rcl_logging_mutex);
   return status;
 }
 
 rcl_ret_t rcl_logging_fini()
 {
-  pthread_mutex_lock(&g_rcl_logging_mutex);
-  if (!g_rcl_logging_initialized) {
-    RCUTILS_LOG_WARN_NAMED("rcl.logging", "rcl_logging_init is not called yet.");
-    pthread_mutex_unlock(&g_rcl_logging_mutex);
-    return RCL_RET_NOT_INIT;
-  }
-
   rcl_ret_t status = RCL_RET_OK;
+  uint64_t current_count = rcutils_atomic_fetch_add_uint64_t(&__rcl_logging_init_count, -1);
+
+  if (current_count != 1) {
+    return status;
+  }
 
   rcutils_logging_set_output_handler(rcutils_logging_console_output_handler);
 
@@ -140,11 +126,6 @@ rcl_ret_t rcl_logging_fini()
   if (RCL_RET_OK == status && g_rcl_logging_ext_lib_enabled) {
     status = rcl_logging_external_shutdown();
   }
-
-  if (RCL_RET_OK == status) {
-    g_rcl_logging_initialized = false;
-  }
-  pthread_mutex_unlock(&g_rcl_logging_mutex);
   return status;
 }
 
--- a/rcl/src/rcl/logging.c
+++ b/rcl/src/rcl/logging.c
@@ -44,7 +44,7 @@ static bool g_rcl_logging_stdout_enabled = false;
 static bool g_rcl_logging_rosout_enabled = false;
 static bool g_rcl_logging_ext_lib_enabled = false;
 
-static atomic_uint_least64_t __rcl_logging_init_count = ATOMIC_VAR_INIT(0);
+static atomic_uint_least64_t g_rcl_logging_init_count = ATOMIC_VAR_INIT(0);
 
 /**
  * An output function that sends to multiple output appenders.
@@ -71,7 +71,7 @@ rcl_logging_init(const rcl_arguments_t * global_args, const rcl_allocator_t * al
 {
   RCL_CHECK_ARGUMENT_FOR_NULL(global_args, RCL_RET_INVALID_ARGUMENT);
   RCL_CHECK_ARGUMENT_FOR_NULL(allocator, RCL_RET_INVALID_ARGUMENT);
-  if (rcutils_atomic_fetch_add_uint64_t(&__rcl_logging_init_count, 1)) {
+  if (rcutils_atomic_fetch_add_uint64_t(&g_rcl_logging_init_count, 1)) {
     return RCL_RET_OK;
   }
   RCUTILS_LOGGING_AUTOINIT
@@ -114,8 +114,8 @@ rcl_ret_t rcl_logging_fini()
 {
   rcl_ret_t status = RCL_RET_OK;
 
-  if (rcutils_atomic_load_uint64_t(&__rcl_logging_init_count) > 0) {
-    uint64_t current_count = rcutils_atomic_fetch_add_uint64_t(&__rcl_logging_init_count, -1);
+  if (rcutils_atomic_load_uint64_t(&g_rcl_logging_init_count) > 0) {
+    uint64_t current_count = rcutils_atomic_fetch_add_uint64_t(&g_rcl_logging_init_count, -1);
     if (current_count != 1) {
       return RCL_RET_OK;
     }
@@ -131,6 +131,7 @@ rcl_ret_t rcl_logging_fini()
   if (RCL_RET_OK == status && g_rcl_logging_ext_lib_enabled) {
     status = rcl_logging_external_shutdown();
   }
+
   return status;
 }
 
--- a/rcl/src/rcl/logging.c
+++ b/rcl/src/rcl/logging.c
@@ -21,6 +21,7 @@ extern "C"
 #include <inttypes.h>
 #include <stdint.h>
 #include <string.h>
+#include <pthread.h>
 
 #include "./arguments_impl.h"
 #include "rcl/allocator.h"
@@ -31,6 +32,7 @@ extern "C"
 #include "rcl/macros.h"
 #include "rcutils/logging.h"
 #include "rcutils/time.h"
+#include "rcutils/logging_macros.h"
 
 #define RCL_LOGGING_MAX_OUTPUT_FUNCS (4)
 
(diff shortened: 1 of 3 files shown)


rcl/src/rcl/logging.c
```diff
@@ -21,6 +21,7 @@ extern "C"
 #include <inttypes.h>
 #include <stdint.h>
 #include <string.h>
+#include <pthread.h>
```
ivanpauno: This won't work on windows.

rcl/src/rcl/logging.c
```diff
@@ -64,10 +69,17 @@ rcl_logging_ext_lib_output_handler(
   const char * format, va_list * args);
 
 rcl_ret_t
-rcl_logging_configure(const rcl_arguments_t * global_args, const rcl_allocator_t * allocator)
+rcl_logging_init(const rcl_arguments_t * global_args, const rcl_allocator_t * allocator)
 {
+  pthread_mutex_lock(&g_rcl_logging_mutex);
+  if (g_rcl_logging_initialized) {
+    RCUTILS_LOG_WARN_NAMED("rcl.logging", "rcl_logging_init is already done.");
```
ivanpauno: You can't log if logging is not inited.
Use `fprintf` in this case.

rcl/src/rcl/logging.c
```diff
@@ -64,10 +69,17 @@ rcl_logging_ext_lib_output_handler(
   const char * format, va_list * args);
 
 rcl_ret_t
-rcl_logging_configure(const rcl_arguments_t * global_args, const rcl_allocator_t * allocator)
+rcl_logging_init(const rcl_arguments_t * global_args, const rcl_allocator_t * allocator)
 {
+  pthread_mutex_lock(&g_rcl_logging_mutex);
```
ivanpauno: Do we really need a mutex?
Maybe, using an atomic?
We have multiplatform support for atomics https://github.com/ros2/rcutils/blob/cd971e01c18d9646c80d73dec7da94af2a8e489b/include/rcutils/stdatomic_helper.h#L88.

rcl/src/rcl/logging.c
fujitatomoya: my bad, that was my out of radar. got it.

rcl/src/rcl/logging.c
fujitatomoya: got it. i will use fprint.

rcl/src/rcl/logging.c
fujitatomoya: yeah, okay using atomic counter provided rcutils.

and supporting the following case as well.

> we should ensure that the logger is not finished until the last context is finished. So we should probably do some kind of "reference" counting, instead of just a boolean indicating if it was already init or not.

rcl/src/rcl/logging.c
```diff
@@ -77,6 +81,7 @@ rcl_logging_configure(const rcl_arguments_t * global_args, const rcl_allocator_t
   g_rcl_logging_ext_lib_enabled = !global_args->impl->log_ext_lib_disabled;
   rcl_ret_t status = RCL_RET_OK;
   g_rcl_logging_num_out_handlers = 0;
+  (void) rcutils_atomic_fetch_add_uint64_t(&__rcl_logging_init_count, 1);
```
fujitatomoya: if multiple contexts exist, this will be called more than once. so that there should be no warning.

rcl/src/rcl/logging.c
```diff
@@ -107,6 +112,12 @@ rcl_logging_configure(const rcl_arguments_t * global_args, const rcl_allocator_t
 rcl_ret_t rcl_logging_fini()
 {
   rcl_ret_t status = RCL_RET_OK;
+  uint64_t current_count = rcutils_atomic_fetch_add_uint64_t(&__rcl_logging_init_count, -1);
```
fujitatomoya: using atomic is not exactly thread safe, but expecting good enough for this case. besides we do not really want to introduce mutex in here.

rcl/src/rcl/logging.c
```diff
@@ -64,10 +67,11 @@ rcl_logging_ext_lib_output_handler(
   const char * format, va_list * args);
 
 rcl_ret_t
-rcl_logging_configure(const rcl_arguments_t * global_args, const rcl_allocator_t * allocator)
+rcl_logging_init(const rcl_arguments_t * global_args, const rcl_allocator_t * allocator)
 {
   RCL_CHECK_ARGUMENT_FOR_NULL(global_args, RCL_RET_INVALID_ARGUMENT);
   RCL_CHECK_ARGUMENT_FOR_NULL(allocator, RCL_RET_INVALID_ARGUMENT);
+
```
fujitatomoya: unnecessary line.

fujitatomoya:
@ivanpauno 

thanks for enlighten me, could you check once again?

ivanpauno:
> using atomic is not exactly thread safe, but expecting good enough for this case. besides we do not really want to introduce mutex in here.

I think that in this case (doing an atomic fetch add followed by a guard that ensures that the following code is executed only under an specific previous value of the atomic) should be thread safe, right?

fujitatomoya:
@ivanpauno 

i did reconsider the code with expectation with rcl logging init/fini always succeed.
i believe that this should be good enough, but if you want to have 100% exclusive including rcl_logging_init/fini failure cases, i need either rwlock or mutext to keep the other threads wait.

could you share your thoughts?

rcl/src/rcl/logging.c
```diff
@@ -107,6 +112,12 @@ rcl_logging_configure(const rcl_arguments_t * global_args, const rcl_allocator_t
 rcl_ret_t rcl_logging_fini()
 {
   rcl_ret_t status = RCL_RET_OK;
+  uint64_t current_count = rcutils_atomic_fetch_add_uint64_t(&__rcl_logging_init_count, -1);
+
+  if (current_count != 1) {
```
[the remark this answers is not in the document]
ivanpauno: Yes, sorry you're right. But `current_count` sounds confusing, maybe `previous_count`?

rcl/src/rcl/logging.c
```diff
@@ -71,7 +71,9 @@ rcl_logging_init(const rcl_arguments_t * global_args, const rcl_allocator_t * al
 {
   RCL_CHECK_ARGUMENT_FOR_NULL(global_args, RCL_RET_INVALID_ARGUMENT);
   RCL_CHECK_ARGUMENT_FOR_NULL(allocator, RCL_RET_INVALID_ARGUMENT);
-
+  if (rcutils_atomic_fetch_add_uint64_t(&__rcl_logging_init_count, 1)) {
+    return RCL_RET_OK;
```
ivanpauno: I wonder if we should log something here, to not return silently.
Maybe, an user might expect to be able to configure a logging of two contexts differently.

ivanpauno:
> i did reconsider the code with expectation with rcl logging init/fini always succeed.

I think that's fine, I'm not sure if we should log something or in some cases indicate an error.
@hidmic @wjwwood for more feedback.

> i believe that this should be good enough, but if you want to have 100% exclusive including rcl_logging_init/fini failure cases, i need either rwlock or mutext to keep the other threads wait.

I think there is no need of an extra rwlock/mutex, but a second reviewer may think different.

rcl/src/rcl/logging.c
fujitatomoya: okay, that is one of the cases, let's add it to notify the user.

rcl/src/rcl/logging.c
```diff
@@ -64,10 +67,13 @@ rcl_logging_ext_lib_output_handler(
   const char * format, va_list * args);
 
 rcl_ret_t
-rcl_logging_configure(const rcl_arguments_t * global_args, const rcl_allocator_t * allocator)
+rcl_logging_init(const rcl_arguments_t * global_args, const rcl_allocator_t * allocator)
 {
   RCL_CHECK_ARGUMENT_FOR_NULL(global_args, RCL_RET_INVALID_ARGUMENT);
   RCL_CHECK_ARGUMENT_FOR_NULL(allocator, RCL_RET_INVALID_ARGUMENT);
+  if (rcutils_atomic_fetch_add_uint64_t(&__rcl_logging_init_count, 1)) {
```
wjwwood: Why are we counting and not just using a flag? This function is marked as not thread-safe, and `rcl_logging_fini` as well. So we don't have to handle the case where this is called multiple times concurrently. It's up to the code above this to ensure it isn't called in a thread unsafe way.

rcl/src/rcl/logging.c
```diff
@@ -107,6 +113,16 @@ rcl_logging_configure(const rcl_arguments_t * global_args, const rcl_allocator_t
 rcl_ret_t rcl_logging_fini()
 {
   rcl_ret_t status = RCL_RET_OK;
+
+  if (rcutils_atomic_load_uint64_t(&__rcl_logging_init_count) > 0) {
+    uint64_t current_count = rcutils_atomic_fetch_add_uint64_t(&__rcl_logging_init_count, -1);
```
wjwwood: Assuming we will stick with reference counting...

The condition above becomes true if this line is executed when `__rcl_logging_init_count` is zero, which can happen since the above condition and this fetch/add are not an atomic action together. You should probably not use an unsigned int, and you should probably not have a two step check, instead preferring to always subtract and add one later if it ended up being less than zero.

ivanpauno:
@wjwwood I suggested doing reference counting, because the creation/destruction of each context calls `rcl_logging_configure` and `rcl_logging_fini`: https://github.com/ros2/rcl/blob/7537f8304493e9b9568748b5601f5c1b7fa463f0/rcl/src/rcl/init.c#L125
https://github.com/ros2/rcl/blob/7537f8304493e9b9568748b5601f5c1b7fa463f0/rcl/src/rcl/init.c#L188

If in `rclcpp` you do:
```cpp
rclcpp::init(argc, argv);
...
{
  auto context1 = std::make_shared<rclcpp::Context >();
  context1->init(0, nullptr);
}
RCLCPP_INFO(...); // Logging was finished when the above context shutdown.
...
```

I think that's a problem.

rcl/src/rcl/logging.c
```diff
@@ -43,6 +44,8 @@ static bool g_rcl_logging_stdout_enabled = false;
 static bool g_rcl_logging_rosout_enabled = false;
 static bool g_rcl_logging_ext_lib_enabled = false;
 
+static atomic_uint_least64_t __rcl_logging_init_count = ATOMIC_VAR_INIT(0);
```
wjwwood: Don't use `__`, that's reserved by the C/C++ standard. Follow the style in this file and use `g_`.

rcl/src/rcl/logging.c
```diff
@@ -115,7 +131,6 @@ rcl_ret_t rcl_logging_fini()
   if (RCL_RET_OK == status && g_rcl_logging_ext_lib_enabled) {
     status = rcl_logging_external_shutdown();
   }
-
```
wjwwood: nitpick: drop this change, its not needed

rcl/src/rcl/logging.c
wjwwood: Ok, I see why a counter, but why atomic?

rcl/src/rcl/logging.c
wjwwood: I agree, but instead I'd prefer for this to fail instead (return an error code). That way the caller can decide how to react to this. As-is the caller cannot tell if the function succeeded or "succeeded" but didn't use the given settings.

wjwwood: In fact, I think the better solution would look like this (from the caller side):

```c++
Context::Context(...)
{
  // ...
  std::lock_guard<std::mutex> lock(g_logging_mutex);  // ensures thread-safety
  if (rcl_logging_is_initialized()) {
    auto ret = rcl_logging_init(...);
    if (ret != RCL_RET_OK) {
      // error handling
    }
  } else {
    auto ret = rcl_logging_increase_ref_count();  // safe because of lock above
    if (ret != RCL_RET_OK) {
      // error handling
    }
  }
  // ...
}

Context::~Context()
{
  // ...
  std::lock_guard<std::mutex> lock(g_logging_mutex);  // ensures thread-safety
  auto ret = rcl_logging_decrease_ref_count();  // will implicitly fini() if ref_count is zero
  if (ret != RCL_RET_OK) {
    // error handling
  }
  // ...
}
```

So that it is more explicit what's happening.

Or in the case where something is using logging but doesn't have what is required to initialize it (requires something else initialized it first):

```c++
{
  // ...
  std::lock_guard<std::mutex> lock(g_logging_mutex);  // ensures thread-safety
  if (rcl_logging_increase_ref_count() == RCL_LOGGING_NOT_INITIALIZED) {
    // error handling because it was unexpectedly not initialized...
  }
  // ...
}
```

Not sure there's a use case for the second example, but either way I think having init return OK but not actually consider the input arguments is probably not good. But at the same time you need to see if it is initialized and increase the reference count some how.

----

Also, out of scope for this pr, but logging should probably not require the use of global... This is only a problem because we're mapping non-globals (multiple context) to the initialization of a global. If each context could create its own logging instance this would not be a problem.

rcl/src/rcl/logging.c
wjwwood: This comment is not needed anymore since we're not concerned with thread-safety (I thought that was what you guys were doing here), but I think you don't need the atomic in that case.

ivanpauno:
> Also, out of scope for this pr, but logging should probably not require the use of global... This is only a problem because we're mapping non-globals (multiple context) to the initialization of a global. If each context could create its own logging instance this would not be a problem.

I agree, but that would modify completly the logging API. e.g:
```cpp
RCLCPP_INFO(context.get_logger(), ...);
```

And we wouldn't necessary want a logger per `Context`, so having an independent object for it would be reasonable. Something like having a singleton logger that's used automatically, and you can create a new one manually. Logging initialization should probably be decoupled of context initialization.

The question is: who is responsible of configuring the logger?
Currently, each context can "configure" logging, but it's kind of not working as the same globals are shared between different contexts.
I think that doing reference counting is a clever way of fixing the problem without modifying the API, but we should probably reconsider the API later.

----

About the above [solution](https://github.com/ros2/rcl/pull/560#discussion_r373282259) proposed, I think that solving the problem in an upper layer is fine, though we will have to solve the problem independently in each client.
My idea was to solve it in rcl, to avoid duplication. I think there are two ways of solving the problem:
- `rcl_logging_configure` and `rcl_logging_fini` do reference counting internally.
- The usual caller of `rcl_logging_configure` and `rcl_logging_fini` do reference counting, avoiding to call them more than once (the usual caller is `rcl_init` and `rcl_shutdown`).

I think that the second option is probable more reasonable, as the caller should be responsible of not initing the logger more than once.
If we solve the problem in `rcl`, the reference counting should be thread-safe, if not, the upper layer needs a global logging lock (that's fine too, but IMO it's a bit confusing).

So, my idea is something like this:


```c
atomic_uint_least64_t g_logging_ref_count = ATOMIC_VAR_INIT(0);

rcl_init(...)
{
  ...
  if (0u == rcutils_atomic_fetch_add_uint64_t(&g_logging_ref_count, 1)) {
     rcl_logging_configure(...);
  } else {
    // log a warn if logging arguments were passed again.
  }
  ...
} 

rcl_shutdown(...)
{
  ...
  if (1u == rcutils_atomic_fetch_add_uint64_t(&g_logging_ref_count, -1)) {
     rcl_logging_fini(...);
  }
  ...
}
```

@wjwwood Is that a reasonable solution?

wjwwood:
> My idea was to solve it in rcl, to avoid duplication.

That's fine, but...

> If we solve the problem in rcl, the reference counting should be thread-safe...

I don't think that can be done safely without a mutex, and `rcl` should not ever use OS primitives like a mutex, which is part of the design goal of `rcl` to not do threading, and other OS related tasks (see: https://docs.ros2.org/eloquent/developer_overview.html#internal-api-architecture-overview).

Using just atomics will not be sufficient in my opinion, or at least I haven't seen a patch that achieves that.

If you think it can be done, then that's fine I'll review a patch for that, but then the documentation needs to be updated to indicate that it is thread-safe.

----

> the reference counting should be thread-safe, if not, the upper layer needs a global logging lock (that's fine too, but IMO it's a bit confusing).

My opinion is that this will be required, regardless of whether or not it is confusing, though I don't see why it is confusing to be honest.

hidmic:
> Using just atomics will not be sufficient in my opinion, or at least I haven't seen a patch that achieves that.

I don't see why an atomic counter would not ensure thread-safety in this case, though it's arguable whether we should be using them at all along with global state instead of pushing loggers inside contexts. But then the question becomes, does it make sense to have a logger per context? If not, why are contexts initializing logging in the first place? Should logging initialization be done once per process, like signal handlers?

wjwwood:
> I don't see why an atomic counter would not ensure thread-safety in this case,

Well, if you're mutating global state then fini and init cannot be called concurrently. I don't see how the atomic counter can protect you. We can review and discuss a concrete pull request more easily. Do you guys think the current patch is thread-safe? If so I'll re-review it with that in mind, though the doc strings still claim that they are not thread-safe.

> But then the question becomes, does it make sense to have a logger per context?

The implications are annoying, but I think it makes sense. Alternatively, perhaps the context should not initialize the logging, as you suggest.

> If not, why are contexts initializing logging in the first place?

As to "why", init (`rcl_init`) used to be global, and it asked the user to pass the argv/argc too. When it was refactored to allow for non-global init, the argv/argc were kept, so they could be used (optionally) for standard ROS command line stuff, even for non-global context usage. Therefore I guess it made sense to initialize logging in the context since initializing logging also needed argv/argc. However, I don't think it's required for that to be the case, i.e. I think logging could be initialized separately from the context if needed.

> Should logging initialization be done once per process, like signal handlers?

Maybe. The signal handlers are initialized only if the global `rclcpp::init()` is used, otherwise it is up to the user to "install the signal handler". So it would be similar, if the user wanted to avoid globals with the context (`rclcpp::init()`), then they would need to initialize logging manually.

If having library globals is an issue for the user (e.g. due to destruction order issues) then they would just have to forego using our signal handlers and make their own and avoid using the global versions of our functions. For logging this isn't as easy for the user to do since logging calls are made through out the code, so they cannot easily avoid them all.

Also, in my opinion, comparing logging to signal handling is kind of flawed, because the logger could technically be used without globals (think about our deliberations over log4cxx vs spdlog) and ideally our logging API would allow for non-global usage. Conversely, I do not believe the signal handling is technically feasible without using globals.

While important, I think this is well out of scope for this pull request. For this pull request we just need to decide how to make it thread-safe, either in rcl or rclcpp depending on how the atomic issue is resolved.

hidmic:
> because the logger could technically be used without globals (think about our deliberations over log4cxx vs spdlog) and ideally our logging API would allow for non-global usage. Conversely, I do not believe the signal handling is technically feasible without using globals.

For sure. I meant to be initialized on a per process basis. But that doesn't have to be the case really. Logging could be initialized separately, against either global or non-global storage, and made available globally or associated explicitly with contexts or nodes, or simply passed around. Not all of those combinations are sane though.

> For this pull request we just need to decide how to make it thread-safe, either in rcl or rclcpp depending on how the atomic issue is resolved.

Fully agreed.

rcl/src/rcl/logging.c
fujitatomoya: got it, will be fixed into the next commit.

rcl/src/rcl/logging.c
fujitatomoya: done

fujitatomoya:
@wjwwood 

> Do you guys think the current patch is thread-safe?

No, atomic counter is atomic access, it cannot take care of the failure cases. for supporting failure cases, we surely will need mutex or rwlock to keep the everyone else exclusive in specific time window.

fujitatomoya:
@ivanpauno @wjwwood @hidmic 

what you guys really want to proceed? Taking care of thread-safe via rclcpp?

IMOP, i believe that current implementation is reasonable at certain level but expecting logging init and fini never fails.

ivanpauno:
> Well, if you're mutating global state then fini and init cannot be called concurrently. I don't see how the atomic counter can protect you. We can review and discuss a concrete pull request more easily. Do you guys think the current patch is thread-safe? If so I'll re-review it with that in mind, though the doc strings still claim that they are not thread-safe.

I proposed an altarnative in https://github.com/ros2/rcl/pull/573, just to clarify my idea with code.
I think it's not a definitive solution, `rcl_init` should probably not be responsible of initializing the loggers. But it's just an easy way to avoid the double initialization/early finishing, and warn the user when they tried to configure the logger more than once.

ivanpauno:
I've merged the alternative PRs, sorry for the delay.
Thanks for having work on this @fujitatomoya !

Issue #469, opened 2019-07-01
Memory leak in test_subscription__rmw_fastrtps_cpp


vmayoral:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - source
- DDS implementation:
  - FastRTPS

#### Actual behavior

When launching the tests compiled with Asan enabled:

<details><summary><b>test_subscription__rmw_fastrtps_cpp</b> result</summary>

```bash
./test_subscription__rmw_fastrtps_cpp
Running main() from /opt/ros2_asan_ws/install-asan/gtest_vendor/src/gtest_vendor/src/gtest_main.cc
[==========] Running 2 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 2 tests from TestSubscriptionFixture__rmw_fastrtps_cpp
[ RUN      ] TestSubscriptionFixture__rmw_fastrtps_cpp.test_subscription_nominal
[       OK ] TestSubscriptionFixture__rmw_fastrtps_cpp.test_subscription_nominal (1026 ms)
[ RUN      ] TestSubscriptionFixture__rmw_fastrtps_cpp.test_subscription_nominal_string
[       OK ] TestSubscriptionFixture__rmw_fastrtps_cpp.test_subscription_nominal_string (1014 ms)
[----------] 2 tests from TestSubscriptionFixture__rmw_fastrtps_cpp (2040 ms total)

[----------] Global test environment tear-down
[==========] 2 tests from 1 test case ran. (2040 ms total)
[  PASSED  ] 2 tests.

=================================================================
==12118==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 144 byte(s) in 1 object(s) allocated from:
    #0 0x7f9770f9ab50 in __interceptor_malloc (/usr/lib/x86_64-linux-gnu/libasan.so.4+0xdeb50)
    #1 0x7f97706e6465 in __default_allocate /opt/ros2_asan_ws/src/ros2/rcutils/src/allocator.c:35
    #2 0x7f9770c7825d in rcl_subscription_init /opt/ros2_asan_ws/src/ros2/rcl/rcl/src/rcl/subscription.c:153
    #3 0x55979db3bde8 in TestSubscriptionFixture__rmw_fastrtps_cpp_test_subscription_nominal_Test::TestBody() /opt/ros2_asan_ws/src/ros2/rcl/rcl/test/rcl/test_subscription.cpp:135
[...]

PR #564, opened 2020-01-31
Rcl issues 469 test arguments

y-okumura-isp:
Related with #469.

This is PR about problems in test_arguments found by using asan.

- rcl_arguments
  - memory leak and initializtion problem about external_log_config_file handling
- rcl_yaml_param_parser
  - memory leak

There are still 2 problems(seems complicated).

### DIFF
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -1102,7 +1102,8 @@ rcl_arguments_fini(
     }
 
     if (NULL != args->impl->external_log_config_file) {
-      args->impl->allocator.deallocate(args->impl->external_log_config_file, args->impl->allocator.state);
+      args->impl->allocator.deallocate(args->impl->external_log_config_file,
+                                       args->impl->allocator.state);
       args->impl->external_log_config_file = NULL;
     }
 
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -984,6 +984,7 @@ rcl_arguments_copy(
   args_out->impl->parameter_overrides = NULL;
   args_out->impl->parameter_files = NULL;
   args_out->impl->num_param_files_args = 0;
+  args_out->impl->external_log_config_file = NULL;
 
   if (args->impl->num_unparsed_args) {
     // Copy unparsed args
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -338,10 +338,6 @@ rcl_parse_arguments(
     return ret;
   }
   rcl_arguments_impl_t * args_impl = args_output->impl;
-  args_impl->log_level = -1;
-  args_impl->log_stdout_disabled = false;
-  args_impl->log_rosout_disabled = false;
-  args_impl->log_ext_lib_disabled = false;
 
   if (argc == 0) {
     // there are no arguments to parse
@@ -2095,13 +2091,10 @@ _rcl_allocate_initialized_arguments_impl(rcl_arguments_t * args, rcl_allocator_t
     return RCL_RET_BAD_ALLOC;
   }
 
-  // TODO(y-okumura-isp): log_level and log_*disabled are initialized in rcl_parse_arguments()
-  //                      but not in rcl_arguments_copy().
-  //                      Check they are only forgotten or there is some reason.
   rcl_arguments_impl_t * args_impl = args->impl;
   args_impl->num_remap_rules = 0;
   args_impl->remap_rules = NULL;
-  // args_impl->log_level = -1;
+  args_impl->log_level = -1;
   args_impl->external_log_config_file = NULL;
   args_impl->unparsed_args = NULL;
   args_impl->num_unparsed_args = 0;
@@ -2110,9 +2103,9 @@ _rcl_allocate_initialized_arguments_impl(rcl_arguments_t * args, rcl_allocator_t
   args_impl->parameter_overrides = NULL;
   args_impl->parameter_files = NULL;
   args_impl->num_param_files_args = 0;
-  // args_impl->log_stdout_disabled = false;
-  // args_impl->log_rosout_disabled = false;
-  // args_impl->log_ext_lib_disabled = false;
+  args_impl->log_stdout_disabled = false;
+  args_impl->log_rosout_disabled = false;
+  args_impl->log_ext_lib_disabled = false;
   args_impl->allocator = *allocator;
 
   return RCL_RET_OK;
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -962,12 +962,10 @@ rcl_arguments_copy(
 
   rcl_allocator_t allocator = args->impl->allocator;
 
-  do {
-    rcl_ret_t ret = _rcl_allocate_initialized_arguments_impl(args_out, &allocator);
-    if (RCL_RET_OK != ret) {
-      return ret;
-    }
-  } while(false);
+  rcl_ret_t ret = _rcl_allocate_initialized_arguments_impl(args_out, &allocator);
+  if (RCL_RET_OK != ret) {
+    return ret;
+  }
 
   if (args->impl->num_unparsed_args) {
     // Copy unparsed args
@@ -1014,7 +1012,7 @@ rcl_arguments_copy(
     args_out->impl->num_remap_rules = args->impl->num_remap_rules;
     for (int i = 0; i < args->impl->num_remap_rules; ++i) {
       args_out->impl->remap_rules[i] = rcl_get_zero_initialized_remap();
-      rcl_ret_t ret = rcl_remap_copy(
+      ret = rcl_remap_copy(
         &(args->impl->remap_rules[i]), &(args_out->impl->remap_rules[i]));
       if (RCL_RET_OK != ret) {
         if (RCL_RET_OK != rcl_arguments_fini(args_out)) {
@@ -1102,8 +1100,8 @@ rcl_arguments_fini(
     }
 
     if (NULL != args->impl->external_log_config_file) {
-      args->impl->allocator.deallocate(args->impl->external_log_config_file,
-                                       args->impl->allocator.state);
+      args->impl->allocator.deallocate(
+        args->impl->external_log_config_file, args->impl->allocator.state);
       args->impl->external_log_config_file = NULL;
     }
 
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -1117,6 +1117,11 @@ rcl_arguments_fini(
       args->impl->parameter_files = NULL;
     }
 
+    if (NULL != args->impl->external_log_config_file) {
+      args->impl->allocator.deallocate(args->impl->external_log_config_file, args->impl->allocator.state);
+      args->impl->external_log_config_file = NULL;
+    }
+
     args->impl->allocator.deallocate(args->impl, args->impl->allocator.state);
     args->impl = NULL;
     return ret;
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -299,6 +299,16 @@ _atob(
   const char * str,
   bool * val);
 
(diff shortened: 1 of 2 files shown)


rcl_yaml_param_parser/src/parser.c
```diff
@@ -1345,7 +1345,9 @@ static rcutils_ret_t parse_key(
             break;
           }
 
-          ret = rem_name_from_ns(ns_tracker, NS_TYPE_NODE, allocator);
+	  allocator.deallocate(node_name_ns, allocator.state);
+
+	  ret = rem_name_from_ns(ns_tracker, NS_TYPE_NODE, allocator);
```
clalancette: There's some weird indentation happening here; probably tabs instead of spaces.

clalancette:
> Can you please provide the output that asan was complaining about for these problems? It will make it easier for us to understand if this is actually fixing the issue. Thanks.

Ah, sorry, I didn't follow the link.  It's in the attached issue.  I'll continue reviewing.

rcl/src/rcl/arguments.c
```diff
@@ -958,6 +958,7 @@ rcl_arguments_copy(
   args_out->impl->parameter_overrides = NULL;
   args_out->impl->parameter_files = NULL;
   args_out->impl->num_param_files_args = 0;
+  args_out->impl->external_log_config_file = NULL;
```
clalancette: So this is basically duplicating the same code that is at https://github.com/ros2/rcl/blob/7537f8304493e9b9568748b5601f5c1b7fa463f0/rcl/src/rcl/arguments.c#L330 (which is why it was probably missed).

I'm going to suggest that instead of adding this line, we add in a method that is something like `rcl_allocate_and_zero`, which will both do the allocation and setting all of these to default values.  That way, we can't forget something like this in the future.

y-okumura-isp:
@clalancette

Thank you for review.

> Can you please provide the output that asan was complaining about for these problems?

Here is the asan output related with 0c37b206eca59a60392c53718d7182bb6039cfe8 (about external_log_config_file).

```
    #0 0x7fe06c8c4b50 in __interceptor_malloc (/usr/lib/x86_64-linux-gnu/libasan.so.4+0xdeb50)
    #1 0x7fe06c0fc885 in __default_allocate /ros2_asan/src/ros2/rcutils/src/allocator.c:35
    #2 0x7fe06c10f57a in rcutils_strndup /ros2_asan/src/ros2/rcutils/src/strdup.c:42
    #3 0x7fe06c10f4f9 in rcutils_strdup /ros2_asan/src/ros2/rcutils/src/strdup.c:33
    #4 0x7fe06c56d854 in _rcl_parse_external_log_config_file /ros2_asan/src/ros2/rcl/rcl/src/rcl/arguments.c:1959
    #5 0x7fe06c55e41a in rcl_parse_arguments /ros2_asan/src/ros2/rcl/rcl/src/rcl/arguments.c:518
    #6 0x55d5ce713cfe in are_known_ros_args(std::vector<char const*, std::allocator<char const*> >) /ros2_asan/src/ros2/rcl/rcl/test/rcl/test_arguments.cpp:116
    #7 0x55d5ce71cb97 in TestArgumentsFixture__rmw_fastrtps_cpp_check_known_vs_unknown_args_Test::TestBody() /ros2_asan/src/ros2/rcl/rcl/test/rcl/test_arguments.cpp:168
    (snip gtest traces)
```


And for 142ef3532b9e4fca20d0d3fcc88bd970d8260bab (about node_name_ns deallocate).

```
   #1 0x7fcbe6fc9885 in __default_allocate /ros2_asan/src/ros2/rcutils/src/allocator.c:35
   #2 0x7fcbe6fdc57a in rcutils_strndup /ros2_asan/src/ros2/rcutils/src/strdup.c:42
   #3 0x7fcbe6fdc4f9 in rcutils_strdup /ros2_asan/src/ros2/rcutils/src/strdup.c:33
   #4 0x7fcbe71fdcdb in parse_key /ros2_asan/src/ros2/rcl/rcl_yaml_param_parser/src/parser.c:1337
   #5 0x7fcbe71ff055 in parse_file_events /ros2_asan/src/ros2/rcl/rcl_yaml_param_parser/src/parser.c:1481
   #6 0x7fcbe7200751 in rcl_parse_yaml_file /ros2_asan/src/ros2/rcl/rcl_yaml_param_parser/src/parser.c:1694
   #7 0x7fcbe743a706 in _rcl_parse_param_file /ros2_asan/src/ros2/rcl/rcl/src/rcl/arguments.c:1919
   #8 0x7fcbe7429bc1 in rcl_parse_arguments /ros2_asan/src/ros2/rcl/rcl/src/rcl/arguments.c:455
   #9 0x5610d1924cfe in are_known_ros_args(std::vector<char const*, std::allocator<char const*> >) /ros2_asan/src/ros2/rcl/rcl/test/rcl/test_arguments.cpp:116
   #10 0x5610d192c345 in TestArgumentsFixture__rmw_fastrtps_cpp_check_known_vs_unknown_args_Test::TestBody() /ros2_asan/src/ros2/rcl/rcl/test/rcl/test_arguments.cpp:158
   (snip gtest traces)
```

rcl_yaml_param_parser/src/parser.c
y-okumura-isp: Thanks. I fix it (bca8dad79f057385a5756e0ee28a3d45ce447478)

rcl/src/rcl/arguments.c
y-okumura-isp: I have the same opinion and do it (183ad826d14e7257d72913bfee36b6360602176d).

Some variables are set in rcl_parse_arguments() but not in rcl_arguments_copy() such as
- log_level
- log_stdout_disabled
- log_rosout_disabled
- log_ext_lib_disabled

Do you know they are only forgotten or there is some reason?
I left it as is because I don't understand rcl_arguments_copy use case in detail.

clalancette: >     * 
> 
> 
> Do you know they are only forgotten or there is some reason?
> I left it as is because I don't understand rcl_arguments_copy use case in detail.

It's not entirely clear to me, but I believe we want to copy them as well.

y-okumura-isp: @clalancette Thank you for comment. I fixed.

ros-discourse:

https://discourse.ros.org/t/introducing-ros2-sanitizer-report-and-analysis/9287/6

rcl/src/rcl/arguments.c
```diff
@@ -941,23 +936,12 @@ rcl_arguments_copy(
...
-  args_out->impl->num_remap_rules = 0;
-  args_out->impl->remap_rules = NULL;
-  args_out->impl->unparsed_args = NULL;
-  args_out->impl->num_unparsed_args = 0;
-  args_out->impl->unparsed_ros_args = NULL;
-  args_out->impl->num_unparsed_ros_args = 0;
-  args_out->impl->parameter_overrides = NULL;
-  args_out->impl->parameter_files = NULL;
-  args_out->impl->num_param_files_args = 0;
+  do {
```
clalancette: Any particular reason for this `do{ ... } while (false)` loop?

rcl_yaml_param_parser/src/parser.c
```diff
@@ -1345,7 +1345,9 @@ static rcutils_ret_t parse_key(
             break;
           }
 
-          ret = rem_name_from_ns(ns_tracker, NS_TYPE_NODE, allocator);
+          allocator.deallocate(node_name_ns, allocator.state);
+
+	  ret = rem_name_from_ns(ns_tracker, NS_TYPE_NODE, allocator);
```
clalancette: Indentation seems weird here again.

rcl_yaml_param_parser/src/parser.c
```diff
@@ -1345,7 +1345,9 @@ static rcutils_ret_t parse_key(
             break;
           }
 
-          ret = rem_name_from_ns(ns_tracker, NS_TYPE_NODE, allocator);
+          allocator.deallocate(node_name_ns, allocator.state);
```
clalancette: Isn't there another leak up above if `find_node` returns something other than `RCUTILS_RET_OK`?

rcl/src/rcl/arguments.c
y-okumura-isp: Thank you for comment.
I wanted to narrow the scope of "ret".

If "ret" is defined in the function scope, it seems to be rcl_arguemnts_copy return value.
But there are lines which directly return values, ex `return RCL_RET_BAD_ALLOC;` or `return return RCL_RET_OK;`. I feel "why not return ret!?".

As I want to use "ret" very temporarily, so I use `do{ ..} while(false)`

How do you think these ideas?

(1) simply use "ret" (line 991 is also affected)
(2) change variable name such as "_ret".
(3) bellow

```
if (RCL_RET_OK != _rcl_allocate_initialized_arguments_impl(args_out, &allocator)) {
  return ret;
}
```

rcl_yaml_param_parser/src/parser.c
y-okumura-isp: How about `deallocate` right after find_node?

```
ret = find_node(node_name_ns, params_st, node_idx);
allocator.deallocate(node_name_ns, allocator.state);
if (RCUTILS_RET_OK != ret) {
  break;
}
```

rcl/src/rcl/arguments.c
clalancette: The problem with 3) is that you won't have the correct value to return.  You could do assignment in the `if` statement, but I generally find that hard to read.

In this case, I think we just have to accept that this is a limitation of C and just leave `ret` as function scoped.  Since we are reducing the complexity of this function anyway, I don't think it will be hard to read.

rcl_yaml_param_parser/src/parser.c
clalancette: Yeah, that should fix it.

y-okumura-isp:
A supplement to the PR above.

I fixed above change requests.
Also I fixed my old change(arguments.c:1105-1106) because ament_uncrustify claims.
So I rewrote it such as https://github.com/ros2/rcl/blob/master/rcl/src/rcl/arguments.c#L1007.

```
// ament_uncrustify claim
--- src/ros2/rcl/rcl/src/rcl/arguments.c
+++ src/ros2/rcl/rcl/src/rcl/arguments.c.uncrustify
@@ -1104 +1104 @@
-                                       args->impl->allocator.state);
+        args->impl->allocator.state);


// my changes at arguments.c:1105-1106
-      args->impl->allocator.deallocate(args->impl->external_log_config_file,
-                                       args->impl->allocator.state);
+      args->impl->allocator.deallocate(
+        args->impl->external_log_config_file, args->impl->allocator.state);
```

Sorry for not accustomed to ROS 2 coding style, I'm little confused.

I thought [ROS 2 follows Google C++ coding style](https://index.ros.org/doc/ros2/Contributing/Code-Style-Language-Versions/#id2), and in https://google.github.io/styleguide/cppguide.html#Function_Calls,
subsequent lines should be aligned with the first argument.
And my editor with google-c-style alignes as such.
But ament_uncrustify claims second argument should be aligned more shallow...

Anyway I hope it's right change.

clalancette:
> I thought [ROS 2 follows Google C++ coding style](https://index.ros.org/doc/ros2/Contributing/Code-Style-Language-Versions/#id2)

We do use the Google C++ coding style, mostly.  The problem is that uncrustify doesn't totally support that style, so there are sometimes differences between what uncrustify wants and what the Google style would suggest.  As long as uncrustify passes, that is the definitive source for us.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9551/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5254/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7799/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9474/)

clalancette:
All of the test failures seen here were in other builds, so they weren't caused by this PR.  Thus I'm going to go ahead and merge this.  Thanks for the contribution.

PR #573, opened 2020-02-07
Ensure no redundant rcl_logging initialization and finalization (alternative)

ivanpauno:
Alternative to https://github.com/ros2/rcl/pull/560.

### DIFF
--- a/rcl/include/rcl/logging.h
+++ b/rcl/include/rcl/logging.h
@@ -90,23 +90,6 @@ RCL_PUBLIC
 RCL_WARN_UNUSED
 bool rcl_logging_rosout_enabled();
 
-/// Return `true` if argv contains logging arguments.
-/**
- * Attribute          | Adherence
- * ------------------ | -------------
- * Allocates Memory   | No
- * Thread-Safe        | No
- * Uses Atomics       | No
- * Lock-Free          | Yes
- *
- * \param argc
- * \param argv can be `NULL`. In that case, the return value is `false`.
- */
-RCL_PUBLIC
-RCL_WARN_UNUSED
-bool
-rcl_contains_logging_arguments(int argc, char const * const * argv);
-
 #ifdef __cplusplus
 }
 #endif
--- a/rcl/include/rcl/logging.h
+++ b/rcl/include/rcl/logging.h
@@ -90,6 +90,23 @@ RCL_PUBLIC
 RCL_WARN_UNUSED
 bool rcl_logging_rosout_enabled();
 
+/// Return `true` if argv contains logging arguments.
+/**
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param argc
+ * \param argv can be `NULL`. In that case, the return value is `false`.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+bool
+rcl_contains_logging_arguments(int argc, char const * const * argv);
+
 #ifdef __cplusplus
 }
 #endif
--- a/rcl/src/rcl/init.c
+++ b/rcl/src/rcl/init.c
@@ -19,19 +19,24 @@ extern "C"
 
 #include "rcl/init.h"
 
+#include "rcl/arguments.h"
+#include "rcl/error_handling.h"
+#include "rcl/logging.h"
+
+#include "rcutils/logging_macros.h"
+#include "rcutils/stdatomic_helper.h"
+
+#include "rmw/error_handling.h"
+
+#include "tracetools/tracetools.h"
+
 #include "./arguments_impl.h"
 #include "./common.h"
 #include "./context_impl.h"
 #include "./init_options_impl.h"
-#include "rcl/arguments.h"
-#include "rcl/error_handling.h"
-#include "rcl/logging.h"
-#include "rcutils/logging_macros.h"
-#include "rcutils/stdatomic_helper.h"
-#include "rmw/error_handling.h"
-#include "tracetools/tracetools.h"
 
-static atomic_uint_least64_t __rcl_next_unique_id = ATOMIC_VAR_INIT(1);
+static atomic_uint_least64_t g_rcl_next_unique_id = ATOMIC_VAR_INIT(1);
+static atomic_uint_least64_t g_logging_ref_count = ATOMIC_VAR_INIT(0);
 
 rcl_ret_t
 rcl_init(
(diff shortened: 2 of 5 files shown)


rcl/src/rcl/logging.c
```diff
@@ -201,6 +201,37 @@ rcl_logging_ext_lib_output_handler(
   }
 }
 
+bool
+rcl_contains_logging_arguments(int argc, char const * const * argv)
```
hidmic: @ivanpauno I'm not particularly fond of coupling logging and arguments this much. But I'm a bit divided, as I can imagine why you chose keeping this function here. @wjwwood thoughts?

rcl/src/rcl/init.c
```diff
@@ -184,13 +201,22 @@ rcl_shutdown(rcl_context_t * context)
...
-  rcl_ret_t rcl_ret = rcl_logging_fini();
-  RCUTILS_LOG_ERROR_EXPRESSION_NAMED(
-    RCL_RET_OK != rcl_ret, ROS_PACKAGE_NAME,
-    "Failed to fini logging, rcl_ret_t: %i, rcl_error_str: %s", rcl_ret,
-    rcl_get_error_string().str);
-  rcl_reset_error();
+  size_t previous_logging_ref_count =
+    rcutils_atomic_fetch_add_uint64_t(&g_logging_ref_count, -1);
+
+  if (1u == previous_logging_ref_count) {
```
hidmic: @ivanpauno it'd be nice to double check if `rcutils` atomics include memory barriers as well. I've mostly assumed so, but just in case.

rcl/include/rcl/logging.h
```diff
@@ -90,6 +90,23 @@ RCL_PUBLIC
 RCL_WARN_UNUSED
 bool rcl_logging_rosout_enabled();
 
+/// Return `true` if argv contains logging arguments.
+/**
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
```
hidmic: @ivanpauno why is it not thread safe?

rcl/src/rcl/init.c
ivanpauno: Yes: 
- no windows: https://github.com/ros2/rcutils/blob/d7bd0575d8eda3e2b19757350cb9169ffbbbcae4/include/rcutils/stdatomic_helper/gcc/stdatomic.h#L368,
- windows:
https://github.com/ros2/rcutils/blob/d7bd0575d8eda3e2b19757350cb9169ffbbbcae4/include/rcutils/stdatomic_helper/win32/stdatomic.h#L270
https://docs.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-interlockedexchangeadd64

rcl/include/rcl/logging.h
ivanpauno: Bad copy-past, will fix it.

rcl/src/rcl/logging.c
ivanpauno: Well, I think that moving to `arguments.c` makes sense.

rcl/src/rcl/init.c
```diff
@@ -122,23 +127,35 @@ rcl_init(
...
-  ret = rcl_logging_configure(&context->global_arguments, &allocator);
-  if (RCL_RET_OK != ret) {
-    fail_ret = ret;
-    RCUTILS_LOG_ERROR_NAMED(
-      ROS_PACKAGE_NAME,
-      "Failed to configure logging: %s",
-      rcutils_get_error_string().str);
-    goto fail;
+  if (0u == rcutils_atomic_fetch_add_uint64_t(&g_logging_ref_count, 1)) {
+    ret = rcl_logging_configure(&context->global_arguments, &allocator);
```
wjwwood: ⚡️ context switch here, then:

- in thread 2, start `rcl_shutdown()`
  - my theory is that `rcl_shutdown()` fill fail because though `g_logging_ref_count` is > 0, trying to do `rcl_logging_fini()` will fail as it has not yet been configured

My conclusion is that `rcl_init()` is not thread-safe with `rcl_shutdown()`.

rcl/src/rcl/logging.c
wjwwood: I don't have a strong opinion.

rcl/src/rcl/init.c
ivanpauno: mmm, I see what you meant now.

So, I always considered that the upper layer will ensure that each `context` calls `rcl_init` and `rcl_shutdown` in a thread safe manner (`rcl_init` and `rcl_shutdown` are still no thread-safe).
That's the case now in [rclcpp](https://github.com/ros2/rclcpp/blob/7c1721a0b390be8242a6b824489d0bc861f6a0ad/rclcpp/src/rclcpp/context.cpp#L85) (it's not the case in [rclpy](https://github.com/ros2/rclpy/blob/822f19406de40f9e7df0c73aa4d52837a42bcfde/rclpy/rclpy/__init__.py#L73), but that's definitely wrong).

Under that situation, using an atomic count adds protection between `rcl_init` and `rcl_shutdown` calls on different objects, and avoids the need of a "global" mutex.[*]

My main reason for pushing this alternative, it's that it solves the problem in a more "minimalistic" way (only one PR, without adding functions like `rcl_logging_increase_ref_count` that we might want to deprecate later). It's easily to backport too.

----

[*] If calls in each context between `rcl_init` and `rcl_shutdown` are protected by an upper layer, the following situations are possible:
- Only one context: It's thread safe because the upper layer ensures that.
- Two contexts, init1-init2-shutdown1-shutdown2: The atomic ensures thread safety between init1 and init 2. The upper layer ensures thread safety between init1 and shutdown1.
- Two contexts, init1-shutdown2-shutdown1: Calling shutdown in a non-inited context is [handled correctly](https://github.com/ros2/rcl/blob/19d53bc9e5110f4b66f5f57cce16fd5404daa18b/rcl/src/rcl/init.c#L174).

wjwwood: But what about this case:

init1-shutdown1-init2-...

Where shutdown1 is interrupted after adjusting the atomic count, but before it does log fini, and context switches to init2 which then increments it, finding that according to the count logging was not initialized, and tries to initialize logging when it already has been. Either the "initialize while already initialized" case will fail, or if that silently passes, when shutdown1 continues it will shutdown logging and leave init2 initialized but without logging.

I remain unconvinced that this is thread-safe even when using separate context objects.

ivanpauno: Oh :man_facepalming:, I completed miss that case. Thanks William!

wjwwood: It's ok, this stuff is tricky. I've only learned to be very suspicious of creating thread-safety with just atomics after messing it up many, many times. :)

Maybe it's possible to salvage this approach with enough iteration, but honestly I think a global mutex provided by rclpy/rclcpp is the safest thing to do to protect the init/shutdown and therefore logging init.

ivanpauno: > Maybe it's possible to salvage this approach with enough iteration, but honestly I think a global mutex provided by rclpy/rclcpp is the safest thing to do to protect the init/shutdown and therefore logging init.

Yes, I agree.
@fujitatomoya do you want to iterate and open PRs in rclpy and rclcpp? If not, I can take it.

fujitatomoya: @ivanpauno 

you can go ahead to take them! thanks for your effort!

PR #580, opened 2020-02-20
Clock API improvements

hidmic:
This pull requests improves `rcl_clock_*` API documentation and adds missing error checks.

*Note:* to be rebased, not squashed, before merging so both commits are kept.

### DIFF
--- a/rcl/include/rcl/time.h
+++ b/rcl/include/rcl/time.h
@@ -173,11 +173,13 @@ rcl_clock_valid(rcl_clock_t * clock);
  * Attribute          | Adherence
  * ------------------ | -------------
  * Allocates Memory   | Yes [1]
- * Thread-Safe        | No
+ * Thread-Safe        | No [2]
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
  * <i>[1] If `clock_type` is `RCL_ROS_TIME`</i>
+ * <i>[2] Function is reentrant, but concurrent calls on the same `clock` object are not safe.
+ *        Thread-safety is also affected by that of the `allocator` object.</i>
  *
  * \param[in] clock_type the type identifying the time source to provide
  * \param[in] clock the handle to the clock which is being initialized
@@ -208,9 +210,13 @@ rcl_clock_init(
  * Attribute          | Adherence
  * ------------------ | -------------
  * Allocates Memory   | No
- * Thread-Safe        | No
+ * Thread-Safe        | No [1]
  * Uses Atomics       | No
  * Lock-Free          | Yes
+ *
+ * <i>[1] Function is reentrant, but concurrent calls on the same `clock` object are not safe.
+ *        Thread-safety is also affected by that of the `allocator` object associated with the
+ *        `clock` object.</i>
  *
  * \param[in] clock the handle to the clock which is being finalized
  * \return `RCL_RET_OK` if the time source was successfully finalized, or
@@ -232,9 +238,12 @@ rcl_clock_fini(
  * Attribute          | Adherence
  * ------------------ | -------------
  * Allocates Memory   | Yes
- * Thread-Safe        | No
+ * Thread-Safe        | No [1]
  * Uses Atomics       | No
  * Lock-Free          | Yes
+ *
+ * <i>[2] Function is reentrant, but concurrent calls on the same `clock` object are not safe.
+ *        Thread-safety is also affected by that of the `allocator` object.</i>
  *
  * \param[in] clock the handle to the clock which is being initialized
  * \param[in] allocator The allocator to use for allocations
@@ -262,9 +271,13 @@ rcl_ros_clock_init(
  * Attribute          | Adherence
  * ------------------ | -------------
  * Allocates Memory   | No
- * Thread-Safe        | No
+ * Thread-Safe        | No [1]
  * Uses Atomics       | No
  * Lock-Free          | Yes
+ *
+ * <i>[1] Function is reentrant, but concurrent calls on the same `clock` object are not safe.
+ *        Thread-safety is also affected by that of the `allocator` object associated with the
+ *        `clock` object.</i>
  *
  * \param[in] clock the handle to the clock which is being initialized
  * \return `RCL_RET_OK` if the time source was successfully finalized, or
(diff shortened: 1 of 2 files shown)


rcl/include/rcl/time.h
```diff
@@ -255,6 +310,18 @@ rcl_steady_clock_init(
  * It is specifically setting up a steady time source. It is expected to be
  * paired with the init fuction.
  *
+ * This function is not thread-safe with any other function operating on the same
+ * clock object.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
```
dirk-thomas: The `Yes` needs a `[1]` note.

rcl/include/rcl/time.h
```diff
@@ -161,6 +169,15 @@ rcl_clock_valid(rcl_clock_t * clock);
 /**
  * This will allocate all necessary internal structures, and initialize variables.
  *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes [1]
+ * Thread-Safe        | No
```
dirk-thomas: Can you clarify why the function isn't (same for the others)?

rcl/include/rcl/time.h
```diff
@@ -183,6 +200,17 @@ rcl_clock_init(
  * Passing a clock with type RCL_CLOCK_UNINITIALIZED will result in
  * RCL_RET_INVALID_ARGUMENT being returned.
  *
+ * This function is not thread-safe with any other function operating on the same
+ * clock object.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
```
dirk-thomas: Where?

rcl/include/rcl/time.h
```diff
@@ -349,6 +454,18 @@ rcl_clock_get_now(rcl_clock_t * clock, rcl_time_point_value_t * time_point_value
...
+ * nor `rcl_clock_remove_jump_callback` functions when used on the same
+ * clock object.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
```
dirk-thomas: Should we mention that this only covers the function itself? The callbacks invoked by this function could do anything.

Same for `rcl_disable_ros_time_override` and `rcl_set_ros_time_override`.

rcl/include/rcl/time.h
hidmic: Good catch.

rcl/include/rcl/time.h
hidmic: I was unsure about this actually, as I have not found an explicit statement of what we mean by thread-safe here. I meant it in the strictest sense i.e. concurrent access to the same object (which is an argument purely because of a language limitation) is safe. If by thread-safe we mean concurrent access for different objects or with no regard for objects is safe (i.e. no shared, global state), then all these functions here are, including those that add and remove callbacks. 

I wonder if `Yes` and `No` for values are enough. Maybe a third `Conditional` value would help us in being a bit more precise.

rcl/include/rcl/time.h
hidmic: Nowhere, wrong value. Fixed in 03d4a769cf9515c6101a32b27c381af9322d6706.

rcl/include/rcl/time.h
hidmic: Good point. Done in 03d4a769cf9515c6101a32b27c381af9322d6706.

rcl/include/rcl/time.h
dirk-thomas: Maybe @wjwwood can clarify how the `Thread-Safe` flag is used throughout the existing API.

clalancette: I don't really think we need a third state for Thread-Safe.  This one clearly isn't thread-safe; if you pass the same object into this method from two threads at the same time, that will leave the `clock` object in an undefined state.  So it looks to me like "Thread-Safe: No" is appropriate here.  The same goes for all of the APIs below.

hidmic: I thought the same, but it's also true that there're varying degrees of thread (un)safety. You can manage different clocks in different threads without issues. Functions that access global state don't even allow that.

dirk-thomas: And I think that kind of differentiated information would be very valuable for the user.

hidmic: @wjwwood bump

wjwwood: Sorry, I thought I replied to this, but maybe I was dreaming, I was talking about it with the Apex folks recently too.

The idea is that you can call it from multiple threads concurrently, but also generally that it isn't affected by other functions being called concurrently. If there's any threading considerations like "can be called concurrently but not on the same object" or "can be called concurrently but not concurrently with fini on the same object, etc...".

Basically, I consider the footnote to be more important than the Yes or No. If you say "Yes" without a note, that would mean to me that the function is most likely pure functional and has no global, or input/output, side effects. If it accesses global state or modifies an object passed by non-const pointer, then you should consider having a No or Yes with a note.

hidmic: Alright, so it could be either so long as there's a note. I'll add one.

hidmic: There, see ca78af86b84e3f7ab84ac08300aa341dc2a3e36a. It's a bit verbose but it is accurate.

rcl/include/rcl/time.h
```diff
@@ -161,6 +169,18 @@ rcl_clock_valid(rcl_clock_t * clock);
...
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes [1]
+ * Thread-Safe        | No [2]
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * <i>[1] If `clock_type` is `RCL_ROS_TIME`</i>
+ * <i>[2] Function is reentrant, but concurrent calls on the same `clock` object are not safe.
+ *        Thread-safety is also affected by that of the `allocator` object.</i>
```
wjwwood: Maybe we should make it clear elsewhere but the allocator calls are assumed to be thread safe. For now I think this can stay but it’s basically true everywhere we use an allocator but I don’t know if we mention it explicitly like this.

hidmic: Yeah, I don't think we're being this explicit elsewhere. To avoid so much repeat though, I wonder if we'd be better off documenting a basic thread-safety target for the library to attain e.g. _all functions are reentrant but concurrent use of the same objects isn't safe_, and then further qualifying it on a case-by-case basis e.g. the default allocator can be used concurrently, this function is fully thread-safe, this function is not reentrant, etc.

hidmic:
Thanks for the reviews guys! Merging.

Issue #581, opened 2020-02-26
Expose QoS settings for /rosout publisher


de-vri-es:
I have noticed that sometimes during node startup, some messages don't arrive at `/rosout` subscribers. I could never replicate this on my own Linux hardware, but I have witnessed it on a Windows PC. Log messages from the first 30 to 100ms would not arrive at a subscriber.

I imagine this could be fixed by changing the QoS settings of the publisher. However, I don't believe we have access to the publisher or it's QoS settings at the moment. If it is true that changing the QoS settings could prevent loosing messages during startup (within reason), could these settings be exposed?

Log messages during node startup can be especially interesting, so it's not nice to loose them.

dirk-thomas:
While exposing the QoS settings would be great I think the default still needs to be different. Currently it uses `rmw_qos_profile_default` but something like `rmw_qos_profile_parameters` which changes the depth to 1000 and the duratbility to transient local would be more appropriate.

See #582 for the proposal. @de-vri-es It would be great if you could give this patch a try and comment if that resolves your issue.

de-vri-es:
That does sound like a good idea. I'm not very familiar with compiling on Windows, which is where I witnessed the issue, but I'll try my best to test it.

ivanpauno:
Deleted the `bug` label after https://github.com/ros2/rcl/pull/582 and #587, the defaults qos are now nicer.
I will left this open, as exposing the qos would still be an improvement.

de-vri-es:
Sounds good. Thanks all for the quick action  :)

fujitatomoya:
@dirk-thomas @ivanpauno @de-vri-es 
CC: @Barry-Xu-2018 

anybody working on this? if not, we could take care of this thread.

ivanpauno:
> anybody working on this? if not, we could take care of this thread.

AFAIK, nobody is working on this.

Usually, people assign themselves when they are working on something (though they might forget :joy:).
We don't usually address issues marked as `help wanted`. Of course, PRs targeting a `help wanted` issue are more than welcomed.

Ada-King:
Hi, @ivanpauno @fujitatomoya  I made a PR for this issue, now only rcl/rmw layer was implemented. As for the rclcpp layer, i have already finished a design draft, which we can discuss later, For now, can you take a quick look for [253](https://github.com/ros2/rmw/pull/253) && [722](https://github.com/ros2/rcl/pull/722) in your spare time?

ivanpauno:
[...]

PR #582, opened 2020-02-26
make rosout publisher transient local with a depth of 1000

dirk-thomas:
Addresses changing the default QoS described in #581.

### DIFF
--- a/rcl/src/rcl/logging_rosout.c
+++ b/rcl/src/rcl/logging_rosout.c
@@ -173,6 +173,8 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(
   const rosidl_message_type_support_t * type_support =
     rosidl_typesupport_c__get_message_type_support_handle__rcl_interfaces__msg__Log();
   rcl_publisher_options_t options = rcl_publisher_get_default_options();
+  options.qos.depth = 1000;
+  options.qos.durability = RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL;
   new_entry.publisher = rcl_get_zero_initialized_publisher();
   status =
     rcl_publisher_init(&new_entry.publisher, node, type_support, ROSOUT_TOPIC_NAME, &options);


rcl/src/rcl/logging_rosout.c
```diff
@@ -173,6 +173,8 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(
   const rosidl_message_type_support_t * type_support =
     rosidl_typesupport_c__get_message_type_support_handle__rcl_interfaces__msg__Log();
   rcl_publisher_options_t options = rcl_publisher_get_default_options();
+  options.qos.depth = 1000;
```
ivanpauno: why 1000? why don't just using KEEP_ALL?

dirk-thomas: Since that might result in infinite memory usage? The goal is to cover the race to allow subscribers coming up shortly after a node starts publishing rosout messages to receive - not a for alate joiner an hour later to get all rosout message from the beginning.

hidmic: I actually agree with your previous comment https://github.com/ros2/rcl/issues/581#issuecomment-591519046, exposing this to the user would be ideal and likely the best solution. But this isn't an unreasonable default.

wjwwood: > Since that might result in infinite memory usage?

That's not how `KEEP_ALL` works exactly. There's a DDS setting we don't currently expose, `MAX_SAMPLES` and `MAX_SAMPLES_PER_INSTANCE` which determines when `KEEP_ALL` stops storing things in the history cache.

wjwwood:
Is transient local what you want? Imagine this scenario:

- you start a GUI showing rosout messages
- start a node that produces ~20 rosout messages
- several hours go by
- you close the GUI and open it again

Is it reasonable to now receive the (now hours old) original messages? In the meantime the node has had to store these ~20 messages forever. Now imagine it published 1000+, now it has been storing 1000 messages the whole time and resending all of them each time you open the GUI.

Is it not already the case that in ROS 1 if you open a rosout viewer that it only sees messages which occur after opening the GUI (and waiting for discovery and/or negotiation of endpoints)?

It sounds like to me that it's a consequence of the middleware's behavior. Personally, I think transient local is not a great default, and instead users can/should turn it on as they need it in certain situations.

rcl/src/rcl/logging_rosout.c
dirk-thomas: How does `KEEP_ALL` currently behave then without these additional options exposed or set to custom values?

ivanpauno:
Maybe, making use of lifespan qos policy combined with the current qos settings (transient local, depth 100) would make sense.

hidmic:
> If not, just reverting to the default qos and exposing the qos settings.

This. Exposing would be best, then we can go and guess what is reasonable for our user base as a default.

dirk-thomas:
> Maybe, making use of lifespan qos policy combined with the current qos settings (transient local, depth 100) would make sense.

That sounds interesting. Please consider to create a pull request for this suggestions.

wjwwood:
I like the lifespan idea too.

PR #584, opened 2020-02-27
Fix a bug that goal can be gone before client gets result.

Barry-Xu-2018:
Allow state change from success to abort while goal expired.

Related issue https://github.com/ros2/rclcpp/issues/861


Barry-Xu-2018:
> I think we can avoid changing the state machine, see [ros2/rclcpp#1005 (review)](https://github.com/ros2/rclcpp/pull/1005#pullrequestreview-367546980) for an explanation.

Agree.  
I close this PR.

PR #591, opened 2020-03-10
Waitset timestamp fields

iluetkeb:
Both the standard wait_set and internal structures got the timestamps.
Currently, the fields are not used.

Also see https://github.com/ros2/rmw/pull/200

### DIFF
--- a/rcl/include/rcl/wait.h
+++ b/rcl/include/rcl/wait.h
@@ -22,6 +22,7 @@ extern "C"
 
 #include <stdbool.h>
 #include <stddef.h>
+#include <rcutils/time.h>
 
 #include "rcl/client.h"
 #include "rcl/guard_condition.h"
@@ -33,6 +34,7 @@ extern "C"
 #include "rcl/types.h"
 #include "rcl/visibility_control.h"
 
+
 struct rcl_wait_set_impl_t;
 
 /// Container for subscription's, guard condition's, etc to be waited on.
@@ -40,18 +42,22 @@ typedef struct rcl_wait_set_t
 {
   /// Storage for subscription pointers.
   const rcl_subscription_t ** subscriptions;
+  rcutils_time_point_value_t * subscriptions_timestamps;
   size_t size_of_subscriptions;
   /// Storage for guard condition pointers.
   const rcl_guard_condition_t ** guard_conditions;
   size_t size_of_guard_conditions;
   /// Storage for timer pointers.
   const rcl_timer_t ** timers;
+  rcutils_time_point_value_t * timers_timestamps;
   size_t size_of_timers;
   /// Storage for client pointers.
   const rcl_client_t ** clients;
+  rcutils_time_point_value_t * clients_timestamps;
   size_t size_of_clients;
   /// Storage for service pointers.
   const rcl_service_t ** services;
+  rcutils_time_point_value_t * services_timestamps;
   size_t size_of_services;
   /// Storage for event pointers.
   const rcl_event_t ** events;
--- a/rcl/include/rcl/wait.h
+++ b/rcl/include/rcl/wait.h
@@ -22,6 +22,7 @@ extern "C"
 
 #include <stdbool.h>
 #include <stddef.h>
+
 #include <rcutils/time.h>
 
 #include "rcl/client.h"
@@ -33,7 +34,6 @@ extern "C"
 #include "rcl/event.h"
 #include "rcl/types.h"
 #include "rcl/visibility_control.h"
-
 
 struct rcl_wait_set_impl_t;
 
--- a/rcl/include/rcl/wait.h
+++ b/rcl/include/rcl/wait.h
@@ -272,6 +272,10 @@ rcl_wait_set_clear(rcl_wait_set_t * wait_set);
  *
  * This can be called on an uninitialized (zero initialized) wait set.
  *
+ * If RCL_RET_BAD_ALLOC is returned, i.e. allocation failed, the wait_set
+ * may be *partially* allocated and therefore, it must still be cleaned up
+ * with _fini.
+ *
  * <hr>
  * Attribute          | Adherence
  * ------------------ | -------------
--- a/rcl/src/rcl/wait.c
+++ b/rcl/src/rcl/wait.c
@@ -33,32 +33,7 @@ extern "C"
 
 #include "./context_impl.h"
 
-typedef struct rcl_wait_set_impl_t
-{
-  // number of subscriptions that have been added to the wait set
-  size_t subscription_index;
-  rmw_subscriptions_t rmw_subscriptions;
-  // number of guard_conditions that have been added to the wait set
-  size_t guard_condition_index;
-  rmw_guard_conditions_t rmw_guard_conditions;
-  // number of clients that have been added to the wait set
-  size_t client_index;
-  rmw_clients_t rmw_clients;
-  // number of services that have been added to the wait set
-  size_t service_index;
-  rmw_services_t rmw_services;
-  // number of events that have been added to the wait set
-  size_t event_index;
-  rmw_events_t rmw_events;
-
-  rmw_wait_set_t * rmw_wait_set;
-  // number of timers that have been added to the wait set
-  size_t timer_index;
-  // context with which the wait set is associated
-  rcl_context_t * context;
-  // allocator used in the wait set
-  rcl_allocator_t allocator;
-} rcl_wait_set_impl_t;
+#include "./wait_set_impl.h"
 
 rcl_wait_set_t
 rcl_get_zero_initialized_wait_set()
--- a/rcl/src/rcl/wait.c
+++ b/rcl/src/rcl/wait.c
@@ -710,6 +710,8 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
       is_ready, ROS_PACKAGE_NAME, "Subscription in wait set is ready");
     if (!is_ready) {
       wait_set->subscriptions[i] = NULL;
+    } else {
+      wait_set->subscriptions_timestamps[i] = wait_set->impl->rmw_subscriptions.timestamps[i];
     }
   }
   // Set corresponding rcl guard_condition handles NULL.
--- a/rcl/src/rcl/wait.c
+++ b/rcl/src/rcl/wait.c
@@ -264,6 +264,20 @@ rcl_wait_set_get_allocator(const rcl_wait_set_t * wait_set, rcl_allocator_t * al
     } \
   } while (false)
 
+/* The RMW structures with a timestamp are twice as large. */
+#define SET_CLEAR_RMW_TS(Type, RMWStorage, RMWCount) \
+  do { \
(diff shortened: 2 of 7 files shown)


iluetkeb:
@dirk-thomas @wjwwood are you generally okay with the field additions to the existing structs? if yes, I could proceed with further work that's actually using them.

iluetkeb:
Hey @JaimeMartin @MiguelCompany this is the start of timestamp information in rmw/rcl. It is intended that these fields are filled by the middleware with the *arrival* timestamp of messages. I would like to start a discussion with you guys on how to best provide this information.

ralph-lange:
Connects to https://github.com/ros2/design/issues/259.

rcl/include/rcl/wait.h
```diff
@@ -33,25 +34,30 @@ extern "C"
 #include "rcl/types.h"
 #include "rcl/visibility_control.h"
 
+
```
wjwwood: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

rcl/include/rcl/wait.h
```diff
@@ -22,6 +22,7 @@ extern "C"
 
 #include <stdbool.h>
 #include <stddef.h>
+#include <rcutils/time.h>
```
wjwwood: This include should be between the standard headers and this package's own headers, like this:

```suggestion

#include <rcutils/time.h>
```

See:

https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

> 1. dir2/foo2.h.
> 2. A blank line
> 3. C system headers (more precisely: headers in angle brackets with the .h extension), e.g. <unistd.h>, <stdlib.h>.
> 4. A blank line
> 5. C++ standard library headers (without file extension), e.g. <algorithm>, <cstddef>.
> 6. A blank line
> 7. Other libraries' .h files.
> 8. Your project's .h files.

MiguelCompany:
> Hey @JaimeMartin @MiguelCompany this is the start of timestamp information in rmw/rcl. It is intended that these fields are filled by the middleware with the _arrival_ timestamp of messages. I would like to start a discussion with you guys on how to best provide this information.

Would it be enough to have a get_first_unread_sample_info on the subscribers that could be called from [rmw_wait](https://github.com/ros2/rmw_fastrtps/blob/0134bcf52244cbbb6e7e8ac631de391beaa85f17/rmw_fastrtps_shared_cpp/src/rmw_wait.cpp#L92) when necessary?

iluetkeb:
> Would it be enough to have a get_first_unread_sample_info on the subscribers that could be called from rmw_wait when necessary?

That should work and would probably provide the most accurate information, however we are a bit concerned about performance. Reading the sample info seems like it would require deserialization. Also, since one of the other things we have identified is that we might potentially need to call rmw_wait more often, we should keep this as performant as possible.

One other idea I had is that you could keep the listener attached to the objects across calls to wait, and then store the current time within the callback that gets notified upon changes. This would likely incur almost no overhead and still provide good-enough information.

rcl/include/rcl/wait.h
iluetkeb: Done.

rcl/include/rcl/wait.h
iluetkeb: Done.

iluetkeb:
> We should consider any necessary updates to the tests and documentation as well.

As said in https://github.com/ros2/rmw/pull/200, I think we need to test everything in rcl, since it does the relevant work.

However, in `test_wait.cpp`, I do not see any tests which are checking the *pre-existing* fields. For example, in `wait_set_is_valid`, there is no looking at the actual contents of the wait_set. Similarly, there is a test `test_resize_to_zero`, which looks at the counts but not the arrays inside the wait_set.

@wjwwood is that intentionally? I mean, I'd be happy to add a test for the contents, but I'm wondering what the current lack of such tests should tell me.

wjwwood:
> @wjwwood is that intentionally? I mean, I'd be happy to add a test for the contents, but I'm wondering what the current lack of such tests should tell me.

I don't think it is intentional, but `wait_set_is_valid` was added recently. I think @jacobperron added it in this pr https://github.com/ros2/rcl/pull/538, but I can't tell just from the context if that's intentional or an oversight.

jacobperron:
For the test I added, I'm fairly certain I just copied one of the existing tests and modified it minimally to exercise the API I added.

iluetkeb:
Alright, @wjwwood @jacobperron thanks for the info. I'll add a test looking at all the fields, then.

rcl/test/CMakeLists.txt
```diff
@@ -221,6 +221,15 @@ function(test_target_function)
...
 
+  rcl_add_custom_gtest(test_wait_set_impl${target_suffix}
+    SRCS rcl/test_wait_set_impl.cpp
+    ENV ${rmw_implementation_env_var}
+    APPEND_LIBRARY_DIRS ${extra_lib_dirs}
+    LIBRARIES ${PROJECT_NAME}
+    AMENT_DEPENDENCIES ${rmw_implementation} "osrf_testing_tools_cpp"
+  )
+
+
```
wjwwood: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

rcl/include/rcl/wait.h
```diff
@@ -266,6 +272,10 @@ rcl_wait_set_clear(rcl_wait_set_t * wait_set);
  *
  * This can be called on an uninitialized (zero initialized) wait set.
  *
+ * If RCL_RET_BAD_ALLOC is returned, i.e. allocation failed, the wait_set
+ * may be *partially* allocated and therefore, it must still be cleaned up
+ * with _fini.
```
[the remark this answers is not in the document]
iluetkeb: This is just documenting existing behavior. I noticed this when examining the existing allocation macros -- when they fail, they just set the *field currently being allocated* to NULL and then return RET_BAD_ALLOC. All existing fields of the structure are left as they are and thus need to be deallocated as usual.

I agree with you that this is sub-optimal behavior, but it was there before, and ist not trivial to handle otherwise, so I don't think I can change this right now but at least I wanted to document it.

rcl/src/rcl/wait.c
```diff
@@ -250,6 +232,26 @@ rcl_wait_set_get_allocator(const rcl_wait_set_t * wait_set, rcl_allocator_t * al
...
+    if (NULL != wait_set->Type ## s) { \
+      memset( \
+        (void *)wait_set->Type ## s, \
+        0, \
+        sizeof(rcl_ ## Type ## _t *) * wait_set->size_of_ ## Type ## s); \
+      memset( \
+        (void *)wait_set->Type ## s_timestamps, \
+        0, \
+        sizeof(rcutils_time_point_value_t *) * wait_set->size_of_ ## Type ## s); \
+      wait_set->impl->Type ## _index = 0; \
```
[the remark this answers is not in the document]
iluetkeb: This should be doable, I'll look at that next week.

rcl/src/rcl/wait.c
```diff
@@ -285,6 +301,42 @@ rcl_wait_set_get_allocator(const rcl_wait_set_t * wait_set, rcl_allocator_t * al
...
+        (void *)wait_set->Type ## s_timestamps, \
+        sizeof(rcutils_time_point_value_t *) * Type ## s_size, \
+        allocator.state); \
+      RCL_CHECK_FOR_NULL_WITH_MSG( \
+        wait_set->Type ## s_timestamps, "allocating memory failed", { \
+        allocator.deallocate((void *)wait_set->Type ## s, allocator.state); \
+        wait_set->Type ## s = NULL; return RCL_RET_BAD_ALLOC;}); \
+      wait_set->size_of_ ## Type ## s = Type ## s_size; \
+      memset((void *)wait_set->Type ## s, 0, sizeof(rcl_ ## Type ## _t *) * Type ## s_size); \
+      ExtraRealloc \
```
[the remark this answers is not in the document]
iluetkeb: This is a bit harder because of the error handling. I'll have a look.

rcl/test/CMakeLists.txt
wjwwood: This was marked as resolved, but was not actually.

wjwwood:
```suggestion
```

iluetkeb:
Alright folks, this now has a test for presence of timestamps, and in https://github.com/ros2/rmw_fastrtps/pull/359 we have an implementation of timestamps for subscriptions, clients and services that passes the test.

Karsten1987:
CI: 
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10005/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5606/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8140/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9889/)

Karsten1987:
@iluetkeb ironically, I don't have access to the fork of boschresearch, so I can't push a fix on your branch. I think what's missing here is the following:

```
diff --git a/rcl/test/CMakeLists.txt b/rcl/test/CMakeLists.txt
index fb31b27..470337f 100644
--- a/rcl/test/CMakeLists.txt
+++ b/rcl/test/CMakeLists.txt
@@ -219,7 +219,7 @@ function(test_target_function)
     ENV ${rmw_implementation_env_var}
     LIBRARIES ${PROJECT_NAME}
-    AMENT_DEPENDENCIES ${rmw_implementation} "osrf_testing_tools_cpp"
+    AMENT_DEPENDENCIES ${rmw_implementation} "osrf_testing_tools_cpp" "test_msgs"
   )

   # This condition runs the timestmap test only when we know the RMW implementation
```

iluetkeb:
That include is not actually necessary, since I moved the relevant test to `test_wait_set_timestamp.c`. I'm wondering why it doesn't complain on my box, but anyway, it's removed now.

I also added you to the correct team for our forks.

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10025/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5621/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8156/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9930/)

rotu:
I'm sorry I'm coming to this late. It seems like a bad idea, and I think should be strongly reconsidered. https://github.com/ros2/design/issues/259#issuecomment-611633199

iluetkeb:
@rotu I have commented in the design issue and I hope I could address your concerns. 

It *is* a bit late now, in my case literally because I'm in Europe, where it's currently half past eight :wink: So I will log off after this comment, and tomorrow being Good Friday, I won't work. If you have further concerns, I can only contribute further on Saturday or (preferably) after Easter. Hope that's enough! Happy Holidays to you all.

iluetkeb:
Closed in favor of https://github.com/ros2/rcl/pull/619

PR #597, opened 2020-03-12
Change naming style for private functions

Blast545:
To keep consistency with other functions only available in this method, added "_" prefix on functions only available to this module.

Edit: Removed underscore from functions internal to the module and added the `static` keyword.

### DIFF
--- a/rcl/src/rcl/time.c
+++ b/rcl/src/rcl/time.c
@@ -31,24 +31,24 @@ typedef struct rcl_ros_clock_storage_t
 } rcl_ros_clock_storage_t;
 
 // Implementation only
-rcl_ret_t
-_rcl_get_steady_time(void * data, rcl_time_point_value_t * current_time)
+static rcl_ret_t
+rcl_get_steady_time(void * data, rcl_time_point_value_t * current_time)
 {
   (void)data;  // unused
   return rcutils_steady_time_now(current_time);
 }
 
 // Implementation only
-rcl_ret_t
-_rcl_get_system_time(void * data, rcl_time_point_value_t * current_time)
+static rcl_ret_t
+rcl_get_system_time(void * data, rcl_time_point_value_t * current_time)
 {
   (void)data;  // unused
   return rcutils_system_time_now(current_time);
 }
 
 // Internal method for zeroing values on init, assumes clock is valid
-void
-_rcl_init_generic_clock(rcl_clock_t * clock)
+static void
+rcl_init_generic_clock(rcl_clock_t * clock)
 {
   clock->type = RCL_CLOCK_UNINITIALIZED;
   clock->jump_callbacks = NULL;
@@ -59,12 +59,12 @@ _rcl_init_generic_clock(rcl_clock_t * clock)
 
 // The function used to get the current ros time.
 // This is in the implementation only
-rcl_ret_t
-_rcl_get_ros_time(void * data, rcl_time_point_value_t * current_time)
+static rcl_ret_t
+rcl_get_ros_time(void * data, rcl_time_point_value_t * current_time)
 {
   rcl_ros_clock_storage_t * t = (rcl_ros_clock_storage_t *)data;
   if (!t->active) {
-    return _rcl_get_system_time(data, current_time);
+    return rcl_get_system_time(data, current_time);
   }
   *current_time = rcutils_atomic_load_uint64_t(&(t->current_time));
   return RCL_RET_OK;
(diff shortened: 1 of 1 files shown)


Blast545:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9626/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5319/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7847/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9529/)

rcl/src/rcl/time.c
```diff
@@ -32,23 +32,23 @@ typedef struct rcl_ros_clock_storage_t
 
 // Implementation only
 rcl_ret_t
-rcl_get_steady_time(void * data, rcl_time_point_value_t * current_time)
+_rcl_get_steady_time(void * data, rcl_time_point_value_t * current_time)
```
ivanpauno: What does private function mean in `C`?
This function should be `static`, there's no naming convention needed.

According to the C++ standard, names starting with `_` or names containing `__` are reserved to the implementation. See https://wiki.sei.cmu.edu/confluence/display/cplusplus/DCL51-CPP.+Do+not+declare+or+define+a+reserved+identifier#DCL51-CPP.

I would rather delete leading underscores that add more.
We could use another convention for static functions, like `s_`, to make the name more obvious.

ivanpauno: See old related discussion: https://github.com/ros2/rclcpp/issues/874

Blast545: I agree with you that by using `static` is enough so that the function is used only in the module. 

I saw other functions with the leading underscore and wanted them to use the same naming convention. I can change this PR to remove the underscore and use the static keyword in all in the functions internal to the module.

ivanpauno: > I saw other functions with the leading underscore and wanted them to use the same naming convention

Having a naming convention is good but using a leading underscore is not standard compliant, so I would avoid that particular convention.
For the moment, I would just add `static` where missing. We can iterate on a correct naming convention later, maybe using `s_` or maybe just using no prefix (i.e.: deleting current leading underscores).

Blast545: Should I remove the leading underscore for the functions that already had it before my commit?

ivanpauno: IMO, yes. But it would be great to check what others think about it.

jacobperron: Given that leading underscores are reserved in the C standard, I think we should avoid it in function names and just use static (and maybe a comment as we've done in this case).
From [C99 standard (7.1.3)](http://www.open-std.org/jtc1/sc22/WG14/www/docs/n1256.pdf):

> - All identifiers that begin with an underscore and either an uppercase letter or another
underscore are always reserved for any use.
> - All identifiers that begin with an underscore are always reserved for use as identifiers
with file scope in both the ordinary and tag name spaces.

tfoote: "Private" in this case is indicating that it's not a public API intended to be stable for external reuse: https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#public-api-declaration It's about documentation more than enforcement.

jacobperron: The function in question can't be used externally anyways since it's not declared in a header.

Blast545: Yes, the functions can't be used externally so it shouldn't be a problem related to API. I opened the issue because it's odd browsing through the code and not having some uniformity regarding names. e.g: `rcl_init_generic_clock` and `_rcl_clock_generic_fini`. I'll proceed removing the underscores and adding the static keyword instead.

ivanpauno: > The function in question can't be used externally anyways since it's not declared in a header.

IIRC, that's not 100% true. If the function isn't marked as `static`, you can declare it in another executable without definition, then link again this library and use it (that probably doesn't apply to Windows DLLs, because of their export/import stuff).
Functions not declared in a library header shouldn't be considered part of the API though.

> I'll proceed removing the underscores and adding the static keyword instead.

:+1:

jacobperron: That's right, thanks for the clarification. Hopefully no one is adding a declaration anywhere.

Blast545:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9645/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5330/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7872/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9540/)

Blast545:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9934/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5544/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8082/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9783/)

PR #598, opened 2020-03-12
Test that nodes are returned with correct multiplicity

rotu:
Signed-off-by: Dan Rose <dan@digilabs.io>

PR #601, opened 2020-03-17
Add missing node destruction

ivanpauno:
Follow-up of https://github.com/ros2/rcl/pull/598.

I didn't realize that the code was cleaning the node.


ivanpauno:
@rotu FYI.

ivanpauno:
> Though I wonder, why putting all these rcl_node_t instances in the heap?

It really doesn't make sense.
I will see if I have time to improve the test, if not, I will just run CI.

rotu:
I also support breaking up the big "do everything" test case into multiple test cases; or a parameterized test case.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9673/)

PR #607, opened 2020-04-07
Use keystore root as security root directory, and not contexts folder

ivanpauno:
Address comment https://github.com/ros2/design/pull/274#discussion_r403312815.

PR #609, opened 2020-04-09
Fix security directory lookup for '/' security contexts

ivanpauno:
Follow-up of https://github.com/ros2/rcl/pull/607.
It wasn't working correctly for the root security context: '/'.

Added test covering this corner case.

### DIFF
--- a/rcl/src/rcl/security.c
+++ b/rcl/src/rcl/security.c
@@ -111,16 +111,16 @@ char * exact_match_lookup(
 {
   // Perform an exact match for the context name in directory <root dir>.
   char * secure_root = NULL;
+  char * contexts_dir = NULL;
+  contexts_dir = rcutils_join_path(ros_secure_root_env, "contexts", *allocator);
   // "/" case when root namespace is explicitly passed in
   if (0 == strcmp(name, "/")) {
-    secure_root = rcutils_strdup(ros_secure_root_env, *allocator);
+    secure_root = contexts_dir;
   } else {
     char * relative_path = NULL;
-    char * contexts_dir = NULL;
     // Get native path, ignore the leading forward slash
     // TODO(ros2team): remove the hard-coded length, use the length of the root namespace instead
     relative_path = rcutils_to_native_path(name + 1, *allocator);
-    contexts_dir = rcutils_join_path(ros_secure_root_env, "contexts", *allocator);
     secure_root = rcutils_join_path(contexts_dir, relative_path, *allocator);
     allocator->deallocate(relative_path, allocator->state);
     allocator->deallocate(contexts_dir, allocator->state);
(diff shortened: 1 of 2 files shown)


rcl/src/rcl/security.c
```diff
@@ -111,16 +111,16 @@ char * exact_match_lookup(
...
   } else {
     char * relative_path = NULL;
-    char * contexts_dir = NULL;
     // Get native path, ignore the leading forward slash
     // TODO(ros2team): remove the hard-coded length, use the length of the root namespace instead
     relative_path = rcutils_to_native_path(name + 1, *allocator);
-    contexts_dir = rcutils_join_path(ros_secure_root_env, "contexts", *allocator);
     secure_root = rcutils_join_path(contexts_dir, relative_path, *allocator);
     allocator->deallocate(relative_path, allocator->state);
     allocator->deallocate(contexts_dir, allocator->state);
```
mikaelarguedas, line 126: This should be moved to the same scope as the allocation, otherwise it's not deallocated for "/"

ivanpauno: why?

In the `if` case, the `contexts_dir` pointer is just being copied to `secure_root`, so deallocating it would deallocate the return value.
Only in the `else` case deallocation is needed.

clalancette: Yeah, in the "/" case, we actually do *not* want to deallocate it, since we want to return the heap-allocated pointer.  In all other cases, we want to use it and then deallocate it, because we'll be returning something else.  I don't love the asymmetry (it will be hard to follow later), but I do think it is correct.

mikaelarguedas: Oh you're totally right I read too fast

ivanpauno:
CI:
- Focal [Build Status](https://ci.ros2.org/job/ci_linux/10039/)
- AArch64 Bionic [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/5633/)
- macOS [Build Status](https://ci.ros2.org/job/ci_osx/8168/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/9949/)

ruffsl:
A tangential question: these two CLI invocations are equivalent, yes?

```
export ROS_SECURITY_ROOT_DIRECTORY=~/my_keystore
export ROS_SECURITY_ENABLE=true
export ROS_SECURITY_STRATEGY=Enforce
---
ros2 run my_package my_node --security-context="/"
# secure_root == "/home/bob"
---
ros2 run my_package my_node
# secure_root == "/home/bob"
```
Wasn't sure where to find the beginnings of this corner case when `--security-context="/"` is absent.

ivanpauno:
> A tangential question: these two CLI invocations are equivalent, yes?

Yes, both are the same.

Issue #611, opened 2020-04-09
Nightly failure on aarch64 rcl.TestTwoExecutablesAfterShutdown.test_services (from rcl.test_services__rmw_cyclonedds_cpp_Release.launch_tests)


tfoote:
This has failed two nightlys in a row

https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_release/1126/testReport/rcl/TestTwoExecutablesAfterShutdown/test_services/

I've been unable to reproduce this on an aws aarch64 bionic machine but have been able to reproduce it in CI on focal

https://ci.ros2.org/job/ci_linux-aarch64/5630

Using this gist: https://gist.github.com/tfoote/dc6c33491124e863f554531b87b0492e

And reducing the build and test to rcl.

<details>

```
13:08:22 30: Test command: /home/jenkins-agent/workspace/ci_linux-aarch64/venv/bin/python3 "-u" "/home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/ament_cmake_test/share/ament_cmake_test/cmake/run_test.py" "/home/jenkins-agent/workspace/ci_linux-aarch64/ws/build/rcl/test_results/rcl/test_services__rmw_cyclonedds_cpp.xunit.xml" "--package-name" "rcl" "--output-file" "/home/jenkins-agent/workspace/ci_linux-aarch64/ws/build/rcl/launch_test/CHANGEME.txt" "--env" "RMW_IMPLEMENTATION=rmw_cyclonedds_cpp" "--append-env" "LD_LIBRARY_PATH=/home/jenkins-agent/workspace/ci_linux-aarch64/ws/build/rcl" "--command" "/home/jenkins-agent/workspace/ci_linux-aarch64/venv/bin/python3" "-m" "launch_testing.launch_test" "/home/jenkins-agent/workspace/ci_linux-aarch64/ws/build/rcl/test/test/test_services__rmw_cyclonedds_cpp_Release.py" "--junit-xml=/home/jenkins-agent/workspace/ci_linux-aarch64/ws/build/rcl/test_results/rcl/test_services__rmw_cyclonedds_cpp.xunit.xml" "--package-name=rcl"
13:08:22 30: Test timeout computed to be: 15
13:08:22 30: -- run_test.py: extra environment variables:
13:08:22 30:  - RMW_IMPLEMENTATION=rmw_cyclonedds_cpp
13:08:22 30: -- run_test.py: extra environment variables to append:
[...]

PR #613, opened 2020-04-09
Reduce the internal timeouts

tfoote:
Debugging #611 these timeouts were so long that the overall test timeout was being triggered before these internal timeouts were triggered.

1000 retries at 100ms -> 10 seconds each. I cut it down to 1 second for each to establish the connection.

Reducing these helps narrow down the focus on where things are actually failing.

### DIFF
--- a/rcl/test/rcl/client_fixture.cpp
+++ b/rcl/test/rcl/client_fixture.cpp
@@ -179,7 +179,7 @@ int main(int argc, char ** argv)
     });
 
     // Wait until server is available
-    if (!wait_for_server_to_be_available(&node, &client, 1000, 100)) {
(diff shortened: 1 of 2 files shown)


tfoote:
This is an incremental change and will save us time on failed tests. Putting this up for review.

The test failures appear to come from a lower level dependency unrelated to my timeout changes. 

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10177/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5732/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8268/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10060/)

PR #617, opened 2020-04-14
Update security environment variables

ruffsl:
Related: https://github.com/ros2/sros2/issues/199

### DIFF
--- a/rcl/src/rcl/security.c
+++ b/rcl/src/rcl/security.c
@@ -106,13 +106,13 @@ rcl_get_enforcement_policy(rmw_security_enforcement_policy_t * policy)
 
 char * exact_match_lookup(
   const char * name,
-  const char * ros_secure_root_env,
+  const char * ros_secure_keystore_env,
   const rcl_allocator_t * allocator)
 {
   // Perform an exact match for the enclave name in directory <root dir>.
   char * secure_root = NULL;
   char * enclaves_dir = NULL;
-  enclaves_dir = rcutils_join_path(ros_secure_root_env, "enclaves", *allocator);
+  enclaves_dir = rcutils_join_path(ros_secure_keystore_env, "enclaves", *allocator);
   // "/" case when root namespace is explicitly passed in
   if (0 == strcmp(name, "/")) {
     secure_root = enclaves_dir;
@@ -161,13 +161,13 @@ char * rcl_get_secure_root(
   }
 
   // found a usable environment variable, copy into our memory before overwriting with next lookup
-  char * ros_secure_root_env = rcutils_strdup(env_buf, *allocator);
+  char * ros_secure_keystore_env = rcutils_strdup(env_buf, *allocator);
 
   char * secure_root = NULL;
   if (ros_secure_enclave_override) {
-    secure_root = rcutils_strdup(ros_secure_root_env, *allocator);
+    secure_root = rcutils_strdup(ros_secure_keystore_env, *allocator);
   } else {
-    secure_root = exact_match_lookup(name, ros_secure_root_env, allocator);
+    secure_root = exact_match_lookup(name, ros_secure_keystore_env, allocator);
   }
 
   if (NULL == secure_root || !rcutils_is_directory(secure_root)) {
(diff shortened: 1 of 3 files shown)


rcl/src/rcl/security.c
```diff
@@ -157,33 +157,33 @@ char * rcl_get_secure_root(
...
 
   // found a usable environment variable, copy into our memory before overwriting with next lookup
-  char * ros_secure_root_env = rcutils_strdup(env_buf, *allocator);
+  char * ros_secure_keystore_env = rcutils_strdup(env_buf, *allocator);
 
   char * secure_root = NULL;
-  if (ros_secure_directory_override) {
-    secure_root = rcutils_strdup(ros_secure_root_env, *allocator);
+  if (ros_secure_enclave_override) {
+    secure_root = exact_match_lookup(ros_secure_keystore_env, ros_secure_keystore_env, allocator);
```
ruffsl, line 168: @ivanpauno < actually, this doesn't look quite right. We should store the string from `ROS_SECURITY_ENCLAVE_OVERRIDE` into something like `ros_secure_enclave_override_env` instead of `ros_secure_keystore_env`. The `ros_secure_keystore_env` should be set to the contents of `ROS_SECURITY_KEYSTORE` regardless if `ROS_SECURITY_ENCLAVE_OVERRIDE` is set as well.

ivanpauno: why?

ivanpauno: Before, the override was a path, and not just a name.
The change looks good to me, if this is the desired behavior now.

rcl/src/rcl/security.c
```diff
@@ -132,58 +132,67 @@ char * rcl_get_secure_root(
...
+    ros_secure_enclave_override = false;
   }
+  char * ros_secure_enclave_override_env = rcutils_strdup(env_buf, *allocator);
 
-  // found a usable environment variable, copy into our memory before overwriting with next lookup
-  char * ros_secure_root_env = rcutils_strdup(env_buf, *allocator);
+  // check if keystore environment variable is empty
+  if (rcutils_get_env(ROS_SECURITY_KEYSTORE_VAR_NAME, &env_buf)) {
+    return NULL;
+  }
```
kyrofa, line 158: `rcutils_get_env()` returns non-NULL if an error was encountered (in which case it returns an error string). Would it be worth showing that error to the user?

rcl/src/rcl/security.c
kyrofa: As I understand it, the override is now a name/identity within the given keystore, which allows for future keystores that are not filesystem-based.

ivanpauno: Ok, I think this also matches the behavior change described here https://github.com/ros2/design/pull/274/commits/98200d8e3b273194afc3170f3a9b65eaab689226, which I didn't see before. LGTM!

rcl/src/rcl/security.c
ruffsl: I was just keeping with the existing return NULL after any error. Are there any existing patterns else where in RCL that do more transparent error handling that we could do in a follow up PR?

ivanpauno: you can do:
```c
char * error_str = rcutils_get_env(ROS_SECURITY_KEYSTORE_VAR_NAME, &env_buf);
if (error_str) {
  RCUTILS_LOG_ERROR("rcutils_get_env failed: %s\n", error_str);
  return NULL;
}
```

ruffsl: See 1ee7094

rcl/src/rcl/security.c
```diff
@@ -132,58 +132,75 @@ char * rcl_get_secure_root(
...
+  bool ros_secure_enclave_override = true;
 
   // find out if either of the configuration environment variables are set
   const char * env_buf = NULL;
   if (NULL == name) {
     return NULL;
   }
-
-  if (rcutils_get_env(ROS_SECURITY_DIRECTORY_OVERRIDE, &env_buf)) {
+
```
mikaelarguedas, line 142:
```suggestion

```

PR #619, opened 2020-04-14
Message info with timestamps support in rcl

iluetkeb:
Second part of implementation for https://github.com/ros2/design/issues/259

This zero-initializes the new timestamps in rmw before calling 'take'.

It requires https://github.com/ros2/rmw/pull/214 to compile.

As discussed, this replaces https://github.com/ros2/rcl/pull/591


rcl/test/CMakeLists.txt
```diff
@@ -206,6 +206,23 @@ function(test_target_function)
...
+  rcl_add_custom_gtest(test_message_info${target_suffix}
+    SRCS rcl/test_message_info.cpp
+    ENV ${rmw_implementation_env_var}
+    APPEND_LIBRARY_DIRS ${extra_lib_dirs}
+    LIBRARIES ${PROJECT_NAME}
+    AMENT_DEPENDENCIES ${rmw_implementation} "osrf_testing_tools_cpp" "test_msgs"
+    ${AMENT_GTEST_ARGS}
+  )
+
+
```
wjwwood, line 225: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

rcl/test/CMakeLists.txt
```diff
@@ -206,6 +206,23 @@ function(test_target_function)
     AMENT_DEPENDENCIES ${rmw_implementation} "osrf_testing_tools_cpp" "test_msgs"
   )
 
+  set(AMENT_GTEST_ARGS "SKIP_TEST")
+  # TODO(wjwwood): remove this when the graph API works properly for connext dynamic
```
wjwwood, line 210: This TODO doesn't look right, and the condition probably isn't want you want? Also the condition below it looks wrong.

rcl/test/rcl/test_message_info.cpp
```diff
@@ -0,0 +1,336 @@
...
+#include <gtest/gtest.h>
+
+#include <chrono>
+#include <string>
+#include <thread>
+
+#include "rcl/subscription.h"
+
+#include "rcl/rcl.h"
+#include "rcl/graph.h"
```
wjwwood, line 25: Do you need all of these includes?

rcl/test/rcl/test_message_info.cpp
```diff
@@ -0,0 +1,336 @@
...
+      if (wait_set.subscriptions[i] && wait_set.subscriptions[i] == subscription) {
+        success = true;
+        return;
+      }
+    }
+  } while (iteration < max_tries);
+  success = false;
+}
+
+/* Basic nominal test of a subscription
```
wjwwood, line 123: Is this right?

iluetkeb:
> I wonder if you even need the new test file, and if you could not add assertions to the test_subscription test case.

These tests currently break on all middlewares that do not yet have timestamps, so I wanted a test that I can conditionally enable only when the middleware is known to have them implemented. AFAICT, this is only possible on the TestSuite level, so I needed a new test.

Conceptually it also seemed cleaner, since I'm not just testing pub/sub but also services.

> Even in the implementations that don't support the feature it should be checked that the timestamp is zero, indicating that it is not supported.

Okay, that's a good point. I will have a look to say how I could do that cleanly.

rcl/test/rcl/test_message_info.cpp
iluetkeb: Yes, I'm using graph.h to check whether the service is up. It seemed more appropriate than having another of those 1s delays in there to make sure the connection is established ;-) If there's a better way, please let me know!

iluetkeb:
I have now removed the support for services from this PR and will make a second PR to implement that part.

That means we will have two different sets of PRs in flight, for all of the relevant packages. This is a serious PITA and I think future improvements to the RMW and its implementations would seriously benefit from having both rmw and its implementations in one repo.

rcl/test/CMakeLists.txt
iluetkeb: Done.

wjwwood:
> This is a serious PITA and I think future improvements to the RMW and its implementations would seriously benefit from having both rmw and its implementations in one repo.

Big 👎 from me. This makes it basically impossible to get one rmw implementation without another one when developing. And it makes it so that making your own rmw implementation is treated differently, i.e. which rmw implementations are good enough to be in the repository? If we deprecate one should it be moved out of the repository? etc...

rcl/test/rcl/test_subscription.cpp
```diff
@@ -197,9 +199,29 @@ TEST_F(CLASSNAME(TestSubscriptionFixture, RMW_IMPLEMENTATION), test_subscription
...
+    EXPECT_TRUE(message_info.source_timestamp.sec <= message_info.received_timestamp.sec);
+    EXPECT_NE(0u, message_info.received_timestamp.nsec);
+    EXPECT_TRUE(pre_publish_time.tv_nsec <= (long)message_info.received_timestamp.nsec);
+    EXPECT_TRUE(message_info.source_timestamp.nsec <= message_info.received_timestamp.nsec);
+  #else
+    EXPECT_EQ(0u, message_info.source_timestamp.sec);
+    EXPECT_EQ(0u, message_info.source_timestamp.nsec);
+    EXPECT_EQ(0u, message_info.received_timestamp.sec);
+    EXPECT_EQ(0u, message_info.received_timestamp.nsec);
+  #endif
```
wjwwood, line 220: 👍 this looks good, thanks.

rcl/test/rcl/test_subscription.cpp
```diff
@@ -179,6 +179,12 @@ TEST_F(CLASSNAME(TestSubscriptionFixture, RMW_IMPLEMENTATION), test_subscription
   //                probably using the count_subscriptions busy wait mechanism
   //                until then we will sleep for a short period of time
   std::this_thread::sleep_for(std::chrono::milliseconds(1000));
+#ifdef RMW_TIMESTAMPS_SUPPORTED
+  timespec pre_publish_ts;
+  EXPECT_EQ(0, clock_gettime(CLOCK_REALTIME, &pre_publish_ts)) << " clock_gettime failed";
```
wjwwood, line 184: This may not be portable, I'd recommend using `std::chrono::system_clock::now()` here instead.

iluetkeb: Right! I used rcutils now.

iluetkeb:
@wjwwood This is complete from my side now.

rcl/test/rcl/test_subscription.cpp
```diff
@@ -179,6 +179,10 @@ TEST_F(CLASSNAME(TestSubscriptionFixture, RMW_IMPLEMENTATION), test_subscription
   //                probably using the count_subscriptions busy wait mechanism
   //                until then we will sleep for a short period of time
   std::this_thread::sleep_for(std::chrono::milliseconds(1000));
+#ifdef RMW_TIMESTAMPS_SUPPORTED
+  rcl_time_point_value_t pre_publish_time;
+  EXPECT_EQ(0, rcutils_system_time_now(&pre_publish_time)) << " could not get system time failed";
```
wjwwood, line 184:
```suggestion
  EXPECT_EQ(RCUTILS_RET_OK, rcutils_system_time_now(&pre_publish_time))
    << " could not get system time failed";  
```

You might want to run `ament_uncrustify` over that, I don't know if I got the style just right.

iluetkeb: Took me two tries, but I think I got it now ,-)

ros-discourse:

https://discourse.ros.org/t/ros2-middleware-change-proposal/15863/7

Issue #585, opened 2020-02-29
rcl test `test_publisher_init_fini` does not `fini`


rotu:
The tests in `test_publisher_init_fini` create a publisher (`rcl_publisher_init`) without finalizing it (`rcl_publisher_fini`). This causes a failed assertion in CycloneDDS in debug mode when tearing down the node (`rcl_node_fini`), and possibly a memory leak. It's clear from the name of the test that the test was *intended* to have fini calls, but it's not clear whether or not RCL and/or CycloneDDS should issue a warning/error in this case.

https://github.com/ros2/rcl/blob/4b9c0a30beea07960cc9dd666dcf36ec1a18065f/rcl/test/rcl/test_publisher.cpp#L199-L204

```
#3  0x00007f60aba81412 in __GI___assert_fail (assertion=0x7f60ab818d1a "ddsrt_hh_iter_first (gv->sertopics, &it) == NULL", file=0x7f60ab818642 "../../src/eclipse-cyclonedds/cyclonedds/src/core/ddsi/src/q_init.c", line=1780, function=0x7f60ab818dd1 "void rtps_fini(struct ddsi_domaingv *)") at assert.c:101
#4  0x00007f60ab7a306b in rtps_fini (gv=0x10ad010) at ../../src/eclipse-cyclonedds/cyclonedds/src/core/ddsi/src/q_init.c:1780
#5  0x00007f60ab7dfa82 in dds_domain_free (vdomain=0x10acd70) at ../../src/eclipse-cyclonedds/cyclonedds/src/core/ddsc/src/dds_domain.c:269
#6  0x00007f60ab7e440a in dds_entity_deriver_delete (e=0x10acd70) at ../../src/eclipse-cyclonedds/cyclonedds/src/core/ddsc/src/dds__types.h:189
#7  0x00007f60ab7e5159 in really_delete_pinned_closed_locked (e=0x10acd70, delstate=DIS_EXPLICIT) at ../../src/eclipse-cyclonedds/cyclonedds/src/core/ddsc/src/dds_entity.c:503
#8  0x00007f60ab7e4e07 in dds_delete_impl_pinned (e=0x10acd70, delstate=DIS_EXPLICIT) at ../../src/eclipse-cyclonedds/cyclonedds/src/core/ddsc/src/dds_entity.c:400
[...]

PR #625, opened 2020-04-20
Documentation rcl_pub/etc_fini() must come before rcl_node_fini()

sloretz:
Context https://github.com/ros2/rcl/issues/585#issuecomment-616760593

In short, this documents that the client library using `rcl` must ensure entities created from nodes are finalized before the node itself is finalized.

### DIFF
--- a/rcl/include/rcl/client.h
+++ b/rcl/include/rcl/client.h
@@ -174,7 +174,7 @@ rcl_client_init(
  * Lock-Free          | Yes
  *
  * \param[inout] client handle to the client to be finalized
- * \param[in] node handle to the node used to create the client
+ * \param[in] node a valid (not finalized) handle to the node used to create the client
  * \return `RCL_RET_OK` if client was finalized successfully, or
  * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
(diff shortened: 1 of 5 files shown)


sloretz:
CI (testing just `rcl`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10324/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5853/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8409/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10233/)

ahcorde:
The new docbloc is used to re-define the node param in some `_fini()` funtions. But the text said `create`. But I think the right work is `destroy`, right ?

https://github.com/ros2/rcl/blob/13c672352f5af62479b17b23045a3191f36f98fe/rcl/include/rcl/service.h#L174-L187

sloretz:
> The new docbloc is used to re-define the node param in some `_fini()` funtions. But the text said `create`. But I think the right work is `destroy`, right ?

Not just any node will do. It must be the node that was "used to create" the entity. Ideas on making the wording clearer?

PR #627, opened 2020-04-21
Implement service info structure with timestamps.

iluetkeb:
This implements the RCL part of https://github.com/ros2/rmw/issues/219

It requires  https://github.com/ros2/rmw/pull/217 and https://github.com/ros2/rmw_implementation/pull/96 to build. An example implementation is in https://github.com/ros2/rmw_fastrtps/pull/369.

### DIFF
--- a/rcl/include/rcl/client.h
+++ b/rcl/include/rcl/client.h
@@ -287,7 +287,7 @@ RCL_WARN_UNUSED
 rcl_ret_t
 rcl_take_response(
   const rcl_client_t * client,
-  rmw_request_id_t * request_header,
+  rmw_service_info_t * request_header,
   void * ros_response);
 
 /// Get the name of the service that this client will request a response from.
--- a/rcl/include/rcl/client.h
+++ b/rcl/include/rcl/client.h
@@ -285,10 +285,20 @@ rcl_send_request(const rcl_client_t * client, const void * ros_request, int64_t 
 RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_ret_t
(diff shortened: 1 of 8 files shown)


iluetkeb:
> I'd prefer to replace the existing functions, but I know there's not time for that. I guess we'll have to do two deprecation cycles over three releases to get back to just rcl_take_X...

I already did a fair amount of work in that direction, but I figured I needed a quicker way. I'm happy to continue in that direction, then maybe we can get this in earlier.

wjwwood:
So, the only thing we could do in the near future that would be better, is to deprecate the old signature and update our use of it in rclpy and rclcpp. That way we could remove it in the next version, rename the current one and deprecate its original name, so that in f+2 turtle we could be at the ideal state... Not deprecating it now just means we have to live with it (renaming or not) for an additional cycle, but that's not the end of the world.

So, while not optimal, I think the current state is fine, tbh.

iluetkeb:
What do I need to do to deprecate it?

wjwwood:
> What do I need to do to deprecate it?

Documentation and there are compiler attributes for that.

> Oh, does this mean I would have to update all the current usages? There are very many of those, I don't think I'll manage that in time.

Yes.

iluetkeb:
btw, I'm currently still working on the rclpy stuff. it's implemented, but there is a double-free somewhere that I'm hunting down.

iluetkeb:
So. This should be ready.

ros-discourse:

https://discourse.ros.org/t/ros2-middleware-change-proposal/15863/7

PR #631, opened 2020-04-23
Convert sleep_for into appropriate logic

Barry-Xu-2018:
Signed-off-by: Barry Xu <Barry.Xu@sony.com>

### DIFF
--- a/rcl/test/rcl/common.cpp
+++ b/rcl/test/rcl/common.cpp
@@ -1,0 +1,237 @@
+// Copyright 2020 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "common.hpp"
+
+#include <chrono>
+#include <thread>
+
+#include "rcutils/logging_macros.h"
+
+#include "osrf_testing_tools_cpp/scope_exit.hpp"
+#include "rcl/error_handling.h"
+#include "rcl/graph.h"
+
+bool
+wait_for_server_to_be_available(
+  rcl_node_t * node,
+  rcl_client_t * client,
+  size_t max_tries,
+  int64_t period_ms)
+{
+  size_t iteration = 0;
+  while (iteration < max_tries) {
+    ++iteration;
+    bool is_ready;
+    rcl_ret_t ret = rcl_service_server_is_available(node, client, &is_ready);
+    if (ret != RCL_RET_OK) {
+      RCUTILS_LOG_ERROR_NAMED(
+        ROS_PACKAGE_NAME,
+        "Error in rcl_service_server_is_available: %s",
+        rcl_get_error_string().str);
+      return false;
+    }
+    if (is_ready) {
+      return true;
+    }
+    std::this_thread::sleep_for(std::chrono::milliseconds(period_ms));
+  }
+  return false;
+}
+
+bool
+wait_for_client_to_be_ready(
+  rcl_client_t * client,
+  rcl_context_t * context,
+  size_t max_tries,
+  int64_t period_ms)
+{
+  rcl_wait_set_t wait_set = rcl_get_zero_initialized_wait_set();
+  rcl_ret_t ret =
+    rcl_wait_set_init(&wait_set, 0, 0, 0, 1, 0, 0, context, rcl_get_default_allocator());
+  if (ret != RCL_RET_OK) {
+    RCUTILS_LOG_ERROR_NAMED(
+      ROS_PACKAGE_NAME, "Error in wait set init: %s", rcl_get_error_string().str);
+    return false;
+  }
+  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+  {
+    if (rcl_wait_set_fini(&wait_set) != RCL_RET_OK) {
+      RCUTILS_LOG_ERROR_NAMED(
+        ROS_PACKAGE_NAME, "Error in wait set fini: %s", rcl_get_error_string().str);
+      throw std::runtime_error("error while waiting for client");
+    }
+  });
+  size_t iteration = 0;
+  while (iteration < max_tries) {
+    ++iteration;
+    if (rcl_wait_set_clear(&wait_set) != RCL_RET_OK) {
+      RCUTILS_LOG_ERROR_NAMED(
+        ROS_PACKAGE_NAME, "Error in wait_set_clear: %s", rcl_get_error_string().str);
+      return false;
+    }
+    if (rcl_wait_set_add_client(&wait_set, client, NULL) != RCL_RET_OK) {
+      RCUTILS_LOG_ERROR_NAMED(
+        ROS_PACKAGE_NAME, "Error in wait_set_add_client: %s", rcl_get_error_string().str);
+      return false;
+    }
+    ret = rcl_wait(&wait_set, RCL_MS_TO_NS(period_ms));
+    if (ret == RCL_RET_TIMEOUT) {
+      continue;
+    }
+    if (ret != RCL_RET_OK) {
+      RCUTILS_LOG_ERROR_NAMED(ROS_PACKAGE_NAME, "Error in wait: %s", rcl_get_error_string().str);
+      return false;
+    }
+    for (size_t i = 0; i < wait_set.size_of_clients; ++i) {
+      if (wait_set.clients[i] && wait_set.clients[i] == client) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
+bool
+wait_for_service_to_be_ready(
+  rcl_service_t * service,
+  rcl_context_t * context,
+  size_t max_tries,
+  int64_t period_ms)
+{
+  rcl_wait_set_t wait_set = rcl_get_zero_initialized_wait_set();
+  rcl_ret_t ret =
+    rcl_wait_set_init(&wait_set, 0, 0, 0, 0, 1, 0, context, rcl_get_default_allocator());
+  if (ret != RCL_RET_OK) {
+    RCUTILS_LOG_ERROR_NAMED(
+      ROS_PACKAGE_NAME, "Error in wait set init: %s", rcl_get_error_string().str);
+    return false;
+  }
+  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+  {
+    if (rcl_wait_set_fini(&wait_set) != RCL_RET_OK) {
+      RCUTILS_LOG_ERROR_NAMED(
+        ROS_PACKAGE_NAME, "Error in wait set fini: %s", rcl_get_error_string().str);
+      throw std::runtime_error("error waiting for service to be ready");
+    }
+  });
+  size_t iteration = 0;
+  while (iteration < max_tries) {
+    ++iteration;
+    if (rcl_wait_set_clear(&wait_set) != RCL_RET_OK) {
+      RCUTILS_LOG_ERROR_NAMED(
+        ROS_PACKAGE_NAME, "Error in wait_set_clear: %s", rcl_get_error_string().str);
+      return false;
+    }
+    if (rcl_wait_set_add_service(&wait_set, service, NULL) != RCL_RET_OK) {
+      RCUTILS_LOG_ERROR_NAMED(
+        ROS_PACKAGE_NAME, "Error in wait_set_add_service: %s", rcl_get_error_string().str);
+      return false;
+    }
+    ret = rcl_wait(&wait_set, RCL_MS_TO_NS(period_ms));
+    if (ret == RCL_RET_TIMEOUT) {
+      continue;
+    }
+    if (ret != RCL_RET_OK) {
+      RCUTILS_LOG_ERROR_NAMED(ROS_PACKAGE_NAME, "Error in wait: %s", rcl_get_error_string().str);
+      return false;
+    }
+    for (size_t i = 0; i < wait_set.size_of_services; ++i) {
+      if (wait_set.services[i] && wait_set.services[i] == service) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
+bool
+wait_for_established_subscription(
+  const rcl_publisher_t * publisher,
+  size_t max_tries,
+  int64_t period_ms)
+{
+  size_t iteration = 0;
+  rcl_ret_t ret = RCL_RET_OK;
+  size_t subscription_count = 0;
+  while (iteration < max_tries) {
+    ++iteration;
+    ret = rcl_publisher_get_subscription_count(publisher, &subscription_count);
+    if (ret != RCL_RET_OK) {
+      RCUTILS_LOG_ERROR_NAMED(
+        ROS_PACKAGE_NAME,
+        "Error in rcl_publisher_get_subscription_count: %s",
+        rcl_get_error_string().str);
+      return false;
+    }
+    if (subscription_count == 1) {
+      return true;
+    }
+    std::this_thread::sleep_for(std::chrono::milliseconds(period_ms));
+  }
+  return false;
+}
+
+bool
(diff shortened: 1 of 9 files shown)


rcl/test/rcl/test_subscription.cpp
```diff
@@ -77,6 +77,29 @@ class CLASSNAME (TestSubscriptionFixture, RMW_IMPLEMENTATION) : public ::testing
...
+  bool & success)
+{
+  size_t iteration = 0;
+  rcl_ret_t ret = RCL_RET_OK;
+  size_t subscription_count = 0;
+  do {
+    ++iteration;
+    ret = rcl_publisher_get_subscription_count(publisher, &subscription_count);
+    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+    if (subscription_count > 0) {
```
fujitatomoya, line 94: subscription_count could be greater than 1? i think that we are expecting subscription_count == 1.

rcl/test/rcl/test_service.cpp
```diff
@@ -76,6 +77,30 @@ class CLASSNAME (TestServiceFixture, RMW_IMPLEMENTATION) : public ::testing::Tes
...
+    ++iteration;
+    ret = rcl_service_server_is_available(node, client, &is_ready);
+    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+    if (is_ready) {
+      success = true;
+      return;
+    }
+    std::this_thread::sleep_for(std::chrono::milliseconds(period_ms));
+  } while (iteration < max_tries);
+  success = false;
```
hidmic, line 101: @Barry-Xu-2018 consider returning this `bool` instead of `void` and a mutable reference for an output argument.

rcl/test/rcl/test_service.cpp
```diff
@@ -76,6 +77,30 @@ class CLASSNAME (TestServiceFixture, RMW_IMPLEMENTATION) : public ::testing::Tes
   }
 };
 
+void
+wait_for_server_to_be_available(
+  rcl_node_t * node,
+  rcl_client_t * client,
+  size_t max_tries,
```
hidmic, line 84: @Barry-Xu-2018 I believe `max_retries` would be more accurate with the corresponding change in the loop to fulfill that contract.

rcl/test/rcl/test_service.cpp
hidmic: Same below.

rcl/test/rcl/test_service.cpp
hidmic: Same below.

rcl/test/rcl/test_service.cpp
Barry-Xu-2018: @hidmic 

Thanks for your comments.  
If using **returning bool**, there is a macro expansion issue (Not decide to update the original logic **assert** code or using other ways).

In wait_for_server_to_be_available(), **assert** code is used.
```
  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;         ( contains a "return" statement inside)

    macro expanded:

  switch (0) case 0: default: if (const ::testing::AssertionResult gtest_ar = ( :: testing :: internal :: EqHelper < ( sizeof ( :: testing :: internal :: IsNullLiteralHelper ( 0 ) ) == 1 ) > :: Compare ( "0" , "ret" , 0 , ret ) )) ;
  else return ::testing::internal::AssertHelper( :: testing :: TestPartResult :: kFatalFailure , __FILE__ , 111 , gtest_ar . failure_message ( ) ) = ::testing::Message() << rcutils_get_error_string().str;
  

    build error:

/root/workspace/src/ros2/rcl/rcl/test/rcl/test_service.cpp: In function ‘bool wait_for_service_to_be_ready(rcl_service_t*, rcl_context_t*, size_t, int64_t)’:
/root/workspace/src/ros2/rcl/rcl/test/rcl/test_service.cpp:112:56: error: void value not ignored as it ought to be
                                                        ^~~
```

rcl/test/rcl/test_service.cpp
Barry-Xu-2018: @hidmic 

I think max_retries maybe make caller confusion. 
The value of max_retries is valid after trying once. So retry once is not the same as try once.  

```
ex:
    wait_for_subscription_to_be_ready(&subscription, context_ptr, 10, 100, success)

About 10,                                                                 
max_tries: it means to do something for 10 times
max_retries: it means to do something for 1 times (try) and 10 times (retry)
```

So it is better to use max_tries.  
What do you think ?

rcl/test/rcl/test_service.cpp
hidmic: I see. And yes, that happens because `ASSERT_*()` macros in `gtest` require the container function to return `void`. 

You can solve that by not using them there in the first place. I agree it's good to keep printing error strings, so maybe instead of a `bool`, you can return `rcl_ret_t` and do `ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str` on it.

rcl/test/rcl/test_service.cpp
hidmic: In that case, you must ban `max_tries = 0` since this code will always at least try once.

rcl/test/rcl/test_service.cpp
Barry-Xu-2018: Thanks for your suggestion.

rcl/test/rcl/test_service.cpp
Barry-Xu-2018: > In that case, you must ban `max_tries = 0` since this code will always at least try once.

Yes.  I update codes.

Barry-Xu-2018:
@hidmic 

Please check update codes at commit `0af91e4`.

Besides, I find a problem.  
While test `test_subscription` by RWM `cyclonedds`, there is a problem related to `wait_set`.  
I am investigating this.   

```
[ RUN      ] TestSubscriptionFixture__rmw_cyclonedds_cpp.test_subscription_nominal_string
1587879829.752939 [0] test_subsc: using network interface enp0s31f6 (udp/192.168.0.200) selected arbitrarily from: enp0s31f6, docker0

  'Security was requested but the Cyclone DDS being used does not have security support enabled. Recompile Cyclone DDS with the '-DENABLE_SECURITY=ON' CMake option, at /root/workspace/src/ros2/rmw_cyclonedds/rmw_cyclonedds_cpp/src/rmw_node.cpp:857'

with this new error message:

  'failed to attach dummy guard condition for blocking on empty waitset, at /root/workspace/src/ros2/rmw_cyclonedds/rmw_cyclonedds_cpp/src/rmw_node.cpp:2640'

<<<

  'failed to attach dummy guard condition for blocking on empty waitset, at /root/workspace/src/ros2/rmw_cyclonedds/rmw_cyclonedds_cpp/src/rmw_node.cpp:2640'

with this new error message:

  'wait_set is null, at /root/workspace/src/ros2/rmw_cyclonedds/rmw_cyclonedds_cpp/src/rmw_node.cpp:2664'

<<<
/root/workspace/src/ros2/rcl/rcl/test/rcl/test_subscription.cpp:300: Failure
Expected equality of these values:
  0
  ret
    Which is: 900
wait_set is null, at /root/workspace/src/ros2/rmw_cyclonedds/rmw_cyclonedds_cpp/src/rmw_node.cpp:2664
[  FAILED  ] TestSubscriptionFixture__rmw_cyclonedds_cpp.test_subscription_nominal_string (10 ms)
```

Barry-Xu-2018:
@hidmic 

I find the cause for above problem.  
Please look at fixing commit `bbf300c`

rcl/test/rcl/test_subscription.cpp
```diff
@@ -77,7 +77,34 @@ class CLASSNAME (TestSubscriptionFixture, RMW_IMPLEMENTATION) : public ::testing
...
+      return ret;
+    }
+    if (subscription_count == 1) {
+      success = true;
+      return ret;
+    }
+    std::this_thread::sleep_for(std::chrono::milliseconds(period_ms));
+  }
+  success = false;
+  return ret;
```
hidmic, line 104: @Barry-Xu-2018 I still can't fathom why you need an additional boolean, but others may think otherwise.

clalancette: I agree with @hidmic; we can just use the `rcl_ret_t` of the function.  No need for the additional `success` variable, it doesn't really add any new information.  This also goes for `wait_for_server_to_be_available` above.

rcl/test/rcl/test_service.cpp
```diff
@@ -87,32 +115,39 @@ wait_for_service_to_be_ready(
   rcl_wait_set_t wait_set = rcl_get_zero_initialized_wait_set();
   rcl_ret_t ret =
     rcl_wait_set_init(&wait_set, 0, 0, 0, 0, 1, 0, context, rcl_get_default_allocator());
-  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
-  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
-  {
-    rcl_ret_t ret = rcl_wait_set_fini(&wait_set);
-    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
```
ivanpauno, line 94: why was this deleted? When returning early, the wait set is not being cleaned.

rcl/test/rcl/test_subscription.cpp
```diff
@@ -88,32 +115,39 @@ wait_for_subscription_to_be_ready(
   rcl_wait_set_t wait_set = rcl_get_zero_initialized_wait_set();
   rcl_ret_t ret =
     rcl_wait_set_init(&wait_set, 1, 0, 0, 0, 0, 0, context, rcl_get_default_allocator());
-  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
-  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
-  {
-    rcl_ret_t ret = rcl_wait_set_fini(&wait_set);
-    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
-  });
```
ivanpauno, line 96: same about this other `scope_exit`. IMO, it's still needed.

rcl/test/rcl/test_service.cpp
```diff
@@ -76,7 +77,34 @@ class CLASSNAME (TestServiceFixture, RMW_IMPLEMENTATION) : public ::testing::Tes
...
+      return ret;
+    }
+    std::this_thread::sleep_for(std::chrono::milliseconds(period_ms));
+  }
+  success = false;
+  return ret;
+}
+
+rcl_ret_t
 wait_for_service_to_be_ready(
```
ivanpauno, line 108: This function is repeated also here: https://github.com/ros2/rcl/blob/65feb5c7255a41da0311664fbb8104a2f086af2e/rcl/test/rcl/service_fixture.cpp#L32.

It would be good to move it to a common header in `rcl/test`.
Though, it may be out of the scope of the PR.

rcl/test/rcl/test_service.cpp
Barry-Xu-2018: Sorry, it's my fault.

rcl/test/rcl/test_service.cpp
Barry-Xu-2018: Yes. Good suggestion.   
Update this.

rcl/test/rcl/test_subscription.cpp
Barry-Xu-2018: Thanks.  Update codes.

rcl/test/rcl/common.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+#ifndef RCL__COMMON_HPP_
+#define RCL__COMMON_HPP_
+
+#include "rcl/client.h"
+#include "rcl/service.h"
+#include "rcl/rcl.h"
+
+bool
+wait_for_server_to_be_available(
```
ivanpauno, line 23: Please, add a minimal doc block to each function (it's used just for testing, but it'll help for future maintainers)

rcl/test/rcl/common.hpp
```diff
@@ -0,0 +1,56 @@
...
+  int64_t period_ms);
+
+bool
+wait_for_established_subscription(
+  const rcl_publisher_t * publisher,
+  size_t max_tries,
+  int64_t period_ms);
+
+bool
+wait_for_subscription_to_be_ready(
```
ivanpauno, line 50: I think this one (and maybe others) are only used in one file right now, so moving it to `common` is not strictly needed.
It's still something that might be useful for a new test, so I think that it's fine to move it here.

rcl/test/rcl/common.hpp
```diff
@@ -0,0 +1,56 @@
+// Copyright 2020 Open Source Robotics Foundation, Inc.
```
ivanpauno, line 1: maybe the name of the file could be `wait_for_entity_helpers`, or sth like that

rcl/test/rcl/common.cpp
```diff
@@ -0,0 +1,237 @@
...
+    ++iteration;
+    ret = rcl_publisher_get_subscription_count(publisher, &subscription_count);
+    if (ret != RCL_RET_OK) {
+      RCUTILS_LOG_ERROR_NAMED(
+        ROS_PACKAGE_NAME,
+        "Error in rcl_publisher_get_subscription_count: %s",
+        rcl_get_error_string().str);
+      return false;
+    }
+    if (subscription_count == 1) {
```
hidmic, line 178: @Barry-Xu-2018 consider 
```suggestion
    if (subscription_count > 0) {
```
instead.

rcl/test/rcl/common.hpp
Barry-Xu-2018: Okay. Add doc bock for each function.

rcl/test/rcl/common.hpp
Barry-Xu-2018: OKay

rcl/test/rcl/common.cpp
Barry-Xu-2018: Okay.

rcl/test/rcl/wait_for_entity_helpers.hpp
```diff
@@ -12,45 +12,55 @@
...
+/// Wait a publisher to get one or more established subscriptions
+/// by a maximum trying times and a milliseconds period
 bool
 wait_for_established_subscription(
   const rcl_publisher_t * publisher,
   size_t max_tries,
   int64_t period_ms);
 
+/// Wait a subscription to be ready with a context
+/// by a maximum trying times and a milliseconds period
```
ivanpauno, line 58:
```suggestion
/// Wait a subscription to be ready, i.e. a message is ready to be handled,
/// by trying at least `times` times with a `period_ms` period.
```

Barry-Xu-2018: @ivanpauno 

All comments have been updated.  
Please review again.

Barry-Xu-2018:
@clalancette @fujitatomoya  

Do you have any comments ?

Barry-Xu-2018:
After checking failure for `Fpr__rcl__ubuntu_focal_amd64`, the failure testcase `rcl.TestServiceFixture__rmw_fastrtps_cpp.test_service_nominal` isn't related to this modification.   
Besides, I find closed build #19 #18 also includes the same failure testcase.  
So I don't do further investigation.

ivanpauno:
CI up to rcl:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10711/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6123/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8724/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10610/)

ivanpauno:
@Barry-Xu-2018 can you check Windows build errors? thanks!

Barry-Xu-2018:
@ivanpauno 

> can you check Windows build errors?  

I fixed it.

ivanpauno:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10622/)

ivanpauno:
@Barry-Xu-2018 can you rebase?

Barry-Xu-2018:
@ivanpauno 

> can you rebase?

Done.

ivanpauno:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10644/)

ivanpauno:
Thanks for the contribution @Barry-Xu-2018, going in!

PR #640, opened 2020-05-01
Make test_two_timers_ready_before_timeout more reliable

ivanpauno:
Fixes https://ci.ros2.org/view/nightly/job/nightly_win_rep/1835/testReport/junit/rcl/TestTimerFixture/test_timer_not_ready/ and similar failures on CI.

Same rationale as https://github.com/ros2/system_tests/pull/430.

### DIFF
--- a/rcl/test/rcl/test_timer.cpp
+++ b/rcl/test/rcl/test_timer.cpp
@@ -117,11 +117,11 @@ TEST_F(TestTimerFixture, test_two_timers) {
   rcl_timer_t timer2 = rcl_get_zero_initialized_timer();
 
   ret = rcl_timer_init(
-    &timer, &clock, this->context_ptr, RCL_MS_TO_NS(5), nullptr, rcl_get_default_allocator());
+    &timer, &clock, this->context_ptr, RCL_MS_TO_NS(50), nullptr, rcl_get_default_allocator());
   ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
 
   ret = rcl_timer_init(
-    &timer2, &clock, this->context_ptr, RCL_MS_TO_NS(20), nullptr, rcl_get_default_allocator());
+    &timer2, &clock, this->context_ptr, RCL_MS_TO_NS(1000), nullptr, rcl_get_default_allocator());
   ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
 
   rcl_wait_set_t wait_set = rcl_get_zero_initialized_wait_set();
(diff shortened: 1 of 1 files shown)


ivanpauno:
I've tried locally, and it still fails.
I will take a closer look to it.

ivanpauno:
CI using `--retest-until-fail 10 --packages-select rcl --ctest-args -R test_timer`, to make sure the test is actually more reliable after this change:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10542/)

ivanpauno:
there seems to be a real bug here!
I'll have to dig a bit more

clalancette:
@ivanpauno I think you were on the right track here, you just didn't go far enough and relax the timing constraints on some of the other tests.  a64f2ca does that.  Let's see how this does in CI:


* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10591/)

ivanpauno:
Oh yes, you're right.
I thought it was the same test failing, but it was a new one

rcl/test/rcl/test_timer.cpp
```diff
@@ -141,7 +141,7 @@ TEST_F(TestTimerFixture, test_two_timers) {
     ret = rcl_wait_set_fini(&wait_set);
     EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
   });
-  ret = rcl_wait(&wait_set, RCL_MS_TO_NS(10));
+  ret = rcl_wait(&wait_set, RCL_MS_TO_NS(100));
```
ivanpauno, line 144: shouldn't this one be less than `time2` period (100ms).

rcl/test/rcl/test_timer.cpp
```diff
@@ -296,7 +296,7 @@ TEST_F(TestTimerFixture, test_canceled_timer) {
     ret = rcl_wait_set_fini(&wait_set);
     EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
   });
-  ret = rcl_wait(&wait_set, RCL_MS_TO_NS(1));
+  ret = rcl_wait(&wait_set, RCL_MS_TO_NS(100));
```
ivanpauno, line 299: I think this one should be greater than `timer` period (500ms).

rcl/test/rcl/test_timer.cpp
clalancette: Bah, transcoding error.  The timer2 period should be 1000ms.  Will fix.

rcl/test/rcl/test_timer.cpp
clalancette: I misunderstood the test before.

In some sense, these numbers don't matter at all.  We create the timer, and then immediately cancel it, so it will never expire.  From that perspective, we can just do the shortest wait possible (1 millisecond), and it doesn't matter how long that takes; the timer will never expire, so the below stuff will always be fine.  I'm actually going to revert this set of changes as it isn't necessary.

clalancette:
I've made the fixes from review, rebased and squashed everything together.  Here's another CI run:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10593/)

clalancette:
> github doesn't allow me to click `Approve`, as I created the PR, but I approve joy.

I've also approved it :).  The Windows CI run is green; I'm going to run one more CI on all platforms, then I'll merge:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10696/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6109/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8705/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10594/)

clalancette:
All right, all platforms happy (macOS failures are on the nightly as well).  Merging this, thanks @ivanpauno .

PR #642, opened 2020-05-04
Initialize service timestamps to 0 and test.

iluetkeb:
@ivanpauno 2nd part of the service timestamp fix.

### DIFF
--- a/rcl/src/rcl/client.c
+++ b/rcl/src/rcl/client.c
@@ -303,6 +303,8 @@ rcl_take_response_with_info(
   RCL_CHECK_ARGUMENT_FOR_NULL(ros_response, RCL_RET_INVALID_ARGUMENT);
 
   bool taken = false;
+  request_header->source_timestamp = 0;
+  request_header->received_timestamp = 0;
   if (rmw_take_response(
       client->impl->rmw_handle, request_header, ros_response, &taken) != RMW_RET_OK)
   {
--- a/rcl/test/rcl/test_service.cpp
+++ b/rcl/test/rcl/test_service.cpp
@@ -182,6 +182,9 @@ TEST_F(CLASSNAME(TestServiceFixture, RMW_IMPLEMENTATION), test_service_nominal) 
   client_request.uint8_value = 1;
   client_request.uint32_value = 2;
   int64_t sequence_number;
+  rcutils_time_point_value_t start_timestamp;
+  // take timestamp before sending request
+  EXPECT_EQ(RCUTILS_RET_OK, rcutils_system_time_now(&start_timestamp));
   ret = rcl_send_request(&client, &client_request, &sequence_number);
   EXPECT_EQ(sequence_number, 1);
   test_msgs__srv__BasicTypes_Request__fini(&client_request);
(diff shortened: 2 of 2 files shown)


rcl/test/rcl/test_service.cpp
```diff
@@ -211,8 +214,22 @@ TEST_F(CLASSNAME(TestServiceFixture, RMW_IMPLEMENTATION), test_service_nominal)
 
     EXPECT_EQ(1, service_request.uint8_value);
     EXPECT_EQ(2UL, service_request.uint32_value);
+#ifdef RMW_TIMESTAMPS_SUPPORTED
```
ivanpauno, line 217: is this set by default?

rcl/test/rcl/test_service.cpp
```diff
@@ -211,8 +214,22 @@ TEST_F(CLASSNAME(TestServiceFixture, RMW_IMPLEMENTATION), test_service_nominal)
...
+    EXPECT_GE(header.source_timestamp, start_timestamp);
+#ifdef RMW_RECEIVED_TIMESTAMP_SUPPORTED
```
ivanpauno, line 219: is this set by default?

rcl/src/rcl/client.c
```diff
@@ -303,6 +303,8 @@ rcl_take_response_with_info(
   RCL_CHECK_ARGUMENT_FOR_NULL(ros_response, RCL_RET_INVALID_ARGUMENT);
 
   bool taken = false;
+  request_header->source_timestamp = 0;
+  request_header->received_timestamp = 0;
```
ivanpauno, line 307: It would be nice to have in `rmw` a `rmw_get_zero_initialized_service_info` function, though I don't mind strongly.

rcl/test/rcl/test_service.cpp
ivanpauno: I see there's logic for this in the CMakeLists.txt, NVM

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10621/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6057/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8650/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10519/)

rcl/src/rcl/client.c
ivanpauno: This suggestion can be done in a follow up, for now let's go ahead with this and https://github.com/ros2/rmw_fastrtps/pull/378.

iluetkeb: > It would be nice to have in rmw a rmw_get_zero_initialized_service_info function, 

Would this still be acceptable now, i.e. after the API freeze? I can easily add it, just wondering whether it would be taken.

ivanpauno: > Would this still be acceptable now, i.e. after the API freeze? I can easily add it, just wondering whether it would be taken.

In general, new API isn't accepted after the freeze. But in this case, it seems to be trivial enough (for me) and it will avoid future errors. @jacobperron do you think it's ok to add that now?

jacobperron: I'd rather not change/add any new API now, unless there is really good reason to do so.

PR #646, opened 2020-05-08
Add sleep before getting publisher info.

nuclearsandwich:
This test appears to be timing/load sensitive. I can reproduce
intermittent failures locally when stressing my host and the more stress
the faster the reproduction. Since adding the sleep I have not seen any
failures to find the expected publisher info locally.

This seems to resolve unreliable test failures such as [this one](https://ci.ros2.org/job/nightly_linux-aarch64_repeated/1185/testReport/projectroot/test/test_info_by_topic__rmw_cyclonedds_cpp/). There still appear to be occasional segfaults when using rmw_fastrtps_cpp or rmw_fastrtps_dynamic_cpp which I don't think are related.

### DIFF
--- a/rcl/test/rcl/test_info_by_topic.cpp
+++ b/rcl/test/rcl/test_info_by_topic.cpp
@@ -18,8 +18,10 @@
 
 #include <gtest/gtest.h>
 
+#include <chrono>
 #include <memory>
 #include <string>
+#include <thread>
 
 #include "rcl/error_handling.h"
 #include "rcl/graph.h"
@@ -357,6 +359,7 @@ TEST_F(
     &subscription_options);
   ASSERT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
   const std::string fqdn = std::string("/") + this->topic_name;
+  std::this_thread::sleep_for(std::chrono::milliseconds(250));
   // Get publishers info by topic
   rmw_topic_endpoint_info_array_t topic_endpoint_info_array_pub =
     rmw_get_zero_initialized_topic_endpoint_info_array();
--- a/rcl/test/rcl/test_info_by_topic.cpp
+++ b/rcl/test/rcl/test_info_by_topic.cpp
@@ -362,31 +362,6 @@ TEST_F(
   ASSERT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
   const std::string fqdn = std::string("/") + this->topic_name;
   ASSERT_TRUE(wait_for_established_subscription(&publisher, 10, 100));
-  //size_t max_tries = 10;
-  //size_t iteration = 0;
-  //size_t publisher_count = 0;
-  //do {
-  //      	ret = rcl_count_publishers(&this->node, fqdn.c_str(), &publisher_count);
-  //      	if (publisher_count >= 1) {
-  //      		break;
-  //      	}
-  //      	++iteration;
-  //      	std::this_thread::sleep_for(std::chrono::milliseconds(100));
-  //} while (iteration < max_tries || publisher_count < 1);
-  //      ASSERT_EQ(publisher_count, 1u) << "Unexpected publisher count";
-  //max_tries = 10;
-  //iteration = 0;
-  //size_t subscriber_count = 0;
-  //do {
-  //      	ret = rcl_count_subscribers(&this->node, fqdn.c_str(), &subscriber_count);
-  //      	if (subscriber_count >= 1) {
-  //      		break;
-  //      	}
-  //      	++iteration;
-  //      	std::this_thread::sleep_for(std::chrono::milliseconds(100));
-  //} while (iteration < max_tries);
-  //      ASSERT_EQ(subscriber_count, 1u) << "Unexpected subscriber count";
-  //ASSERT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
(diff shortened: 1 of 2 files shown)


nuclearsandwich:
@jacobperron I paged you since I know you'd been looking at timing-sensitive tests in other parts of the stack. I don't know whether it's expected that there's a time after the creation of the publisher and subscriber where they may not be added to the count.

jacobperron:
> don't know whether it's expected that there's a time after the creation of the publisher and subscriber where they may not be added to the count

I wouldn't think to expect a delay, but I suppose it can vary based on what's happening in the RMW implementation. Does the failure only occur with `rmw_cyclonedds_cpp`?

rcl/test/rcl/test_info_by_topic.cpp
```diff
@@ -357,6 +359,7 @@ TEST_F(
     &subscription_options);
   ASSERT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
   const std::string fqdn = std::string("/") + this->topic_name;
+  std::this_thread::sleep_for(std::chrono::milliseconds(250));
```
wjwwood, line 362: If this does help, then I'd recommend polling in a loop, because this only requires a minor hiccup in the machine to still flake.

nuclearsandwich:
> I wouldn't expect a delay, but I suppose it can vary based on what's happening in the RMW implementation. Does the failure only occur with `rmw_cyclonedds_cpp`?

I have to go back and check my test logs. I think I've seen this failure with Fast-RTPS as well but I also get occasional segfaults with Fast-RTPS (as does the build farm).

nuclearsandwich:
In all my test runs I only ever saw this exact failure with cyclonedds.
```
38222-[62.103s] 14: [ RUN      ] TestInfoByTopicFixture__rmw_cyclonedds_cpp.test_rcl_get_publishers_subscription_info_by_topic
38223-[62.107s] 14: /home/steven/osrf/bugcrunch/ws/src/ros2/rcl/rcl/test/rcl/test_info_by_topic.cpp:380: Failure
38224:[62.108s] 14: Expected equality of these values:
38225-[62.108s] 14:   topic_endpoint_info_array_sub.size
38226-[62.108s] 14:     Which is: 0
38227-[62.108s] 14:   1u
38228-[62.108s] 14:     Which is: 1
```

fujitatomoya:
@nuclearsandwich 

i would use rcl_count_publishers & rcl_count_subscribers to make sure if publication and subscription are ready instead of sleeping. if it can count writer/reader from rmw graph cache, which means that it can get writer/reader info for publication and subscription.

nuclearsandwich:
> i would use rcl_count_publishers & rcl_count_subscribers to make sure if publication and subscription are ready instead of sleeping. if it can count writer/reader from rmw graph cache, which means that it can get writer/reader info for publication and subscription.

Thanks for the suggestion. If I understand correctly this would essentially be creating the publisher and subscriber and polling to determine that they were actually created before trying to use them?

fujitatomoya:
@nuclearsandwich 

No, rcl_count_publishers & rcl_count_subscribers does not include polling, it is one shot to count the endpoint. so as @wjwwood mentioned, polling would be required in the test code to confirm the publication and subscription are ready.

i would do the following,

1. create publication and subscription with test topic.
2. wait until publisher count and subscriber count with test topic becomes greater than 0.
3. issue rcl_get_publishers_info_by_topic/rcl_get_subscriptions_info_by_topic.

BTW, helper functions are getting organized https://github.com/ros2/rcl/pull/631, so this might get along with it.

nuclearsandwich:
I appreciate the guidance @fujitatomoya. I'm stepping out of my usual wheelhouse for this one. I took a look at #631 and rebased this PR onto it. Which means that I'll rebase this once more when #631 actually merges. Using the wait for entity helpers does show improvement.

fujitatomoya:
@nuclearsandwich 

could you rebase? https://github.com/ros2/rcl/pull/631 is merged.

rcl/test/rcl/test_info_by_topic.cpp
```diff
@@ -357,6 +361,8 @@ TEST_F(
     &subscription_options);
   ASSERT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
   const std::string fqdn = std::string("/") + this->topic_name;
+  ASSERT_TRUE(wait_for_established_subscription(&publisher, 10, 100));
+  ASSERT_TRUE(wait_for_subscription_to_be_ready(&subscription, &this->context, 10, 100));
```
fujitatomoya, line 365: I am not sure this needs to be called here. I believe that we can confirm that graph cache is constructed vi wait_for_established_subscription.

iuhilnehc-ynos: This case tests both `rcl_get_publishers_info_by_topic` and `rcl_get_subscriptions_info_by_topic`, so I think we need to add a new wait_helper(such as `wait_for_established_publisher`) which is similar to `wait_for_established_subscription`. (make sure two directions are established.)

NOTE: `wait_for_subscription_to_be_ready` is used for a subscriber to wait to get message data from a publisher.
it means there should be a `rcl_publish` calling before (or at a new thread) and then `rcl_take` to get message data.
So we don't use it here.

clalancette:
This looks to be superseded by #859, so I'm going to close it out.  Feel free to reopen if you think otherwise.

PR #649, opened 2020-05-14
Increase rcl_lifecycle test coverage and add more safety checks

brawner:
As part of the push to Quality Level 1, this adds more more tests of the rcl_lifecycle API. While writing these tests, I found a few instances where more checks were needed.

If I didn't understand parts of the API and added checks where they weren't desired, feel free to make comments below. Alternatively, if there are more thorough checks that people feel are necessary I can also add those in. These were pretty much what I found passing around bad parameters.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10750/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6145/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8752/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10639/)
* Coverage [Build Status](https://ci.ros2.org/job/ci_linux_coverage/94/)

### DIFF
--- a/rcl_lifecycle/src/default_state_machine.c
+++ b/rcl_lifecycle/src/default_state_machine.c
@@ -692,25 +692,25 @@ rcl_lifecycle_init_default_state_machine(
 
 // Semicolon handles the "a label can only be part of a statement..." error
 fail:;
-  // If rcl_lifecycle_transition_map_fini() fails, it will clobber the error string here, twice
-  // Here, we concatenate the error strings if that happens
+  // If rcl_lifecycle_transition_map_fini() fails, it will clobber the error string here.
+  // Concatenate the error strings if that happens
   const char * current_error = (rcl_error_is_set()) ? rcl_get_error_string().str : "";
-  rcl_reset_error();
 
   if (rcl_lifecycle_transition_map_fini(&state_machine->transition_map, allocator) != RCL_RET_OK) {
-    const char * fini_error = (rcl_error_is_set()) ? rcl_get_error_string().str : "";
-    rcl_reset_error();
-
+    const char * fini_error = "";
+    if (rcl_error_is_set()) {
+      fini_error = rcl_get_error_string().str;
+      rcl_reset_error();
+    }
     RCL_SET_ERROR_MSG_WITH_FORMAT_STRING(
       "Freeing transition map failed while handling a previous error. Leaking memory!"
       "\nOriginal error:\n\t%s\nError encountered in rcl_lifecycle_transition_map_fini():\n\t%s\n",
       current_error, fini_error);
-  } else if (strcmp(current_error, "") != 0) {
-    RCL_SET_ERROR_MSG(current_error);
-  } else {
-    RCL_SET_ERROR_MSG("Unspecified error in default_state_machine _register_transitions()");
   }
 
+  if (!rcl_error_is_set()) {
+    RCL_SET_ERROR_MSG("Unspecified error in default_state_machine _register_transitions()");
+  }
   return RCL_RET_ERROR;
 }
 
--- a/rcl_lifecycle/src/default_state_machine.c
+++ b/rcl_lifecycle/src/default_state_machine.c
@@ -708,7 +708,7 @@ fail:;
   } else if (strcmp(current_error, "") != 0) {
     RCL_SET_ERROR_MSG(current_error);
   } else {
-    RCL_SET_ERROR_MSG("Unspecified in default_state_machine _register_transitions()");
+    RCL_SET_ERROR_MSG("Unspecified error in default_state_machine _register_transitions()");
   }
 
   return RCL_RET_ERROR;
--- a/rcl_lifecycle/src/default_state_machine.c
+++ b/rcl_lifecycle/src/default_state_machine.c
@@ -650,7 +650,7 @@ _register_transitions(
   return ret;
 }
 
-// default implementation as despicted on
+// default implementation as depicted on
 // design.ros2.org
 rcl_ret_t
 rcl_lifecycle_init_default_state_machine(
--- a/rcl_lifecycle/src/default_state_machine.c
+++ b/rcl_lifecycle/src/default_state_machine.c
@@ -691,6 +691,7 @@ rcl_lifecycle_init_default_state_machine(
   return ret;
 
 fail:
+  // if rcl_lifecycle_transition_map_fini() fails, it will clobber the error string here, twice
   if (rcl_lifecycle_transition_map_fini(&state_machine->transition_map, allocator) != RCL_RET_OK) {
     RCL_SET_ERROR_MSG("could not free lifecycle transition map. Leaking memory!\n");
   }
--- a/rcl_lifecycle/src/default_state_machine.c
+++ b/rcl_lifecycle/src/default_state_machine.c
@@ -690,11 +690,25 @@ rcl_lifecycle_init_default_state_machine(
 
   return ret;
 
-fail:
-  // if rcl_lifecycle_transition_map_fini() fails, it will clobber the error string here, twice
+// Semicolon handles the "a label can only be part of a statement..." error
+fail:;
+  // If rcl_lifecycle_transition_map_fini() fails, it will clobber the error string here, twice
+  // Here, we concatenate the error strings if that happens
+  const char * current_error = (rcl_error_is_set()) ? rcl_get_error_string().str : "";
+  rcl_reset_error();
+
   if (rcl_lifecycle_transition_map_fini(&state_machine->transition_map, allocator) != RCL_RET_OK) {
-    RCL_SET_ERROR_MSG("could not free lifecycle transition map. Leaking memory!\n");
+    const char * fini_error = (rcl_error_is_set()) ? rcl_get_error_string().str : "";
+    rcl_reset_error();
+
+    RCL_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "Freeing transition map failed while handling a previous error. Leaking memory!"
+      "\nOriginal error:\n\t%s\nError encountered in rcl_lifecycle_transition_map_fini():\n\t%s\n",
+      current_error, fini_error);
+  } else {
+    RCL_SET_ERROR_MSG(current_error);
   }
+
   return RCL_RET_ERROR;
 }
 
--- a/rcl_lifecycle/src/default_state_machine.c
+++ b/rcl_lifecycle/src/default_state_machine.c
@@ -657,6 +657,8 @@ rcl_lifecycle_init_default_state_machine(
   rcl_lifecycle_state_machine_t * state_machine, const rcutils_allocator_t * allocator)
 {
   rcl_ret_t ret = RCL_RET_ERROR;
+  // Used for concatenating error messages in the fail: block.
+  const char * fail_error_message = "";
 
   // ***************************
   // register all primary states
@@ -690,11 +692,12 @@ rcl_lifecycle_init_default_state_machine(
 
   return ret;
 
-// Semicolon handles the "a label can only be part of a statement..." error
-fail:;
+fail:
   // If rcl_lifecycle_transition_map_fini() fails, it will clobber the error string here.
   // Concatenate the error strings if that happens
-  const char * current_error = (rcl_error_is_set()) ? rcl_get_error_string().str : "";
+  if (rcl_error_is_set()) {
+    fail_error_message = rcl_get_error_string().str;
+  }
 
   if (rcl_lifecycle_transition_map_fini(&state_machine->transition_map, allocator) != RCL_RET_OK) {
     const char * fini_error = "";
@@ -705,7 +708,7 @@ fail:;
     RCL_SET_ERROR_MSG_WITH_FORMAT_STRING(
       "Freeing transition map failed while handling a previous error. Leaking memory!"
       "\nOriginal error:\n\t%s\nError encountered in rcl_lifecycle_transition_map_fini():\n\t%s\n",
-      current_error, fini_error);
+      fail_error_message, fini_error);
   }
 
   if (!rcl_error_is_set()) {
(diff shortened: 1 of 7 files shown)


brawner:
It looks like there is a bit more I can do to boost coverage. I'll work on this more tomorrow.

brawner:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10769/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6161/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8767/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10659/)

brawner:
Much better coverage, but com_interface and default_state_machine are dragging it down.


Name  ↓ | Classes | Lines | Conditionals
-- | -- | -- | --
com_interface.c | 100% 1/1 | 77% 66/86 | 66/86
default_state_machine.c | 100% 1/1 | 80% 147/184  | 147/184
rcl_lifecycle.c | 100% 1/1  | 92% 167/181 |  167/181
transition_map.c | 100% 1/1 |  97% 77/79  | 77/79

rcl_lifecycle/src/rcl_lifecycle.c
```diff
@@ -58,6 +60,10 @@ rcl_lifecycle_state_init(
     RCL_SET_ERROR_MSG("state pointer is null\n");
     return RCL_RET_ERROR;
   }
+  if (!label) {
+    RCL_SET_ERROR_MSG("State label is null\n");
+    return RCL_RET_ERROR;
```
ahcorde, line 65:
```suggestion
    return RCL_RET_INVALID_ARGUMENT;
```

rcl_lifecycle/src/rcl_lifecycle.c
```diff
@@ -118,7 +124,22 @@ rcl_lifecycle_transition_init(
 
   if (!transition) {
     RCL_SET_ERROR_MSG("transition pointer is null\n");
-    return RCL_RET_OK;
+    return RCL_RET_ERROR;
```
ahcorde, line 127:
```suggestion
    return RCL_RET_INVALID_ARGUMENT;
```

rcl_lifecycle/src/rcl_lifecycle.c
```diff
@@ -118,7 +124,22 @@ rcl_lifecycle_transition_init(
...
+  }
+
+  if (!label) {
+    RCL_SET_ERROR_MSG("label pointer is null\n");
+    return RCL_RET_ERROR;
```
ahcorde, line 132:
```suggestion
    return RCL_RET_INVALID_ARGUMENT;
```

rcl_lifecycle/src/rcl_lifecycle.c
```diff
@@ -118,7 +124,22 @@ rcl_lifecycle_transition_init(
...
+  }
+
+  if (!start) {
+    RCL_SET_ERROR_MSG("start state pointer is null\n");
+    return RCL_RET_ERROR;
```
ahcorde, line 137:
```suggestion
    return RCL_RET_INVALID_ARGUMENT;
```

rcl_lifecycle/src/rcl_lifecycle.c
```diff
@@ -118,7 +124,22 @@ rcl_lifecycle_transition_init(
...
+  }
+
+  if (!goal) {
+    RCL_SET_ERROR_MSG("goal state pointer is null\n");
+    return RCL_RET_ERROR;
```
ahcorde, line 142:
```suggestion
    return RCL_RET_INVALID_ARGUMENT;
```

rcl_lifecycle/src/rcl_lifecycle.c
```diff
@@ -140,7 +161,7 @@ rcl_lifecycle_transition_fini(
   const rcl_allocator_t * allocator)
 {
   if (!allocator) {
-    RCL_SET_ERROR_MSG("can't initialize transition, no allocator given\n");
+    RCL_SET_ERROR_MSG("can't finalize transition, no allocator given\n");
     return RCL_RET_ERROR;
```
ahcorde, line 165:
```suggestion
    return RCL_RET_INVALID_ARGUMENT;
```

rcl_lifecycle/src/rcl_lifecycle.c
```diff
@@ -192,6 +213,14 @@ rcl_lifecycle_state_machine_init(
   bool default_states,
   const rcl_allocator_t * allocator)
 {
+  if (!state_machine) {
+    RCL_SET_ERROR_MSG("State machine is null\n");
+    return RCL_RET_ERROR;
```
ahcorde, line 218:
```suggestion
    return RCL_RET_INVALID_ARGUMENT;
```

rcl_lifecycle/src/rcl_lifecycle.c
```diff
@@ -192,6 +213,14 @@ rcl_lifecycle_state_machine_init(
...
+  }
+  if (!node_handle) {
+    RCL_SET_ERROR_MSG("Node handle is null\n");
+    return RCL_RET_ERROR;
```
ahcorde, line 222:
```suggestion
    return RCL_RET_INVALID_ARGUMENT;
```

rcl_lifecycle/src/rcl_lifecycle.c
```diff
@@ -192,6 +213,14 @@ rcl_lifecycle_state_machine_init(
...
+  }
   if (!allocator) {
     RCL_SET_ERROR_MSG("can't initialize state machine, no allocator given\n");
     return RCL_RET_ERROR;
```
ahcorde, line 226:
```suggestion
    return RCL_RET_INVALID_ARGUMENT;
```

rcl_lifecycle/src/transition_map.c
```diff
@@ -53,6 +53,11 @@ rcl_lifecycle_transition_map_fini(
   rcl_lifecycle_transition_map_t * transition_map,
   const rcutils_allocator_t * allocator)
 {
+  if (!allocator) {
+    RCL_SET_ERROR_MSG("can't free transition map, no allocator given\n");
+    return RCL_RET_ERROR;
```
ahcorde, line 58:
```suggestion
    return RCL_RET_INVALID_ARGUMENT;
```

brawner:
I'm not sure I can change the error codes of already existing functions this late in the game. While, I agree invalid_argument is better than error, I chose RCL_RET_ERROR to match the rest of the code.

rcl_lifecycle/src/default_state_machine.c
```diff
@@ -691,6 +691,7 @@ rcl_lifecycle_init_default_state_machine(
   return ret;
 
 fail:
+  // if rcl_lifecycle_transition_map_fini() fails, it will clobber the error string here, twice
```
Karsten1987, line 694: is that meant as a TODO here? If `fini()` fails and sets its own error message, I think we can either reset it here or concatenate. I think overwriting the error message actually yields a warning.

brawner: No it was an issue I had found, and didn't have a great way of resolving it. I didn't know concatenation would work, I'll add that in. 

It will indeed print a warning if the error message is set and then it proceeds to set another error message. However, I didn't want to lose the trace of errors messages as they fall from one failure to the next (so it would have been impossible to see the original failure case).

rcl_lifecycle/src/default_state_machine.c
```diff
@@ -690,11 +690,25 @@ rcl_lifecycle_init_default_state_machine(
...
-    RCL_SET_ERROR_MSG("could not free lifecycle transition map. Leaking memory!\n");
+    const char * fini_error = (rcl_error_is_set()) ? rcl_get_error_string().str : "";
+    rcl_reset_error();
+
+    RCL_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "Freeing transition map failed while handling a previous error. Leaking memory!"
+      "\nOriginal error:\n\t%s\nError encountered in rcl_lifecycle_transition_map_fini():\n\t%s\n",
+      current_error, fini_error);
+  } else {
+    RCL_SET_ERROR_MSG(current_error);
```
Karsten1987, line 709: If I see this correctly here, this would mean we always set an error message, even if set to empty string. Do we need that? We could pick up the previous error message only when fini actually fails, couldn't we?

brawner:
Checking one last time

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10835/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6209/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8825/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10720/)

rcl_lifecycle/src/default_state_machine.c
brawner: Ah, good catch. Added a check for the empty string and a default error message if one hadn't been set for some reason.

rcl_lifecycle/src/default_state_machine.c
```diff
@@ -690,10 +690,27 @@ rcl_lifecycle_init_default_state_machine(
 
   return ret;
 
-fail:
+// Semicolon handles the "a label can only be part of a statement..." error
+fail:;
```
Karsten1987, line 694:
```suggestion
fail:
```

Or was that semicolon intended?

rcl_lifecycle/src/default_state_machine.c
```diff
@@ -690,10 +690,27 @@ rcl_lifecycle_init_default_state_machine(
...
+  // If rcl_lifecycle_transition_map_fini() fails, it will clobber the error string here, twice
+  // Here, we concatenate the error strings if that happens
+  const char * current_error = (rcl_error_is_set()) ? rcl_get_error_string().str : "";
+  rcl_reset_error();
+
   if (rcl_lifecycle_transition_map_fini(&state_machine->transition_map, allocator) != RCL_RET_OK) {
-    RCL_SET_ERROR_MSG("could not free lifecycle transition map. Leaking memory!\n");
+    const char * fini_error = (rcl_error_is_set()) ? rcl_get_error_string().str : "";
```
Karsten1987, line 701: if I interpret this correctly this line will always result in an empty string, given that you reset the error message just above.

rcl_lifecycle/src/default_state_machine.c
```diff
@@ -690,10 +690,27 @@ rcl_lifecycle_init_default_state_machine(
...
+    rcl_reset_error();
+
+    RCL_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "Freeing transition map failed while handling a previous error. Leaking memory!"
+      "\nOriginal error:\n\t%s\nError encountered in rcl_lifecycle_transition_map_fini():\n\t%s\n",
+      current_error, fini_error);
+  } else if (strcmp(current_error, "") != 0) {
+    RCL_SET_ERROR_MSG(current_error);
```
Karsten1987, line 709: Sorry to iterate again over this, but I think that `else if` branch is not needed. If I see this correctly, we'll fetch the error message before just to reset it again in this branch. 
Why not just retrieving the existing error message within the first `if` branch? 

something like this:
```c++
if (rcl_lifecycle_transition_map_fini(&state_machine->transition_map, allocator) != RCL_RET_OK) {
  const char * fini_error = "";
  if (rcl_error_is_set()) {
     fini_error = rcl_get_error_string().str;
     rcl_reset_error();
  }
  "Freeing transition map failed while handling a previous error. Leaking memory!"
  "\nOriginal error:\n\t%s\nError encountered in rcl_lifecycle_transition_map_fini():\n\t%s\n",
  current_error, fini_error);
}

if (!rcl_error_is_set()) {
  RCL_SET_ERROR_MSG("Unspecified error ..");
}

```

brawner: The problem originally was that the error message was already set. But if `rcl_lifecycle_transition_map_fini` failed, it would set its own message, and then the message would get set a third time inside this first if loop. So the goal was to keep track of `current_message`, and append fini_error if it existed and throw a combined error that was hopefully more helpful than either of them individually.

However, if `rcl_lifecycle_transition_map_fini` succeeds, we still want the user to know what caused the code to reach `fail:` in the first place, which is in `current_message` , but only if rcl_error_is_set() was true in line 697.

Originally when I submitted this, I just relied on rcutils to log the transition of error messages, since it warns if a new error message is set while a previous one had already existed. However, since I'm pretty new to the ROS 2 C codebase, I'm really not confident about any particular approach here :man_shrugging:

rcl_lifecycle/src/default_state_machine.c
brawner: The compiler wouldn't let me declare `current_error` directly after `fail:` without it. The semicolon effectively creates a new statement. Should I move the variable declarations to the top?

brawner:
Address @Karsten1987's feedback and am running CI again.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10870/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6234/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8855/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10760/)

brawner:
Thank you for your reviews and help!

hidmic:
@brawner @Karsten1987 this patch is causing test failures in `rclcpp_lifecycle` tests. See https://ci.ros2.org/view/nightly/job/nightly_linux_release/1555/. It seems [these additional checks](https://github.com/ros2/rcl/pull/649/files#diff-ad8f79383e6e7ca8c0e4bc3b11502bc6R135-R143) in `rcl_lifecycle_transition_init()` are not consistent with what `rclcpp_lifecycle` [expects of this API](https://github.com/ros2/rclcpp/blob/87bb9f9758dce239d37256ae91d58ce6325f405b/rclcpp_lifecycle/src/transition.cpp#L156).

Karsten1987:
Looks like a transition can be _lazy_ initialized where start and goal state don't have to be present. Which means we should remove the nullptr checks for these two.

PR #652, opened 2020-05-15
warn about unused return value for set_logger_level

### LINKED ISSUE #651: Ignored return value from rcl_logging_external_set_logger_level
The return value of `rcl_logging_external_set_logger_level` is ignored at: https://github.com/ros2/rcl/blob/f92e52a2bb6956fb754256959d35fa5e809f7839/rcl/src/rcl/logging.c#L98

dirk-thomas:
Redo of dropped #650. Fixes #651.

### DIFF
--- a/rcl/include/rcl/macros.h
+++ b/rcl/include/rcl/macros.h
@@ -23,7 +23,7 @@ extern "C"
 /// Ignored return values of functions with this macro will emit a warning.
 #define RCL_WARN_UNUSED RCUTILS_WARN_UNUSED
 
-#define RCL_UNUSED(x) RCUTILS_UNUSED(x)
+#define RCL_UNUSED(x) (void)(x)
 
 #ifdef __cplusplus
 }
--- a/rcl/include/rcl/macros.h
+++ b/rcl/include/rcl/macros.h
@@ -20,14 +20,10 @@ extern "C"
 {
 #endif
 
-#ifndef _WIN32
 /// Ignored return values of functions with this macro will emit a warning.
-# define RCL_WARN_UNUSED __attribute__((warn_unused_result))
-#else
-# define RCL_WARN_UNUSED _Check_return_
-#endif
+#define RCL_WARN_UNUSED RCUTILS_WARN_UNUSED
 
-#define RCL_UNUSED(x) (void)(x)
+#define RCL_UNUSED(x) RCUTILS_UNUSED(x)
 
 #ifdef __cplusplus
 }
--- a/rcl/src/rcl/logging.c
+++ b/rcl/src/rcl/logging.c
@@ -95,7 +95,13 @@ rcl_logging_configure(const rcl_arguments_t * global_args, const rcl_allocator_t
   if (g_rcl_logging_ext_lib_enabled) {
     status = rcl_logging_external_initialize(config_file, g_logging_allocator);
     if (RCL_RET_OK == status) {
-      rcl_logging_external_set_logger_level(NULL, default_level);
+      // TODO(dirk-thomas) the return value should be typed and compared to
+      // constants instead of zero
+      int logging_status = rcl_logging_external_set_logger_level(
+        NULL, default_level);
+      if (logging_status != 0) {
+        status = RCL_RET_ERROR;
+      }
       g_rcl_logging_out_handlers[g_rcl_logging_num_out_handlers++] =
         rcl_logging_ext_lib_output_handler;
     }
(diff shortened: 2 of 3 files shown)


rcl/src/rcl/logging.c
```diff
@@ -95,7 +95,13 @@ rcl_logging_configure(const rcl_arguments_t * global_args, const rcl_allocator_t
   if (g_rcl_logging_ext_lib_enabled) {
     status = rcl_logging_external_initialize(config_file, g_logging_allocator);
     if (RCL_RET_OK == status) {
-      rcl_logging_external_set_logger_level(NULL, default_level);
+      // TODO(dirk-thomas) the return value should be typed and compared to
+      // constants instead of zero
```
dirk-thomas, line 99: While this mismatch is unfortunately it is better then not enforcing to check the return value.

dirk-thomas:
CI builds up to `rcl` with only FastRTPS, testing the three touched packages:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10788/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6174/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8781/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10677/)

One full CI build on Linux without any tests to check for compiler warnings:
* [Build Status](https://ci.ros2.org/job/ci_linux/10789/)

rcl/include/rcl/macros.h
```diff
@@ -20,14 +20,10 @@ extern "C"
...
-#ifndef _WIN32
 /// Ignored return values of functions with this macro will emit a warning.
-# define RCL_WARN_UNUSED __attribute__((warn_unused_result))
-#else
-# define RCL_WARN_UNUSED _Check_return_
-#endif
+#define RCL_WARN_UNUSED RCUTILS_WARN_UNUSED
 
-#define RCL_UNUSED(x) (void)(x)
+#define RCL_UNUSED(x) RCUTILS_UNUSED(x)
```
hidmic, line 26: @dirk-thomas this change seems unrelated. No need to split in a separate PR, a different commit + rebase-merge would do too.

rcl/src/rcl/logging.c
hidmic: @dirk-thomas mind to explain? `rcl_logging_external_set_logger_level()` returns an `rcl_ret_t` value.

rcl/include/rcl/macros.h
dirk-thomas: Reverted in a1c8a5b1d0c80946805b81e5016cf7977172bf63.

dirk-thomas:
> mind to explain? `rcl_logging_external_set_logger_level()` returns an `rcl_ret_t` value.

Afaik the different implementations have a different return type since they don't depend on `rcl` and therefore can't use the typed in the interface in this package. I would say very awkward (if not broken) design. This PR just doesn't touch that interface but only makes sure that users do check the return value.

clalancette:
> Afaik the different implementations have a different return type since they don't depend on `rcl` and therefore can't use the typed in the interface in this package. I would say very awkward (if not broken) design. This PR just doesn't touch that interface but only makes sure that users do check the return value.

Yeah, we've been discussing this a bit over at https://github.com/ros2/rcl_logging/pull/33#issuecomment-629416344 .  The more I think about it, I think the right solution here is to define an "rcl_logging_interface" package which just has a header file which defines the interface and the return values.  Then `rcl_logging_spdlog` and friends depend on that package, and implement the interface, while `rcl` also depends on that package, and calls the interface.

But its enough churn that I think we should do it post-Foxy.  While the current setup is fragile, the APIs haven't changed since Dashing, so its not like this is likely to change much.

rcl/include/rcl/macros.h
```diff
@@ -20,12 +20,8 @@ extern "C"
 {
 #endif
 
-#ifndef _WIN32
 /// Ignored return values of functions with this macro will emit a warning.
-# define RCL_WARN_UNUSED __attribute__((warn_unused_result))
-#else
-# define RCL_WARN_UNUSED _Check_return_
-#endif
+#define RCL_WARN_UNUSED RCUTILS_WARN_UNUSED
```
hidmic, line 24: @dirk-thomas same here about unrelated changes. Again, I'd be perfectly fine with this changes in a separate commit so they don't get collapsed into the same commit about unused return values.

dirk-thomas: I think this should stay together since this change makes sure that the signature is actually the same.

hidmic: BTW https://github.com/ros2/rcl/pull/653 does both.

PR #653, opened 2020-05-15
Consolidate macro duplication

rotu:
Signed-off-by: Dan Rose <dan@digilabs.io>

### DIFF
--- a/rcl/include/rcl/macros.h
+++ b/rcl/include/rcl/macros.h
@@ -15,22 +15,10 @@
 #ifndef RCL__MACROS_H_
 #define RCL__MACROS_H_
 
-#ifdef __cplusplus
-extern "C"
-{
-#endif
+#include "rcutils/macros.h"
 
-#ifndef _WIN32
-/// Ignored return values of functions with this macro will emit a warning.
-# define RCL_WARN_UNUSED __attribute__((warn_unused_result))
-#else
-# define RCL_WARN_UNUSED _Check_return_
-#endif
+#define RCL_WARN_UNUSED RCUTILS_WARN_UNUSED
 
-#define RCL_UNUSED(x) (void)(x)
-
-#ifdef __cplusplus
-}
-#endif
+#define RCL_UNUSED(x) RCUTILS_UNUSED(x)
 
 #endif  // RCL__MACROS_H_
--- a/rcl/include/rcl/macros.h
+++ b/rcl/include/rcl/macros.h
@@ -17,9 +17,18 @@
 
 #include "rcutils/macros.h"
 
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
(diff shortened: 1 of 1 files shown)


rotu:
I don't know why the duplication exists (maybe the RCL version should be deprecated instead)

dirk-thomas:
This needs to be rebased since ##652 already changes some of the touched code.

rotu:
Backmerged.

rcl/include/rcl/macros.h
```diff
@@ -15,18 +15,11 @@
 #ifndef RCL__MACROS_H_
 #define RCL__MACROS_H_
 
-#ifdef __cplusplus
-extern "C"
-{
-#endif
```
dirk-thomas, line 21: Why should this be removed?

rotu: Because there is nothing to specify the linkage of. And were there any, we’d want to use whatever linkage is specified by `rcutils`

dirk-thomas: Almost all C headers in the `rcl` packages do use `extern "C"` - no matter if they do contain things which require it or not. The rational is that since the package is providing a C-only API. Always stating it explicitly ensures that it is being used - even if additional code is added to a header later - since it might otherwise be overlooked.

Therefore a :-1: from me to remove it. I would rather propose to add it to the headers where it is currently not.

fujitatomoya: > since it might otherwise be overlooked.

i believe this is reasonable.

rotu: I think it's needless clutter and misleading to keep it in. If a macro in this `macros.h` file declared a function, that enclosing `extern "c"` would falsely suggest that the function uses C linkage.

rotu:
All feedback has been considered and addressed or dismissed. Please merged.

rcl/include/rcl/macros.h
jacobperron: I don't feel strongly about it, but leaving seems harmless to me.

Anyways, removing it seems orthogonal to this PR and could be proposed separately.

clalancette: I also don't feel super strongly about it, but on balance I would rather leave the `extern "C"` in.  It just keeps all of the headers consistent, and if we did ever introduce a C-linkage symbol in it, it would prevent it from being mangled.

dirk-thomas: @rotu With the majority of feedback leaning toward keeping the `extern "C"` blocks please revert this part of the patch so we can merge the deduplication.

dirk-thomas: I reverted the removal of `extern C` in b8f11aa0ef7672fed01d298d5915fbb619ccf341.

PR #664, opened 2020-05-28
Implement a generic way to change logging levels

iuhilnehc-ynos:
This PR is related to [rclcpp_1098](https://github.com/ros2/rclcpp/issues/1098)

### DIFF
--- a/rcl/include/rcl/arguments.h
+++ b/rcl/include/rcl/arguments.h
@@ -16,6 +16,7 @@
 #define RCL__ARGUMENTS_H_
 
 #include "rcl/allocator.h"
+#include "rcl/log_level.h"
 #include "rcl/macros.h"
 #include "rcl/types.h"
 #include "rcl/visibility_control.h"
@@ -342,6 +343,34 @@ rcl_remove_ros_arguments(
   rcl_allocator_t allocator,
   int * nonros_argc,
   const char ** nonros_argv[]);
+
+/// Return log level parsed from the command line.
+/**
+ * Log level are parsed directly from command line arguments.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] arguments An arguments structure that has been parsed.
+ * \param[out] log_level Log level settings as parsed from command line arguments.
+ *   This structure must be finalized by the caller.
+ *   The output is NULL if no log were parsed.
+ * \return `RCL_RET_OK` if everything goes correctly, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if any function arguments are invalid, or
+ * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_arguments_get_log_level(
+  const rcl_arguments_t * arguments,
+  rcl_log_level_t ** log_level);
 
 /// Copy one arguments structure into another.
 /**
--- a/rcl/include/rcl/arguments.h
+++ b/rcl/include/rcl/arguments.h
@@ -359,7 +359,7 @@ rcl_remove_ros_arguments(
  * \param[in] arguments An arguments structure that has been parsed.
  * \param[out] log_level Log level settings as parsed from command line arguments.
  *   This structure must be finalized by the caller.
- *   The output is NULL if no log were parsed.
+ *   The output is NULL if no logs were parsed.
  * \return `RCL_RET_OK` if everything goes correctly, or
  * \return `RCL_RET_INVALID_ARGUMENT` if any function arguments are invalid, or
  * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
--- a/rcl/include/rcl/arguments.h
+++ b/rcl/include/rcl/arguments.h
@@ -78,9 +78,9 @@ rcl_get_zero_initialized_arguments(void);
  * Parameter override rule parsing is supported via `-p/--param` flags e.g. `--param name:=value`
  * or `-p name:=value`.
  *
- * The default log level will be parsed as `--log-level level`, where `level` is a name
- * representing one of the log levels in the `RCUTILS_LOG_SEVERITY` enum, e.g. `info`, `debug`,
- * `warn`, not case sensitive.
+ * The default log level will be parsed as `--log-level level` and logger levels will be parsed as
+ * multiple `--log-level name:=level`, where `level` is a name representing one of the log levels
+ * in the `RCUTILS_LOG_SEVERITY` enum, e.g. `info`, `debug`, `warn`, not case sensitive.
  * If multiple of these rules are found, the last one parsed will be used.
  *
  * If an argument does not appear to be a valid ROS argument e.g. a `-r/--remap` flag followed by
--- a/rcl/include/rcl/arguments.h
+++ b/rcl/include/rcl/arguments.h
@@ -344,9 +344,9 @@ rcl_remove_ros_arguments(
   int * nonros_argc,
   const char ** nonros_argv[]);
 
-/// Return log level parsed from the command line.
+/// Return log levels parsed from the command line.
 /**
- * Log level are parsed directly from command line arguments.
+ * Log levels are parsed directly from command line arguments.
  *
  * <hr>
  * Attribute          | Adherence
@@ -357,20 +357,19 @@ rcl_remove_ros_arguments(
  * Lock-Free          | Yes
  *
  * \param[in] arguments An arguments structure that has been parsed.
- * \param[out] log_level Log level settings as parsed from command line arguments.
+ * \param[out] log_levels Log levels as parsed from command line arguments.
  *   This structure must be finalized by the caller.
  *   The output is NULL if no logs were parsed.
  * \return `RCL_RET_OK` if everything goes correctly, or
  * \return `RCL_RET_INVALID_ARGUMENT` if any function arguments are invalid, or
- * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
- * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ * \return `RCL_RET_BAD_ALLOC` if allocating memory failed.
  */
 RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_ret_t
-rcl_arguments_get_log_level(
+rcl_arguments_get_log_levels(
   const rcl_arguments_t * arguments,
-  rcl_log_level_t ** log_level);
+  rcl_log_levels_t ** log_levels);
 
 /// Copy one arguments structure into another.
 /**
--- a/rcl/include/rcl/arguments.h
+++ b/rcl/include/rcl/arguments.h
@@ -358,8 +358,7 @@ rcl_remove_ros_arguments(
  *
  * \param[in] arguments An arguments structure that has been parsed.
  * \param[out] log_levels Log levels as parsed from command line arguments.
- *   This structure must be finalized by the caller.
- *   The output is NULL if no logs were parsed.
+ *   The output must be finished by the caller if the function successes.
  * \return `RCL_RET_OK` if everything goes correctly, or
  * \return `RCL_RET_INVALID_ARGUMENT` if any function arguments are invalid, or
  * \return `RCL_RET_BAD_ALLOC` if allocating memory failed.
@@ -369,7 +368,7 @@ RCL_WARN_UNUSED
 rcl_ret_t
 rcl_arguments_get_log_levels(
   const rcl_arguments_t * arguments,
-  rcl_log_levels_t ** log_levels);
+  rcl_log_levels_t * log_levels);
 
 /// Copy one arguments structure into another.
 /**
--- a/rcl/include/rcl/log_level.h
+++ b/rcl/include/rcl/log_level.h
@@ -84,7 +84,9 @@ rcl_get_zero_initialized_log_levels();
  * \param[in] logger_count Number of logger settings to be allocated.
  *  This reserves memory for logger_settings, but doesn't initialize it.
  * \return `RCL_RET_OK` if the structure was initialized successfully, or
- * \return `RCL_RET_INVALID_ARGUMENT` if log_levels or allocator are invalid, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if log_levels is NULL, or
+ *  log_levels contains initialized memory, or
+ *  allocator is invalid, or
  * \return `RCL_RET_BAD_ALLOC` if allocating memory failed.
  */
 RCL_PUBLIC
@@ -105,9 +107,12 @@ rcl_log_levels_init(
  *
  * \param[in] src The structure to be copied.
  *  Its allocator is used to copy memory into the new structure.
- * \param[out] dst A zero-initialized log levels structure to be copied into.
+ * \param[out] dst A log levels structure to be copied into.
  * \return `RCL_RET_OK` if the structure was copied successfully, or
- * \return `RCL_RET_INVALID_ARGUMENT` if src or dst are invalid, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if src is NULL, or
+ *  if src allocator is invalid, or
+ *  if dst is NULL, or
+ *  if dst contains already allocated memory, or
  * \return `RCL_RET_BAD_ALLOC` if allocating memory failed.
  */
 RCL_PUBLIC
@@ -127,7 +132,8 @@ rcl_log_levels_copy(const rcl_log_levels_t * src, rcl_log_levels_t * dst);
  *
  * \param[in] log_levels The structure which its resources have to be deallocated.
  * \return `RCL_RET_OK` if the memory was successfully freed, or
- * \return `RCL_RET_INVALID_ARGUMENT` if log_levels is invalid.
+ * \return `RCL_RET_INVALID_ARGUMENT` if log_levels is NULL, or
+ *  if ist allocator is invalid and the structure contains initialized memory.
  */
 RCL_PUBLIC
 rcl_ret_t
@@ -145,7 +151,7 @@ rcl_log_levels_fini(rcl_log_levels_t * log_levels);
  *
  * \param[in] log_levels The structure to be shrunk.
  * \return `RCL_RET_OK` if the memory was successfully shrunk, or
- * \return `RCL_RET_INVALID_ARGUMENT` if log_levels is invalid, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if log_levels is NULL or if its allocator is invalid, or
  * \return `RCL_RET_BAD_ALLOC` if reallocating memory failed.
  */
 RCL_PUBLIC
@@ -162,12 +168,16 @@ rcl_log_levels_shrink_to_size(rcl_log_levels_t * log_levels);
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param[in] log_levels The structure to be added.
- * \param[in] logger_name Name for the logger.
- * \param[in] log_level Minimum log level severity.
+ * \param[in] log_levels The structure where to set the logger log level.
+ * \param[in] logger_name Name for the logger, a copy of it will be stored in the structure.
+ * \param[in] log_level Minimum log level severity to be set for logger_name.
  * \return `RCL_RET_OK` if add logger setting successfully, or
  * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
- * \return `RCL_RET_INVALID_ARGUMENT` if log_levels or logger_name are invalid.
+ * \return `RCL_RET_INVALID_ARGUMENT` if log_levels is NULL, or
+ *   if log_levels was not initialized, or
+ *   if log_levels allocator is invalid, or
+ *   if logger_name is NULL, or
+ * \return `RCL_RET_ERROR` if the log_levels structure is already full.
  */
 RCL_PUBLIC
 rcl_ret_t
--- a/rcl/include/rcl/log_level.h
+++ b/rcl/include/rcl/log_level.h
@@ -25,22 +25,25 @@ extern "C"
 {
 #endif
 
-/// A logger item to specify a name and log_level
+/// A logger item to specify a name and a log level
 typedef struct rcl_logger_setting_t
 {
   /// name for the logger.
   char * name;
   /// level for the logger.
-  int level;
+  rcl_log_severity_t level;
 } rcl_logger_setting_t;
 
-/// Hold log_level.
+/// Hold default logger level and other logger setting.
 typedef struct rcl_log_level_t
 {
-  /// Default log level (represented by `RCUTILS_LOG_SEVERITY` enum) or -1 if not specified.
-  int default_log_level;
-  struct rcl_logger_setting_t * logger_settings;   ///<  Array of logger
-  size_t num_loggers;       ///< Number of loggers
+  /// Default logger level
+  rcl_log_severity_t default_logger_level;
+  /// Array of logger setting
+  struct rcl_logger_setting_t * logger_settings;
+  /// Number of logger settings
+  size_t num_logger_settings;
+  /// Allocator used to allocate objects in this struct
   rcl_allocator_t allocator;
 } rcl_log_level_t;
 
@@ -51,7 +54,7 @@ typedef struct rcl_log_level_t
 RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_log_level_t *
-rcl_log_level_init(const rcutils_allocator_t allocator);
+rcl_log_level_init(const rcl_allocator_t allocator);
 
 /// \brief Copy log level structure, allocate memory by using allocator of rcl_log_level_t
 /// \param[in] log_level points to the log level struct to be copied
@@ -62,7 +65,7 @@ rcl_log_level_t * rcl_log_level_copy(const rcl_log_level_t * log_level);
 
 
 /// \brief Free log level structure
-/// \param[in] log_level points to the populated log level structure
+/// \param[in] log_level points to the log level structure to be freed
 RCL_PUBLIC
 void
 rcl_log_level_fini(rcl_log_level_t * log_level);
--- a/rcl/include/rcl/log_level.h
+++ b/rcl/include/rcl/log_level.h
@@ -53,6 +53,17 @@ typedef struct rcl_log_levels_t
 } rcl_log_levels_t;
 
 /// Return a rcl_log_levels_t struct with members initialized to zero value.
+/**
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | Yes
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \return a rcl_log_levels_t struct with members initialized to zero value.
+ */
 RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_log_levels_t
@@ -64,7 +75,7 @@ rcl_get_zero_initialized_log_levels();
  * Attribute          | Adherence
  * ------------------ | -------------
  * Allocates Memory   | Yes
- * Thread-Safe        | Yes
+ * Thread-Safe        | No
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
@@ -114,7 +125,7 @@ rcl_log_levels_copy(const rcl_log_levels_t * src, rcl_log_levels_t * dst);
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param[in] log_levels The members of structure to be deallocated.
+ * \param[in] log_levels The structure which its resources have to be deallocated.
  * \return `RCL_RET_OK` if the memory was successfully freed, or
  * \return `RCL_RET_INVALID_ARGUMENT` if log_levels is invalid.
  */
@@ -127,7 +138,7 @@ rcl_log_levels_fini(rcl_log_levels_t * log_levels);
  * <hr>
  * Attribute          | Adherence
  * ------------------ | -------------
- * Allocates Memory   | No
+ * Allocates Memory   | Yes
  * Thread-Safe        | No
  * Uses Atomics       | No
  * Lock-Free          | Yes
--- a/rcl/include/rcl/log_level.h
+++ b/rcl/include/rcl/log_level.h
@@ -53,7 +53,7 @@ RCL_WARN_UNUSED
 rcl_log_level_t *
 rcl_log_level_init(const rcutils_allocator_t allocator);
 
-/// \brief Copy log level structure
+/// \brief Copy log level structure, allocate memory by using allocator of rcl_log_level_t
 /// \param[in] log_level points to the log level struct to be copied
 /// \return a pointer to the copied log level structure on success or NULL on failure
 RCL_PUBLIC
--- a/rcl/include/rcl/log_level.h
+++ b/rcl/include/rcl/log_level.h
@@ -25,25 +25,30 @@ extern "C"
 {
 #endif
 
-/// A logger item to specify a name and a log level
+/// typedef for RCUTILS_LOG_SEVERITY;
+typedef enum RCUTILS_LOG_SEVERITY rcl_log_severity_t;
+
+/// A logger item to specify a name and a log level.
 typedef struct rcl_logger_setting_t
 {
-  /// name for the logger.
-  char * name;
-  /// level for the logger.
+  /// Name for the logger.
+  const char * name;
+  /// Minimum log level severity of the logger.
   rcl_log_severity_t level;
 } rcl_logger_setting_t;
 
 /// Hold default logger level and other logger setting.
 typedef struct rcl_log_levels_t
 {
-  /// Default logger level
+  /// Minimum default logger level severity.
   rcl_log_severity_t default_logger_level;
-  /// Array of logger setting
+  /// Array of logger setting.
   struct rcl_logger_setting_t * logger_settings;
-  /// Number of logger settings
+  /// Number of logger settings.
   size_t num_logger_settings;
-  /// Allocator used to allocate objects in this struct
+  /// Capacity of logger settings.
+  size_t capacity_logger_settings;
+  /// Allocator used to allocate objects in this struct.
   rcl_allocator_t allocator;
 } rcl_log_levels_t;
 
@@ -53,18 +58,18 @@ typedef struct rcl_log_levels_t
  * Attribute          | Adherence
  * ------------------ | -------------
  * Allocates Memory   | Yes
- * Thread-Safe        | No
+ * Thread-Safe        | Yes
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param[in] allocator memory allocator to be used.
- * \param[in] logger_count to allocate the logger setting count of log levels.
- * \return a pointer to log level structure on success or NULL on failure.
+ * \param[in] allocator Memory allocator to be used.
+ * \param[in] logger_count Number of logger settings to be allocated.
+ * \return a pointer to the allocated log levels structure on success or NULL on failure.
  */
 RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_log_levels_t *
-rcl_log_levels_init(const rcl_allocator_t allocator, size_t logger_count);
+rcl_log_levels_init(const rcl_allocator_t * allocator, size_t logger_count);
 
 /// Copy one log levels structure into another.
 /**
@@ -76,9 +81,9 @@ rcl_log_levels_init(const rcl_allocator_t allocator, size_t logger_count);
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param[in] log_levels The structure to be copied.
+ * \param[in] src The structure to be copied.
  *  Its allocator is used to copy memory into the new structure.
- * \param[out] log_levels_out A zero-initialized log levels structure to be copied into.
+ * \param[out] dst A zero-initialized log levels structure to be copied into.
  * \return `RCL_RET_OK` if the structure was copied successfully, or
  * \return `RCL_RET_INVALID_ARGUMENT` if log_levels or log_levels_out are invalid, or
  * \return `RCL_RET_BAD_ALLOC` if allocating memory failed.
@@ -86,7 +91,7 @@ rcl_log_levels_init(const rcl_allocator_t allocator, size_t logger_count);
 RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_ret_t
-rcl_log_levels_copy(const rcl_log_levels_t * log_levels, rcl_log_levels_t * log_levels_out);
+rcl_log_levels_copy(const rcl_log_levels_t * src, rcl_log_levels_t * dst);
 
 /// Free log levels structure.
 /**
@@ -105,6 +110,46 @@ rcl_log_levels_copy(const rcl_log_levels_t * log_levels, rcl_log_levels_t * log_
 RCL_PUBLIC
 rcl_ret_t
 rcl_log_levels_fini(rcl_log_levels_t * log_levels);
+
+/// Shrink log levels structure.
+/**
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] log_levels The structure to be shrunk.
+ * \return `RCL_RET_OK` if the memory was successfully shrunk, or
+ * \return `RCL_RET_BAD_ALLOC` if reallocating memory failed, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if log_levels is invalid.
+ */
+RCL_PUBLIC
+rcl_ret_t
+rcl_log_levels_shrink_to_size(rcl_log_levels_t * log_levels);
+
+/// Add logger setting with a name and a level.
+/**
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] log_levels The structure to be added.
+ * \param[in] logger_name Name for the logger.
+ * \param[in] log_level Minimum log level severity.
+ * \return `RCL_RET_OK` if add logger setting successfully, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if log_levels is invalid.
+ */
+RCL_PUBLIC
+rcl_ret_t
+rcl_log_levels_add_logger_setting(
+  rcl_log_levels_t * log_levels, const char * logger_name, rcl_log_severity_t log_level);
 
 #ifdef __cplusplus
 }
--- a/rcl/include/rcl/log_level.h
+++ b/rcl/include/rcl/log_level.h
@@ -52,6 +52,12 @@ typedef struct rcl_log_levels_t
   rcl_allocator_t allocator;
 } rcl_log_levels_t;
 
+/// Return a rcl_log_levels_t struct with members initialized to zero value.
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_log_levels_t
+rcl_get_zero_initialized_log_levels();
+
 /// Initialize a log levels structure.
 /**
  * <hr>
@@ -62,14 +68,19 @@ typedef struct rcl_log_levels_t
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param[in] allocator Memory allocator to be used.
+ * \param[in] log_levels The structure to be initialized.
+ * \param[in] allocator Memory allocator to be used and assigned into log_levels.
  * \param[in] logger_count Number of logger settings to be allocated.
- * \return a pointer to the allocated log levels structure on success or NULL on failure.
+ *  This reserves memory for logger_settings, but doesn't initialize it.
+ * \return `RCL_RET_OK` if the structure was initialized successfully, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if log_levels or allocator are invalid, or
+ * \return `RCL_RET_BAD_ALLOC` if allocating memory failed.
  */
 RCL_PUBLIC
 RCL_WARN_UNUSED
-rcl_log_levels_t *
-rcl_log_levels_init(const rcl_allocator_t * allocator, size_t logger_count);
+rcl_ret_t
+rcl_log_levels_init(
+  rcl_log_levels_t * log_levels, const rcl_allocator_t * allocator, size_t logger_count);
 
 /// Copy one log levels structure into another.
 /**
@@ -85,7 +96,7 @@ rcl_log_levels_init(const rcl_allocator_t * allocator, size_t logger_count);
  *  Its allocator is used to copy memory into the new structure.
  * \param[out] dst A zero-initialized log levels structure to be copied into.
  * \return `RCL_RET_OK` if the structure was copied successfully, or
- * \return `RCL_RET_INVALID_ARGUMENT` if log_levels or log_levels_out are invalid, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if src or dst are invalid, or
  * \return `RCL_RET_BAD_ALLOC` if allocating memory failed.
  */
 RCL_PUBLIC
@@ -93,7 +104,7 @@ RCL_WARN_UNUSED
 rcl_ret_t
 rcl_log_levels_copy(const rcl_log_levels_t * src, rcl_log_levels_t * dst);
 
-/// Free log levels structure.
+/// Reclaim resources held inside rcl_log_levels_t structure.
 /**
  * <hr>
  * Attribute          | Adherence
@@ -103,7 +114,7 @@ rcl_log_levels_copy(const rcl_log_levels_t * src, rcl_log_levels_t * dst);
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param[in] log_levels The structure to be deallocated.
+ * \param[in] log_levels The members of structure to be deallocated.
  * \return `RCL_RET_OK` if the memory was successfully freed, or
  * \return `RCL_RET_INVALID_ARGUMENT` if log_levels is invalid.
  */
@@ -123,8 +134,8 @@ rcl_log_levels_fini(rcl_log_levels_t * log_levels);
  *
  * \param[in] log_levels The structure to be shrunk.
  * \return `RCL_RET_OK` if the memory was successfully shrunk, or
- * \return `RCL_RET_BAD_ALLOC` if reallocating memory failed, or
- * \return `RCL_RET_INVALID_ARGUMENT` if log_levels is invalid.
+ * \return `RCL_RET_INVALID_ARGUMENT` if log_levels is invalid, or
+ * \return `RCL_RET_BAD_ALLOC` if reallocating memory failed.
  */
 RCL_PUBLIC
 rcl_ret_t
@@ -135,7 +146,7 @@ rcl_log_levels_shrink_to_size(rcl_log_levels_t * log_levels);
  * <hr>
  * Attribute          | Adherence
  * ------------------ | -------------
- * Allocates Memory   | No
+ * Allocates Memory   | Yes
  * Thread-Safe        | No
  * Uses Atomics       | No
  * Lock-Free          | Yes
@@ -144,7 +155,8 @@ rcl_log_levels_shrink_to_size(rcl_log_levels_t * log_levels);
  * \param[in] logger_name Name for the logger.
  * \param[in] log_level Minimum log level severity.
  * \return `RCL_RET_OK` if add logger setting successfully, or
- * \return `RCL_RET_INVALID_ARGUMENT` if log_levels is invalid.
+ * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if log_levels or logger_name are invalid.
  */
 RCL_PUBLIC
 rcl_ret_t
--- a/rcl/include/rcl/log_level.h
+++ b/rcl/include/rcl/log_level.h
@@ -1,0 +1,74 @@
+// Copyright 2020 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCL__LOG_LEVEL_H_
+#define RCL__LOG_LEVEL_H_
+
+#include "rcl/allocator.h"
+#include "rcl/macros.h"
+#include "rcl/types.h"
+#include "rcl/visibility_control.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+/// A logger item to specify a name and log_level
+typedef struct rcl_logger_setting_t
+{
+  /// name for the logger.
+  char * name;
+  /// level for the logger.
+  int level;
+} rcl_logger_setting_t;
+
+/// Hold log_level.
+typedef struct rcl_log_level_t
+{
+  /// Default log level (represented by `RCUTILS_LOG_SEVERITY` enum) or -1 if not specified.
+  int default_log_level;
+  struct rcl_logger_setting_t * logger_settings;   ///<  Array of logger
+  size_t num_loggers;       ///< Number of loggers
+  rcl_allocator_t allocator;
+} rcl_log_level_t;
+
+
+/// \brief Initialize log level structure
+/// \param[in] allocator memory allocator to be used
+/// \return a pointer to log level structure on success or NULL on failure
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_log_level_t *
+rcl_log_level_init(const rcutils_allocator_t allocator);
+
+/// \brief Copy log level structure
+/// \param[in] log_level points to the log level struct to be copied
+/// \return a pointer to the copied log level structure on success or NULL on failure
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_log_level_t * rcl_log_level_copy(const rcl_log_level_t * log_level);
+
+
+/// \brief Free log level structure
+/// \param[in] log_level points to the populated log level structure
+RCL_PUBLIC
+void
+rcl_log_level_fini(rcl_log_level_t * log_level);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif  // RCL__LOG_LEVEL_H_
--- a/rcl/include/rcl/log_level.h
+++ b/rcl/include/rcl/log_level.h
@@ -35,7 +35,7 @@ typedef struct rcl_logger_setting_t
 } rcl_logger_setting_t;
 
 /// Hold default logger level and other logger setting.
-typedef struct rcl_log_level_t
+typedef struct rcl_log_levels_t
 {
   /// Default logger level
   rcl_log_severity_t default_logger_level;
@@ -45,30 +45,66 @@ typedef struct rcl_log_level_t
   size_t num_logger_settings;
   /// Allocator used to allocate objects in this struct
   rcl_allocator_t allocator;
-} rcl_log_level_t;
+} rcl_log_levels_t;
 
-
-/// \brief Initialize log level structure
-/// \param[in] allocator memory allocator to be used
-/// \return a pointer to log level structure on success or NULL on failure
+/// Initialize a log levels structure.
+/**
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] allocator memory allocator to be used.
+ * \param[in] logger_count to allocate the logger setting count of log levels.
+ * \return a pointer to log level structure on success or NULL on failure.
+ */
 RCL_PUBLIC
 RCL_WARN_UNUSED
-rcl_log_level_t *
-rcl_log_level_init(const rcl_allocator_t allocator);
+rcl_log_levels_t *
+rcl_log_levels_init(const rcl_allocator_t allocator, size_t logger_count);
 
-/// \brief Copy log level structure, allocate memory by using allocator of rcl_log_level_t
-/// \param[in] log_level points to the log level struct to be copied
-/// \return a pointer to the copied log level structure on success or NULL on failure
+/// Copy one log levels structure into another.
+/**
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] log_levels The structure to be copied.
+ *  Its allocator is used to copy memory into the new structure.
+ * \param[out] log_levels_out A zero-initialized log levels structure to be copied into.
+ * \return `RCL_RET_OK` if the structure was copied successfully, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if log_levels or log_levels_out are invalid, or
+ * \return `RCL_RET_BAD_ALLOC` if allocating memory failed.
+ */
 RCL_PUBLIC
 RCL_WARN_UNUSED
-rcl_log_level_t * rcl_log_level_copy(const rcl_log_level_t * log_level);
+rcl_ret_t
+rcl_log_levels_copy(const rcl_log_levels_t * log_levels, rcl_log_levels_t * log_levels_out);
 
-
-/// \brief Free log level structure
-/// \param[in] log_level points to the log level structure to be freed
+/// Free log levels structure.
+/**
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] log_levels The structure to be deallocated.
+ * \return `RCL_RET_OK` if the memory was successfully freed, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if log_levels is invalid.
+ */
 RCL_PUBLIC
-void
-rcl_log_level_fini(rcl_log_level_t * log_level);
+rcl_ret_t
+rcl_log_levels_fini(rcl_log_levels_t * log_levels);
 
 #ifdef __cplusplus
 }
--- a/rcl/include/rcl/types.h
+++ b/rcl/include/rcl/types.h
@@ -114,4 +114,7 @@ typedef rmw_ret_t rcl_ret_t;
 /// typedef for rmw_serialized_message_t;
 typedef rmw_serialized_message_t rcl_serialized_message_t;
 
+/// typedef for RCUTILS_LOG_SEVERITY;
+typedef enum RCUTILS_LOG_SEVERITY rcl_log_severity_t;
+
 #endif  // RCL__TYPES_H_
--- a/rcl/include/rcl/types.h
+++ b/rcl/include/rcl/types.h
@@ -114,7 +114,4 @@ typedef rmw_ret_t rcl_ret_t;
 /// typedef for rmw_serialized_message_t;
 typedef rmw_serialized_message_t rcl_serialized_message_t;
 
-/// typedef for RCUTILS_LOG_SEVERITY;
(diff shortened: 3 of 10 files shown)


clalancette:
Thanks for taking this on.  Just note that we are in the middle of releasing Foxy, so I probably won't have time to look at this until after that is released (June 5th).  If you don't hear from me after that date, feel free to give me a ping and I can take a look then.

iuhilnehc-ynos:
@clalancette
Sorry to disturb you. If you have time, please help review this PR.

rcl/src/rcl/arguments.c
```diff
@@ -139,6 +139,32 @@ rcl_arguments_get_param_overrides(
...
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(arguments, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(arguments->impl, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(log_level, RCL_RET_INVALID_ARGUMENT);
+
+  if (NULL != *log_level) {
+    RCL_SET_ERROR_MSG("Output log level pointer is not null. May leak memory.");
+    return RCL_RET_INVALID_ARGUMENT;
+  }
+  *log_level = NULL;
```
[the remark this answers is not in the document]
iuhilnehc-ynos: @fujitatomoya 
Yes, it must be NULL here. I'll update it.

rcl/include/rcl/log_level.h
```diff
@@ -0,0 +1,74 @@
...
+
+/// \brief Initialize log level structure
+/// \param[in] allocator memory allocator to be used
+/// \return a pointer to log level structure on success or NULL on failure
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_log_level_t *
+rcl_log_level_init(const rcutils_allocator_t allocator);
+
+/// \brief Copy log level structure
```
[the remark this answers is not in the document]
iuhilnehc-ynos: seems reasonable

rcl/src/rcl/arguments.c
```diff
@@ -139,6 +139,32 @@ rcl_arguments_get_param_overrides(
...
+  if (NULL != arguments->impl->log_level) {
+    *log_level = rcl_log_level_copy(arguments->impl->log_level);
+    if (NULL == *log_level) {
+      return RCL_RET_BAD_ALLOC;
+    }
+  }
+  return RCL_RET_OK;
+}
+
+#define RCL_LOG_LEVEL_FLAG_COMMA ","
```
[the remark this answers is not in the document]
iuhilnehc-ynos: > define right in the middle of the program file is not good for code style.

I just followed the definition at https://github.com/ros2/rcl/blob/85ef95514d6ee47bb8285009add84e956595c018/rcl/src/rcl/arguments.c#L209-L210

Maybe I need to move both of them at the beginning of this file.

rcl/src/rcl/arguments.c
```diff
@@ -1596,17 +1636,58 @@ rcl_ret_t
...
-  }
-  RCL_SET_ERROR_MSG("Argument does not use a valid severity level");
-  return RCL_RET_ERROR;
+  size_t colon_len = strlen(RCL_LOG_LEVEL_FLAG_COMMA);
+  size_t seperator_len = strlen(RCL_LOG_LEVEL_FLAG_SEPERATOR);
+  rcutils_ret_t ret = RCUTILS_RET_OK;
+  const char * p = arg;
+
+  do {
+    char * colon = strstr(p, RCL_LOG_LEVEL_FLAG_COMMA);
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Thanks, I'll update it

rcl/src/rcl/arguments.c
```diff
@@ -1596,17 +1636,58 @@ rcl_ret_t
...
   RCL_CHECK_ARGUMENT_FOR_NULL(arg, RCL_RET_INVALID_ARGUMENT);
   RCL_CHECK_ARGUMENT_FOR_NULL(log_level, RCL_RET_INVALID_ARGUMENT);
 
-  rcutils_ret_t ret = rcutils_logging_severity_level_from_string(arg, allocator, log_level);
-  if (RCUTILS_RET_OK == ret) {
-    return RCL_RET_OK;
-  }
-  RCL_SET_ERROR_MSG("Argument does not use a valid severity level");
-  return RCL_RET_ERROR;
+  size_t colon_len = strlen(RCL_LOG_LEVEL_FLAG_COMMA);
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Thanks.

rcl/src/rcl/logging.c
```diff
@@ -63,17 +63,33 @@ rcl_logging_configure_with_output_handler(
...
     g_logging_allocator = *allocator;
-  int default_level = global_args->impl->log_level;
+  int default_level = -1;
+
+  rcl_log_level_t * log_level = global_args->impl->log_level;
   const char * config_file = global_args->impl->external_log_config_file;
   g_rcl_logging_stdout_enabled = !global_args->impl->log_stdout_disabled;
   g_rcl_logging_rosout_enabled = !global_args->impl->log_rosout_disabled;
   g_rcl_logging_ext_lib_enabled = !global_args->impl->log_ext_lib_disabled;
+
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Updated

rcl/src/rcl/arguments.c
```diff
@@ -139,6 +144,28 @@ rcl_arguments_get_param_overrides(
...
+rcl_arguments_get_log_level(
+  const rcl_arguments_t * arguments,
+  rcl_log_level_t ** log_level)
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(arguments, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(arguments->impl, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(log_level, RCL_RET_INVALID_ARGUMENT);
+
+  if (NULL != *log_level) {
+    RCL_SET_ERROR_MSG("Output log level pointer is not null. May leak memory.");
```
clalancette, line 157: Technically, the second part of this sentence isn't true.  Since we return an error, then we won't actually lose a pointer.

```suggestion
    RCL_SET_ERROR_MSG("Output log level pointer is not null.");
```

rcl/src/rcl/arguments.c
```diff
@@ -139,6 +144,28 @@ rcl_arguments_get_param_overrides(
...
+    return RCL_RET_INVALID_ARGUMENT;
+  }
+  if (NULL != arguments->impl->log_level) {
+    *log_level = rcl_log_level_copy(arguments->impl->log_level);
+    if (NULL == *log_level) {
+      return RCL_RET_BAD_ALLOC;
+    }
+  }
```
clalancette, line 165: So if `arguments->impl->log_level` is NULL, then we'll leave `*log_level` as NULL.  Is that expected?  What happens when we go to free?

rcl/src/rcl/arguments.c
```diff
@@ -1596,17 +1634,58 @@ rcl_ret_t
...
+  do {
+    char * comma = strstr(p, RCL_LOG_LEVEL_FLAG_COMMA);
+    char * item;
+    if (comma == NULL) {
+      item = rcutils_strdup(p, allocator);
+    } else {
+      item = rcutils_strndup(p, comma - p, allocator);
+    }
+
+    char * seperator = strstr(item, RCL_LOG_LEVEL_FLAG_SEPERATOR);
```
clalancette, line 1656:
```suggestion
    char * separator = strstr(item, RCL_LOG_LEVEL_FLAG_SEPARATOR);
```

rcl/src/rcl/arguments.c
```diff
@@ -40,6 +40,11 @@ extern "C"
 {
 #endif
 
+#define RCL_ENABLE_FLAG_PREFIX "--enable-"
+#define RCL_DISABLE_FLAG_PREFIX "--disable-"
+#define RCL_LOG_LEVEL_FLAG_COMMA ","
+#define RCL_LOG_LEVEL_FLAG_SEPERATOR "="
```
clalancette, line 46:
```suggestion
#define RCL_LOG_LEVEL_FLAG_SEPARATOR "="
```

rcl/src/rcl/arguments.c
```diff
@@ -1596,17 +1634,58 @@ rcl_ret_t
...
   RCL_CHECK_ARGUMENT_FOR_NULL(log_level, RCL_RET_INVALID_ARGUMENT);
 
-  rcutils_ret_t ret = rcutils_logging_severity_level_from_string(arg, allocator, log_level);
-  if (RCUTILS_RET_OK == ret) {
-    return RCL_RET_OK;
-  }
-  RCL_SET_ERROR_MSG("Argument does not use a valid severity level");
-  return RCL_RET_ERROR;
+  size_t comma_len = strlen(RCL_LOG_LEVEL_FLAG_COMMA);
+  size_t seperator_len = strlen(RCL_LOG_LEVEL_FLAG_SEPERATOR);
```
clalancette, line 1643:
```suggestion
  size_t separator_len = strlen(RCL_LOG_LEVEL_FLAG_SEPARATOR);
```

rcl/src/rcl/arguments.c
```diff
@@ -1596,17 +1634,58 @@ rcl_ret_t
...
+  const char * p = arg;
+
+  do {
+    char * comma = strstr(p, RCL_LOG_LEVEL_FLAG_COMMA);
+    char * item;
+    if (comma == NULL) {
+      item = rcutils_strdup(p, allocator);
+    } else {
+      item = rcutils_strndup(p, comma - p, allocator);
+    }
```
clalancette, line 1654: There's some error cases here that may not be covered.  I'm wondering about things like the following:

* `--log-level rcl=debug,`
* `--log-level rcl=debug,,`
* `--log-level rcl=`
* `--log-level rcl=,`

(there may be more, these are just some of the ones that came to mind)

Can you write tests for these cases to make sure we properly error and don't leak memory?

rcl/src/rcl/log_level.c
```diff
@@ -0,0 +1,105 @@
...
+#include "rcl/error_handling.h"
+#include "rcutils/allocator.h"
+#include "rcutils/strdup.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#define MAX_NUM_LOGGER_SETTING_ENTRIES 256U
```
clalancette, line 26: We've been (slowly) trying to remove these arbitrary limits from the code.  Can you make it so that this isn't an arbitrary limit, but is just based on how many the user passed in?

iuhilnehc-ynos:
@clalancette 
I think I need to rebase my branch.

iuhilnehc-ynos:
@clalancette 

> @clalancette
> I think I need to rebase my branch.

After checking, the issue is related to **rosdep key**, not rebase.
```
RuntimeError: Could not resolve the rosdep key 'rcl_logging_interface'
```

Could you help check it?

clalancette:
Yeah, I think we just got unlucky when this CI was run.  It should succeed now, I'll give it a kick to see for sure.

clalancette:

iuhilnehc-ynos:

iuhilnehc-ynos:
@clalancette 

I tested on my local environment. 
```
100% tests passed, 0 tests failed out of 89

Label Time Summary:
copyright     =   0.33 sec*proc (1 test)
cppcheck      =   2.09 sec*proc (1 test)
cpplint       =   4.62 sec*proc (1 test)
gtest         =  90.24 sec*proc (77 tests)
lint_cmake    =   0.26 sec*proc (1 test)
linter        =   8.94 sec*proc (6 tests)
uncrustify    =   0.76 sec*proc (1 test)
xfail         =   1.09 sec*proc (1 test)
xmllint       =   0.88 sec*proc (1 test)

Total Test time (real) = 102.24 sec
Finished <<< rcl [1min 42s]

Summary: 1 package finished [1min 43s]
```

rcl/test/rcl/test_log_level.cpp
```diff
@@ -0,0 +1,214 @@
...
+#define EXPECT_INVALID_RET_FOR_ARGUMENTS_LOG_LEVEL(...) \
+  { \
+    rcl_ret_t ret; \
+    rcl_arguments_t local_arguments = rcl_get_zero_initialized_arguments(); \
+    const char * local_argv[] = {__VA_ARGS__}; \
+    unsigned int local_argc = (sizeof(local_argv) / sizeof(const char *)); \
+    ret = rcl_parse_arguments( \
+      local_argc, local_argv, rcl_get_default_allocator(), &local_arguments); \
+    ASSERT_EQ(RCL_RET_INVALID_ROS_ARGS, ret); \
+  }
```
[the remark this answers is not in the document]
iuhilnehc-ynos: @clalancette 
Not using a function so gtest failure output shows the line number where the macro is used.
And It seems that function taking ... is not the better way to use here.

rcl/include/rcl/arguments.h
```diff
@@ -343,6 +344,34 @@ rcl_remove_ros_arguments(
...
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] arguments An arguments structure that has been parsed.
+ * \param[out] log_level Log level settings as parsed from command line arguments.
+ *   This structure must be finalized by the caller.
+ *   The output is NULL if no log were parsed.
```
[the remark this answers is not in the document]
iuhilnehc-ynos: I'll update it.

rcl/test/rcl/test_log_level.cpp
```diff
@@ -0,0 +1,214 @@
...
+  rcl_log_level_t * log_level = NULL;
+  GET_LOG_LEVEL_FROM_ARGUMENTS(
+    log_level, "process_name", "--ros-args",
+    "--log-level", "debug");
+  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+  {
+    rcl_log_level_fini(log_level);
+  });
+  EXPECT_EQ((int)RCUTILS_LOG_SEVERITY_DEBUG, log_level->default_log_level);
+  EXPECT_EQ((size_t)0, log_level->num_loggers);
```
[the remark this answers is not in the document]
iuhilnehc-ynos: I'll update it.

rcl/test/rcl/test_log_level.cpp
```diff
@@ -0,0 +1,214 @@
...
+{
+public:
+  void SetUp()
+  {
+  }
+
+  void TearDown()
+  {
+  }
+};
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Remove it

rcl/test/rcl/test_log_level.cpp
```diff
@@ -0,0 +1,214 @@
...
+  EXPECT_INVALID_RET_FOR_ARGUMENTS_LOG_LEVEL(
+    "process_name", "--ros-args", "--log-level",
+    "rcl=");
+
+  EXPECT_INVALID_RET_FOR_ARGUMENTS_LOG_LEVEL(
+    "process_name", "--ros-args", "--log-level",
+    "rcl=,");
+}
+
+#define GET_LOG_LEVEL_FROM_ARGUMENTS(log_level, ...) \
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Same above

rcl/src/rcl/log_level.c
```diff
@@ -0,0 +1,105 @@
...
+fail:
+  rcl_log_level_fini(out_log_level);
+  return NULL;
+}
+
+void
+rcl_log_level_fini(
+  rcl_log_level_t * log_level)
+{
+  rcutils_allocator_t allocator = log_level->allocator;
```
fujitatomoya, line 89: need to check if log_level is not NULL just like rcl_log_level_init and rcl_log_level_copy.

rcl/test/rcl/test_log_level.cpp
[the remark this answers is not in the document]
fujitatomoya: i think that it should print the ASSER_EQ line instead of where macro is used. the place macro is used is not the place the problem actually happens.

rcl/include/rcl/log_level.h
```diff
@@ -0,0 +1,74 @@
...
+#include "rcl/macros.h"
+#include "rcl/types.h"
+#include "rcl/visibility_control.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+/// A logger item to specify a name and log_level
```
ivanpauno, line 28:
```suggestion
/// A logger item to specify a name and a log level
```

rcl/include/rcl/log_level.h
```diff
@@ -0,0 +1,74 @@
...
+typedef struct rcl_logger_setting_t
+{
+  /// name for the logger.
+  char * name;
+  /// level for the logger.
+  int level;
+} rcl_logger_setting_t;
+
+/// Hold log_level.
+typedef struct rcl_log_level_t
```
ivanpauno, line 38: Naming is confusing IMO.
It's not  clear that `rcl_log_level_t` is an array of `rcl_logger_setting_t` (the later has a `level` member, which makes things more confusing).

Maybe:

`rcl_logger_setting_t` -> `rcl_logger_setting_t`
`rcl_log_level_t` -> `rcl_logger_setting_array_t`

or

`rcl_logger_setting_t`
`rcl_log_level_t` -> `rcl_loggers_settings_t`

I'm not sure what is the right naming

rcl/include/rcl/log_level.h
```diff
@@ -0,0 +1,74 @@
...
+{
+  /// Default log level (represented by `RCUTILS_LOG_SEVERITY` enum) or -1 if not specified.
```
ivanpauno, line 40: If this can only be vales in an enum, why not using that type?
Unfortunately, the type was names with the wrong casing, but adding a typedef in rcl sounds like a good idea:
```c
typedef RCUTILS_LOG_SEVERITY rcl_log_severity_t;
```

we could also rename `severity` to `level`

rcl/include/rcl/log_level.h
```diff
@@ -0,0 +1,74 @@
...
+} rcl_log_level_t;
+
+
+/// \brief Initialize log level structure
+/// \param[in] allocator memory allocator to be used
+/// \return a pointer to log level structure on success or NULL on failure
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_log_level_t *
+rcl_log_level_init(const rcutils_allocator_t allocator);
```
ivanpauno, line 54: I would recommend having two functions:
```c
rcl_log_level_t rcl_get_zero_initialized_log_level();

rcl_ret_t
rcl_log_level_init(rcl_log_level_t * log_level, const rcl_allocator_t * allocator);
```

This approach allows having a stack allocated `rcl_log_level_t` (of course, it's member would still be heap allocated) and better error handling in the init function.

ivanpauno: Notes:
- Use `rcl_allocator_t` and not `rcutils_allocator_t`.
- We have some functions passing allocators by value, and some others by value. We have more of the first kind so I would like to standardize to that style.

rcl/include/rcl/log_level.h
```diff
@@ -0,0 +1,74 @@
...
+
+/// \brief Copy log level structure, allocate memory by using allocator of rcl_log_level_t
+/// \param[in] log_level points to the log level struct to be copied
+/// \return a pointer to the copied log level structure on success or NULL on failure
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_log_level_t * rcl_log_level_copy(const rcl_log_level_t * log_level);
```
ivanpauno, line 61: For better error handling, I would recommend:

```c
rcl_ret_t
rcl_log_level_copy(const rcl_log_level_t * src, rcl_log_level_t * dst, const rcl_allocator_t * allocator);
```

An allocator should be passed too

rcl/include/rcl/log_level.h
```diff
@@ -0,0 +1,74 @@
...
+
+
+/// \brief Free log level structure
+/// \param[in] log_level points to the populated log level structure
+RCL_PUBLIC
+void
```
ivanpauno, line 67: rcl_ret_t?

It doesn't matter if the function doesn't fail now, but it might potentially return an error in the future.

rcl/include/rcl/log_level.h
```diff
@@ -0,0 +1,74 @@
...
+rcl_log_level_fini(rcl_log_level_t * log_level);
```
ivanpauno, line 68: allocator?

rcl/test/rcl/test_log_level.cpp
```diff
@@ -0,0 +1,206 @@
...
+    unsigned int local_argc = (sizeof(local_argv) / sizeof(const char *)); \
+    ret = rcl_parse_arguments( \
+      local_argc, local_argv, rcl_get_default_allocator(), &local_arguments); \
+    ASSERT_EQ(RCL_RET_INVALID_ROS_ARGS, ret); \
+  }
+
+TEST(CLASSNAME(TestLogLevelFixture, RMW_IMPLEMENTATION), error_log_level) {
+  EXPECT_INVALID_RET_FOR_ARGUMENTS_LOG_LEVEL(
+    "process_name", "--ros-args", "--log-level",
+    "=debug");
```
ivanpauno, line 49: I think we usually avoid using `=` in command line arguments, because `bash` or other shells might use them for something else.
We tend to use `:=` in its place.

I'm not sure if using `=` here is ok, or if we should use `:=`.

rcl/test/rcl/test_log_level.cpp
```diff
@@ -0,0 +1,206 @@
...
+  EXPECT_EQ(static_cast<size_t>(1), log_level->num_loggers);
+  EXPECT_STREQ("rcl", log_level->logger_settings[0].name);
+  EXPECT_EQ(static_cast<int>(RCUTILS_LOG_SEVERITY_DEBUG), log_level->logger_settings[0].level);
+}
+
+TEST(CLASSNAME(TestLogLevelFixture, RMW_IMPLEMENTATION), logger_with_default_log_level) {
+  rcl_log_level_t * log_level = NULL;
+  GET_LOG_LEVEL_FROM_ARGUMENTS(
+    log_level, "process_name", "--ros-args",
+    "--log-level", "rcl=debug,debug");
```
ivanpauno, line 163: I'm also unsure about `,` usage.
We could also allow multiple `--log-level` arguments, and use something like:
```
--ros-args --log-level info --log-level rcl:=debug --log-level rclcpp:=debug
```

of course, it's much more verbose. I'm not sure of what's the best option.

rcl/include/rcl/log_level.h
ivanpauno: I see that the `rcl_log_level_t` struct internally stores an allocator, disregard.

rcl/include/rcl/log_level.h
ivanpauno: I see that the `src` `rcl_log_level_t` struct internally stores an allocator, so the extra allocator argument isn't needed.

clalancette:
> I think that before going ahead with this, we should update the related [design document](https://github.com/ros2/design/blob/gh-pages/articles/150_ros_command_line_arguments.md) and ask for feedback. There are different options on how the command line arguments could look, and it would be good to get consensus about that first.

Yeah, that's a good point.  We should probably go through that, and then revisit this PR with that in mind.

> This is a partial review, I haven't actually reviewed the implementation.

I just did a bunch of work here to rebase this onto master and fix the memory leaks that I was able to find with valgrind.  But since some of parsing may change depending on the outcome of the design review, we may want to hold off going much further.

fujitatomoya:
> I think that before going ahead with this, we should update the related design document and ask for feedback. There are different options on how the command line arguments could look, and it would be good to get consensus about that first.

that makes sense, let's do that.

iuhilnehc-ynos:
@clalancette 
Thanks for pointing errors and updating them.

@ivanpauno @fujitatomoya 
Yes, it's better to get a consensus on the design file. I'll update it.

rcl/include/rcl/log_level.h
iuhilnehc-ynos: It needs a variable with heap allocated here.
```
typedef struct rcl_arguments_impl_t
...
  rcl_log_level_t * log_level;                // not define with `rcl_log_level_t log_level;`
}

args_impl->log_level = rcl_log_level_init(allocator);
```

use the same style referred to 
https://github.com/ros2/rcl/blob/b9cfc243bf018091b5c5bcb5585d19591daf4ae6/rcl/src/rcl/arguments.c#L280

if updating with your recommendation, rcl_log_level_t rcl_get_zero_initialized_log_level seems not working.

rcl/include/rcl/log_level.h
iuhilnehc-ynos: refer to
https://github.com/ros2/rcl/blob/b9cfc243bf018091b5c5bcb5585d19591daf4ae6/rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h#L51-L53

rcl/include/rcl/log_level.h
iuhilnehc-ynos: actually I like to use rcl_log_level_t, because user use '--log-level' to control it

rcl/include/rcl/log_level.h
```diff
@@ -0,0 +1,74 @@
...
+  int level;
+} rcl_logger_setting_t;
+
+/// Hold log_level.
+typedef struct rcl_log_level_t
+{
+  /// Default log level (represented by `RCUTILS_LOG_SEVERITY` enum) or -1 if not specified.
+  int default_log_level;
+  struct rcl_logger_setting_t * logger_settings;   ///<  Array of logger
+  size_t num_loggers;       ///< Number of loggers
```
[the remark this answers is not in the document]
iuhilnehc-ynos: I'll update it

rcl/include/rcl/log_level.h
```diff
@@ -0,0 +1,74 @@
...
+/// \brief Copy log level structure, allocate memory by using allocator of rcl_log_level_t
+/// \param[in] log_level points to the log level struct to be copied
+/// \return a pointer to the copied log level structure on success or NULL on failure
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_log_level_t * rcl_log_level_copy(const rcl_log_level_t * log_level);
+
+
+/// \brief Free log level structure
+/// \param[in] log_level points to the populated log level structure
```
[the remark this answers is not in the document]
iuhilnehc-ynos: I'll update it

rcl/include/rcl/log_level.h
iuhilnehc-ynos: refer to
https://github.com/ros2/rcl/blob/b9cfc243bf018091b5c5bcb5585d19591daf4ae6/rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h#L57

rcl/test/rcl/test_log_level.cpp
```diff
@@ -21,41 +21,22 @@
...
-    "rcl=debug,,");
-
-  EXPECT_INVALID_RET_FOR_ARGUMENTS_LOG_LEVEL(
-    "process_name", "--ros-args", "--log-level",
-    "rcl=");
-
-  EXPECT_INVALID_RET_FOR_ARGUMENTS_LOG_LEVEL(
-    "process_name", "--ros-args", "--log-level",
-    "rcl=,");
+  ASSERT_EQ(RCL_RET_INVALID_ROS_ARGS, setup_and_parse_log_level_args("=debug"));
```
iuhilnehc-ynos, line 34: @clalancette
If adding an error test case with
("process_name", "--ros-args", "--log-level", "debug" "--log-level", "rcl=debug,,"), this function doesn't work.

rcl/include/rcl/log_level.h
iuhilnehc-ynos: How about updating with
```
/// A logger item to specify a name and a log level
typedef struct rcl_logger_setting_t
{
  /// name for the logger.
  char * name;
  /// level for the logger.
  rcl_log_severity_t level;
} rcl_logger_setting_t;

/// Hold default logger level and other logger setting.
typedef struct rcl_log_level_t
{
  /// Default logger level
  rcl_log_severity_t default_logger_level;
  /// Array of logger setting
  struct rcl_logger_setting_t * logger_settings;
  /// Number of logger settings
  size_t num_logger_settings;
  rcl_allocator_t allocator;
} rcl_log_level_t;

typedef struct rcl_arguments_impl_t {
    rcl_log_level_t * log_level_settings;
}
```

iuhilnehc-ynos:
incorrectly signed off, pushed forcibly

rcl/include/rcl/arguments.h
```diff
@@ -343,6 +344,34 @@ rcl_remove_ros_arguments(
...
+ * \return `RCL_RET_INVALID_ARGUMENT` if any function arguments are invalid, or
+ * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_arguments_get_log_level(
+  const rcl_arguments_t * arguments,
+  rcl_log_level_t ** log_level);
```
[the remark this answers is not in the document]
ivanpauno: Ok, I really don't like the previous API (and it's the only place we wrote something like that).

One usually follows either "user allocated" semantics:
```c
rcl_ret_t
rcl_arguments_get_log_level(
  const rcl_arguments_t * arguments,
  rcl_log_level_t * log_level);
```

or "function allocated" semantics:
```c
rcl_log_level_t *
rcl_arguments_get_log_level(
  const rcl_arguments_t * arguments);
```

Options that aren't either one or the other doesn't look good to me.

In this case, functions in the same API are inconsistent between each other:
`rcl_arguments_get_log_level` use this weird "function allocated with error code" semantics, but `rcl_log_level_copy` follows the normal "function allocated" semantics. That doesn't make sense.

rcl/include/rcl/log_level.h
ivanpauno: `rcl_log_level_t` sounds like a single log level, i.e. it sounds like a typedef of https://github.com/ros2/rcutils/blob/049167533d473cca5c8ae8347becdd2209503927/include/rcutils/logging.h#L165.

It should be clear in the name that it's a collection of logger settings IMO.

rcl/include/rcl/log_level.h
ivanpauno: > I think the reason why using int directly is that the rcutils_logging_set_logger_level uses 'int' as an argument.

mmm, I didn't see that. Ok, using `int` directly sounds good.

rcl/include/rcl/log_level.h
ivanpauno: You could still heap allocate the struct:

```c
args_impl->log_levels = malloc(sizeof(rcl_log_levels_t));
if (NULL == args_impl->log_levels) {
  ... handle error
}
*args_impl->log_levels = rcl_get_zero_initialized_log_level();
rcl_ret_t ret = rcl_log_level_init(args_impl->log_levels, allocator);
...
```

Considering that previous API was already written like this https://github.com/ros2/rcl/pull/664#discussion_r447684454, I think it's fine to keep the same style.

rcl/include/rcl/log_level.h
ivanpauno: https://github.com/ros2/rcl/blob/b9cfc243bf018091b5c5bcb5585d19591daf4ae6/rcl/include/rcl/arguments.h#L388-L390
?

That's likely the only fini function without return code.

rcl/include/rcl/log_level.h
ivanpauno: https://github.com/ros2/rcl/blob/7723308e7c7f7515590bd709525fd840b4aed254/rcl/include/rcl/arguments.h#L366-L369
?

rcl/include/rcl/log_level.h
ivanpauno: `rcl_log_level_t` -> `rcl_log_levels_t`
works for me

iuhilnehc-ynos: Updated

rcl/include/rcl/log_level.h
iuhilnehc-ynos: Updated with this style

rcl/include/rcl/log_level.h
iuhilnehc-ynos: Updated with this style

rcl/src/rcl/log_level.c
iuhilnehc-ynos: I lost this comment, updated

rcl/src/rcl/logging.c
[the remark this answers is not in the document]
iuhilnehc-ynos: Updated

rcl/test/rcl/test_log_level.cpp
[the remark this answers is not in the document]
ivanpauno: > Not using a function so gtest failure output shows the line number where the macro is used.

There's some ways to avoid that problem https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#adding-traces-to-assertions.

I agree that macros should be avoided, but as part of a test implementation I don't mind much.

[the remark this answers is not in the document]
ivanpauno: This macro was already deleted from the code, so ignore the above comment ...

rcl/test/rcl/test_log_level.cpp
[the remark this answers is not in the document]
ivanpauno: see https://github.com/ros2/rcl/pull/664#discussion_r448324383

rcl/include/rcl/log_level.h
```diff
@@ -0,0 +1,113 @@
...
+ *  Its allocator is used to copy memory into the new structure.
+ * \param[out] log_levels_out A zero-initialized log levels structure to be copied into.
+ * \return `RCL_RET_OK` if the structure was copied successfully, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if log_levels or log_levels_out are invalid, or
+ * \return `RCL_RET_BAD_ALLOC` if allocating memory failed.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_log_levels_copy(const rcl_log_levels_t * log_levels, rcl_log_levels_t * log_levels_out);
```
ivanpauno, line 89:
```suggestion
rcl_log_levels_copy(const rcl_log_levels_t * src, rcl_log_levels_t * dst);
```

Nothing more clear than `src` and `dst` :wink:

rcl/include/rcl/log_level.h
```diff
@@ -0,0 +1,113 @@
...
+  rcl_allocator_t allocator;
+} rcl_log_levels_t;
+
+/// Initialize a log levels structure.
+/**
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
```
ivanpauno, line 56: why is this not thread safe?

rcl/include/rcl/log_level.h
```diff
@@ -0,0 +1,113 @@
...
+ * Lock-Free          | Yes
+ *
+ * \param[in] allocator memory allocator to be used.
+ * \param[in] logger_count to allocate the logger setting count of log levels.
+ * \return a pointer to log level structure on success or NULL on failure.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_log_levels_t *
+rcl_log_levels_init(const rcl_allocator_t allocator, size_t logger_count);
```
ivanpauno, line 67:
```suggestion
rcl_log_levels_init(const rcl_allocator_t * allocator, size_t logger_count);
```

rcl/include/rcl/types.h
```diff
@@ -114,4 +114,7 @@ typedef rmw_ret_t rcl_ret_t;
 /// typedef for rmw_serialized_message_t;
 typedef rmw_serialized_message_t rcl_serialized_message_t;
 
+/// typedef for RCUTILS_LOG_SEVERITY;
+typedef enum RCUTILS_LOG_SEVERITY rcl_log_severity_t;
```
ivanpauno, line 118: please, add this to `log_level.h`

rcl/src/rcl/arguments.c
```diff
@@ -40,6 +40,10 @@ extern "C"
 {
 #endif
 
+#define RCL_ENABLE_FLAG_PREFIX "--enable-"
+#define RCL_DISABLE_FLAG_PREFIX "--disable-"
+#define RCL_LOG_LEVEL_FLAG_SEPARATOR ":="
```
ivanpauno, line 45: I'm not sure you need this define ...
We already have logic to parse that

rcl/src/rcl/arguments.c
ivanpauno: This seems to be solved in line 153 now

rcl/src/rcl/arguments.c
```diff
@@ -139,11 +143,45 @@ rcl_arguments_get_param_overrides(
...
+  rcl_ret_t ret = rcl_log_levels_copy(arguments->impl->log_levels, *log_levels);
+  if (RCL_RET_OK != ret) {
+    rcl_ret_t log_levels_ret = rcl_log_levels_fini(*log_levels);
+    if (RCL_RET_OK != log_levels_ret) {
+      ret = log_levels_ret;
+      RCL_SET_ERROR_MSG("Error while finalizing log levels due to another error");
+    }
+    *log_levels = NULL;
+    return ret;
+  }
```
ivanpauno, line 175: this is wrong.

you should never need to call a fini function after a init/copy function failed.
init/copy functions should take care of cleaning up resources on failure.

note: there might be other places where we do this, but it's not the correct thing to do.

rcl/src/rcl/arguments.c
```diff
@@ -639,6 +679,20 @@ rcl_parse_arguments(
     }
   }
 
+  // Shrink logger settings of log levels
```
ivanpauno, line 682: Please create a function for this:

```c
rcl_ret_t rcl_log_levels_shrink_to_size(const char * name, rcl_log_severity_t level);
```

In that way, things are much easier to unit test.

rcl/src/rcl/arguments.c
```diff
@@ -1597,17 +1662,72 @@ rcl_ret_t
...
+  RCL_CHECK_ARGUMENT_FOR_NULL(log_level->logger_settings, RCL_RET_INVALID_ARGUMENT);
+
+  size_t separator_len = strlen(RCL_LOG_LEVEL_FLAG_SEPARATOR);
+  rcutils_ret_t ret = RCUTILS_RET_OK;
+  char * name = NULL;
+  char * level = NULL;
+  int logger_level = 0;
+  char * separator = strstr(arg, RCL_LOG_LEVEL_FLAG_SEPARATOR);
+
+  if (separator) {
```
ivanpauno, line 1678: Please, use the [lexer](https://github.com/ros2/rcl/blob/f6736211c90067f2d93778ef98c4b47d4f5791ee/rcl/include/rcl/lexer.h).
See for example https://github.com/ros2/rcl/blob/f6736211c90067f2d93778ef98c4b47d4f5791ee/rcl/src/rcl/arguments.c#L1734.

rcl/src/rcl/arguments.c
```diff
@@ -1597,17 +1662,72 @@ rcl_ret_t
...
+
+    ret = rcutils_logging_severity_level_from_string(
+      level,
+      allocator,
+      &logger_level);
+    allocator.deallocate(level, allocator.state);
+    if (RCUTILS_RET_OK == ret) {
+      logger_log_level->name = name;
+      logger_log_level->level = (rcl_log_severity_t)logger_level;
+      log_level->num_logger_settings += 1;
```
ivanpauno, line 1713: consider extending `rcl_log_levels_t` API to handle this:

```c
rcl_ret_t rcl_log_levels_add_logger_setting(const char * logger_name, rcl_log_severity_t log_level);
```

this will make things much easier to unit test.

rcl/src/rcl/log_level.c
```diff
@@ -0,0 +1,102 @@
...
+#include "rcl/log_level.h"
+
+#include "rcl/error_handling.h"
+#include "rcutils/allocator.h"
+#include "rcutils/strdup.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
```
ivanpauno, line 24: this is only needed in pure C header files

rcl/src/rcl/log_level.c
```diff
@@ -0,0 +1,102 @@
...
+
+rcl_log_levels_t *
+rcl_log_levels_init(const rcl_allocator_t allocator, size_t logger_count)
+{
+  RCL_CHECK_ALLOCATOR_WITH_MSG(&allocator, "invalid allocator", return NULL);
+  rcl_log_levels_t * log_levels = allocator.allocate(sizeof(rcl_log_levels_t), allocator.state);
+  if (NULL == log_levels) {
+    RCL_SET_ERROR_MSG("Error allocating mem");
```
ivanpauno, line 32:
```suggestion
    RCL_SET_ERROR_MSG("Error allocating memory");
```

rcl/src/rcl/arguments.c
ivanpauno: consider adding a `capacity` member to the struct too

rcl/src/rcl/log_level.c
```diff
@@ -0,0 +1,102 @@
...
+  return log_levels;
+}
+
+rcl_ret_t
+rcl_log_levels_copy(const rcl_log_levels_t * log_levels, rcl_log_levels_t * log_levels_out)
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(log_levels, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(log_levels_out, RCL_RET_INVALID_ARGUMENT);
+
+  rcutils_allocator_t allocator = log_levels->allocator;
```
ivanpauno, line 62:
```suggestion
  rcl_allocator_t * allocator = &log_levels->allocator;
```

You will need to dereference the pointer elsewhere

rcl/src/rcl/arguments.c
```diff
@@ -139,11 +143,45 @@ rcl_arguments_get_param_overrides(
...
+  RCL_CHECK_ARGUMENT_FOR_NULL(arguments->impl, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(arguments->impl->log_levels, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(log_levels, RCL_RET_INVALID_ARGUMENT);
+
+  if (NULL != *log_levels) {
+    RCL_SET_ERROR_MSG("Output log level pointer is not null.");
+    return RCL_RET_INVALID_ARGUMENT;
+  }
+
+  *log_levels = rcl_log_levels_init(
```
ivanpauno, line 161: You shouldn't need to init before copying.

rcl/src/rcl/log_level.c
```diff
@@ -0,0 +1,102 @@
...
+      }
+      return NULL;
+    }
+  }
+
+  return log_levels;
+}
+
+rcl_ret_t
+rcl_log_levels_copy(const rcl_log_levels_t * log_levels, rcl_log_levels_t * log_levels_out)
```
ivanpauno, line 57: as mentioned in other comments:

* you shouldn't need to call init before calling copy.
* the function should make sure to free resources in `log_levels_out` if it previously allocated resources.
* If something fails:
  * All allocated resources by the function should be freed (weak guarantee). It seems to already be doing this correctly
  * `log_levels_out` should not be modified (strong guarantee). It's ok to not satisfy this item, though preferred.

iuhilnehc-ynos:
> we would expect the resulting log level to be INFO.

Currently, in that case, there will be two logger settings stored in log_levels, and then call twice `rcutils_logging_set_logger_level` in order.

rcl/include/rcl/log_level.h
```diff
@@ -0,0 +1,113 @@
...
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+/// A logger item to specify a name and a log level
+typedef struct rcl_logger_setting_t
+{
+  /// name for the logger.
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Updated

rcl/include/rcl/log_level.h
```diff
@@ -0,0 +1,113 @@
...
+  char * name;
+  /// level for the logger.
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Updated

rcl/include/rcl/log_level.h
```diff
@@ -0,0 +1,113 @@
...
+#include "rcl/macros.h"
+#include "rcl/types.h"
+#include "rcl/visibility_control.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+/// A logger item to specify a name and a log level
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Updated

rcl/include/rcl/log_level.h
```diff
@@ -0,0 +1,113 @@
...
+/// Hold default logger level and other logger setting.
+typedef struct rcl_log_levels_t
+{
+  /// Default logger level
+  rcl_log_severity_t default_logger_level;
+  /// Array of logger setting
+  struct rcl_logger_setting_t * logger_settings;
+  /// Number of logger settings
+  size_t num_logger_settings;
+  /// Allocator used to allocate objects in this struct
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Updated

rcl/include/rcl/log_level.h
```diff
@@ -0,0 +1,113 @@
...
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] allocator memory allocator to be used.
+ * \param[in] logger_count to allocate the logger setting count of log levels.
+ * \return a pointer to log level structure on success or NULL on failure.
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Updated

rcl/include/rcl/log_level.h
iuhilnehc-ynos: Updated

rcl/include/rcl/log_level.h
iuhilnehc-ynos: Updated

rcl/include/rcl/log_level.h
iuhilnehc-ynos: Updated

rcl/include/rcl/types.h
iuhilnehc-ynos: Updated

rcl/src/rcl/arguments.c
iuhilnehc-ynos: Use lexer

rcl/src/rcl/arguments.c
iuhilnehc-ynos: Updated

rcl/src/rcl/arguments.c
iuhilnehc-ynos: Updated

rcl/src/rcl/log_level.c
iuhilnehc-ynos: Updated

rcl/src/rcl/log_level.c
iuhilnehc-ynos: Updated

rcl/src/rcl/arguments.c
iuhilnehc-ynos: I don't see there is a problem here. What's your suggestion?

iuhilnehc-ynos:
> Thanks, this is starting to shape up now and look good. I ended up pushing some additional error tests to test_log_level.cpp.
> 
> One test that I think is missing now is a test for the command-line precedence order. That is, if a user passes `--ros-args --log-level rcl:=debug --log-level rcl:=info`, we would expect the resulting log level to be INFO. @iuhilnehc-ynos could you please add a test for that?

@clalancette 
Updated.

rcl/src/rcl/arguments.c
iuhilnehc-ynos: @ivanpauno 
misunderstood your meaning.
Updated.

rcl/src/rcl/log_level.c
iuhilnehc-ynos: Updated

rcl/src/rcl/log_level.c
```diff
@@ -0,0 +1,166 @@
...
+    allocator->deallocate(log_levels->logger_settings, allocator->state);
+    log_levels->logger_settings = NULL;
+    log_levels->capacity_logger_settings = 0;
+  } else if (log_levels->num_logger_settings < log_levels->capacity_logger_settings) {
+    log_levels->logger_settings = rcutils_reallocf(
+      log_levels->logger_settings,
+      sizeof(rcl_logger_setting_t) * log_levels->num_logger_settings, allocator);
+    if (NULL == log_levels->logger_settings) {
+      return RCL_RET_BAD_ALLOC;
+    }
```
ivanpauno, line 122: I don't think that `rcl_log_levels_shrink_to_size` should automatically deallocate the memory on error.
Instead, the old settings should still be valid, and the user call fini if desired.

```suggestion
    new_logger_settings = allocator->reallocate(
      log_levels->logger_settings,
      sizeof(rcl_logger_setting_t) * log_levels->num_logger_settings,
      allocator->state);
    if (NULL == new_logger_settings) {
      return RCL_RET_BAD_ALLOC;
    }
    log_levels->logger_settings = new_logger_settings;
```

rcl/src/rcl/log_level.c
```diff
@@ -0,0 +1,166 @@
...
+
+rcl_ret_t
+rcl_log_levels_add_logger_setting(
+  rcl_log_levels_t * log_levels, const char * logger_name, rcl_log_severity_t log_level)
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(log_levels, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(log_levels->logger_settings, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(logger_name, RCL_RET_INVALID_ARGUMENT);
+
+  // check if there exists a same name logger setting
```
ivanpauno, line 136:
```suggestion
  // check if there exists a logger with the same name
```

rcl/src/rcl/log_level.c
```diff
@@ -0,0 +1,166 @@
...
+  rcl_logger_setting_t * logger_setting = NULL;
+  for (size_t i = 0; i < log_levels->num_logger_settings; ++i) {
+    if (strcmp(log_levels->logger_settings[i].name, logger_name) == 0) {
+      logger_setting = &log_levels->logger_settings[i];
+      if (log_levels->logger_settings[i].level != log_level) {
+        RCUTILS_LOG_WARN_NAMED(
```
ivanpauno, line 142:
```suggestion
        RCUTILS_LOG_DEBUG_NAMED(
```

replacing a logger setting is expected IMO

rcl/src/rcl/log_level.c
```diff
@@ -0,0 +1,166 @@
...
+          ROS_PACKAGE_NAME, "Minimum log level of logger [%s] will be replaced from %d to %d",
+          logger_name, logger_setting->level, log_level);
+      }
+      log_levels->allocator.deallocate((void *)logger_setting->name, log_levels->allocator.state);
+      break;
```
ivanpauno, line 147: why deallocating it?
IMO, it's simpler to override the log level and return

```suggestion
          logger_name, logger_setting->level, log_level);
          logger_setting->level = log_level;
      }
      return;
```

rcl/src/rcl/log_level.c
```diff
@@ -0,0 +1,166 @@
...
+      RCL_SET_ERROR_MSG("No capacity to store a logger setting");
+      return RCL_RET_ERROR;
+    }
+
+    logger_setting =
+      &log_levels->logger_settings[log_levels->num_logger_settings];
+    log_levels->num_logger_settings += 1;
+  }
+
+  logger_setting->name = logger_name;
```
ivanpauno, line 162: This is borrowing `logger_name`, I would prefer making a copy here.
Either case, behavior should be clear in documentation.

rcl/test/rcl/test_log_level.cpp
```diff
@@ -0,0 +1,192 @@
...
+  });
+  EXPECT_EQ(RCUTILS_LOG_SEVERITY_DEBUG, log_levels->default_logger_level);
+  EXPECT_EQ(0ul, log_levels->num_logger_settings);
+}
+
+TEST(TestLogLevel, logger_log_level_debug) {
+  rcl_log_levels_t * log_levels = NULL;
+  GET_LOG_LEVEL_FROM_ARGUMENTS(
+    log_levels, "process_name", "--ros-args",
+    "--log-level", "rcl:=debug");
```
ivanpauno, line 110: add test with a logger name including dots

rcl/src/rcl/arguments.c
```diff
@@ -1593,21 +1646,149 @@ _rcl_parse_remap_begin_remap_rule(
...
+    RCL_SET_ERROR_MSG("failed to get start of logger name");
+    return RCL_RET_ERROR;
+  }
+
+  rcl_ret_t ret = rcl_lexer_lookahead2_peek(lex_lookahead, &lexeme);
+  if (RCL_RET_OK != ret) {
+    return ret;
+  }
+
+  while (RCL_LEXEME_SEPARATOR != lexeme) {
```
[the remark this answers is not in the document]
iuhilnehc-ynos: logger name can be named as anything.
Current logic just cares the position of SEPARATOR(:=), and then get the name by (name_end-name_start),
so I think it's not necessary to check RCL_LEXEME_DOT here.

I add a test case in which the logger name contains '.', and it passed.

rcl/src/rcl/arguments.c
```diff
@@ -1593,21 +1646,149 @@ _rcl_parse_remap_begin_remap_rule(
...
+      goto cleanup;
+    }
+
+    ret = rcl_lexer_lookahead2_expect(&lex_lookahead, RCL_LEXEME_SEPARATOR, NULL, NULL);
+    if (RCL_RET_WRONG_LEXEME == ret) {
+      ret = RCL_RET_INVALID_LOG_LEVEL_RULE;
+      goto cleanup;
+    }
+
+    const char * level = rcl_lexer_lookahead2_get_text(&lex_lookahead);
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Updated

rcl/include/rcl/log_level.h
```diff
@@ -0,0 +1,158 @@
...
+  size_t num_logger_settings;
+  /// Capacity of logger settings.
+  size_t capacity_logger_settings;
+  /// Allocator used to allocate objects in this struct.
+  rcl_allocator_t allocator;
+} rcl_log_levels_t;
+
+/// Initialize a log levels structure.
+/**
+ * <hr>
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Updated

rcl/src/rcl/log_level.c
```diff
@@ -0,0 +1,166 @@
...
+
+  dst->default_logger_level = src->default_logger_level;
+  dst->num_logger_settings = src->num_logger_settings;
+  dst->capacity_logger_settings = src->capacity_logger_settings;
+  dst->allocator = src->allocator;
+  for (size_t i = 0; i < src->num_logger_settings; ++i) {
+    dst->logger_settings[i].name =
+      rcutils_strdup(src->logger_settings[i].name, *allocator);
+    if (NULL == dst->logger_settings[i].name) {
+      if (RCL_RET_OK != rcl_log_levels_fini(dst)) {
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Thank you. Updated

rcl/src/rcl/log_level.c
iuhilnehc-ynos: Updated

rcl/src/rcl/log_level.c
iuhilnehc-ynos: Updated

rcl/src/rcl/log_level.c
iuhilnehc-ynos: Updated

rcl/src/rcl/log_level.c
iuhilnehc-ynos: Updated

rcl/src/rcl/log_level.c
iuhilnehc-ynos: Updated

rcl/test/rcl/test_log_level.cpp
iuhilnehc-ynos: Updated

rcl/src/rcl/log_level.c
```diff
@@ -0,0 +1,166 @@
...
+
+rcl_ret_t
+rcl_log_levels_copy(const rcl_log_levels_t * src, rcl_log_levels_t * dst)
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(src, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(dst, RCL_RET_INVALID_ARGUMENT);
+  const rcl_allocator_t * allocator = &src->allocator;
+  RCL_CHECK_ALLOCATOR_WITH_MSG(allocator, "invalid allocator", return RCL_RET_INVALID_ARGUMENT);
+
+  dst->logger_settings = allocator->allocate(
```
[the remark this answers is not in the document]
iuhilnehc-ynos: I'll take your suggestion. Updated

rcl/src/rcl/arguments.c
```diff
@@ -1593,21 +1651,166 @@ _rcl_parse_remap_begin_remap_rule(
...
+  if (logger_level) {
+    allocator->deallocate(logger_level, allocator->state);
+  }
+  if (RCL_RET_OK != ret) {
+    if (RCL_RET_OK != rcl_lexer_lookahead2_fini(&lex_lookahead)) {
+      RCUTILS_LOG_ERROR_NAMED(ROS_PACKAGE_NAME, "Failed to fini lookahead2 after error occurred");
+    }
+  } else {
+    ret = rcl_lexer_lookahead2_fini(&lex_lookahead);
+  }
```
ivanpauno, line 1811: I find this a bit more clear:
```suggestion
  rcl_ret_t rv = rcl_lexer_lookahead2_fini(&lex_lookahead);
  if (RCL_RET_OK != rv) {
    if (RCL_RET_OK != ret) {
      RCUTILS_LOG_ERROR_NAMED(ROS_PACKAGE_NAME, "Failed to fini lookahead2 after error occurred");
    } else {
      ret = rv;
    }
  }
```
not a big difference though

rcl/src/rcl/arguments_impl.h
```diff
@@ -50,8 +51,8 @@ typedef struct rcl_arguments_impl_t
   /// Length of remap_rules.
   int num_remap_rules;
 
-  /// Default log level (represented by `RCUTILS_LOG_SEVERITY` enum) or -1 if not specified.
-  int log_level;
+  /// Log levels parsed from arguments.
+  rcl_log_levels_t * log_levels;
```
ivanpauno, line 55: as commented elsewhere, this could be directly:
```suggestion
  rcl_log_levels_t log_levels;
```

rcl/test/rcl/test_log_level.cpp
```diff
@@ -0,0 +1,207 @@
...
+  EXPECT_EQ(RCUTILS_LOG_SEVERITY_INFO, log_levels->logger_settings[0].level);
+}
+
+int main(int argc, char ** argv)
+{
+  ::testing::InitGoogleTest(&argc, argv);
+  int ret = RUN_ALL_TESTS();
+  // We need to do rcutils_logging_shutdown() since the tests above may
+  // have initialized logging.
+  if (rcutils_logging_shutdown() != RCUTILS_RET_OK) {
```
ivanpauno, line 203: if you need this, please write a TestCase clase with the proper setup and teardown, instead of writing a main

clalancette: I added that.

To be clear, we only want to do this once at the end of the program.  Is there a particular reason to do it in a class rather than in a main?  It seems like a whole lot more boilerplate code to me and the benefit is unclear to me, since we don't need the class otherwise.

ivanpauno: > To be clear, we only want to do this once at the end of the program. Is there a particular reason to do it in a class rather than in a main? It seems like a whole lot more boilerplate code to me and the benefit is unclear to me, since we don't need the class otherwise.

I didn't see that comment. Both ways are fine, I typically avoid writing a main in `gtest` but there's no big difference.

rcl/include/rcl/log_level.h
```diff
@@ -52,6 +52,12 @@ typedef struct rcl_log_levels_t
   rcl_allocator_t allocator;
 } rcl_log_levels_t;
 
+/// Return a rcl_log_levels_t struct with members initialized to zero value.
+RCL_PUBLIC
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Updated

rcl/include/rcl/log_level.h
```diff
@@ -62,14 +68,19 @@ typedef struct rcl_log_levels_t
  * Uses Atomics       | No
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Updated

rcl/include/rcl/log_level.h
```diff
@@ -123,8 +134,8 @@ rcl_log_levels_fini(rcl_log_levels_t * log_levels);
  *
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Updated

rcl/include/rcl/log_level.h
```diff
@@ -103,7 +114,7 @@ rcl_log_levels_copy(const rcl_log_levels_t * src, rcl_log_levels_t * dst);
  * Uses Atomics       | No
  * Lock-Free          | Yes
  *
- * \param[in] log_levels The structure to be deallocated.
+ * \param[in] log_levels The members of structure to be deallocated.
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Updated

rcl/src/rcl/arguments.c
```diff
@@ -328,11 +328,16 @@ rcl_parse_arguments(
     ret = RCL_RET_BAD_ALLOC;
     goto fail;
   }
-  args_impl->log_levels = rcl_log_levels_init(&allocator, argc);
-  if (NULL == args_impl->log_levels) {
+  rcl_log_levels_t * log_levels = allocator.allocate(sizeof(rcl_log_levels_t), allocator.state);
```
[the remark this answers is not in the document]
iuhilnehc-ynos: OK, updated

rcl/src/rcl/arguments.c
```diff
@@ -328,11 +328,16 @@ rcl_parse_arguments(
...
+  if (NULL == log_levels) {
     ret = RCL_RET_BAD_ALLOC;
     goto fail;
   }
+  ret = rcl_log_levels_init(log_levels, &allocator, argc);
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Updated

rcl/src/rcl/arguments.c
```diff
@@ -1738,35 +1744,48 @@ _rcl_parse_log_level(
...
+      goto cleanup;
+    }
+
+    ret = rcl_lexer_lookahead2_expect(&lex_lookahead, RCL_LEXEME_EOF, NULL, NULL);
+    if (RCL_RET_OK != ret) {
       ret = RCL_RET_INVALID_LOG_LEVEL_RULE;
       goto cleanup;
     }
+
+    logger_level = rcutils_strndup(level_token, level_token_length, *allocator);
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Updated

rcl/src/rcl/log_level.c
```diff
@@ -19,15 +19,25 @@
...
-    return NULL;
-  }
+  static rcl_log_levels_t log_levels = {
+    .default_logger_level = RCUTILS_LOG_SEVERITY_UNSET,
+    .logger_settings = NULL,
+    .num_logger_settings = 0,
+    .capacity_logger_settings = 0,
+    .allocator = {NULL},
+  };
+  return log_levels;
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Updated

rcl/src/rcl/log_level.c
```diff
@@ -19,15 +19,25 @@
...
+}
+
+rcl_ret_t
+rcl_log_levels_init(
+  rcl_log_levels_t * log_levels, const rcl_allocator_t * allocator, size_t logger_count)
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(log_levels, RCL_RET_INVALID_ARGUMENT);
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Updated

rcl/src/rcl/log_level.c
```diff
@@ -57,6 +63,9 @@ rcl_log_levels_copy(const rcl_log_levels_t * src, rcl_log_levels_t * dst)
   RCL_CHECK_ARGUMENT_FOR_NULL(dst, RCL_RET_INVALID_ARGUMENT);
   const rcl_allocator_t * allocator = &src->allocator;
   RCL_CHECK_ALLOCATOR_WITH_MSG(allocator, "invalid allocator", return RCL_RET_INVALID_ARGUMENT);
+  if (dst->logger_settings != NULL) {
+    return RCL_RET_INVALID_ARGUMENT;
```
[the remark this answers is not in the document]
iuhilnehc-ynos: Updated

rcl/src/rcl/arguments.c
```diff
@@ -139,22 +139,52 @@ rcl_arguments_get_param_overrides(
...
+  RCL_CHECK_ARGUMENT_FOR_NULL(log_levels, RCL_RET_INVALID_ARGUMENT);
+  const rcl_allocator_t * allocator = &arguments->impl->allocator;
+  RCL_CHECK_ALLOCATOR_WITH_MSG(allocator, "invalid allocator", return RCL_RET_INVALID_ARGUMENT);
+
+  if (NULL != *log_levels) {
+    RCL_SET_ERROR_MSG("Output log level pointer is not null.");
+    return RCL_RET_INVALID_ARGUMENT;
+  }
+
+  *log_levels = allocator->zero_allocate(1, sizeof(rcl_log_levels_t), allocator->state);
```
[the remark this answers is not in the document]
iuhilnehc-ynos: OK, updated

rcl/src/rcl/arguments.c
iuhilnehc-ynos: Updated

rcl/src/rcl/arguments_impl.h
iuhilnehc-ynos: Updated

rcl/src/rcl/log_level.c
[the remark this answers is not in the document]
iuhilnehc-ynos: If using this style
```
  return (const rcl_log_levels_t) {
           .default_logger_level = RCUTILS_LOG_SEVERITY_UNSET,
           .logger_settings = NULL,
           .num_logger_settings = 0,
           .capacity_logger_settings = 0,
           .allocator = {NULL},
  };
```

cpplint failed as 'You don't need a ; after a }'
The PR about this issue (https://github.com/google/styleguide/issues/311) is not merged.

If we want to support this style, https://github.com/ament/ament_lint.git used by ROS2 needs to be updated.
So I will not update it this time.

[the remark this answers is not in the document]
ivanpauno: Done https://github.com/ros2/rcl/pull/664/commits/c8dab5f4ab3421a3f6c7eb21b027be75c9cf6c94

rcl/src/rcl/arguments.c
```diff
@@ -966,16 +962,14 @@ rcl_arguments_fini(
       args->impl->num_remap_rules = 0;
     }
 
-    if (args->impl->log_levels) {
-      rcl_ret_t log_levels_ret = rcl_log_levels_fini(args->impl->log_levels);
+    if (args->impl->log_levels.logger_settings) {
```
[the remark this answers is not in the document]
ivanpauno: Done https://github.com/ros2/rcl/pull/664/commits/d795983ff89e6ba13c7d5a67f9682d114050a866

ivanpauno:
There was a memory leak, I've fixed that in https://github.com/ros2/rcl/pull/664/commits/bb5e6ec5be856f70ad0404043df9e1977656dfa6 (I've actually done a API modification to make things more intuitive).
I've also addressed my remaining comments and improved documentation.

I think this is ready to go.
We could also improve coverage of `rcl_log_levels_t` API, I'm not sure if we want to do that here.

@clalancette @iuhilnehc-ynos for a final review.

clalancette:
I rebased it onto the latest master to deal with the new API changes in `rmw`.  Rpr job is *probably* going to fail now.  I'm going to do one more review here, and assuming that is good I'll run CI.

ivanpauno:
> Rpr job is probably going to fail now.

Yes, I'm planning to make some releases today to fix that.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11399/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6619/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9345/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11292/)

PR #665, opened 2020-05-28
compare the return value to constants

iuhilnehc-ynos:

### DIFF
--- a/rcl/src/rcl/logging.c
+++ b/rcl/src/rcl/logging.c
@@ -88,11 +88,13 @@ rcl_logging_configure_with_output_handler(
   if (g_rcl_logging_ext_lib_enabled) {
     status = rcl_logging_external_initialize(config_file, g_logging_allocator);
     if (RCL_RET_OK == status) {
-      status = rcl_logging_external_set_logger_level(NULL, default_level);
-      if (RCL_RET_OK == status) {
-        g_rcl_logging_out_handlers[g_rcl_logging_num_out_handlers++] =
-          rcl_logging_ext_lib_output_handler;
+      rcl_ret_t logging_status = rcl_logging_external_set_logger_level(
+        NULL, default_level);
+      if (logging_status != RCL_RET_OK) {
+        status = RCL_RET_ERROR;
       }
+      g_rcl_logging_out_handlers[g_rcl_logging_num_out_handlers++] =
+        rcl_logging_ext_lib_output_handler;
     }
   }
   rcutils_logging_set_output_handler(output_handler);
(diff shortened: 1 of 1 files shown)


rcl/src/rcl/logging.c
```diff
@@ -88,15 +88,11 @@ rcl_logging_configure_with_output_handler(
...
     status = rcl_logging_external_initialize(config_file, g_logging_allocator);
     if (RCL_RET_OK == status) {
-      // TODO(dirk-thomas) the return value should be typed and compared to
-      // constants instead of zero
-      int logging_status = rcl_logging_external_set_logger_level(
-        NULL, default_level);
-      if (logging_status != 0) {
-        status = RCL_RET_ERROR;
+      status = rcl_logging_external_set_logger_level(NULL, default_level);
+      if (RCL_RET_OK == status) {
```
fujitatomoya, line 92: This changes a logic, only when log level can be set, then register handler?
Previously registering handler anyway even if log level cannot be set.

iuhilnehc-ynos: Yes. And I think this is the correct logic.

iuhilnehc-ynos: @fujitatomoya 
After consideration, I shouldn't update the original logic while refactoring the source code.
I will update it.

fujitatomoya: @iuhilnehc-ynos

I guess that original logic is meant to, I'd like to hear from others.

clalancette:
I'd actually like to hold off on a change like this.  Over in https://github.com/ros2/rcl_logging/pull/33#issuecomment-629416344 , we've been discussing refactoring this interface, and the types that `rcl_logging_external_set_logger_level()` will return will almost certainly change.

We aren't working on it at the moment, but feel free to start working on that if you'd like.  Otherwise, I will probably start taking a look at fixing this up after Foxy is released.

iuhilnehc-ynos:
> I'd actually like to hold off on a change like this. Over in [ros2/rcl_logging#33 (comment)](https://github.com/ros2/rcl_logging/pull/33#issuecomment-629416344) , we've been discussing refactoring this interface, and the types that `rcl_logging_external_set_logger_level()` will return will almost certainly change.

Thanks for your explanation. I got it and close this PR right now.

> We aren't working on it at the moment, but feel free to start working on that if you'd like. Otherwise, I will probably start taking a look at fixing this up after Foxy is released.

Actually, I updated this function because of #664 ( I noticed there is a TODO inside rcl_logging_configure_with_output_handler).
If I have time, I'd like to do that.

PR #667, opened 2020-06-01
Ensure rcl_publisher_init() fails safely.

hidmic:


hidmic:
Any objection to merging as-is? I don't think cross-platform CI would add much.

jacobperron:
I agree, I think we can merge it with the PR check passing.

Issue #669, opened 2020-06-02
rcl_get_localhost_only function not returning a rmw_localhost_only_t value


Blast545:
While adding tests for the `rcl` package I realized the `localhost.c` function `rcl_get_localhost_only` uses this logic to return the possible value of the localhost_only variable:

https://github.com/ros2/rcl/blob/e47042cba4f32f77b361d0f87d905b9386cadde8/rcl/src/rcl/localhost.c#L44

Which only returns `0` or `1`, but the current possible values for this variable are `0-1-2`, as implemented in [rmw/localhost.h](https://github.com/ros2/rmw/blob/eaf5932723efa0be3fda4847b69e09d3753f0a6b/rmw/include/rmw/localhost.h#L26):

```
typedef enum RMW_PUBLIC_TYPE rmw_localhost_only_t
{
  /// Uses ROS_LOCALHOST_ONLY environment variable.
  RMW_LOCALHOST_ONLY_DEFAULT = 0,
  /// Forces using only localhost.
  RMW_LOCALHOST_ONLY_ENABLED = 1,
  /// Forces disabling localhost only.
  RMW_LOCALHOST_ONLY_DISABLED = 2,
} rmw_localhost_only_t;
```

Blast545:
After checking the expected usage of this function, it is expected that the result of this function to be a boolean value because the result coming from this is then fed to `rmw_create_node()` that takes `localhost_only` value as a boolean.

I think the solution to this would be changing the `rcl_get_localhost_only` signature to return boolean instead.

Blast545:
Closed with #670

PR #670, opened 2020-06-02
Fix conversions between rmw_localhost_only_t and bool

Blast545:
When initializing a node, the current code tries to load the localhost_only variable from the environment variable. This PR takes the value already parsed from the initialized context passed to the node.

Also fixes wrong result conversion with the `rcl_get_localhost_only` function.

### DIFF
--- a/rcl/src/rcl/init.c
+++ b/rcl/src/rcl/init.c
@@ -162,10 +162,6 @@ rcl_init(
       fail_ret = ret;
       goto fail;
     }
-  } else if (RMW_LOCALHOST_ONLY_DISABLED == *localhost_only) {
-    // If the user sets the rmw init option to RMW_LOCALHOST_ONLY_DISABLED
-    // manually, save the boolean variable as 0
-    *localhost_only = 0;
   }
 
   if (context->global_arguments.impl->enclave) {
--- a/rcl/src/rcl/init.c
+++ b/rcl/src/rcl/init.c
@@ -162,6 +162,10 @@ rcl_init(
       fail_ret = ret;
       goto fail;
     }
+  } else if (RMW_LOCALHOST_ONLY_DISABLED == *localhost_only) {
+    // If the user sets the rmw init option to RMW_LOCALHOST_ONLY_DISABLED
+    // manually, save the boolean variable as 0
+    *localhost_only = 0;
   }
 
   if (context->global_arguments.impl->enclave) {
--- a/rcl/src/rcl/localhost.c
+++ b/rcl/src/rcl/localhost.c
@@ -41,9 +41,9 @@ rcl_get_localhost_only(rmw_localhost_only_t * localhost_only)
       get_env_error_str);
     return RCL_RET_ERROR;
   }
-  *localhost_only = ros_local_host_env_val != NULL &&
+  *localhost_only = (ros_local_host_env_val != NULL &&
     strcmp(
-    ros_local_host_env_val,
-    "1") == 0 ? RMW_LOCALHOST_ONLY_ENABLED : RMW_LOCALHOST_ONLY_DISABLED;
+      ros_local_host_env_val,
+      "1") == 0) ? RMW_LOCALHOST_ONLY_ENABLED : RMW_LOCALHOST_ONLY_DISABLED;
   return RCL_RET_OK;
 }
--- a/rcl/src/rcl/localhost.c
+++ b/rcl/src/rcl/localhost.c
@@ -41,6 +41,9 @@ rcl_get_localhost_only(rmw_localhost_only_t * localhost_only)
       get_env_error_str);
     return RCL_RET_ERROR;
   }
-  *localhost_only = ros_local_host_env_val != NULL && strcmp(ros_local_host_env_val, "1") == 0;
+  *localhost_only = ros_local_host_env_val != NULL &&
+    strcmp(
+    ros_local_host_env_val,
+    "1") == 0 ? RMW_LOCALHOST_ONLY_ENABLED : RMW_LOCALHOST_ONLY_DISABLED;
   return RCL_RET_OK;
 }
--- a/rcl/src/rcl/node.c
+++ b/rcl/src/rcl/node.c
@@ -263,11 +263,7 @@ rcl_node_init(
   RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME, "Using domain ID of '%zu'", domain_id);
   node->impl->actual_domain_id = domain_id;
 
-  if (RMW_LOCALHOST_ONLY_DEFAULT == localhost_only) {
-    if (RMW_RET_OK != rcl_get_localhost_only(&localhost_only)) {
-      goto fail;
-    }
-  }
+  localhost_only = *(&context->impl->init_options.impl->rmw_init_options.localhost_only);
 
   node->impl->rmw_node_handle = rmw_create_node(
     &(node->context->impl->rmw_context),
--- a/rcl/src/rcl/node.c
+++ b/rcl/src/rcl/node.c
@@ -267,7 +267,8 @@ rcl_node_init(
 
   node->impl->rmw_node_handle = rmw_create_node(
     &(node->context->impl->rmw_context),
-    name, local_namespace_, domain_id, localhost_only);
+    name, local_namespace_, domain_id,
+    localhost_only == RMW_LOCALHOST_ONLY_ENABLED);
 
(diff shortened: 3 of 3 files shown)


rcl/src/rcl/init.c
```diff
@@ -162,6 +162,10 @@ rcl_init(
       fail_ret = ret;
       goto fail;
     }
+  } else if (RMW_LOCALHOST_ONLY_DISABLED == *localhost_only) {
+    // If the user sets the rmw init option to RMW_LOCALHOST_ONLY_DISABLED
+    // manually, save the boolean variable as 0
+    *localhost_only = 0;
```
ivanpauno, line 168: why?
`context->impl->init_options.impl->rmw_init_options.localhost_only` is of type `rmw_localhost_only_t`

what needs to be fixed is `rcl_get_localhost_only`

rcl/src/rcl/node.c
```diff
@@ -263,11 +263,7 @@ rcl_node_init(
   RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME, "Using domain ID of '%zu'", domain_id);
   node->impl->actual_domain_id = domain_id;
 
-  if (RMW_LOCALHOST_ONLY_DEFAULT == localhost_only) {
-    if (RMW_RET_OK != rcl_get_localhost_only(&localhost_only)) {
-      goto fail;
-    }
-  }
+  localhost_only = *(&context->impl->init_options.impl->rmw_init_options.localhost_only);
```
ivanpauno, line 266: why taking a reference and then dereferencing?

ivanpauno: Line 270 needs to be fixed, as `rmw_create_node` is expecting a bool. It should be:

```c
name, local_namespace_, domain_id, localhost_only == RMW_LOCALHOST_ONLY_ENABLED);
```

rcl/src/rcl/init.c
Blast545: `rcl_get_localhost_only` is not called if a user sets `localhost_only=RMW_LOCALHOST_ONLY_DISABLED` in the `rmw_init_options_t` struct. That's the case being covered by this portion of code.

rcl/src/rcl/node.c
Blast545: > why taking a reference and then dereferencing?

Isn't the dereference applied to the `localhost_only` element at the end?

Blast545: > Line 270 needs to be fixed, as `rmw_create_node` is expecting a bool. It should be:

I'll add this!

rcl/src/rcl/init.c
ivanpauno: Yes, in that case `rcl_get_localhost_only` is not being called, and there's no need to modify `localhost_only` value either (i.e. when `context->impl->init_options.impl->rmw_init_options.localhost_only == RMW_LOCALHOST_ONLY_DISABLED` things are working well).

They are not working well when `context->impl->init_options.impl->rmw_init_options.localhost_only == RMW_LOCALHOST_ONLY_DEFAULT` because of [this line](https://github.com/ros2/rcl/blob/e47042cba4f32f77b361d0f87d905b9386cadde8/rcl/src/rcl/localhost.c#L44). That needs to be corrected.

rcl/src/rcl/node.c
ivanpauno: > Isn't the dereference applied to the localhost_only element at the end?

`&a->b` takes the address of member `b` of the struct pointed by `a`.
`*(&a->b)` is equivalent to `a->b`

rcl/src/rcl/init.c
Blast545: > They are not working well when `context->impl->init_options.impl->rmw_init_options.localhost_only == RMW_LOCALHOST_ONLY_DEFAULT` because of [this line](https://github.com/ros2/rcl/blob/e47042cba4f32f77b361d0f87d905b9386cadde8/rcl/src/rcl/localhost.c#L44). That needs to be corrected.

That line is returning 1 or 0 properly (although its signature is misleading and has to be changed, left an issue opened for that).

But I do think it's not OK that all the functions receiving that enum have to make the boolean comparison against `RMW_LOCALHOST_ONLY_DISABLED` before using the value and IMO this should be addressed in this function where the context is initialized.

I'll remove this because the check you suggested me for `node.c` solves the case I was addressing but it would be great to come back at this at some point.

ivanpauno: > That line is returning 1 or 0 properly

It's returning `1` or `0` and setting that to a `rmw_localhost_only_t` which doesn't expect `1/0`, so that line is wrong. It should be:

```c
  *localhost_only = (ros_local_host_env_val != NULL && strcmp(ros_local_host_env_val, "1") == 0) ? RMW_LOCALHOST_ONLY_ENABLED ? RMW_LOCALHOST_ONLY_DISABLED;
```

>  (although its signature is misleading and has to be changed, left an issue opened for that).

The signature is ok and shouldn't be changed. The idea is that in the future we will have the following functions:

```c
rcl_init_options_enable_localhost_only(rcl_init_options_t * init_options);
rcl_init_options_disable_localhost_only(rcl_init_options_t * init_options);
```

There would be equivalents in `rclcpp/rclpy` for those.
If the default constructed `init_options` isn't modified with those functions, localhost only will be set according to the environment variable (as by default it's value is RMW_LOCALHOST_ONLY_DEFAULT).
If the user explicitly enabled/disabled it, the environment variable won't be checked.

As a consequence, at `rcl` level the variable can be `default/enable/disable`, and at `rmw` implementation level can be only `enable/disable`.

Blast545: I'll open a PR for that in the `rcl_get_localhost_only` function. This PR is about passing the context setting to the node, mind giving your thoughts on it?

ivanpauno: > I'll open a PR for that in the `rcl_get_localhost_only` function. This PR is about passing the context setting to the node, mind giving your thoughts on it?

I don't see why opening another PR, as I think this change is incorrect and the only problem is in `rcl_get_localhost_only`.
Can you explain what this extra `else/if` does?

> This PR is about passing the context setting to the node

Can you explain what you mean?

Blast545: > I don't see why opening another PR, as I think this change is incorrect and the only problem is in `rcl_get_localhost_only`.

IMO there are two problems:

1. The one fixed in this PR, `localhost_only` being loaded from environment variable in `node.c` although this variable was already loaded in init.c

2. rcl_get_localhost_only returning boolean instead of  rmw_localhost_only_t.

This PR fixes problem 1. 

> Can you explain what this extra `else/if` does?

Based on my conversation with you, that else/if was removed, please check the latest commit of this PR.

ivanpauno: > rcl_get_localhost_only returning boolean instead of rmw_localhost_only_t.

It's not returning a boolean, it's returning `rmw_localhost_only_t` and doing the conversion incorrectly.
IMO, we should fix the logic in this same PR as today is the deadline for merging things before Foxy release (anyways we can backport the fix in a patch release, so it's not a big deal).

> Based on my conversation with you, that else/if was removed, please check the latest commit of this PR.

I didn't see the commit that removed that :joy:.

Blast545:
Latest commit should close #669

rcl/src/rcl/localhost.c
```diff
@@ -41,6 +41,9 @@ rcl_get_localhost_only(rmw_localhost_only_t * localhost_only)
       get_env_error_str);
     return RCL_RET_ERROR;
   }
-  *localhost_only = ros_local_host_env_val != NULL && strcmp(ros_local_host_env_val, "1") == 0;
+  *localhost_only = ros_local_host_env_val != NULL &&
+    strcmp(
+    ros_local_host_env_val,
+    "1") == 0 ? RMW_LOCALHOST_ONLY_ENABLED : RMW_LOCALHOST_ONLY_DISABLED;
```
ivanpauno, line 47: nit: can you use parentheses? the precedence of the operators isn't obvious

Blast545:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11008/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6343/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8967/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10933/)

PR #671, opened 2020-06-03
Add check rcl_node_options_copy invalid out

Blast545:
Found this scenario when adding tests for `rcl`, the attached test code in this PR causes a not handled exception when trying to call `rcl_node_options_fini` for `not_ini_options`.

I consider this a bug because `rcl_node_options_copy` seems to be a valid way to initialize `rcl_node_options_t` objects.

EDIT: Adding a check and changing the test to reflect that expected usage does not manage receiving a not initialized `rcl_node_options_t` object.

### DIFF
--- a/rcl/src/rcl/node_options.c
+++ b/rcl/src/rcl/node_options.c
@@ -50,6 +50,10 @@ rcl_node_options_copy(
     RCL_SET_ERROR_MSG("Attempted to copy options into itself");
     return RCL_RET_INVALID_ARGUMENT;
   }
+  if (NULL != options_out->arguments.impl) {
+    RCL_SET_ERROR_MSG("Options out must be zero initialized");
+    return RCL_RET_INVALID_ARGUMENT;
+  }
   options_out->domain_id = options->domain_id;
   options_out->allocator = options->allocator;
   options_out->use_global_arguments = options->use_global_arguments;
@@ -57,8 +61,6 @@ rcl_node_options_copy(
   if (NULL != options->arguments.impl) {
     rcl_ret_t ret = rcl_arguments_copy(&(options->arguments), &(options_out->arguments));
     return ret;
-  } else {
-    options_out->arguments.impl = NULL;
   }
   return RCL_RET_OK;
 }
--- a/rcl/src/rcl/node_options.c
+++ b/rcl/src/rcl/node_options.c
@@ -57,6 +57,8 @@ rcl_node_options_copy(
   if (NULL != options->arguments.impl) {
     rcl_ret_t ret = rcl_arguments_copy(&(options->arguments), &(options_out->arguments));
     return ret;
+  } else {
+    options_out->arguments.impl = NULL;
   }
   return RCL_RET_OK;
 }
--- a/rcl/test/rcl/test_node.cpp
+++ b/rcl/test/rcl/test_node.cpp
@@ -798,6 +798,5 @@ TEST_F(CLASSNAME(TestNodeFixture, RMW_IMPLEMENTATION), test_rcl_node_options) {
 TEST_F(CLASSNAME(TestNodeFixture, RMW_IMPLEMENTATION), test_rcl_node_options_fail) {
   rcl_node_options_t not_ini_options;
   rcl_node_options_t default_options = rcl_node_get_default_options();
-  EXPECT_EQ(RCL_RET_OK, rcl_node_options_copy(&default_options, &not_ini_options));
-  EXPECT_EQ(RCL_RET_OK, rcl_node_options_fini(&not_ini_options));
+  EXPECT_EQ(RCL_RET_INVALID_ARGUMENT, rcl_node_options_copy(&default_options, &not_ini_options));
 }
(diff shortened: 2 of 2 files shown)


rcl/src/rcl/node_options.c
```diff
@@ -57,6 +57,8 @@ rcl_node_options_copy(
   if (NULL != options->arguments.impl) {
     rcl_ret_t ret = rcl_arguments_copy(&(options->arguments), &(options_out->arguments));
     return ret;
+  } else {
+    options_out->arguments.impl = NULL;
```
brawner, line 61: Looking at this in more depth, we may actually want to just return an error if arguments.impl is not NULL. To avoid memory leaks, `rcl_arguments_copy` returns an error if options_out->arguments.impl is not NULL before copying to it. If `impl` had been allocated before setting it to NULL here, there would be a similar memory leak.

Blast545:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11010/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6345/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8969/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10936/)

rcl/src/rcl/node_options.c
Blast545: You mean checking arguments.impl of the options_out?

brawner: Yah, just like rcl_arguments_copy does here:
https://github.com/ros2/rcl/blob/master/rcl/src/rcl/arguments.c#L791

Blast545: I agree with you about the memory leak, I'll erase this check and add a different check at the beginning of the function to address this.

rcl/test/rcl/test_node.cpp
```diff
@@ -751,6 +751,5 @@ TEST_F(CLASSNAME(TestNodeFixture, RMW_IMPLEMENTATION), test_rcl_node_names) {
 TEST_F(CLASSNAME(TestNodeFixture, RMW_IMPLEMENTATION), test_rcl_node_options_fail) {
   rcl_node_options_t not_ini_options;
```
brawner, line 752: Since `not_ini_options` is uninitialized there is a possibility that the impl pointer could be 0 in some runs, so you might be better off explicitly setting it to a non-null value for this test.

rcl/src/rcl/node_options.c
```diff
@@ -50,6 +50,10 @@ rcl_node_options_copy(
     RCL_SET_ERROR_MSG("Attempted to copy options into itself");
     return RCL_RET_INVALID_ARGUMENT;
   }
+  if (NULL != options_out->arguments.impl) {
```
fujitatomoya, line 53: Why do we need this? the same check will be done via rcl_arguments_copy. If we do not want to update any contents for options_out, we could move rcl_arguments_copy at the 1st.

brawner: `rcl_arguments_copy` requires that the source impl is non-null. This helps with the case when the impl of the source argument is null.

The original motivation was that an uninitialized `rcl_node_options_t` when used as a destination for `rcl_node_options_copy` might cause a segfault in `rcl_node_options_fini` because options_out->arguments.impl was never set to NULL. The current behavior doesn't require the `arguments.impl` of the source in `rcl_node_options_copy` to be non-null. But if it is NULL, then the `arguments.impl` is not touched and it will remain an unitialized pointer.

This check just helps to ensure that the output node_options_t is valid when this function is called.

fujitatomoya: understood.

rcl/test/rcl/test_node.cpp
hidmic: Precisely this @Blast545, not a good idea to use an uninitialized variable.

brawner: I think this is why windows failed. It probably had not_ini_options.impl as NULL, either randomly or intentionally.

Blast545: @hidmic I agree that's a bad idea to leave `not_ini_options` uninitialized before calling the copy method. Just checked the docblock in the header and already says that the destination has to be initialized, so I can reformat the test.

@brawner @hidmic However, this PR is catching a failure for the "not init" scenario, which means that if the variable is init this check won't be catched for Line Coverage. Any Ideas how should I proceed?

brawner: I think all you need to do is assign a pointer to not_ini_options.impl to catch the expected failure on copy.

Blast545:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11213/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6485/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9173/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11130/)

Blast545:
@ros-pull-request-builder

ivanpauno:

PR #673, opened 2020-06-05
Add message lost event

ivanpauno:
Implements https://github.com/ros2/rmw/issues/226.
See https://github.com/ros2/rmw/pull/232 for related changes in RMW.


ivanpauno:
For some reason I didn't understand yet, the written test doesn't generate a message lost event.

PR #672, opened 2020-06-04
print RCL_LOCALHOST_ENV_VAR if error happens via rcutils_get_env.

fujitatomoya:
fix only for error message.

Signed-off-by: Tomoya.Fujita <Tomoya.Fujita@sony.com>

PR #675, opened 2020-06-08
[Foxy] print RCL_LOCALHOST_ENV_VAR if error happens via rcutils_get_env. (#672)

ivanpauno:
Backports #672 to Foxy.


dirk-thomas:
This should be `--ff-only` instead to avoid diverging `foxy` from `master` while we can.

ivanpauno:
> This should be --ff-only instead to avoid diverging foxy from master while we can.

I will do a merge fast-forward of this PR, the commit hash is exactly the same that the one on master. I opened this, so I get a previous approval of doing the backport.

DCO will fail, as if I sign the commit it will change the hash (though, I don't think that cherry-picked commits need to be re-signed).

ivanpauno:
@jacobperron the PR checker doesn't seem to be working correctly.

jacobperron:

jacobperron:
I don't know what's up the the PR job. Could be an issue with Jenkins. Other than this PR, we haven't had any PR jobs triggered within the last 20 hours.

dirk-thomas:
@ros-pull-request-builder retest this please (after wiping the workspace)

PR #682, opened 2020-06-12
Improve enclave validation test coverage.

hidmic:

### DIFF
--- a/rcl/test/rcl/test_validate_enclave_name.cpp
+++ b/rcl/test/rcl/test_validate_enclave_name.cpp
@@ -64,7 +64,7 @@ TEST(TestValidateEnclaveName, test_validation_string) {
     {"/foo/$", RCL_ENCLAVE_NAME_INVALID_CONTAINS_UNALLOWED_CHARACTERS, 5},
     {"/bar#", RCL_ENCLAVE_NAME_INVALID_CONTAINS_UNALLOWED_CHARACTERS, 4},
     {"/foo//bar", RCL_ENCLAVE_NAME_INVALID_CONTAINS_REPEATED_FORWARD_SLASH, 5},
-    {"/1bar", RCL_ENCLAVE_NAME_INVALID_NAME_TOKEN_STARTS_WITH_NUMBER, 1}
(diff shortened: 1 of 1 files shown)


rcl/test/rcl/test_validate_enclave_name.cpp
```diff
@@ -82,4 +85,6 @@ TEST(TestValidateEnclaveName, test_validation_string) {
       "Enclave '" << enclave << "' failed with '" << validation_result << "'.";
     EXPECT_NE(nullptr, rcl_enclave_name_validation_result_string(validation_result)) << enclave;
   }
+  EXPECT_NE(nullptr, rcl_enclave_name_validation_result_string(-1));  // invalid result
```
brawner, line 88: What's the expected error string? Can you EXPECT_STREQ to that?

hidmic: I could. But since the rest of the test doesn't and it's of no consequence to behavior, it's not clear it adds any value.

brawner: You're expecting an error. The value is that you're testing that the correct error is caught.

hidmic: I can't say I agree with re-stating an implementation defined constant in the test, but I don't feel that strongly about it either. See https://github.com/ros2/rcl/pull/682/commits/38103ef9d632b2bc6f7d75cc997d39b3c513b2d5.

brawner: Is it possible that these tests are linked against a different implementation? If so, then I would agree with you.

hidmic: Not really, constants are hard-coded in `rcl`.

hidmic:
CI up to `rcl`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11159/)

brawner:
Per the QD, all PRs need to be tested on all four platforms. Even if it's a simple change like this, it's helpful to prevent new warnings. MSVS, Clang and GCC warn on different things sometimes

PR #683, opened 2020-06-12
Improve wait sets test coverage.

hidmic:

### DIFF
--- a/rcl/test/rcl/test_wait.cpp
+++ b/rcl/test/rcl/test_wait.cpp
@@ -29,6 +29,8 @@
 
 #include "rcutils/logging_macros.h"
 
+#include "./allocator_testing_utils.h"
+
 #ifdef RMW_IMPLEMENTATION
 # define CLASSNAME_(NAME, SUFFIX) NAME ## __ ## SUFFIX
 # define CLASSNAME(NAME, SUFFIX) CLASSNAME_(NAME, SUFFIX)
@@ -36,7 +38,11 @@
 # define CLASSNAME(NAME, SUFFIX) NAME
 #endif
 
+#ifndef _WIN32
 #define TOLERANCE RCL_MS_TO_NS(6)
+#else
+#define TOLERANCE RCL_MS_TO_NS(15)
+#endif
 
 class CLASSNAME (WaitSetTestFixture, RMW_IMPLEMENTATION) : public ::testing::Test
 {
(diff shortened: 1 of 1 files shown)


rcl/test/rcl/test_wait.cpp
```diff
@@ -257,6 +282,50 @@ TEST_F(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), zero_timeout_triggered
   EXPECT_LE(diff, TOLERANCE);
 }
 
+// Test rcl_wait with a negative timeout value (blocking ferver) and an overrun timer
```
brawner, line 285:
```suggestion
// Test rcl_wait with a negative timeout value (blocking forever) and an overrun timer
```

rcl/test/rcl/test_wait.cpp
```diff
@@ -257,6 +282,50 @@ TEST_F(CLASSNAME(WaitSetTestFixture, RMW_IMPLEMENTATION), zero_timeout_triggered
...
+  EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+  {
+    ret = rcl_timer_fini(&timer);
+    EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  });
+  ret = rcl_wait_set_add_timer(&wait_set, &timer, NULL);
+  EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+
+  // Time spent during wait should be negligible.
```
brawner, line 318: If this fails, and `rcl_wait` doesn't return will the test just hang?

Also, I can tell you've already figured this out, but timer based tests are notoriously unreliable on our Windows build farm, and it would be good to have as few of them as possible. It's not clear to me that TOLERANCE is a requirement of whether this test passes or succeeds. With the efforts of performance testing underway, I think measurements of how long rcl_wait takes to return should be put there.

What's the behavior you're trying to assert on this test?

hidmic: > If this fails, and rcl_wait doesn't return will the test just hang?

If wait sets' implementation misbehaves you mean? It may hang, though at that point it's not clear one can trust the given timeout will be honored either. At any rate, the CTest timeout will eventually kick in.

> Also, I can tell you've already figured this out, but timer based tests are notoriously unreliable on our Windows build farm, and it would be good to have as few of them as possible.

Indeed. Not entirely convinced ignoring them is a good idea though.

> It's not clear to me that TOLERANCE is a requirement of whether this test passes or succeeds.

The test needs to assert that `rmw_wait()` didn't block, or, actually, that it was requested to not block. Without mocks, we can only resort to this indirect technique.

> What's the behavior you're trying to assert on this test?

That [this snippet](https://github.com/ros2/rcl/blob/7723308e7c7f7515590bd709525fd840b4aed254/rcl/src/rcl/wait.c#L587) has the intended effect.

rcl/test/rcl/test_wait.cpp
hidmic: See https://github.com/ros2/rcl/pull/683/commits/e7ce68a0cdf9beb11a5e2faee8ba8ff4741dadc4.

rcl/test/rcl/test_wait.cpp
brawner: What sort of event does wait_set wait on here? Can you just ASSERT that the event didn't finish before the function returned? That is, make the event something long but finite, say 1-10 seconds so that wait will return in either case. But it will assert success if the event didn't finish, and failure if it did?

hidmic: Hmm, not sure I follow. This test is covering the situation in which a timer is way past its deadline by the time you wait on it for an X amount of time. In which case the function should realize, adjust its internal timeout and return almost immediately. That's what the test is checking. This [other test](https://github.com/ros2/rcl/blob/7723308e7c7f7515590bd709525fd840b4aed254/rcl/test/rcl/test_wait.cpp#L130) does the same thing for a timer that is on-time.

The scenario you describe is being (indirectly) covered by [this test](https://github.com/ros2/rcl/blob/7723308e7c7f7515590bd709525fd840b4aed254/rcl/test/rcl/test_timer.cpp#L117).

> If this fails, and rcl_wait doesn't return will the test just hang?

We can make that timeout finite. We cannot change the assertion itself though.

brawner: Ok, I think I understand better. I still think having a tolerance that is so tight it has to be different on different platforms sort of shows that the final EXPECT_LE is not cleanly deciding which behavior was followed in the code under test. Can you make timeout longer like 1-10 seconds so that it is very obvious whether it returned immediately or not?

brawner: And make TOLERANCE correspondingly larger

hidmic: Hmm, I'm not sure I follow you here. Which timeout do you mean? The timer period is 0, the wait timeout is -1 (block forever).

hidmic:
CI up to `rcl`:
* Linux (w/ coverage) [Build Status](http://ci.ros2.org/job/ci_linux/11196/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6473/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9157/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11114/)

brawner:
This test has been flakey on Windows, checking repeated
[Build Status](https://ci.ros2.org/job/ci_windows/11119/)

hidmic:
Seems we're in the clear.

PR #685, opened 2020-06-12
Improve rcl clock test coverage.

hidmic:

### DIFF
--- a/rcl/test/rcl/test_time.cpp
+++ b/rcl/test/rcl/test_time.cpp
@@ -24,6 +24,8 @@
 #include "rcl/error_handling.h"
 #include "rcl/time.h"
 
+#include "./allocator_testing_utils.h"
+
 #ifdef RMW_IMPLEMENTATION
 # define CLASSNAME_(NAME, SUFFIX) NAME ## __ ## SUFFIX
 # define CLASSNAME(NAME, SUFFIX) CLASSNAME_(NAME, SUFFIX)
@@ -317,23 +319,6 @@ TEST(CLASSNAME(rcl_time, RMW_IMPLEMENTATION), specific_clock_instantiation) {
 }
 
 TEST(CLASSNAME(rcl_time, RMW_IMPLEMENTATION), rcl_time_difference) {
-  rcl_allocator_t allocator = rcl_get_default_allocator();
-  auto * ros_clock =
-    static_cast<rcl_clock_t *>(allocator.allocate(sizeof(rcl_clock_t), allocator.state));
-  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
-  {
-    allocator.deallocate(ros_clock, allocator.state);
-  });
-  rcl_ret_t retval = rcl_ros_clock_init(ros_clock, &allocator);
-  ASSERT_EQ(retval, RCL_RET_OK) << rcl_get_error_string().str;
-  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
-  {
-    EXPECT_EQ(RCL_RET_OK, rcl_ros_clock_fini(ros_clock)) << rcl_get_error_string().str;
-  });
-  EXPECT_TRUE(ros_clock != nullptr);
-  EXPECT_TRUE(ros_clock->data != nullptr);
-  EXPECT_EQ(ros_clock->type, RCL_ROS_TIME);
-
   rcl_ret_t ret;
   rcl_time_point_t a, b;
 
@@ -354,23 +339,10 @@ TEST(CLASSNAME(rcl_time, RMW_IMPLEMENTATION), rcl_time_difference) {
 
   b.clock_type = RCL_SYSTEM_TIME;
   EXPECT_EQ(rcl_difference_times(&a, &b, &d), RCL_RET_ERROR) << rcl_get_error_string().str;
+  rcl_reset_error();
 }
 
 TEST(CLASSNAME(rcl_time, RMW_IMPLEMENTATION), rcl_time_difference_signed) {
-  rcl_allocator_t allocator = rcl_get_default_allocator();
-  auto * ros_clock =
-    static_cast<rcl_clock_t *>(allocator.allocate(sizeof(rcl_clock_t), allocator.state));
-  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
-  {
-    allocator.deallocate(ros_clock, allocator.state);
-  });
-  rcl_ret_t retval = rcl_ros_clock_init(ros_clock, &allocator);
-  ASSERT_EQ(retval, RCL_RET_OK) << rcl_get_error_string().str;
-  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
-  {
-    EXPECT_EQ(RCL_RET_OK, rcl_ros_clock_fini(ros_clock)) << rcl_get_error_string().str;
-  });
-
   rcl_time_point_t a, b;
   a.nanoseconds = RCL_S_TO_NS(0LL) + 0LL;
   b.nanoseconds = RCL_S_TO_NS(10LL) + 0LL;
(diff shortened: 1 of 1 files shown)


rcl/src/rcl/time.c
```diff
@@ -366,6 +363,10 @@ rcl_set_ros_time_override(
   }
   rcl_time_jump_t time_jump;
   rcl_ros_clock_storage_t * storage = (rcl_ros_clock_storage_t *)clock->data;
+  if (!storage) {
```
brawner, line 366: Could this if statement just be a `RCL_CHECK_ARGUMENT_FOR_NULL` at the top of the function?

rcl/src/rcl/time.c
```diff
@@ -453,19 +454,18 @@ rcl_clock_remove_jump_callback(
   }
 
   // Shrink size of the callback array
-  if (clock->num_jump_callbacks == 1) {
+  if (clock->num_jump_callbacks-- == 1) {
```
brawner, line 457: I think this change may need to go into a separate PR, based on the issue below.

rcl/src/rcl/time.c
```diff
@@ -453,19 +454,18 @@ rcl_clock_remove_jump_callback(
...
     clock->allocator.deallocate(clock->jump_callbacks, clock->allocator.state);
     clock->jump_callbacks = NULL;
   } else {
     rcl_jump_callback_info_t * callbacks = clock->allocator.reallocate(
-      clock->jump_callbacks, sizeof(rcl_jump_callback_info_t) * (clock->num_jump_callbacks - 1),
+      clock->jump_callbacks, sizeof(rcl_jump_callback_info_t) * clock->num_jump_callbacks,
```
brawner, line 462: I'm having a hard time understanding what is happening here. At this point `clock->num_jump_callbacks` is 0. So realloc(ptr, 0) is happening. Why not just free the array? 

It looks like realloc of size 0 is now undefined behavior. I think this `if` statement could be cleaned up.

http://www.open-std.org/jtc1/sc22/wg14/www/docs/summary.htm#dr_400

rcl/test/rcl/test_time.cpp
```diff
@@ -533,21 +499,15 @@ TEST(CLASSNAME(rcl_time, RMW_IMPLEMENTATION), rcl_time_fail_set_jump_callbacks)
...
-  auto * ros_clock =
-    static_cast<rcl_clock_t *>(allocator.allocate(sizeof(rcl_clock_t), allocator.state));
-  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
-  {
-    allocator.deallocate(ros_clock, allocator.state);
-  });
-  rcl_ret_t retval = rcl_ros_clock_init(ros_clock, &allocator);
-  ASSERT_EQ(retval, RCL_RET_OK) << rcl_get_error_string().str;
+  rcl_ret_t ret = rcl_ros_clock_init(&ros_clock, &allocator);
+  ASSERT_EQ(ret, RCL_RET_OK) << rcl_get_error_string().str;
```
brawner, line 505: ros_clock doesn't need to be cleaned up if ret is not OK?

rcl/test/rcl/test_time.cpp
```diff
@@ -584,33 +544,27 @@ TEST(CLASSNAME(rcl_time, RMW_IMPLEMENTATION), rcl_time_forward_jump_callbacks) {
...
```
brawner, line 563: Same question.

rcl/test/rcl/test_time.cpp
```diff
@@ -681,35 +630,30 @@ TEST(CLASSNAME(rcl_time, RMW_IMPLEMENTATION), rcl_clock_add_jump_callback) {
   rcl_jump_callback_t cb = reinterpret_cast<rcl_jump_callback_t>(0xBEEF);
   void * user_data = reinterpret_cast<void *>(0xCAFE);
 
-  EXPECT_EQ(RCL_RET_INVALID_ARGUMENT, rcl_clock_add_jump_callback(clock, threshold, NULL, NULL));
+  EXPECT_EQ(RCL_RET_INVALID_ARGUMENT, rcl_clock_add_jump_callback(&clock, threshold, NULL, NULL));
```
brawner, line 633: To ensure that the invalid_argument refers to a missing jump_callback, can you set `user_data` appropriately here.

rcl/test/rcl/test_time.cpp
```diff
@@ -722,61 +666,80 @@ TEST(CLASSNAME(rcl_time, RMW_IMPLEMENTATION), rcl_clock_remove_jump_callback) {
   void * user_data3 = reinterpret_cast<void *>(0xBEAD);
   void * user_data4 = reinterpret_cast<void *>(0xDEED);
 
-  EXPECT_EQ(RCL_RET_INVALID_ARGUMENT, rcl_clock_remove_jump_callback(clock, NULL, NULL));
+  EXPECT_EQ(RCL_RET_INVALID_ARGUMENT, rcl_clock_remove_jump_callback(&clock, NULL, NULL));
```
brawner, line 669: Set user_data here.

rcl/src/rcl/time.c
brawner: By undefined, I guess it's implementation defined. So this might not be portable.

rcl/src/rcl/time.c
hidmic: It could go on a separate PR, but it goes hand in hand with the test that prevents a regression.

rcl/src/rcl/time.c
hidmic: > At this point `clock->num_jump_callbacks` is `0`

Hmm for `clock->num_jump_callbacks` to be `0` here, it would've had to be `1` when the condition was evaluated, in which case it would've executed the first branch. Note that we're using postfix decrement, not prefix. I can change the condition if it's confusing though.

rcl/test/rcl/test_time.cpp
hidmic: It doesn't. And I _think_ we do strive for that to be the case for every `rcl` "object" (and rightly so).

rcl/test/rcl/test_time.cpp
hidmic: Same as above.

rcl/src/rcl/time.c
brawner: You're right, I was confused when looking at the diff of this logic. The if clause is executed if num_jump_callbacks is 1 when entering the loop. It will never be 0 because this function will have returned before then. The else clause is evaluated when it is strictly greater than 1. Thus reallocate will be called with a size that is never 0.

The only other difference that might be important between the diffs is that if reallocate fails, `num_jump_callbacks` no longer matches the size of `jump_callbacks`. Should num_jump_callbacks be reincremented in that case?

rcl/src/rcl/time.c
brawner: Is this fixing a bug and should it be backported, or is it just some cleanup?

rcl/src/rcl/time.c
hidmic: > Should `num_jump_callbacks` be reincremented in that case?

I thought about it, but the [preceding loop](https://github.com/ros2/rcl/blob/7723308e7c7f7515590bd709525fd840b4aed254/rcl/src/rcl/time.c#L445) probably removed that callback by then. 

Also, a failed reallocation means the function failed to shrink storage, not that the callback itself wasn't removed successfully. A future deallocation wouldn't be affected and a future reallocation would be a no-op.

rcl/src/rcl/time.c
hidmic: Done in https://github.com/ros2/rcl/pull/685/commits/012c120bad0e56a8da83e53ec07d4b3a1c635b60.

rcl/src/rcl/time.c
hidmic: Hmm, that is a good point. I think it could be backported, yeah. It adds safety checks and ensure `rcl_remove_jump_callback()` fails a reallocation safely i.e. that you can continue despite a failed a reallocation.

rcl/test/rcl/test_time.cpp
hidmic: Good point. See https://github.com/ros2/rcl/pull/685/commits/f72b7e87dea794f10c749363b68733632fb084fd.

rcl/test/rcl/test_time.cpp
hidmic: Good point. See https://github.com/ros2/rcl/pull/685/commits/f72b7e87dea794f10c749363b68733632fb084fd.

hidmic:
CI up to `rcl`:
* Linux (w/ coverage) [Build Status](http://ci.ros2.org/job/ci_linux/11184/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6463/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9135/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11098/)

hidmic:
> I agree that the time.c changes should be considered for backporting.

Alright then. I'll rebase-merge instead of squashing so we can cherry pick and backport that commit later on. Does that sound good @brawner @Blast545 ?

brawner:
I'm not a veteran with this code base, so if others are ok with rebase merging then sure. But I would lean toward the side of just separating them into different PRs. I can't exactly tell, but the changes to time.c don's seem related to the test file. By squash merging in separate PRs, the commit links to the PR directly, which has the discussion and review, which may be helpful for anyone looking through the source code.

hidmic:
Tests here would fail without [this change](https://github.com/ros2/rcl/pull/685/files#diff-019482352ce3358b201f3806b28d0377R175) and [these changes](https://github.com/ros2/rcl/pull/685/files#diff-019482352ce3358b201f3806b28d0377R447-R460). The rest are mere cleanup. 

I can nonetheless split them up and make this PR depend on that PR. Discussion already took place here, so that PR would ultimately refer to this PR. But I don't mind either way.

hidmic:
A clean rebase, but I'll re-run the pickiest CI (Windows' :) ) just to be safe.

hidmic:
Alright, going in. Thanks for the reviews guys!

PR #689, opened 2020-06-19
Keep domain id if ROS_DOMAIN_ID is invalid.

hidmic:
Precisely what the title says. This pull request prevents `rcl_get_default_domain_id()` from forcing `0` for a domain id on failure cases. Found while working on https://github.com/ros2/rmw_implementation/pull/106.

### DIFF
--- a/rcl/src/rcl/domain_id.c
+++ b/rcl/src/rcl/domain_id.c
@@ -44,7 +44,7 @@ rcl_get_default_domain_id(size_t * domain_id)
   if (ros_domain_id && strcmp(ros_domain_id, "") != 0) {
     char * end = NULL;
     unsigned long number = strtoul(ros_domain_id, &end, 0);  // NOLINT(runtime/int)
-    if (number == 0UL && *end != NULL) {
+    if (number == 0UL && *end != '\0') {
       RCL_SET_ERROR_MSG("ROS_DOMAIN_ID is not an integral number");
       return RCL_RET_ERROR;
     }
--- a/rcl/src/rcl/domain_id.c
+++ b/rcl/src/rcl/domain_id.c
@@ -42,12 +42,13 @@ rcl_get_default_domain_id(size_t * domain_id)
     return RCL_RET_ERROR;
   }
   if (ros_domain_id && strcmp(ros_domain_id, "") != 0) {
-    unsigned long number = strtoul(ros_domain_id, NULL, 0);  // NOLINT(runtime/int)
-    if (number == 0UL && strspn(ros_domain_id, "0") != strlen(ros_domain_id)) {
+    char * end = NULL;
+    unsigned long number = strtoul(ros_domain_id, &end, 0);  // NOLINT(runtime/int)
+    if (number == 0UL && *end != NULL) {
       RCL_SET_ERROR_MSG("ROS_DOMAIN_ID is not an integral number");
       return RCL_RET_ERROR;
     }
-    if (number == ULONG_MAX && errno == ERANGE) {
+    if ((number == ULONG_MAX && errno == ERANGE) || number > SIZE_MAX) {
       RCL_SET_ERROR_MSG("ROS_DOMAIN_ID is out of range");
       return RCL_RET_ERROR;
     }
--- a/rcl/src/rcl/domain_id.c
+++ b/rcl/src/rcl/domain_id.c
@@ -14,6 +14,7 @@
 
 #include "rcl/domain_id.h"
 
+#include <errno.h>
 #include <limits.h>
 
 #include "rcutils/get_env.h"
@@ -40,10 +41,14 @@ rcl_get_default_domain_id(size_t * domain_id)
       get_env_error_str);
     return RCL_RET_ERROR;
   }
-  if (ros_domain_id) {
+  if (ros_domain_id && strcmp(ros_domain_id, "") != 0) {
     unsigned long number = strtoul(ros_domain_id, NULL, 0);  // NOLINT(runtime/int)
-    if (number == ULONG_MAX) {
-      RCL_SET_ERROR_MSG("failed to interpret ROS_DOMAIN_ID as integral number");
+    if (number == 0UL && strspn(ros_domain_id, "0") != strlen(ros_domain_id)) {
+      RCL_SET_ERROR_MSG("ROS_DOMAIN_ID is not an integral number");
+      return RCL_RET_ERROR;
+    }
+    if (number == ULONG_MAX && errno == ERANGE) {
+      RCL_SET_ERROR_MSG("ROS_DOMAIN_ID is out of range");
       return RCL_RET_ERROR;
     }
     *domain_id = (size_t)number;
--- a/rcl/test/rcl/test_domain_id.cpp
+++ b/rcl/test/rcl/test_domain_id.cpp
@@ -22,15 +22,31 @@
 
 TEST(TestGetDomainId, test_nominal) {
   ASSERT_TRUE(rcutils_set_env("ROS_DOMAIN_ID", "42"));
-  size_t domain_id = 0u;
+  size_t domain_id = RCL_DEFAULT_DOMAIN_ID;
   EXPECT_EQ(RCL_RET_OK, rcl_get_default_domain_id(&domain_id));
   EXPECT_EQ(42u, domain_id);
 
-  ASSERT_TRUE(rcutils_set_env("ROS_DOMAIN_ID", "998446744073709551615"));
-  domain_id = 0u;
+  ASSERT_TRUE(rcutils_set_env("ROS_DOMAIN_ID", ""));
+  domain_id = RCL_DEFAULT_DOMAIN_ID;
+  EXPECT_EQ(RCL_RET_OK, rcl_get_default_domain_id(&domain_id));
+  EXPECT_EQ(RCL_DEFAULT_DOMAIN_ID, domain_id);
+
+  ASSERT_TRUE(rcutils_set_env("ROS_DOMAIN_ID", "0000"));
+  domain_id = RCL_DEFAULT_DOMAIN_ID;
+  EXPECT_EQ(RCL_RET_OK, rcl_get_default_domain_id(&domain_id));
+  EXPECT_EQ(0u, domain_id);
+
+  ASSERT_TRUE(rcutils_set_env("ROS_DOMAIN_ID", "0   not really"));
+  domain_id = RCL_DEFAULT_DOMAIN_ID;
   EXPECT_EQ(RCL_RET_ERROR, rcl_get_default_domain_id(&domain_id));
   rcl_reset_error();
-  EXPECT_EQ(0u, domain_id);
+  EXPECT_EQ(RCL_DEFAULT_DOMAIN_ID, domain_id);
+
+  ASSERT_TRUE(rcutils_set_env("ROS_DOMAIN_ID", "998446744073709551615"));
+  domain_id = RCL_DEFAULT_DOMAIN_ID;
+  EXPECT_EQ(RCL_RET_ERROR, rcl_get_default_domain_id(&domain_id));
+  rcl_reset_error();
+  EXPECT_EQ(RCL_DEFAULT_DOMAIN_ID, domain_id);
 
   EXPECT_EQ(RCL_RET_INVALID_ARGUMENT, rcl_get_default_domain_id(nullptr));
 }


hidmic:
I'll run cross-platform CI once I've addressed the failure with `rmw_cyclonedds_cpp`.

rcl/src/rcl/domain_id.c
```diff
@@ -40,10 +41,14 @@ rcl_get_default_domain_id(size_t * domain_id)
       get_env_error_str);
     return RCL_RET_ERROR;
   }
-  if (ros_domain_id) {
+  if (ros_domain_id && strcmp(ros_domain_id, "") != 0) {
     unsigned long number = strtoul(ros_domain_id, NULL, 0);  // NOLINT(runtime/int)
-    if (number == ULONG_MAX) {
-      RCL_SET_ERROR_MSG("failed to interpret ROS_DOMAIN_ID as integral number");
+    if (number == 0UL && strspn(ros_domain_id, "0") != strlen(ros_domain_id)) {
```
ivanpauno, line 46: I didn't remember that `strtoul` can return `0` on error. That's pretty weird ...

ivanpauno:
@hidmic Consider backporting to Dashing, Eloquent and Foxy if applicable.

hidmic:
CI up to `rcl` and `test_rmw_implementation` against all RMW implementations:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11229/) (canary build)

rcl/src/rcl/domain_id.c
```diff
@@ -40,10 +41,14 @@ rcl_get_default_domain_id(size_t * domain_id)
       get_env_error_str);
     return RCL_RET_ERROR;
   }
-  if (ros_domain_id) {
+  if (ros_domain_id && strcmp(ros_domain_id, "") != 0) {
```
ivanpauno, line 44: can a `else {*domain_id = 0u;}` be added instead of all the PRs addressing rmw implementations?

wjwwood: What if they don't want to use `0` as the default?

rcl/src/rcl/domain_id.c
meyerj: [`strtoul()`](https://en.cppreference.com/w/cpp/string/byte/strtoul) returns `ULONG_MAX` in case of an overflow (depending on the size of `unsigned long` on the specific platform). But an invalid character is not an overflow and `strtoul()` just stops parsing. You would need to add a `char ** str_end` argument instead of `NULL` as the second argument and check whether `str_end` points to the terminating null character.

See also my comment in https://github.com/ros2/rclcpp/pull/1188#discussion_r443083058. `rclcpp` obviously has its own implementation of `rcl_get_default_domain_id()` and does not use this one.

rcl/src/rcl/domain_id.c
ivanpauno: > What if they don't want to use 0 as the default?

We used to force that in the past, and a lot of the rmw implementations aren't handling the default correctly (thus the other PRs are needed).
I don't mind if we want  to change this.

rcl/src/rcl/domain_id.c
ivanpauno: > You would need to add a char ** str_end argument instead of NULL as the second argument and check whether str_end points to the terminating null character.

:+1:

> See also my comment in ros2/rclcpp#1188 (comment). rclcpp obviously has its own implementation of rcl_get_default_domain_id() and does not use this one.

Originally `rcl_get_default_domain_id` function didn't exist, thus it wasn't used in some places.

> See also my comment in ros2/rclcpp#1188 (comment). rclcpp obviously has its own implementation of rcl_get_default_domain_id() and does not use this one.

Actually, `domain_id` in `NodeOptions` is completely ignored now.
I have to delete that field in `rcl` and `rmw` layers (see https://github.com/ros2/rclcpp/issues/910). Luckily it was never exposed to the user in the upper layers.

rcl/src/rcl/domain_id.c
ivanpauno: After this PR, [rcl_node_get_domain_id](https://github.com/ros2/rcl/blob/b9cfc243bf018091b5c5bcb5585d19591daf4ae6/rcl/include/rcl/node.h#L371) started to return 0 when the environment variable is unset.

The purpose of that API was to return the actual domain id that the node was using (see [here](https://github.com/ros2/rcl/blob/acd167e1734c2c15196fb92f76bff1559260eb1a/rcl/src/rcl/node.c#L488)), but the "actual" domain id is now erroneous.

IMO, the behavior before this PR was correct, i.e. using 0 as the default domain id if the environment variable is not set.
I really don't see any value in letting the rmw implementation define the actual domain id, and adding a layer in rmw to query the default domain id actually used seems unnecessary.

@hidmic @wjwwood what do you think?

hidmic: > IMO, the behavior before this PR was correct, i.e. using 0 as the default domain id if the environment variable is not set.

As I've stated elsewhere ([here](https://github.com/ros2/rmw_connext/pull/427)?) I disagree.

> I really don't see any value in letting the rmw implementation define the actual domain id, and adding a layer in rmw to query the default domain id actually used seems unnecessary.

I wonder what's the use case for `rcl_get_node_domain_id`. 

> After this PR, rcl_node_get_domain_id started to return 0 when the environment variable is unset.

Hmm, how's that possible? I would expect it to be `RCL_DEFAULT_DOMAIN_ID` w/o further changes.

hidmic: > Hmm, how's that possible? I would expect it to be RCL_DEFAULT_DOMAIN_ID w/o further changes.

Ok, perhaps I misunderstood. `rcl_get_default_domain_id()` behaves as expected.

> I wonder what's the use case for rcl_get_node_domain_id.

If we really need it, yeah, it'll require new `rmw`  API. Having said that, nothing prevents `rcl` from forcing `0` though:

```c
  domain_id = node->impl->options.domain_id;
  if (RCL_DEFAULT_DOMAIN_ID == domain_id) {
    if (RCL_RET_OK != rcl_get_default_domain_id(&domain_id)) {
      goto fail;
    }
+   if (RCL_DEFAULT_DOMAIN_ID == domain_id) {
+     domain_id = 0u;
+   }
  }
  RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME, "Using domain ID of '%zu'", domain_id);
```

We've backported this to Foxy already, so that may be the only solution there.

ivanpauno: > I wonder what's the use case for rcl_get_node_domain_id.

I think it's good to be able to introspect the domain id, and if it returns SIZE_MAX when it's actually using 0 it might be confusing (I know of some people that were confused).
There was also a log saying `domain_id = -1` which was confusing (I have to double check from where that log came from, the log [here](https://github.com/ros2/rcl/blob/b9cfc243bf018091b5c5bcb5585d19591daf4ae6/rcl/src/rcl/node.c#L263) seem to have the correct format string)).

> Hmm, how's that possible? I would expect it to be RCL_DEFAULT_DOMAIN_ID w/o further changes.

Ups, edited. It started to return SIZE_MAX.

> As I've stated elsewhere (here?) I disagree.

I think you're referring to this:

> IMHO the issue is that both domain ID defaults are conflated. rcl has a default to decide whether it has to check the environment or not, and rmw has a default with a sentinel value. If no ID is passed to rcl and the environment does no provide one either, the rmw default should be honored.
> We can discuss the rmw default though. It could be 0, sure, if we think no rmw can benefit from knowing if that ID is the default and not a user-defined one that happens to have the same value.

There wasn't a RMW_DEFAULT_DOMAIN_ID in the past (see [blame](https://github.com/ros2/rmw/blame/master/rmw/include/rmw/domain_id.h#L19)), that was moved from rcl.
Maybe I should have duplicated the domain id in rcl init options instead of moving the definition of the default, as the default was ONLY checked in `rcl`. What is the value of a default in `rmw`?

> If we really need it, yeah, it'll require new rmw API. Having said that, nothing prevents rcl from forcing 0 though:

That patch looks ok.

hidmic: > There wasn't a RMW_DEFAULT_DOMAIN_ID in the past (see blame), that was moved from rcl.

Interesting. Fun chain of events then. 

> Maybe I should have duplicated the domain id in rcl init options instead of moving the definition of the default, as the default was ONLY checked in rcl. What is the value of a default in rmw?

Well, an `rmw` implementation is expected to provide a default domain ID via [`rmw_init_options_init()`](https://github.com/ros2/rmw/blob/4a1914a3b8c36a0bb59677cdbab9a21fb0da8ddd/rmw/include/rmw/init_options.h#L74). From an API perspective, the sole value I see with having a sentinel ID is to defer picking an actual default (allowing, say, an implementation to pick up a default from a network-wide pool when it is actually needed and not when options are initialized).

> IMHO the issue is that both domain ID defaults are conflated. rcl has a default to decide whether it has to check the environment or not, and rmw has a default with a sentinel value. If no ID is passed to rcl and the environment does no provide one either, the rmw default should be honored.

And that remains true. I see `rcl_init_options_t` using `rmw_init_options_t` domain ID [directly](https://github.com/ros2/rcl/blob/b9cfc243bf018091b5c5bcb5585d19591daf4ae6/rcl/src/rcl/init_options.c#L140). And then implicitly relies on it being `RMW_DEFAULT_DOMAIN_ID == RCL_DEFAULT_DOMAIN_ID` to know whether it has to check the environment or not.

There's absolutely no guarantee that `rmw_init_options_init()` will use `RMW_DEFAULT_DOMAIN_ID` as opposed to `42` or `5`. Which would then break environment checks.

hidmic: > That patch looks ok.

I think that:

- For Foxy, we don't have a choice. `rcl_get_default_domain_id()` has to force a 0 or `rcl_node_get_domain_id()` breaks as you said.
- For Galactic, I think we need `rmw` API **and** fix those assumptions about default domain IDs.

ivanpauno: What about restoring how it used to work in node options?

- Delete RMW_DEFAULT_DOMAIN_ID (or deprecate it).
- Add domain_id field to [rcl_init_options](https://github.com/ros2/rcl/blob/b9cfc243bf018091b5c5bcb5585d19591daf4ae6/rcl/src/rcl/init_options_impl.h#L28).
- The rcl API ([1](https://github.com/ros2/rcl/blob/b9cfc243bf018091b5c5bcb5585d19591daf4ae6/rcl/include/rcl/init_options.h#L172) and [2](https://github.com/ros2/rcl/blob/b9cfc243bf018091b5c5bcb5585d19591daf4ae6/rcl/include/rcl/init_options.h#L150)) should set and get the value in rcl, not directly the one in the rmw counterpart.
  - The default constructed rcl init options are set with `domain_id=RCL_DEFAULT_DOMAIN_ID`.
  - The default constructed rmw init options are set with `domain_id=0`.
  - If the user sets a domain id using [1](https://github.com/ros2/rcl/blob/b9cfc243bf018091b5c5bcb5585d19591daf4ae6/rcl/include/rcl/init_options.h#L172) and [2](https://github.com/ros2/rcl/blob/b9cfc243bf018091b5c5bcb5585d19591daf4ae6/rcl/include/rcl/init_options.h#L150), `rcl_init_options_t::domain_id` will be modified (and not its rmw counterpart).
  - When `rcl_init` is called, if `rcl_init_options_t::domain_id` is `RCL_DEFAULT_DOMAIN_ID` the environment is check and `rmw_init_options_t::domain_id` is set with the correct value. Otherwise, `rmw_init_options_t::domain_id` is a copy of `rcl_init_options_t::domain_id`.
  - `rcl_init_options_get_domain_id` will always return `rcl_init_options_t::domain_id` (this can return the unresolved default domain id).
  - the [node actual domain id](https://github.com/ros2/rcl/blob/b9cfc243bf018091b5c5bcb5585d19591daf4ae6/rcl/src/rcl/node.c#L59) is a copy of the `rmw_init_options::domain_id` that the node used (i.e.: it will always be a resolved number, and not SIZE_MAX).

That will match fairly well how this used to work before the node/participant mapping change (I think we overlooked this details in the middle of some pretty big PRs).

> Well, an rmw implementation is expected to provide a default domain ID via rmw_init_options_init(). From an API perspective, the sole value I see with having a sentinel ID is to defer picking an actual default (allowing, say, an implementation to pick up a default from a network-wide pool when it is actually needed and not when options are initialized).

Sure, but do we need that feature? I think this changes were a bit accidental and that the original purpose of the default domain id wasn't this.
I don't see a strong use case for letting the rmw implementation decide the domain id.

hidmic: > Sure, but do we need that feature?

That's a fair question that the @ros2/team has to answer. 

Personally, when I build a library, since I cannot predict (and likely won't see half of) all use cases, I try to keep it as general as possible unless there's a good reason not to. As opposed to when I build a particular system with a concrete known purpose for which I wouldn't add feature I know I don't need. But that's me.

ivanpauno: > Personally, when I build a library, since I cannot predict (and likely won't see half of) all use cases, I try to keep it as general as possible unless there's a good reason not to. As opposed to when I build a particular system with a concrete known purpose for which I wouldn't add a feature I know I don't need. But that's me.

Ok, we can add a `rmw_get_default_domain_id` and implement it in the three rmw implementations.
Then we can modify `rcl` to do something like:
```c
  domain_id = node->impl->options.domain_id;
  if (RCL_DEFAULT_DOMAIN_ID == domain_id) {
    if (RCL_RET_OK != rcl_get_default_domain_id(&domain_id)) {
      goto fail;
    }
+   if (RCL_DEFAULT_DOMAIN_ID == domain_id) {
+     domain_id = rmw_get_default_domain_id();
+   }
  }
  RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME, "Using domain ID of '%zu'", domain_id);
```

In Foxy we would still need to set it to `0`, as we probably don't want to backport new API unnecessarily.
Does that sound good?

hidmic: Yeap. For a quick and dirty fix forcing to `0` is OK. For the rest, don't trust me. Let's wait for others to comment.

wjwwood: For this point:

> There's absolutely no guarantee that rmw_init_options_init() will use RMW_DEFAULT_DOMAIN_ID as opposed to 42 or 5. Which would then break environment checks.

I think it must use `RMW_DEFAULT_DOMAIN_ID`. When the domain id is in the init options, that is just a default value, the user may override it. If the user wants the middleware to select the domain id, then `RMW_DEFAULT_DOMAIN_ID` should be used.

The purpose of `rmw_init_options_init()` was not to allow the middleware to specify a custom default domain id. Though it could have been that or still could be, though I don't think that makes sense... The user needs to be able to explicitly ask the middleware to pick a domain ID regardless of what is in the init options initially, and that is what the `RMW_DEFAULT_DOMAIN_ID` is about, from my perspective.

----

I think that we should remove the domain id from the node (node options, `actual_domain_id`, etc) and move it into the rmw context object. I think we should keep the `RMW_DEFAULT_DOMAIN_ID` sentinel value, and it can be set in the rmw init options (we could require middlewares initialize it to that value) and then it will always be resolved in the resulting rmw context object.

The reason for keeping the `RMW_DEFAULT_DOMAIN_ID`, in my opinion, is to let the domain id being used be configured in the middleware, maybe based on some middleware specific configuration file (similar to our `_DEFAULT` QoS settings we have) or as @hidmic said based on some dynamic heuristic like a pool of domain id's.

I do not think we need a function to query what the default domain ID would be if we initialized with `RMW_DEFAULT_DOMAIN_ID` (e.g. `rmw_get_default_domain_id()`), but I could be convinced otherwise. I think it would be sufficient to say you need to initialize the system in order to find out what the default is.

In `rcl`, if the environment variable `ROS_DOMAIN_ID` is set, then it should be honored, always. The only time the middleware would choose a domain id, is if the `ROS_DOMAIN_ID` is not set and the `RMW_DEFAULT_DOMAIN_ID` value is used, or if `ROS_DOMAIN_ID` is somehow set to the `RMW_DEFAULT_DOMAIN_ID` value.

I think `rcl_get_default_domain_id()` should be renamed probably, since it does not get the default, instead it gets the value from the environment variable if it is set otherwise it does nothing to the given `size_t *`.

ivanpauno: > I think that we should remove the domain id from the node (node options, actual_domain_id, etc) and move it into the rmw context object. I think we should keep the RMW_DEFAULT_DOMAIN_ID sentinel value, and it can be set in the rmw init options (we could require middlewares initialize it to that value) and then it will always be resolved in the resulting rmw context object.

:+1:

> The reason for keeping the RMW_DEFAULT_DOMAIN_ID, in my opinion, is to let the domain id being used be configured in the middleware, maybe based on some middleware specific configuration file (similar to our _DEFAULT QoS settings we have) or as @hidmic said based on some dynamic heuristic like a pool of domain id's.

:+1:

> I do not think we need a function to query what the default domain ID would be if we initialized with RMW_DEFAULT_DOMAIN_ID (e.g. rmw_get_default_domain_id()), but I could be convinced otherwise. I think it would be sufficient to say you need to initialize the system in order to find out what the default is.

We currently don't have any way to introspect the `domain_id` that the rmw implementation set, we would need either:
- rmw_context_get_domain_id(rmw_context_t * node)
- rmw_node_get_domain_id(rmw_node_t * node)

the first one makes more sense.
We can call that function after initing the `rmw_context` in `rcl`, and store it in the corresponding `rcl_context_t` structure.

Does that make sense?

wjwwood: Why do we need a function in rmw to get that field? Could it not just be a public member of the `rmw_context_t` object? I mean either would work, I'm just wondering if there's a reason to hide this member internally or not.

Otherwise, lgtm.

----

One lingering doubt in my mind, should we leave the domain id in the node just so middlewares that choose to do it that way can still map domain id to nodes rather than contexts? I'm not sure it makes sense to do that, but I'm just asking if anyone thinks we'll need that at some point?

ivanpauno: > Why do we need a function in rmw to get that field? Could it not just be a public member of the rmw_context_t object? I mean either would work, I'm just wondering if there's a reason to hide this member internally or not.

Yeah, a member variable also sounds good.

> One lingering doubt in my mind, should we leave the domain id in the node just so middlewares that choose to do it that way can still map domain id to nodes rather than contexts? I'm not sure it makes sense to do that, but I'm just asking if anyone thinks we'll need that at some point?

My mental idea is that you have to create two contexts if you need different domain ids, regardless if the implementation can have a different domain id for each node or not.
Multi domain id applications are strange (mainly domain bridges?), and support for setting the domain id with node granularity doesn't seem to add much value.

hidmic: It's settled then! 

> My mental idea is that you have to create two contexts if you need different domain ids, regardless if the implementation can have a different domain id for each node or not.
Multi domain id applications are strange (mainly domain bridges?), and support for setting the domain id with node granularity doesn't seem to add much value.

I agree. I had my doubts about this at some point in time, but I see now those were unfounded.

PR #686, opened 2020-06-15
Use RCL_RET_* codes only.

hidmic:
Precisely what the title says. Also, update documentation.

CI up to `rcl` for safety:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11160/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6450/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9107/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11082/)

PR #693, opened 2020-06-22
[foxy backport] Use RCL_RET_* codes only. (#686)

hidmic:


hidmic:

PR #668, opened 2020-06-01
Add tests for rcl package

Blast545:
As the title says, aiming to increase quality level of the package. This covers missing basic functionality not tested already and several bad arguments tests.

PR #694, opened 2020-06-22
[foxy backport] Keep domain id if ROS_DOMAIN_ID is invalid. (#689)

hidmic:
Including domain id tests (some coming from https://github.com/ros2/rcl/pull/668).

### DIFF
--- a/rcl/src/rcl/domain_id.c
+++ b/rcl/src/rcl/domain_id.c
@@ -14,6 +14,7 @@
 
 #include "rcl/domain_id.h"
 
+#include <errno.h>
 #include <limits.h>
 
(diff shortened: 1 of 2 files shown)


hidmic:
About a fraction of https://github.com/ros2/rcl/pull/668 coming in with this PR. That commit bundles a lot of tests. We can backport them all before this PR instead of bringing in only a fraction of it as it is done here. I don't have a strong opinion.

ivanpauno:
This change broke `rcl_node_get_domain_id` behavior as described in https://github.com/ros2/rcl/pull/689#discussion_r448978801.

Issue #687, opened 2020-06-16
rcl_wait() returns early when a timer awakes


hidmic:
## Bug report

**Required Info:**

- Operating System:
  - Windows 10
- Installation type:
  - From source
- Version or commit hash:
  - [master](https://github.com/ros2/rcl/tree/7723308e7c7f7515590bd709525fd840b4aed254)
- DDS implementation:
  - RTI Connext

#### Steps to reproduce issue

Run timer tests using `rmw_connext_cpp`. 

#### Expected behavior
Tests pass.

#### Actual behavior
Often, tests fail when `rcl_wait()` returns early when a timer awakes.

#### Additional information
For one, even though a clock is specified for timers, how [`rmw_wait()`](https://github.com/ros2/rmw/blob/f8d23d06c12b6b65c3a0a849ddc124c19aec6779/rmw/include/rmw/rmw.h#L1066-L1099) abides to the given timeout is not specified. This issue persists for both steady time and system time timers. 

This issue does not occur when using `rmw_fastrtps*_cpp` or  `rmw_cyclonedds_cpp`, as these return at the right time or (most often) past it.

Suspecting a bug in `rmw_connext_cpp`, I could confirm that the expected timeout is passed to RTI Connext's [`WaitSet::wait()`](https://community.rti.com/static/documentation/connext-dds/5.3.1/doc/api/connext_dds/api_cpp2/classdds_1_1core_1_1cond_1_1WaitSet.html#a31d822460851438c3173ba573f54e86e) API which nonetheless returns early for both steady time and system time clocks. However, I believe this is less of a bug in RTI Connext than a misuse of `rmw_wait()` for `rcl` timers' implementation, considering time resolution. The closest achievable timeout on a given OS is not necessarily greater than or equal than any one specified.

hidmic:
One potential workaround would be to relax [the definition of a ready timer](https://github.com/ros2/rcl/blob/7723308e7c7f7515590bd709525fd840b4aed254/rcl/src/rcl/timer.c#L300) using a platform specific time tolerance, and even make that configurable for a user to tune it.

hidmic:
Alternatively, since it's quite likely not every possible timer period can be achieved due to time resolution constrains, the expectation on `rcl` timers' period could be relaxed to be just an average quantity over some time window.

hidmic:
CC @ivanpauno @wjwwood @peterpena

clalancette:
[...]

PR #707, opened 2020-07-03
Make rcl timers track jitter and regulate period.

hidmic:
First stab at #687.

### DIFF
--- a/rcl/src/rcl/timer.c
+++ b/rcl/src/rcl/timer.c
@@ -40,6 +40,10 @@ typedef struct rcl_timer_impl_t
   atomic_uintptr_t callback;
   // This is a duration in nanoseconds.
   atomic_uint_least64_t period;
+  // This is a duration in nanoseconds.
+  atomic_int_least64_t period_correction;
+  // This is a duration in nanoseconds.
+  atomic_int_least64_t period_jitter;
   // This is a time in nanoseconds since an unspecified time.
   atomic_int_least64_t last_call_time;
   // This is a time in nanoseconds since an unspecified time.
@@ -48,6 +52,8 @@ typedef struct rcl_timer_impl_t
   atomic_int_least64_t time_credit;
   // A flag which indicates if the timer is canceled.
   atomic_bool canceled;
+  // The user supplied options.
+  rcl_timer_options_t options;
   // The user supplied allocator.
   rcl_allocator_t allocator;
 } rcl_timer_impl_t;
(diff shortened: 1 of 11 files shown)


rcl/src/rcl/timer.c
```diff
@@ -254,17 +308,36 @@ rcl_timer_call(rcl_timer_t * timer)
...
+
   rcl_timer_callback_t typed_callback =
     (rcl_timer_callback_t)rcutils_atomic_load_uintptr_t(&timer->impl->callback);
 
   int64_t next_call_time = rcutils_atomic_load_int64_t(&timer->impl->next_call_time);
-  int64_t period = rcutils_atomic_load_uint64_t(&timer->impl->period);
+  const int64_t expected_time_since_last_call = next_call_time - previous_call_time;
+
+  const int64_t period = rcutils_atomic_load_uint64_t(&timer->impl->period);
+  int64_t period_correction = rcutils_atomic_load_int64_t(&timer->impl->period_correction);
```
ivanpauno, line 322: what's the point of the atomics here? (timer->impl->period_correction and timer->impl->period_jitter)

rcl/src/rcl/timer.c
```diff
@@ -348,6 +422,8 @@ rcl_timer_exchange_period(const rcl_timer_t * timer, int64_t new_period, int64_t
   RCL_CHECK_ARGUMENT_FOR_NULL(timer, RCL_RET_INVALID_ARGUMENT);
   RCL_CHECK_ARGUMENT_FOR_NULL(old_period, RCL_RET_INVALID_ARGUMENT);
   *old_period = rcutils_atomic_exchange_uint64_t(&timer->impl->period, new_period);
+  rcutils_atomic_store(
+    &timer->impl->period_correction, timer->impl->options.initial_period_correction);
```
ivanpauno, line 426: should initial period jitter be reset here?

rcl/test/rcl/test_context.cpp
```diff
@@ -128,4 +128,26 @@ TEST_F(CLASSNAME(TestContextFixture, RMW_IMPLEMENTATION), nominal) {
...
+
+  ret = rcl_context_fini(&context);
+  EXPECT_EQ(ret, RCL_RET_INVALID_ARGUMENT);
+  rcl_reset_error();
+
+  ret = rcl_shutdown(&context);
+  EXPECT_EQ(ret, RCL_RET_OK);
+
+  ret = rcl_context_fini(&context);
+  EXPECT_EQ(ret, RCL_RET_OK);
```
ivanpauno, line 152: are changes in this file related?

rcl/test/rcl/test_events.cpp
```diff
@@ -696,6 +696,40 @@ TEST_P(TestEventFixture, test_pubsub_incompatible_qos)
...
+    &publisher_event,
+    &publisher,
+    unknown_pub_type);
+  EXPECT_EQ(ret, RCL_RET_INVALID_ARGUMENT);
+
+  subscription_event = rcl_get_zero_initialized_event();
+  ret = rcl_subscription_event_init(
+    &subscription_event,
+    &subscription,
+    unknown_sub_type);
```
ivanpauno, line 719: same about changes in this file?

hidmic:
> how are we going to validate that the changes here make things better?

That's the other half of this work. I have yet to put together a benchmark we can run on all platforms to analyze timer frequency and jitter.

rcl/src/rcl/timer.c
```diff
@@ -59,6 +65,50 @@ rcl_get_zero_initialized_timer()
   return null_timer;
 }
 
+#ifndef _WIN32
+#define DEFAULT_INITIAL_PERIOD_JITTER 0
+#else
+#define DEFAULT_INITIAL_PERIOD_JITTER 0
+#endif
```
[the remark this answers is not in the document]
hidmic: It's a placeholder. I hope to have a better estimate as a result of benchmarks on either platform.

rcl/src/rcl/timer.c
hidmic: I didn't because unlike feedback error, I don't expect jitter to change with the setpoint. But now that I think again about this, it might (e.g. due to the configured period becoming larger or smaller than the scheduler's time slot). We should reset it to a reasonable default.

rcl/test/rcl/test_events.cpp
hidmic: Same as above.

rcl/test/rcl/test_context.cpp
hidmic: Yes, they are. This patch changes `rcl_timer_init` API to accept an `rcl_timer_options_t *` argument, so all call sites have to be updated.

---

Alternatively, we could either have a separate `rcl_timer_init_with_options()` overload (though that doesn't scale all that well) OR use macros for versioning and deprecation. In general, I don't think we have a good policy when breaking C API.

clalancette:
I'm going to close this one out, since I don't think we have any plans to continue on this in the near future.  Feel free to reopen if you think otherwise.

PR #708, opened 2020-07-06
Remove domain_id and localhost_only from node_options

ivanpauno:
Both parameters were added to `init_options` during the node/participant mapping change.
This together with the linked PRs completely removes them.

### DIFF
--- a/rcl/test/rcl/test_node.cpp
+++ b/rcl/test/rcl/test_node.cpp
@@ -105,7 +105,7 @@ TEST_F(CLASSNAME(TestNodeFixture, RMW_IMPLEMENTATION), test_rcl_node_accessors) 
   const char * fq_name = "/ns/test_rcl_node_accessors_node";
   rcl_node_options_t default_options = rcl_node_get_default_options();
   ret = rcl_node_init(&invalid_node, name, namespace_, &invalid_context, &default_options);
-  ASSERT_EQ(RCL_RET_OK, ret);
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
   OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
   {
     osrf_testing_tools_cpp::memory_tools::disable_monitoring_in_all_threads();
--- a/rcl/test/rcl/test_node.cpp
+++ b/rcl/test/rcl/test_node.cpp
@@ -94,6 +94,7 @@ TEST_F(CLASSNAME(TestNodeFixture, RMW_IMPLEMENTATION), test_rcl_node_accessors) 
   {
     EXPECT_EQ(RCL_RET_OK, rcl_init_options_fini(&init_options)) << rcl_get_error_string().str;
   });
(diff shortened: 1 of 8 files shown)


rcl/test/rcl/test_node.cpp
```diff
@@ -94,6 +94,7 @@ TEST_F(CLASSNAME(TestNodeFixture, RMW_IMPLEMENTATION), test_rcl_node_accessors)
   {
     EXPECT_EQ(RCL_RET_OK, rcl_init_options_fini(&init_options)) << rcl_get_error_string().str;
   });
+  EXPECT_EQ(RCL_RET_OK, rcl_init_options_set_domain_id(&init_options, 42));
```
fujitatomoya, line 97: minor comment: I'd use ASSERT_EQ to check the API instead, though i do understand this never fails and it is not something critical.

ivanpauno: why? IMHO, `EXPECT` should be used everywhere, except it's clear that not satisfying the condition will lead to a segmentation fault. e.g.:
```cpp
ASSERT_NE(nullptr, pointer);
EXPECT_EQ(SOME_VALUE, *pointer);
```
In that case you obviously need an `ASSERT`.

rcl/test/rcl/test_node.cpp
```diff
@@ -103,19 +104,8 @@ TEST_F(CLASSNAME(TestNodeFixture, RMW_IMPLEMENTATION), test_rcl_node_accessors)
...
-    ASSERT_NE(RCL_RET_OK, ret);
-    // So retry with the default domain id setting (uses the environment as is).
-    default_options.domain_id = rcl_node_get_default_options().domain_id;
-    ret = rcl_node_init(&invalid_node, name, namespace_, &invalid_context, &default_options);
-    ASSERT_EQ(RCL_RET_OK, ret);
-  } else {
-    // This is the normal check (not windows and windows if not opensplice)
-    ASSERT_EQ(RCL_RET_OK, ret);
-  }
+  ASSERT_EQ(RCL_RET_OK, ret);
```
hidmic, line 108: @ivanpauno nit: 
```suggestion
  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
```

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11388/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6611/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9337/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11281/)

rcl/test/rcl/test_node.cpp
fujitatomoya: > In that case you obviously need an ASSERT.

that makes sense to me.

i was just thinking what if `rcl_init_options_set_domain_id` fails, then it proceeds the rest of test. in that case, it would lead un-expected failure.

ivanpauno:
I will start doing releases for Rolling, as I've probably broken several Rpr checkers.

PR #699, opened 2020-06-25
Remove unused internal prototypes

Blast545:
As the title says. These functions prototypes are already declared in the `rcl/remap.h` public header.

The access to `rcl_remap_copy` was changed to `RCL_LOCAL` because it is not part of the public header.

Signed-off-by: Jorge Perez <jjperez@ekumenlabs.com>

PR #709, opened 2020-07-07
Move rcl_remap_copy to public header

Blast545:
Follow up for failing test: https://github.com/ros2/rcl/pull/703#issuecomment-653678953.

I thought this was the only thing needed for making the local function available in the test: https://github.com/ros2/rcl/blob/6f3c3fb10391edaacb2676e2345a754bef1b76f2/rcl/src/rcl/remap_impl.h#L72-L77

I had to make available the function in the internal impl file: https://github.com/ros2/rcl/blob/2f4f0d4af517ca53b38c871ac283e7e395cbc6e3/rcl/src/rcl/remap_impl.h#L72

I think it is not a problem making internal functions available in the impl headers, because AFAIK those functions won't be available when using the package. In particular, for this function, I changed it from PUBLIC to LOCAL with this PR:
#699.

Last edit: After discussion with @ivanpauno and @hidmic I moved the function be available in the public header of `rcl`

### DIFF
--- a/rcl/include/rcl/remap.h
+++ b/rcl/include/rcl/remap.h
@@ -247,6 +247,31 @@ rcl_remap_node_namespace(
   rcl_allocator_t allocator,
   char ** output_namespace);
 
+/// Copy one remap structure into another.
+/**
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] rule The structure to be copied.
+ *  Its allocator is used to copy memory into the new structure.
+ * \param[out] rule_out A zero-initialized rcl_remap_t structure to be copied into.
+ * \return `RCL_RET_OK` if the structure was copied successfully, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if any function arguments are invalid, or
+ * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
+ * \return `RCL_RET_ERROR` if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_remap_copy(
+  const rcl_remap_t * rule,
+  rcl_remap_t * rule_out);
+
 /// Reclaim resources held inside rcl_remap_t structure.
 /**
  * <hr>
--- a/rcl/src/rcl/remap_impl.h
+++ b/rcl/src/rcl/remap_impl.h
@@ -69,7 +69,7 @@ typedef struct rcl_remap_impl_t
  * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
  * \return `RCL_RET_ERROR` if an unspecified error occurs.
  */
-RCL_LOCAL
+RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_ret_t
 rcl_remap_copy(
--- a/rcl/src/rcl/remap_impl.h
+++ b/rcl/src/rcl/remap_impl.h
@@ -51,31 +51,6 @@ typedef struct rcl_remap_impl_t
   rcl_allocator_t allocator;
 } rcl_remap_impl_t;
 
-/// Copy one remap structure into another.
-/**
- * <hr>
- * Attribute          | Adherence
- * ------------------ | -------------
- * Allocates Memory   | Yes
- * Thread-Safe        | No
- * Uses Atomics       | No
- * Lock-Free          | Yes
- *
- * \param[in] rule The structure to be copied.
- *  Its allocator is used to copy memory into the new structure.
- * \param[out] rule_out A zero-initialized rcl_remap_t structure to be copied into.
- * \return `RCL_RET_OK` if the structure was copied successfully, or
- * \return `RCL_RET_INVALID_ARGUMENT` if any function arguments are invalid, or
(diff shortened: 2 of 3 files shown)


Blast545:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11396/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6617/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9343/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11290/)

Blast545:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11397/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6617/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9343/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11290/)

rcl/test/rcl/test_remap.cpp
```diff
@@ -19,9 +19,9 @@
 #include "rcl/remap.h"
 #include "rcl/error_handling.h"
 
+#include "./allocator_testing_utils.h"
 #include "./arg_macros.hpp"
 #include "./arguments_impl.h"
-#include "./allocator_testing_utils.h"
 #include "./remap_impl.h"
```
ivanpauno, line 25: how does `./remap_impl.h` works if the header is not in the same folder?

rcl/src/rcl/remap_impl.h
```diff
@@ -69,7 +69,7 @@ typedef struct rcl_remap_impl_t
  * \return `RCL_RET_BAD_ALLOC` if allocating memory failed, or
  * \return `RCL_RET_ERROR` if an unspecified error occurs.
  */
-RCL_LOCAL
+RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_ret_t
 rcl_remap_copy(
```
ivanpauno, line 75: why not moving this function to a public header?
I don't see why it should be in a private header.

rcl/test/rcl/test_remap.cpp
ivanpauno: I see https://github.com/ros2/rcl/blob/70eba5e788af95c0e9f0dfe71af9ae630a9d4d76/rcl/test/CMakeLists.txt#L182
In this case, it's better to use `remap_impl.h` without `./`.
`./` should be reserved for includes that are actually in the same directory.

rcl/src/rcl/remap_impl.h
Blast545: I think in that case the function will be exposed to the users. And based on the `remap.h` and `arguments.h` code, is not expected that the user will deal with the function

rcl/test/rcl/test_remap.cpp
Blast545: `cpplint` complains if I don't include the directory for `.h` files

rcl/src/rcl/remap_impl.h
ivanpauno: The same applies to:
- https://github.com/ros2/rcl/blob/70eba5e788af95c0e9f0dfe71af9ae630a9d4d76/rcl/include/rcl/remap.h#L42
- https://github.com/ros2/rcl/blob/70eba5e788af95c0e9f0dfe71af9ae630a9d4d76/rcl/include/rcl/remap.h#L268

those functions have no external users and they are public.

I don't see why we should half expose an API.

Blast545: Then shouldn't those be LOCAL as well? IMO there shouldn't be public that's not expected to be used by external users, I might be wrong.

ivanpauno: > Then shouldn't those be LOCAL as well? IMO there shouldn't be public that's not expected to be used by external users, I might be wrong.

You shouldn't have anything LOCAL in a public header neither something public in a private header.
I don't see why this shouldn't be in a public header.
@hidmic for feedback.

hidmic: LOCAL in a public header is definitely wrong, your code will compile but fail to link. PUBLIC in a private header won't do harm, but it is unusual. I'd be inclined towards putting `rcl_remap_copy` in `remap.h`.

Blast545:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11479/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6683/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9417/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11376/)

rcl/include/rcl/remap.h
```diff
@@ -247,6 +247,31 @@ rcl_remap_node_namespace(
...
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] rule The structure to be copied.
+ *  Its allocator is used to copy memory into the new structure.
+ * \param[out] rule_out A zero-initialized rcl_remap_t structure to be copied into.
+ * \return `RCL_RET_OK` if the structure was copied successfully, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if any function arguments are invalid, or
```
hidmic, line 264: @Blast545 ideally, it'd be great this function's documentation was a bit more thorough. But I won't block this PR.

PR #710, opened 2020-07-08
Remove extra null check

Blast545:
This check is not needed, under unexpected `getenv` the function `rcutils_get_env` returns `NULL`.

### DIFF
--- a/rcl/src/rcl/security.c
+++ b/rcl/src/rcl/security.c
@@ -146,9 +146,6 @@ char * rcl_get_secure_root(
     RCUTILS_LOG_ERROR("rcutils_get_env failed: %s\n", get_env_error_str);
     return NULL;
   }
-  if (!env_buf) {
-    return NULL;
-  }
   if (0 == strcmp("", env_buf)) {
     ros_secure_enclave_override = false;
   }


rcl/src/rcl/security.c
```diff
@@ -146,9 +146,6 @@ char * rcl_get_secure_root(
     RCUTILS_LOG_ERROR("rcutils_get_env failed: %s\n", get_env_error_str);
     return NULL;
   }
-  if (!env_buf) {
```
ivanpauno, line 149: at least doing an `assert(env_buf)` would be good.

It's true that `rcutils_get_env` is supposed to return `NULL` if `env_buf` was `NULL` (and thus the check in L145 should be enough), but I don't know why deleting the check is a good idea.

ivanpauno: i.e. deleting the check to increase coverage doesn't sound like the correct approach.

dirk-thomas: From the docs of the `strcmp` function called with `env_buf` below:

> The behavior is undefined if lhs or rhs are not pointers to null-terminated byte strings.

So we should definitely not remove this check / return.

Blast545: @ivanpauno I didn't do this to improve coverage, I saw it and I thought it was dead code because that function doesn't set env_buf to NULL.

@dirk-thomas You are right, I didn't think of that check as a protection for `strcmp`, I'll close this PR.

ivanpauno: > @ivanpauno I didn't do this to improve coverage, I saw it and I thought it was dead code because that function doesn't set env_buf to NULL

You're right, I shouldn't make wrong assumptions.

PR #718, opened 2020-07-17
Move actual domain id from node to context

ivanpauno:
Follow up of discussion https://github.com/ros2/rcl/pull/689#discussion_r448978801.

The function `rcl_node_get_domain_id` could be deprecated.

### DIFF
--- a/rcl/include/rcl/context.h
+++ b/rcl/include/rcl/context.h
@@ -242,8 +242,7 @@ rcl_context_get_instance_id(rcl_context_t * context);
 
 /// Returns the context domain id.
 /**
- * The given context must be non-`NULL` and be valid.
- * If context is `NULL`, then `0` will be returned.
+ * The given context must be non-`NULL` and valid.
  * If context is uninitialized, then it is undefined behavior.
  *
  * <hr>
--- a/rcl/include/rcl/context.h
+++ b/rcl/include/rcl/context.h
@@ -248,9 +248,12 @@ rcl_context_get_instance_id(rcl_context_t * context);
  * Attribute          | Adherence
  * ------------------ | -------------
  * Allocates Memory   | No
- * Thread-Safe        | No
+ * Thread-Safe        | Yes [1]
  * Uses Atomics       | No
  * Lock-Free          | No
+ *
+ * <i>[1] Calling the function asynchronously with `rcl_init` or `rcl_shutdown` can result
+ *  in the function sometimes succeeding and sometimes returning `RCL_RET_INVALID_ARGUMENT`.
  *
  * \param[in] context object from which the domain id should be retrieved.
  * \param[out] domain_id output variable where the domain id will be returned.
--- a/rcl/include/rcl/context.h
+++ b/rcl/include/rcl/context.h
@@ -257,7 +257,7 @@ rcl_context_get_instance_id(rcl_context_t * context);
  *
  * \param[in] context from which the domain id should be retrieved.
  * \param[out] domain_id output variable where the domain id will be returned.
- * \return RCL_RET_INVALID_ARGUMENT if `context` is invalid, or
+ * \return RCL_RET_INVALID_ARGUMENT if `context` is invalid \ref `rcl_context_is_valid`, or
  * \return RCL_RET_INVALID_ARGUMENT if `domain_id` is `NULL`, or
  * \return RCL_RET_OK if the domain id was correctly retrieved.
  */
--- a/rcl/include/rcl/context.h
+++ b/rcl/include/rcl/context.h
@@ -240,6 +240,31 @@ RCL_WARN_UNUSED
 rcl_context_instance_id_t
 rcl_context_get_instance_id(rcl_context_t * context);
 
+/// Returns the context domain id.
+/**
+ * The given context must be non-`NULL` and be valid.
+ * If context is `NULL`, then `0` will be returned.
+ * If context is uninitialized, then it is undefined behavior.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | No
+ *
+ * \param[in] context object from which the domain id should be retrieved.
+ * \param[out] domain_id output variable where the domain id will be returned.
+ * \return RCL_RET_INVALID_ARGUMENT if `context` is invalid, or
+ *  RCL_RET_INVALID_ARGUMENT if `domain_id` is `NULL`, or
+ *  RCL_RET_OK if the domain id was correctly retrieved.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_context_get_domain_id(rcl_context_t * context, size_t * domain_id);
+
 /// Return `true` if the given context is currently valid, otherwise `false`.
 /**
  * If context is `NULL`, then `false` is returned.
--- a/rcl/include/rcl/context.h
+++ b/rcl/include/rcl/context.h
@@ -242,8 +242,7 @@ rcl_context_get_instance_id(rcl_context_t * context);
 
 /// Returns the context domain id.
 /**
- * The given context must be non-`NULL` and valid.
- * If context is uninitialized, then it is undefined behavior.
+ * \pre If context is uninitialized, then it is undefined behavior.
  *
  * <hr>
  * Attribute          | Adherence
--- a/rcl/include/rcl/context.h
+++ b/rcl/include/rcl/context.h
@@ -255,7 +255,7 @@ rcl_context_get_instance_id(rcl_context_t * context);
  * <i>[1] Calling the function asynchronously with `rcl_init` or `rcl_shutdown` can result
  *  in the function sometimes succeeding and sometimes returning `RCL_RET_INVALID_ARGUMENT`.
  *
- * \param[in] context object from which the domain id should be retrieved.
+ * \param[in] context from which the domain id should be retrieved.
  * \param[out] domain_id output variable where the domain id will be returned.
  * \return RCL_RET_INVALID_ARGUMENT if `context` is invalid, or
  * \return RCL_RET_INVALID_ARGUMENT if `domain_id` is `NULL`, or
--- a/rcl/include/rcl/context.h
+++ b/rcl/include/rcl/context.h
@@ -255,8 +255,8 @@ rcl_context_get_instance_id(rcl_context_t * context);
  * \param[in] context object from which the domain id should be retrieved.
  * \param[out] domain_id output variable where the domain id will be returned.
  * \return RCL_RET_INVALID_ARGUMENT if `context` is invalid, or
- *  RCL_RET_INVALID_ARGUMENT if `domain_id` is `NULL`, or
- *  RCL_RET_OK if the domain id was correctly retrieved.
+ * \return RCL_RET_INVALID_ARGUMENT if `domain_id` is `NULL`, or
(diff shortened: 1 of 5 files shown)


rcl/include/rcl/context.h
```diff
@@ -240,6 +240,31 @@ RCL_WARN_UNUSED
 rcl_context_instance_id_t
 rcl_context_get_instance_id(rcl_context_t * context);
 
+/// Returns the context domain id.
+/**
+ * The given context must be non-`NULL` and be valid.
+ * If context is `NULL`, then `0` will be returned.
```
hidmic, line 246: @ivanpauno why this behavior? I'd expect a no-op and `RCL_RET_INVALID_ARGUMENT` as return code.

rcl/include/rcl/context.h
```diff
@@ -240,6 +240,31 @@ RCL_WARN_UNUSED
 rcl_context_instance_id_t
 rcl_context_get_instance_id(rcl_context_t * context);
 
+/// Returns the context domain id.
+/**
+ * The given context must be non-`NULL` and be valid.
```
hidmic, line 245: @ivanpauno consider clarifying what valid means in this context. Initialized and not finalized perhaps? Or initialized and not shutdown? Also, consider using `/pre`.

rcl/include/rcl/context.h
```diff
@@ -240,6 +240,31 @@ RCL_WARN_UNUSED
...
+ * If context is `NULL`, then `0` will be returned.
+ * If context is uninitialized, then it is undefined behavior.
```
hidmic, line 247: @ivanpauno I think this is covered by the first sentence.

rcl/include/rcl/context.h
```diff
@@ -240,6 +240,31 @@ RCL_WARN_UNUSED
...
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
```
hidmic, line 253: @ivanpauno isn't it?

rcl/src/rcl/node.c
```diff
@@ -474,12 +468,10 @@ rcl_node_get_options(const rcl_node_t * node)
...
 {
-  const rcl_node_options_t * node_options = rcl_node_get_options(node);
-  if (!node_options) {
-    return RCL_RET_NODE_INVALID;  // error already set
-  }
+  RCL_CHECK_ARGUMENT_FOR_NULL(node, RCL_RET_NODE_INVALID);
   RCL_CHECK_ARGUMENT_FOR_NULL(domain_id, RCL_RET_INVALID_ARGUMENT);
-  *domain_id = node->impl->actual_domain_id;
+  RCL_CHECK_FOR_NULL_WITH_MSG(node->context, "invalid node", return RCL_RET_NODE_INVALID);
+  *domain_id = node->context->impl->rmw_context.actual_domain_id;
```
hidmic, line 474: @ivanpauno consider using `rcl_context_get_domain_id()` instead.

rcl/include/rcl/context.h
ivanpauno: I copied the docblock of rcl_context_get_instance_id and I then forgot to delete that line.

rcl/include/rcl/context.h
ivanpauno: > initialized and not finalized perhaps? Or initialized and not shutdown?

It means the second one.

> Also, consider using /pre.

This is not a precondition, because the function handles those cases correctly returning an appropriate error code.

ivanpauno: See https://github.com/ros2/rcl/pull/718/commits/85708bed10368e0ddc973c1c44fce2b8d7f513a7.

I'm not sure how to rephrase uninitialized to be more clear.
e.g. see [this other docblock](https://github.com/ros2/rcl/blob/1a7f737fa9484b93d8c3052d37ad3ac55c921f2f/rcl/include/rcl/context.h#L217-L219), that's really confusing.

rcl/src/rcl/node.c
ivanpauno: See https://github.com/ros2/rcl/pull/718/commits/2244f0c5da54c1a81957e04682241f8e810c6286.

rcl/include/rcl/context.h
ivanpauno: See https://github.com/ros2/rcl/pull/718/commits/85708bed10368e0ddc973c1c44fce2b8d7f513a7.

rcl/include/rcl/context.h
```diff
@@ -240,6 +240,29 @@ RCL_WARN_UNUSED
...
+ * Uses Atomics       | No
+ * Lock-Free          | No
+ *
+ * \param[in] context object from which the domain id should be retrieved.
+ * \param[out] domain_id output variable where the domain id will be returned.
+ * \return RCL_RET_INVALID_ARGUMENT if `context` is invalid, or
+ *  RCL_RET_INVALID_ARGUMENT if `domain_id` is `NULL`, or
+ *  RCL_RET_OK if the domain id was correctly retrieved.
```
wjwwood, line 259:
```suggestion
 * \return RCL_RET_INVALID_ARGUMENT if `domain_id` is `NULL`, or
 * \return RCL_RET_OK if the domain id was correctly retrieved.
```

rcl/include/rcl/context.h
wjwwood: Obviously we need to hash out how we document this, but I would say yes here. Even if you tried to init and access at the same time, it would not cause an error, since it is a primitive that can be accessed atomically. But also I'd just say that the precondition prevents calling init and this function at the same time as defined behavior.

rcl/include/rcl/context.h
ivanpauno: Ups, I forgot how to write doxygen correctly :joy:.
See https://github.com/ros2/rcl/pull/718/commits/7aa5594694b54a3aefe76a9b2d6d3b74af94e075.

rcl/include/rcl/context.h
ivanpauno: > Obviously we need to hash out how we document this, but I would say yes here. Even if you tried to init and access at the same time, it would not cause an error, since it is a primitive that can be accessed atomically.

After thinking a bit, I think you're right. But it's unpredictable if you will get `RCL_RET_INVALID_ARGUMENT` as return code or if the function will succeed.

ivanpauno: See https://github.com/ros2/rcl/pull/718/commits/313fbca5e072e6cc810ffd974f23d269d21210b1.

ivanpauno: Maybe `actual_domain_id` should be an atomic? to ensure correct memory ordering.

rcl/include/rcl/context.h
hidmic: Agreed about not being a `/pre`, my bad. That other docblock is insane.

rcl/include/rcl/context.h
```diff
@@ -240,6 +240,32 @@ RCL_WARN_UNUSED
 rcl_context_instance_id_t
 rcl_context_get_instance_id(rcl_context_t * context);
 
+/// Returns the context domain id.
+/**
+ * \pre If context is uninitialized, then it is undefined behavior.
```
hidmic, line 245: @ivanpauno nit: 
```suggestion
 * \remarks If context is uninitialized, then it is undefined behavior.
```

hidmic: I have to say though that this is true for every argument of every C API ever.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11528/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6699/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9433/)

rcl/include/rcl/context.h
wjwwood: It wouldn't harm to make it atomic, as on most systems it will be a noop.

ivanpauno:
The only changes in the last force push are in [this](https://github.com/ros2/rcl/pull/718/commits/2cdd2bae11df8203229e1d55bfac1331a4c6b6cf) commit, no other changes were introduced (I needed to force push because of rebasing).

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11536/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6701/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9436/)
* Windows [Build Status](http://ci.ros2.org/job/ci_osx/1​14​59/)

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11546/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6709/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9445/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11485/)

rcl/include/rcl/context.h
```diff
@@ -240,6 +240,32 @@ RCL_WARN_UNUSED
...
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | Yes [1]
+ * Uses Atomics       | No
+ * Lock-Free          | No
+ *
+ * <i>[1] Calling the function asynchronously with `rcl_init` or `rcl_shutdown` can result
+ *  in the function sometimes succeeding and sometimes returning `RCL_RET_INVALID_ARGUMENT`.
```
hidmic, line 256: @ivanpauno Hmm, not sure if went through this before, but this doesn't seem to imply it's safe.

rcl/include/rcl/context.h
```diff
@@ -240,6 +240,32 @@ RCL_WARN_UNUSED
...
+ *
+ * \param[in] context object from which the domain id should be retrieved.
```
hidmic, line 258: @ivanpauno nit: 
```suggestion
 * \param[in] context from which the domain id should be retrieved.
```

rcl/include/rcl/context.h
```diff
@@ -240,6 +240,32 @@ RCL_WARN_UNUSED
...
+ * \param[out] domain_id output variable where the domain id will be returned.
+ * \return RCL_RET_INVALID_ARGUMENT if `context` is invalid, or
```
hidmic, line 260: @ivanpauno nit: it'd be good to specify what `invalid` means (perhaps refer to the function that checks for its validity?)

rcl/include/rcl/context.h
ivanpauno: See discussion in https://github.com/ros2/rcl/pull/718#discussion_r456521461.

rcl/include/rcl/context.h
ivanpauno: Done in https://github.com/ros2/rcl/pull/718/commits/33fcc78e343dde17b80c407f5b9016ef0ddb9812

rcl/include/rcl/context.h
ivanpauno: See https://github.com/ros2/rcl/pull/718/commits/33fcc78e343dde17b80c407f5b9016ef0ddb9812.

rcl/include/rcl/context.h
hidmic: Alright, yeah, it's unusual but not unsafe.

ivanpauno:
Last few commits only improved docblocks, I will only re-run linux CI to cover that:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11583/)

PR #720, opened 2020-07-20
Add function rcl_event_is_valid

Blast545:
This functions checks if a passed event is a valid already init instance of a publisher/subscriber event. Created to be added in the internal checks of the `rcl_event_get_rmw_handle` and `rcl_take_event`.

Tests added with this PR verify the expected usage for the function.

### DIFF
--- a/rcl/src/rcl/event.c
+++ b/rcl/src/rcl/event.c
@@ -30,9 +30,9 @@ extern "C"
 #include "rmw/event.h"
 
 #include "./common.h"
+#include "./event_impl.h"
 #include "./publisher_impl.h"
 #include "./subscription_impl.h"
-#include "./event_impl.h"
 
 rcl_event_t
 rcl_get_zero_initialized_event()
--- a/rcl/src/rcl/event.c
+++ b/rcl/src/rcl/event.c
@@ -32,12 +32,7 @@ extern "C"
 #include "./common.h"
 #include "./publisher_impl.h"
 #include "./subscription_impl.h"
-
-typedef struct rcl_event_impl_t
-{
-  rmw_event_t rmw_handle;
-  rcl_allocator_t allocator;
-} rcl_event_impl_t;
+#include "./event_impl.h"
 
 rcl_event_t
 rcl_get_zero_initialized_event()
@@ -195,12 +190,10 @@ rcl_event_is_valid(const rcl_event_t * event)
 {
   RCL_CHECK_FOR_NULL_WITH_MSG(event, "event pointer is invalid", return false);
   RCL_CHECK_FOR_NULL_WITH_MSG(event->impl, "event's implementation is invalid", return false);
-  /*
   if (event->impl->rmw_handle.event_type == RMW_EVENT_INVALID) {
     RCUTILS_SET_ERROR_MSG("event's implementation not init");
     return false;
   }
-  */
   return true;
 }
 
--- a/rcl/src/rcl/event.c
+++ b/rcl/src/rcl/event.c
@@ -24,6 +24,7 @@ extern "C"
 #include "rcl/error_handling.h"
 #include "rcl/expand_topic_name.h"
 #include "rcl/remap.h"
+#include "rcutils/allocator.h"
 #include "rcutils/logging_macros.h"
 #include "rmw/error_handling.h"
 #include "rmw/validate_full_topic_name.h"
@@ -194,6 +195,8 @@ rcl_event_is_valid(const rcl_event_t * event)
     RCUTILS_SET_ERROR_MSG("event's implementation not init");
     return false;
   }
+  RCUTILS_CHECK_ALLOCATOR_WITH_MSG(
+    &event->impl->allocator, "not valid allocator", return false);
   return true;
 }
 
--- a/rcl/src/rcl/event.c
+++ b/rcl/src/rcl/event.c
@@ -190,6 +190,20 @@ rcl_event_get_rmw_handle(const rcl_event_t * event)
   }
 }
 
+bool
+rcl_event_is_valid(const rcl_event_t * event)
+{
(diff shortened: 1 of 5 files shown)


Blast545:
CI (Remaining Tier 1 platforms after peer review):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11549/)

ivanpauno:
I think that adding this is a good idea, but can you explain your use case?

ivanpauno:
> I think that adding this is a good idea, but can you explain your use case?

I was thinking that this new function was going to check if the `rmw_implementation` support that's event, I see the goal of this is different. Can you clarify what is the goal?

rcl/src/rcl/event.c
```diff
@@ -190,6 +190,20 @@ rcl_event_get_rmw_handle(const rcl_event_t * event)
   }
 }
 
+bool
+rcl_event_is_valid(const rcl_event_t * event)
+{
+  RCL_CHECK_FOR_NULL_WITH_MSG(event, "event pointer is invalid", return false);
+  RCL_CHECK_FOR_NULL_WITH_MSG(event->impl, "event's implementation is invalid", return false);
+  /*
+  if (event->impl->rmw_handle.event_type == RMW_EVENT_INVALID) {
```
ivanpauno, line 199: I don't see much value of this if we don't check that the `rmw` event is also valid

Blast545:
@ivanpauno The planned use case is using this new function to see if an event is valid to be used or not, similar to the function used to check publishers and subscribers.

rcl/src/rcl/event.c
Blast545: My initial proposal included that check, but I wasn't able to make it work. I think it requires making the typedef of the event impl exposed in another header. (Similar to the other `module_impl.h` headers). As this change required another change (exposing a typedef in a header) I wasn't sure it was desired to include two changes in one PR.

If you think it's not a problem I can make a PR with the whole changes for this to happen.

ivanpauno: > My initial proposal included that check, but I wasn't able to make it work. I think it requires making the typedef of the event impl exposed in another header

I'm not sure what typedef you're referring to.
The definition of the event implementation is defined in this same file https://github.com/ros2/rcl/blob/5d1962f3be6e9b8d6325af42ba9f0bd40b3db7ab/rcl/src/rcl/event.c#L36-L40.

> If you think it's not a problem I can make a PR with the whole changes for this to happen.

Yeah, I think it would be good to have a "final" version of the function on this PR.

Blast545: > The definition of the event implementation is defined in this same file

It is available there, so I can write in the `event.c` the code to check for `rmw_handle.implementation_identifier`. But I can't write the test it out because the type is not available externally. My suggestion for this is using a `event_impl.h` header as with other modules. I haven't tested it, but I think it will solve the issue.

ivanpauno: > My suggestion for this is using a event_impl.h header as with other modules. I haven't tested it, but I think it will solve the issue.

Ah yeah, it's not super nice to use the private definition in the tests but it sounds ok.

Blast545:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11561/)

rcl/src/rcl/event.c
Blast545: Added this check, moved the definition to a private header and added the test scenario for it. Let me know if you think I should change the name of this PR accordingly.

rcl/src/rcl/event.c
```diff
@@ -32,12 +32,7 @@ extern "C"
 #include "./common.h"
 #include "./publisher_impl.h"
 #include "./subscription_impl.h"
-
-typedef struct rcl_event_impl_t
-{
-  rmw_event_t rmw_handle;
-  rcl_allocator_t allocator;
-} rcl_event_impl_t;
+#include "./event_impl.h"
```
ivanpauno, line 35: nit: alpha order

rcl/src/rcl/event.c
```diff
@@ -190,6 +185,18 @@ rcl_event_get_rmw_handle(const rcl_event_t * event)
   }
 }
 
+bool
+rcl_event_is_valid(const rcl_event_t * event)
```
ivanpauno, line 189: It would be nice to also check the validity of `event->impl->rmw_handle.event_type`

rcl/test/rcl/test_events.cpp
```diff
@@ -730,6 +733,35 @@ TEST_F(TestEventFixture, test_bad_get_handle)
   EXPECT_EQ(NULL, rcl_event_get_rmw_handle(NULL));
 }
 
+/*
+ * Test cases for the event_is_valid function
+ */
+TEST_F(TestEventFixture, test_event_is_valid)
+{
+  EXPECT_FALSE(rcl_event_is_valid(nullptr));
+  rcl_reset_error();
```
ivanpauno, line 742:
```suggestion
  EXPECT_TRUE(rcl_error_is_set());
  rcl_reset_error();
```

By checking [rcl_error_is_set](https://github.com/ros2/rcl/blob/966efa3c49258cd011bbe2e78ebd718308bcb615/rcl/include/rcl/error_handling.h#L41), you make sure the implementation is actually setting an error message.

This applies everywhere else before `rcl_reset_error()`.

Blast545:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11588/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6729/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9467/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11550/)

PR #727, opened 2020-07-29
Add fault injection macros to rcl functions

brawner:
This adds the fault injection macros for all packages I've added tests to so far. I also added some fault injection unit tests to client and subscription, but they don't increase coverage that I'm aware of. I'll post CI results shortly.

Depends on #728

### DIFF
--- a/rcl/test/rcl/test_client.cpp
+++ b/rcl/test/rcl/test_client.cpp
@@ -294,6 +294,10 @@ TEST_F(TestClientFixture, test_client_init_fini_maybe_fail)
     if (RCL_RET_OK == ret) {
       EXPECT_TRUE(rcl_client_is_valid(&client));
       ret = rcl_client_fini(&client, this->node_ptr);
+      if (RCL_RET_OK != ret) {
+        // If fault injection caused fini to fail, we should try it again.
+        EXPECT_EQ(RCL_RET_OK, rcl_client_fini(&client, this->node_ptr));
+      }
     } else {
       EXPECT_TRUE(rcl_error_is_set());
       rcl_reset_error();
--- a/rcl/test/rcl/test_client.cpp
+++ b/rcl/test/rcl/test_client.cpp
@@ -17,6 +17,7 @@
 #include "rcl/client.h"
 
 #include "rcl/rcl.h"
+#include "rcutils/testing/fault_injection.h"
 
 #include "test_msgs/srv/basic_types.h"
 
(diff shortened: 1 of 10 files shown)


brawner:
Builds and tests of the following packages `rcutils`, `rcl_logging_spdlog`, `rosidl_typesupport_c`, `rosidl_typesupport_cpp`, `rcl`, `rcl_action`, `rcl_lifecycle`, `rosidl_runtime_c`, `rmw_dds_common`, `rmw`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11916/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6961/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9709/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11850/)

rcl/test/rcl/test_client.cpp
```diff
@@ -277,3 +278,25 @@ TEST_F(TestClientFixture, test_client_bad_arguments) {
...
+    test_msgs, srv, BasicTypes);
+  constexpr char topic_name[] = "chatter";
+  rcl_client_options_t default_client_options = rcl_client_get_default_options();
+
+  RCUTILS_FAULT_INJECTION_TEST(
+  {
+    rcl_client_t client = rcl_get_zero_initialized_client();
+    rcl_ret_t ret = rcl_client_init(
+      &client, this->node_ptr, ts, topic_name, &default_client_options);
+    if (RCL_RET_OK == ret) {
```
hidmic, line 294: @brawner hmm, what if no error is injected within `rcl_client_init` but one makes it into `rcl_client_fini`?

rcl/test/rcl/test_client.cpp
```diff
@@ -277,3 +278,25 @@ TEST_F(TestClientFixture, test_client_bad_arguments) {
...
+      EXPECT_TRUE(rcl_client_is_valid(&client));
+      ret = rcl_client_fini(&client, this->node_ptr);
```
hidmic, line 296: @brawner why not:
```suggestion
      EXPECT_EQ(RCL_RET_OK, rcl_client_fini(&client, this->node_ptr));
```
?

rcl/test/rcl/test_subscription.cpp
```diff
@@ -1065,3 +1066,26 @@ TEST_F(CLASSNAME(TestSubscriptionFixtureInit, RMW_IMPLEMENTATION), test_subscrip
...
+  constexpr char topic[] = "chatter";
+  rcl_subscription_options_t subscription_options = rcl_subscription_get_default_options();
+  rcl_subscription_t subscription = rcl_get_zero_initialized_subscription();
+
+  RCUTILS_FAULT_INJECTION_TEST(
+  {
+    rcl_ret_t ret = rcl_subscription_init(
+      &subscription, this->node_ptr, ts, topic, &subscription_options);
+
+    if (RCL_RET_OK == ret) {
```
hidmic, line 1083: @brawner same question and suggestion as above.

rcl/test/rcl/test_client.cpp
Blast545: What happens in this scenario if `init` doesn't return error but `fini` does?

rcl/test/rcl/test_client.cpp
brawner: I added a second `rcl_client_fini` with an appropriate check for this case.

rcl/test/rcl/test_client.cpp
brawner: While there are few functions in `_fini` that can fail it still is possible with fault injection in some cases. I added a check and a second attempt to finalize with your suggested `EXPECT_EQ` check on the second one.

rcl/test/rcl/test_subscription.cpp
brawner: Resolved similarly.

brawner:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12050/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7073/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9823/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11983/)

rcl/test/rcl/test_client.cpp
hidmic: I will point out that coping with incidental fault injection this way can get cumbersome fast. That's why at some point I suggested disabling fault injection entirely for the duration of these post-call checks. 

I won't block on this though.

Blast545:
Please take a look to PR job before merging

brawner:
@Blast545 Rpr job fails because it uses an older binary version of rcutils. The CI jobs I posted use the current master branches, except for this PR branch, so I think it's all good.

PR #729, opened 2020-07-29
Set transition_map->states/transition size to 0 on fini

brawner:
This was discovered during fault injection tests. But states_size and transititions_size need to be set to 0 on finalization because if finalization is ever called again, the for loop is entered without checking if states is null.


brawner:
Rebased onto master 
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11910/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6953/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9704/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11844/)

brawner:
Thanks for the reviews!

PR #732, opened 2020-07-29
Add mimick tests, publisher 95% coverage

Blast545:
This PR increases `publisher.c` coverage to 95.3 %

Signed-off-by: Jorge Perez <jjperez@ekumenlabs.com>

PR #734, opened 2020-08-06
Complete subscription API test coverage

hidmic:
Precisely what the title says. With these additional tests, `lcov` locally reports 95.3% line coverage.


hidmic:
CI up to `rcl`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11851/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6910/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9663/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11800/)

hidmic:
Note to reviewers. This PR, just like #732, introduces the `mocking_utils` namespace. Where it ends up living it's of no consequence.

rcl/src/rcl/subscription.c
```diff
@@ -72,7 +72,7 @@ rcl_subscription_init(
   rcutils_string_map_t substitutions_map = rcutils_get_zero_initialized_string_map();
   rcutils_ret_t rcutils_ret = rcutils_string_map_init(&substitutions_map, 0, rcutils_allocator);
   if (rcutils_ret != RCUTILS_RET_OK) {
-    RCL_SET_ERROR_MSG(rcutils_get_error_string().str);
+    // RCL_SET_ERROR_MSG(rcutils_get_error_string().str);
```
hidmic, line 75: @brawner @Blast545 this is something we do inconsistently throughout the code base, and while it's correct strictly speaking, it generates plenty overwrite warnings since `rcutils`, `rmw` and `rcl` error handling macros are currently the same.

Should we separate their states and propagate explicitly everywhere? Should we comment out (like here) and leave a note for future revision? Should we change `rcutils` to not complain about overwrites when the message stays the same?

rcl/test/rcl/test_subscription.cpp
```diff
@@ -528,6 +595,139 @@ TEST_F(CLASSNAME(TestSubscriptionFixture, RMW_IMPLEMENTATION), test_subscription
...
+  rmw_ret_t ret = rcl_subscription_init(
+    &subscription, this->node_ptr, ts, topic,
+    &subscription_options);
+  ASSERT_EQ(RMW_RET_OK, ret) << rcl_get_error_string().str;
+
+  test_msgs__msg__Strings * loaned_message = nullptr;
+  void ** type_erased_loaned_message_pointer = reinterpret_cast<void **>(&loaned_message);
+  EXPECT_EQ(
+    RCL_RET_SUBSCRIPTION_INVALID, rcl_take_loaned_message(
+      nullptr, type_erased_loaned_message_pointer, nullptr, nullptr));
```
Blast545, line 615: Here and below for invalid parameters tests using `nullptr` it would be great if you can follow the idea that @brawner gave me here: https://github.com/ros2/rcl/pull/732#discussion_r467264915

rcl/test/rcl/test_subscription.cpp
```diff
@@ -565,28 +765,54 @@ TEST_F(CLASSNAME(TestSubscriptionFixture, RMW_IMPLEMENTATION), test_subscription
...
-      ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
-      ASSERT_EQ(
-        std::string(test_string),
-        std::string(msg_loaned->string_value.data, msg_loaned->string_value.size));
-    } else {
-      ret = rcl_take(&subscription, &msg, nullptr, nullptr);
-      ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
-      ASSERT_EQ(
-        std::string(test_string), std::string(msg.string_value.data, msg.string_value.size));
+    if (!rcl_subscription_can_loan_messages(&subscription)) {
```
Blast545, line 773: Do you think it's a good idea to mock the loaned functions this way? might be a bit misleading. I think it would be helpful adding a comment about the purpose of this block of code for future readers.

rcl/test/rcl/test_subscription.cpp
```diff
@@ -654,10 +899,35 @@ TEST_F(
     rcl_take_serialized_message(&subscription_zero_init, &serialized_msg, nullptr, nullptr));
   rcl_reset_error();
 
+  EXPECT_EQ(
+    RCL_RET_INVALID_ARGUMENT,
+    rcl_take_serialized_message(&subscription, nullptr, nullptr, nullptr));
```
Blast545, line 904: Same comment on `nullptr`

rcl/src/rcl/subscription.c
hidmic: I purposefully didn't want to turn `rcutils` error state into a human-readable backtrace, as formatting is usually not appropriate (`rmw_error_msg.str` has already been formatted to provide file path and line number in a single line) and the single page char array may be exhausted rather quickly once you adopt that convention.

Whether `rcutils` error handling should support such use, that's a discussion I think we should have.

> If the warnings are important, we should try to handle the error strings when we expect them in the code (read+reset).

That's probably the right way given `rcutils` current state.

rcl/src/rcl/subscription.c
```diff
@@ -269,11 +271,8 @@ rcl_take(
   rmw_ret_t ret = rmw_take_with_info(
     subscription->impl->rmw_handle, ros_message, &taken, message_info_local, allocation);
   if (ret != RMW_RET_OK) {
-    RCL_SET_ERROR_MSG(rmw_get_error_string().str);
-    if (RMW_RET_BAD_ALLOC == ret) {
-      return RCL_RET_BAD_ALLOC;
-    }
-    return RCL_RET_ERROR;
+    // RCL_SET_ERROR_MSG(rmw_get_error_string().str);
+    return rcl_convert_rmw_ret_to_rcl_ret(ret);
```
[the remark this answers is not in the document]
Blast545: I added these changes return code changes as part of #738, to group them in a similar category and be able of tracking them for backport.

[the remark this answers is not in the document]
hidmic: I see and agree. I'll rebase on top of `master` once #738 is merged.

rcl/test/rcl/test_subscription.cpp
hidmic: Well, we need a way to test the happy path when using loaned messages for subscription at the `rcl` level. By mocking like this, we eliminate code duplication and the test remains true i.e. it tests what it is supposed to test. Loaning APIs say nothing about how the middleware performs such allocations, so I think we're in the clear.

I can add a comment for sure though.

hidmic:
Comments addressed and branch rebased. PTAL !

rcl/test/rcl/test_subscription.cpp
Blast545: Got it, I agree. Please remember adding the comment with this before merging.

rcl/src/rcl/subscription.c
Blast545: Should we open an issue for this discussion? we'd probably need to check `rcl` at some point to remove the remaining warnings. In any case, I don't think this will block merging this PR.

hidmic: We should.

rcl/test/rcl/test_subscription.cpp
hidmic: Yes! See 2f3875f.

hidmic:
One last CI run up to `rcl`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11874/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6924/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9676/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11815/)

rcl/src/rcl/subscription.c
```diff
@@ -72,7 +72,8 @@ rcl_subscription_init(
   rcutils_string_map_t substitutions_map = rcutils_get_zero_initialized_string_map();
   rcutils_ret_t rcutils_ret = rcutils_string_map_init(&substitutions_map, 0, rcutils_allocator);
   if (rcutils_ret != RCUTILS_RET_OK) {
-    RCL_SET_ERROR_MSG(rcutils_get_error_string().str);
```
brawner, line 75: I think the changes in this file should go in a separate PR since they resolve bugs and will need to be backported.

rcl/src/rcl/subscription.c
brawner: Unless there is evidence setting the error message is causing performance issues, I don't see any need to change it. As performance tests get written, we may seem some crop up and they can be addressed on an individual basis. I've concatenated error messages before, but it's a bit tedious. The current approach of warning when there is a message overwrite has plenty of advantages actually. It let's you set the error message, but it also let's you see the propagation of error messages through the code.

rcl/test/rcl/test_subscription.cpp
```diff
@@ -528,6 +595,139 @@ TEST_F(CLASSNAME(TestSubscriptionFixture, RMW_IMPLEMENTATION), test_subscription
...
+      RCL_RET_UNSUPPORTED, rcl_take_loaned_message(
+        &subscription, type_erased_loaned_message_pointer, nullptr, nullptr));
+    rcl_reset_error();
+
+    rmw_take_loaned_message_with_info_returns = RMW_RET_ERROR;
+    EXPECT_EQ(
+      RCL_RET_ERROR, rcl_take_loaned_message(
+        &subscription, type_erased_loaned_message_pointer, nullptr, nullptr));
+    rcl_reset_error();
+  }
```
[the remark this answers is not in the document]
hidmic: There can be and there is one, just not in the bad loaning (unit) test case but in the regular loaning (integration) test case above. That means that if loaning is supported by the underlying implementation, there will not be a mock that makes `rcl_take_loaned_message` call succeed, but success is something we expect from a real implementation and so we should be covered.

rcl/test/rcl/test_subscription.cpp
```diff
@@ -734,10 +1052,6 @@ TEST_F(CLASSNAME(TestSubscriptionFixtureInit, RMW_IMPLEMENTATION), test_subscrip
   EXPECT_EQ(NULL, rcl_subscription_get_options(nullptr));
   rcl_reset_error();
 
-  EXPECT_EQ(
```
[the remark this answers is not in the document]
hidmic: It has its own test now,  `test_bad_get_publisher_count`.

rcl/src/rcl/subscription.c
hidmic: That's fair point, though at the same time we need the additional tests to ensure those changes have the desired effect. What about splitting the PR in two commits that can be rebased instead of squashed? Once we've settled reviews, of course.

rcl/src/rcl/subscription.c
hidmic: > The current approach of warning when there is a message overwrite has plenty of advantages actually. It let's you set the error message, but it also let's you see the propagation of error messages through the code.

Right, but that's not the spirit of those overwrites, which are actually `rcutils` error handling reporting there was an error with error handling itself. Also, to have libraries print to `stderr` as part of normal operation (in case a chain of errors is handled by upper layers) seems inadequate.

rcl/package.xml
```diff
@@ -33,6 +33,7 @@
   <test_depend>launch_testing_ament_cmake</test_depend>
   <test_depend>osrf_testing_tools_cpp</test_depend>
   <test_depend>test_msgs</test_depend>
+  <test_depend>mimick_vendor</test_depend>
```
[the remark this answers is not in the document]
hidmic: Done in 5becdfee8960a9d06c65fe357231e716481a7fda.

rcl/test/rcl/test_subscription.cpp
```diff
@@ -143,6 +146,11 @@ TEST_F(
   rcl_reset_error();
 }
 
+MOCKING_UTILS_BOOL_OPERATOR_RETURNS_FALSE(rcutils_allocator_t, ==)
```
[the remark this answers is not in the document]
hidmic: Done in d463ffe6e28ea3c003567742addd3cb79667fd7c

rcl/test/rcl/test_subscription.cpp
```diff
@@ -528,6 +595,139 @@ TEST_F(CLASSNAME(TestSubscriptionFixture, RMW_IMPLEMENTATION), test_subscription
   }
 }
 
+/* Test for all failure modes in subscription take with loaned messages function.
+ */
+TEST_F(CLASSNAME(TestSubscriptionFixture, RMW_IMPLEMENTATION), test_bad_take_loaned_message) {
+  const char * topic = "rcl_loan";
```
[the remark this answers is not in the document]
hidmic: Done in bdb331a69964eb75a285ba6c99ab0bfea4906bb3.

rcl/test/rcl/test_subscription.cpp
```diff
@@ -528,6 +595,139 @@ TEST_F(CLASSNAME(TestSubscriptionFixture, RMW_IMPLEMENTATION), test_subscription
   }
 }
 
+/* Test for all failure modes in subscription take with loaned messages function.
+ */
+TEST_F(CLASSNAME(TestSubscriptionFixture, RMW_IMPLEMENTATION), test_bad_take_loaned_message) {
```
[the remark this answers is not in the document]
hidmic: Sounds reasonable. Done in 527f0c3c3df893f50592c94d51530721a90f3e3b

rcl/test/rcl/test_subscription.cpp
```diff
@@ -528,6 +595,139 @@ TEST_F(CLASSNAME(TestSubscriptionFixture, RMW_IMPLEMENTATION), test_subscription
...
+  }
+
+  EXPECT_EQ(
+    RCL_RET_OK,
+    rcl_subscription_fini(&subscription, this->node_ptr)) << rcl_get_error_string().str;
+}
+
+/* Test for all failure modes in subscription return loaned messages function.
+ */
+TEST_F(CLASSNAME(TestSubscriptionFixture, RMW_IMPLEMENTATION), test_bad_return_loaned_message) {
```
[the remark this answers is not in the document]
hidmic: Sounds reasonable too. Done in 527f0c3c3df893f50592c94d51530721a90f3e3b

rcl/test/rcl/test_subscription.cpp
```diff
@@ -565,28 +765,54 @@ TEST_F(CLASSNAME(TestSubscriptionFixture, RMW_IMPLEMENTATION), test_subscription
...
-        std::string(test_string),
-        std::string(msg_loaned->string_value.data, msg_loaned->string_value.size));
-    } else {
-      ret = rcl_take(&subscription, &msg, nullptr, nullptr);
-      ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
-      ASSERT_EQ(
-        std::string(test_string), std::string(msg.string_value.data, msg.string_value.size));
+    if (!rcl_subscription_can_loan_messages(&subscription)) {
+      patch_take.then_call(
+        [](const rmw_subscription_t * subscription,
```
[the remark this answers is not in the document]
hidmic: Yeah, this was `uncrustify`... Renaming some arguments helps a bit. See 00a5098fbfcf36f403d87e80c62fb9661571eb08.

rcl/test/rcl/test_subscription.cpp
```diff
@@ -705,24 +975,72 @@ TEST_F(
     rcl_take_sequence(&subscription, seq_size, &messages, &message_infos_short, nullptr));
   rcl_reset_error();
 
-  // This test fails for rmw_cyclonedds_cpp function rmw_take_sequence
-  // Tracked here: https://github.com/ros2/rmw_cyclonedds/issues/193
-  /*
+  EXPECT_EQ(
+    RCL_RET_INVALID_ARGUMENT,
+    rcl_take_sequence(&subscription, seq_size, nullptr, &message_infos, nullptr));
```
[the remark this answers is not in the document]
hidmic: Makes sense, for the entire test actually. See 3ec8fbf2aeae3c87ea5e7f59ce1cb8ec2b318561.

rcl/src/rcl/subscription.c
brawner: I'm actually really unsure what the best solution is, but I don't think leaving commented code is desirable. I think the simplest solution would be to add a TODO with a link to the issue you create and leave the code as is. If you feel that the extra `RCL_SET_ERROR_MSG` calls should be removed, then remove them as well as adding the TODO.

rcl/src/rcl/subscription.c
hidmic: > If you feel strongly that this should be done by squashing and splitting commits in this PR

No, I explicitly said rebasing, not squashing. Anyhow, if lack of tests in said PR is not a concern, I may as well go with it.

rcl/src/rcl/subscription.c
hidmic: > I don't think leaving commented code is desirable

Yeah, that was intended to trigger the discussion rather than a solution for this issue.

rcl/src/rcl/subscription.c
hidmic: See #739.

brawner: > No, I explicitly said rebasing, not squashing

Sorry, I thought you were going to squash the 10 commits here into two separate commits then rebase on to master, one commit which would be cherry-pickable for backporting, and the other for the unit tests.

hidmic:
Another CI run up to `rcl`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11903/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6946/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9698/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11834/)

hidmic:
PR rebased on top of `master`. @brawner PTAL.

rcl/test/rcl/test_subscription.cpp
```diff
@@ -686,43 +957,92 @@ TEST_F(
...
+    rcl_take_sequence(&subscription, seq_size, &messages, &message_infos_short, allocation));
+  rcl_reset_error();
+
+  EXPECT_EQ(
+    RCL_RET_INVALID_ARGUMENT,
+    rcl_take_sequence(&subscription, seq_size, nullptr, &message_infos, allocation));
   rcl_reset_error();
 
-  // This test fails for rmw_cyclonedds_cpp function rmw_take_sequence
-  // Tracked here: https://github.com/ros2/rmw_cyclonedds/issues/193
```
brawner, line 709: Is this issue now resolved or does the test no longer fail?

hidmic: Issue is not resolved, but test no longer relies on it to work.

hidmic:
With green CI and two approvals, I'll proceed and merge.

nuclearsandwich:
@hidmic these tests are segfaulting on CentOS. I've opened https://github.com/ros2/rcl/issues/749

Issue #574, opened 2020-02-11
Add tests checking when a graph guard condition is triggered


ivanpauno:
#### Feature description

There are currently some tests for this ([here](https://github.com/ros2/rcl/blob/2daa408ea46bb2a4305a6221d73ea400e4c02d62/rcl/test/rcl/test_graph.cpp#L1231)), but those are a bit fragile and aren't checking that the graph guard condition is being triggered for all the possible graph changes.

#### Implementation considerations

I imagine the tests looking like:
```
> create another node
  > assert that graph guard condition is triggered within a timeout
> create a publisher
  > assert that graph guard condition is triggered within a timeout
... etc for other possible graph changes
```
instead of using a thread.

Unrelated note: Some of the current graph tests are sleeping with a timeout. Waiting on a graph guard condition trigger in a loop until the desired change is detected would be more reliable.

(edit) Extended the description.

Barry-Xu-2018:
@ivanpauno 

> Feature description
>
> There's currently nothing testing when a graph guard condition is triggered, tests should be added.

I want to make clear the requirement of this issue.  

After checking test in [rcl/test/rcl/test_graph.cpp](https://github.com/ros2/rcl/blob/master/rcl/test/rcl/test_graph.cpp), existing test cases are to wait graph guard condition triggering after graph is changed (such as calling rcl_publisher_init()).  
e.g.
test case `test_graph_guard_condition_topics` 

https://github.com/ros2/rcl/blob/2daa408ea46bb2a4305a6221d73ea400e4c02d62/rcl/test/rcl/test_graph.cpp#L1271-L1287 

Calling rcl_publisher_init(), rcl_subscription_init(), rcl_subscription_fini() and rcl_publisher_fini() will affect graph.  That is, `graph guard condition` will be triggered. 

While a graph guard condition is triggered, what test do you want to be added ? e.g. check if publisher is added or deleted ?  

> Implementation considerations
>
> Current graph tests could be refactored to first wait a graph condition is triggered, that will also avoid some flakiness of the current tests.

Current test implementation like as below (such as check_graph_state() ) 
```
loop for try 
    Check status (such as the number of publishers, subscribers)
    If not match expected value 
         add graph guard condition to wait_set
         rcl_wait() //Wait for graph changing 
```

Do you mean rcl_wait() should be done (after graph condition trigger) before check status ?

ivanpauno:
> test case test_graph_guard_condition_topics

[...]

PR #736, opened 2020-08-10
Add tests checking when a graph guard condition is triggered

Ada-King:
Related issue : [#574](https://github.com/ros2/rcl/issues/574)

Signed-off-by: Ada-King <Bingtao.Du@sony.com>

### DIFF
--- a/rcl/test/rcl/test_graph.cpp
+++ b/rcl/test/rcl/test_graph.cpp
@@ -1295,6 +1295,32 @@ TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_graph_guard_conditi
   ASSERT_GE(graph_changes_count, 4ul);
 }
 
+/* Test the graph when guard condition is triggered.
+ */
+TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_graph_guard_condition_triggered) {
+  rcl_ret_t ret;
+  int64_t wait_time_out = -1;
+  const rcl_guard_condition_t * graph_guard_condition =
+    rcl_node_get_graph_guard_condition(this->node_ptr);
+  ASSERT_NE(nullptr, graph_guard_condition) << rcl_get_error_string().str;
+  ret = rcl_wait_set_clear(this->wait_set_ptr);
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  ret = rcl_wait_set_add_guard_condition(this->wait_set_ptr, graph_guard_condition, NULL);
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+
+  std::thread trigger_thread(
+    [ = ]() {
+      std::this_thread::sleep_for(std::chrono::milliseconds(200));
+      rcl_ret_t ret =
+      rcl_trigger_guard_condition(const_cast<rcl_guard_condition_t *>(graph_guard_condition));
+      EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+    });
+
+  ret = rcl_wait(this->wait_set_ptr, wait_time_out);
+  EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  trigger_thread.join();
+}
+
 /* Test the rcl_service_server_is_available function.
  */
 TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_rcl_service_server_is_available) {


rcl/test/rcl/test_graph.cpp
```diff
@@ -1295,6 +1295,32 @@ TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_graph_guard_conditi
...
+  ret = rcl_wait_set_clear(this->wait_set_ptr);
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  ret = rcl_wait_set_add_guard_condition(this->wait_set_ptr, graph_guard_condition, NULL);
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+
+  std::thread trigger_thread(
+    [ = ]() {
+      std::this_thread::sleep_for(std::chrono::milliseconds(200));
+      rcl_ret_t ret =
+      rcl_trigger_guard_condition(const_cast<rcl_guard_condition_t *>(graph_guard_condition));
```
wjwwood, line 1315:
```suggestion
        rcl_trigger_guard_condition(const_cast<rcl_guard_condition_t *>(graph_guard_condition));
```

rcl/test/rcl/test_graph.cpp
```diff
@@ -1295,6 +1295,32 @@ TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_graph_guard_conditi
   ASSERT_GE(graph_changes_count, 4ul);
 }
 
+/* Test the graph when guard condition is triggered.
+ */
+TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_graph_guard_condition_triggered) {
+  rcl_ret_t ret;
+  int64_t wait_time_out = -1;
```
wjwwood, line 1302:
```suggestion
  const int64_t wait_time_out = -1;
```

Or consider getting rid of it completely, since it never changes.

rcl/test/rcl/test_graph.cpp
```diff
@@ -1295,6 +1295,32 @@ TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_graph_guard_conditi
...
+    [ = ]() {
+      std::this_thread::sleep_for(std::chrono::milliseconds(200));
+      rcl_ret_t ret =
+      rcl_trigger_guard_condition(const_cast<rcl_guard_condition_t *>(graph_guard_condition));
+      EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+    });
+
+  ret = rcl_wait(this->wait_set_ptr, wait_time_out);
+  EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  trigger_thread.join();
```
wjwwood, line 1321: What is the higher level purpose of this test? From my perspective it's essentially testing that triggering a guard condition will wake up a wait set, but is the goal to do something specific to the graph guard condition?

Reading https://github.com/ros2/rcl/issues/574, I don't think this test really addresses it, because it doesn't check that the graph guard condition is triggered by certain events, nor does it assert anything about behavior that utilizes the graph guard condition (e.g. the GraphListener in rclcpp).

@ivanpauno perhaps you could explain what kind of test you were expecting from https://github.com/ros2/rcl/issues/574?

ivanpauno: > @ivanpauno perhaps you could explain what kind of test you were expecting from #574?

Sorry, my description of the issue in #574 is a bit poor.
AFAIR, my motivation for opening that issue was that some rmw implementations weren't triggering the graph guard condition in all graph changes.
My idea was to add more tests that check that after each function that can have a "graph side effect", the graph guard conditions is triggered within a timeout.
e.g.:
```
-> create a node
  -> check that graph guard condition is triggered within a timeout
-> create a publisher
  -> check that graph guard condition is triggered within a timeout
... etc for everything that can produce a graph change
```

I think that the only existing test is [this one]( https://github.com/ros2/rcl/blob/2daa408ea46bb2a4305a6221d73ea400e4c02d62/rcl/test/rcl/test_graph.cpp#L1231).
That test is a bit fragile, but my basic idea is to make that more extensive to cover all graph changes.
There are some cases that are hard to test, i.e. making sure that a graph guard condition is triggered when a publisher matching an existing subscription is detected (in that case the graph guard condition is triggered twice, once because the publisher was detected and the other because the subscription found a matching publisher).

> What is the higher level purpose of this test? 

I also don't see value in the test added here.

ivanpauno: I have edited #574 to make its purpose more clear.

Barry-Xu-2018:
@clalancette 

This PR is obsoleted.   
There was a new PR (#811) instead of this.   
@Ada-King please close this PR.

Issue #555, opened 2020-01-16
yaml parser raises an error "Sequence should be of same type" when meets .nan in list of doubles


iquater:
ros2 launch raises an exception while parsing list with .nan value

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - binaries
- Version or commit hash:
  - dashing
- DDS implementation:
  -  Fast-RTPS
- Client library (if applicable):
  - rclcpp, rclpy

#### Steps to reproduce issue
Create python or cpp package:
```python
import rclpy
from rclpy.node import Node


class TestNode(Node):
    def __init__(self):
        super().__init__("parameter_holder")

def main(args=None):
    rclpy.init(args=args)
    parameter_holder = TestNode()
    rclpy.spin(parameter_holder)
    parameter_holder.destroy_node()
    rclpy.shutdown()
    return 0
```
launch file:
```python
from launch.launch_description import LaunchDescription
from launch_ros.actions import Node

def generate_launch_description():
    return LaunchDescription([
        Node(package='reprod',
             node_executable='param_holder',
             node_name='parameter_holder',
             output='screen',
             parameters=[
                 "/tmp/test.yaml"
             ])
    ])
```
configuration file
```yaml
parameter_holder:
    ros__parameters:
        test_list: [1.0, 2.0]
        test_list2: [1.0, .nan]
```

#### Expected behavior
Yaml parser should treat .nan as valid double (or float) value.
#### Actual behavior
Yaml parser treats .nan as string and throws exception:

`RuntimeError: Failed to parse yaml params file '/tmp/test.yaml': Sequence should be of same type. Value type 'string' do not belong at line_num 4`

dirk-thomas:
Loosely related to ros2/rosidl#351.

ivanpauno:
@iquater Thanks for opening the issue. This is not currently supported, but it's a desired feature.
A PR adding support to special floating point values is more than welcome.

PR #741, opened 2020-08-12
Fix yaml parser error when meets .nan

Ada-King:
Related issue [555](https://github.com/ros2/rcl/issues/555)

The problem has examined on my host machine.

Signed-off-by: Ada-King <Bingtao.Du@sony.com>

### DIFF
--- a/rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
+++ b/rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
@@ -32,10 +32,6 @@
 #include "rcl_yaml_param_parser/types.h"
 #include "rcl_yaml_param_parser/visibility_control.h"
 
-#ifdef _WIN32
-#define strcasecmp stricmp
-#endif
-
 #ifdef __cplusplus
 extern "C"
 {
--- a/rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
+++ b/rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
@@ -32,6 +32,10 @@
 #include "rcl_yaml_param_parser/types.h"
 #include "rcl_yaml_param_parser/visibility_control.h"
 
+#ifdef _WIN32
+#define strcasecmp stricmp
+#endif
+
 #ifdef __cplusplus
 extern "C"
 {
--- a/rcl_yaml_param_parser/src/parser.c
+++ b/rcl_yaml_param_parser/src/parser.c
@@ -21,6 +21,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <yaml.h>
+#include <ctype.h>
 
 #include "rcl_yaml_param_parser/parser.h"
 #include "rcl_yaml_param_parser/types.h"
@@ -1105,11 +1106,23 @@ static void * get_value(
   {
     errno = 0;
     endptr = NULL;
-    if (strcasecmp(value, ".nan") == 0) {
-      char * tmp_value = (char *)value;
-      snprintf(tmp_value, strlen(tmp_value), "%s", "nan");
+    char * tmp_value = NULL;
+    if ((0 == strcasecmp(value, ".nan")) ||
+      (0 == strcasecmp(value, ".inf")) ||
+      (0 == strcasecmp(value, "+.inf")) ||
+      (0 == strcasecmp(value, "-.inf")) ||
+      (0 == strcasecmp(value, ".infinity")) ||
+      (0 == strcasecmp(value, "+.infinity")) ||
+      (0 == strcasecmp(value, "-.infinity")))
+    {
+      tmp_value = rcutils_strdup(value, allocator);
+      for ( ; !isalpha(*tmp_value); ) {
+        tmp_value += 1;
+      }
+      dval = strtod(tmp_value, &endptr);
+    } else {
+      dval = strtod(value, &endptr);
     }
-    dval = strtod(value, &endptr);
     if ((0 == errno) && (NULL != endptr)) {
       if ((NULL != endptr) && (endptr != value)) {
         if (('\0' != *value) && ('\0' == *endptr)) {
--- a/rcl_yaml_param_parser/src/parser.c
+++ b/rcl_yaml_param_parser/src/parser.c
@@ -1105,9 +1105,9 @@ static void * get_value(
   {
     errno = 0;
     endptr = NULL;
-    if(strcasecmp(value, ".nan") == 0) {
-      char *tmp_value = (char *)value;
-      strcpy(tmp_value, "nan");
+    if (strcasecmp(value, ".nan") == 0) {
+      char * tmp_value = (char *)value;
+      snprintf(tmp_value, strlen(tmp_value), "%s", "nan");
     }
     dval = strtod(value, &endptr);
     if ((0 == errno) && (NULL != endptr)) {
--- a/rcl_yaml_param_parser/src/parser.c
+++ b/rcl_yaml_param_parser/src/parser.c
@@ -1105,6 +1105,10 @@ static void * get_value(
   {
     errno = 0;
     endptr = NULL;
+    if(strcasecmp(value, ".nan") == 0) {
+      char *tmp_value = (char *)value;
+      strcpy(tmp_value, "nan");
+    }
     dval = strtod(value, &endptr);
     if ((0 == errno) && (NULL != endptr)) {
       if ((NULL != endptr) && (endptr != value)) {
--- a/rcl_yaml_param_parser/src/parser.c
+++ b/rcl_yaml_param_parser/src/parser.c
@@ -1106,7 +1106,7 @@ static void * get_value(
   {
     errno = 0;
     endptr = NULL;
-    const char * vernier_ptr = NULL;
+    const char * iter_ptr = NULL;
     if ((0 == strcasecmp(value, ".nan")) ||
       (0 == strcasecmp(value, ".inf")) ||
       (0 == strcasecmp(value, "+.inf")) ||
@@ -1115,10 +1115,10 @@ static void * get_value(
       (0 == strcasecmp(value, "+.infinity")) ||
       (0 == strcasecmp(value, "-.infinity")))
     {
-      for (vernier_ptr = value; !isalpha(*vernier_ptr); ) {
-        vernier_ptr += 1;
+      for (iter_ptr = value; !isalpha(*iter_ptr); ) {
+        iter_ptr += 1;
       }
-      dval = strtod(vernier_ptr, &endptr);
+      dval = strtod(iter_ptr, &endptr);
     } else {
       dval = strtod(value, &endptr);
     }
--- a/rcl_yaml_param_parser/src/parser.c
+++ b/rcl_yaml_param_parser/src/parser.c
@@ -1106,7 +1106,7 @@ static void * get_value(
   {
     errno = 0;
     endptr = NULL;
-    char * tmp_value = NULL;
+    const char * vernier_ptr = NULL;
     if ((0 == strcasecmp(value, ".nan")) ||
       (0 == strcasecmp(value, ".inf")) ||
       (0 == strcasecmp(value, "+.inf")) ||
@@ -1115,11 +1115,10 @@ static void * get_value(
       (0 == strcasecmp(value, "+.infinity")) ||
       (0 == strcasecmp(value, "-.infinity")))
     {
-      tmp_value = rcutils_strdup(value, allocator);
-      for ( ; !isalpha(*tmp_value); ) {
-        tmp_value += 1;
+      for (vernier_ptr = value; !isalpha(*vernier_ptr); ) {
+        vernier_ptr += 1;
       }
-      dval = strtod(tmp_value, &endptr);
+      dval = strtod(vernier_ptr, &endptr);
     } else {
       dval = strtod(value, &endptr);
     }
--- a/rcl_yaml_param_parser/src/parser.c
+++ b/rcl_yaml_param_parser/src/parser.c
@@ -1124,6 +1124,9 @@ static void * get_value(
         iter_ptr += 1;
       }
       dval = strtod(iter_ptr, &endptr);
+      if (*value == '-') {
+        dval = -dval;
+      }
     } else {
       dval = strtod(value, &endptr);
     }
--- a/rcl_yaml_param_parser/src/parser.c
+++ b/rcl_yaml_param_parser/src/parser.c
@@ -1107,13 +1107,18 @@ static void * get_value(
     errno = 0;
     endptr = NULL;
     const char * iter_ptr = NULL;
-    if ((0 == strcasecmp(value, ".nan")) ||
-      (0 == strcasecmp(value, ".inf")) ||
-      (0 == strcasecmp(value, "+.inf")) ||
-      (0 == strcasecmp(value, "-.inf")) ||
-      (0 == strcasecmp(value, ".infinity")) ||
-      (0 == strcasecmp(value, "+.infinity")) ||
-      (0 == strcasecmp(value, "-.infinity")))
+    if ((0 == strcmp(value, ".nan")) ||
+      (0 == strcmp(value, ".NaN")) ||
+      (0 == strcmp(value, ".NAN")) ||
+      (0 == strcmp(value, ".inf")) ||
+      (0 == strcmp(value, ".Inf")) ||
+      (0 == strcmp(value, ".INF")) ||
+      (0 == strcmp(value, "+.inf")) ||
+      (0 == strcmp(value, "+.Inf")) ||
+      (0 == strcmp(value, "+.INF")) ||
+      (0 == strcmp(value, "-.inf")) ||
+      (0 == strcmp(value, "-.Inf")) ||
+      (0 == strcmp(value, "-.INF")))
     {
       for (iter_ptr = value; !isalpha(*iter_ptr); ) {
         iter_ptr += 1;
--- a/rcl_yaml_param_parser/test/special_float.yaml
+++ b/rcl_yaml_param_parser/test/special_float.yaml
@@ -1,0 +1,4 @@
+test_node:
+    ros__parameters:
+        isstring: [string, .nananan, .infinf]
+        nan_inf: [1.1, 2.2, .nan, .NAN, .inf, +.inf, -.INF, +.infinity, -.INFINITY]
--- a/rcl_yaml_param_parser/test/special_float.yaml
+++ b/rcl_yaml_param_parser/test/special_float.yaml
@@ -1,4 +1,4 @@
 test_node:
     ros__parameters:
-        isstring: [string, .nananan, .infinf]
-        nan_inf: [1.1, 2.2, .nan, .NAN, .inf, +.inf, -.INF, +.infinity, -.INFINITY]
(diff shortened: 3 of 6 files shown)


rcl_yaml_param_parser/src/parser.c
```diff
@@ -1105,6 +1105,10 @@ static void * get_value(
   {
     errno = 0;
     endptr = NULL;
+    if (strcasecmp(value, ".nan") == 0) {
+      char * tmp_value = (char *)value;
+      snprintf(tmp_value, strlen(tmp_value), "%s", "nan");
+    }
```
clalancette, line 1111: While this works, it breaks the contract of the function.  That is, value is passed in as `const char * const value`, and so the caller should (rightly) assume that the string won't be modified.

I think in order to do what you want to do here, we need to advance the value pointer by one byte so it points to the `n` at the front.

Ada-King: > While this works, it breaks the contract of the function. That is, value is passed in as const char * const value, and so the caller should (rightly) assume that the string won't be modified.

That make sense, but i am not sure whether  ~~const~~ char is really neccessary.

> I think in order to do what you want to do here, we need to advance the value pointer by one byte so it points to the n at the front.

Sorry, i dont't get it. The `value` pointer is decorated by const, so it's a read-only pointer, how can i change its direction ?

clalancette: > Sorry, i dont't get it. The `value` pointer is decorated by const, so it's a read-only pointer, how can i change its direction ?

We'll have to be a little creative :).

Thinking about it more, I think the best thing to do here is probably to make a copy of the original data, then try to parse that as a double.  That way we keep the current function contract, but can also move as we want.

When you write tests for this, please also consider the case where the user is passing in `nananana` as a string; we'll want to make sure we don't improperly consider that a double.

Also, while we are here, I think we should probably add in support for YAMLs "Inf", just to be complete.

Ada-King:
> Besides the inline comment, can you also please add a test for this case? Thanks.

Ada-King:
> Besides the inline comment, can you also please add a test for this case? Thanks.

Close this by accident... never mind.
The code has been modified according to your suggestion. Please check :)

rcl_yaml_param_parser/test/test_parse_yaml.cpp
```diff
@@ -497,6 +497,54 @@ TEST(test_file_parser, maximum_number_parameters) {
...
+    ASSERT_TRUE(NULL != path) << rcutils_get_error_string().str;
+    OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+    {
+      allocator.deallocate(path, allocator.state);
+    });
+    ASSERT_TRUE(rcutils_exists(path)) << "No test YAML file found at " << path;
+    rcl_params_t * params_hdl = rcl_yaml_node_struct_init(allocator);
+    ASSERT_TRUE(NULL != params_hdl) << rcutils_get_error_string().str;
+    bool res = rcl_parse_yaml_file(path, params_hdl);
+    fprintf(stderr, "%s\n", rcutils_get_error_string().str);
```
clalancette, line 542: Remove the debugging statement here, please.

rcl_yaml_param_parser/test/test_parse_yaml.cpp
```diff
@@ -497,6 +497,54 @@ TEST(test_file_parser, maximum_number_parameters) {
...
+    EXPECT_FALSE(res);
```
clalancette, line 543: Hm, do we really expect this to fail?  Shouldn't this succeed, but make the parameter a string rather than a double?

rcl_yaml_param_parser/test/special_float2.yaml
```diff
@@ -0,0 +1,8 @@
+# config/test_yaml
+---
+
+lidar_ns:
+  lidar_1:
+    ros__parameters:
+      test_nan: [1.1, 2.2, nananana]
```
clalancette, line 7: I think I'm really looking for `.nananan`, which should be parsed as a string, rather than a double.

rcl_yaml_param_parser/test/special_float2.yaml
```diff
@@ -0,0 +1,8 @@
...
+      test_nan2: [3.3, nanananan]
```
fujitatomoya, line 8: what about `infxxx` and `.infxxx`?

rcl_yaml_param_parser/src/parser.c
```diff
@@ -1105,7 +1106,23 @@ static void * get_value(
...
+    char * tmp_value = NULL;
+    if ((0 == strcasecmp(value, ".nan")) ||
+      (0 == strcasecmp(value, ".inf")) ||
+      (0 == strcasecmp(value, "+.inf")) ||
+      (0 == strcasecmp(value, "-.inf")) ||
+      (0 == strcasecmp(value, ".infinity")) ||
+      (0 == strcasecmp(value, "+.infinity")) ||
+      (0 == strcasecmp(value, "-.infinity")))
+    {
+      tmp_value = rcutils_strdup(value, allocator);
```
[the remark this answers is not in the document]
Ada-King: Sorry about this low-level mistake, i will fix it.

rcl_yaml_param_parser/test/test_parse_yaml.cpp
```diff
@@ -497,6 +497,54 @@ TEST(test_file_parser, maximum_number_parameters) {
...
+    });
+    ASSERT_TRUE(rcutils_exists(path)) << "No test YAML file found at " << path;
+    rcl_params_t * params_hdl = rcl_yaml_node_struct_init(allocator);
+    ASSERT_TRUE(NULL != params_hdl) << rcutils_get_error_string().str;
+    OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+    {
+      rcl_yaml_node_struct_fini(params_hdl);
+    });
+    bool res = rcl_parse_yaml_file(path, params_hdl);
+    EXPECT_TRUE(res) << rcutils_get_error_string().str;
```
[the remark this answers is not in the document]
Ada-King: Ok, got it.

rcl_yaml_param_parser/test/test_parse_yaml.cpp
Ada-King: Ok

rcl_yaml_param_parser/test/special_float2.yaml
Ada-King: That's my misreading, i will change it.

rcl_yaml_param_parser/test/special_float2.yaml
Ada-King: I will consider this case.

rcl_yaml_param_parser/src/parser.c
```diff
@@ -1105,7 +1106,22 @@ static void * get_value(
   {
     errno = 0;
     endptr = NULL;
-    dval = strtod(value, &endptr);
+    const char * vernier_ptr = NULL;
```
clalancette, line 1109: Just a question: what's a `vernier_ptr`?  I haven't heard that term before.

rcl_yaml_param_parser/test/special_float.yaml
```diff
@@ -0,0 +1,4 @@
+test_node:
+    ros__parameters:
+        isstring: [string, .nananan, .infinf]
+        nan_inf: [1.1, 2.2, .nan, .NAN, .inf, +.inf, -.INF, +.infinity, -.INFINITY]
```
clalancette, line 4: Yes!  These look like good tests.

rcl_yaml_param_parser/test/test_parse_yaml.cpp
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <stdio.h>
+#include <math.h>
```
clalancette, line 16: While you are here, please remove `stdio.h` (it is no longer used once you rebase), and also change this to `#include <cmath>` (since this is a C++ file).

rcl_yaml_param_parser/test/test_parse_yaml.cpp
```diff
@@ -497,6 +498,51 @@ TEST(test_file_parser, maximum_number_parameters) {
...
+    "test_node", "nan_inf", params_hdl);
+  ASSERT_TRUE(NULL != param_value) << rcutils_get_error_string().str;
+  ASSERT_TRUE(NULL != param_value->double_array_value);
+  ASSERT_EQ(9U, param_value->double_array_value->size);
+  EXPECT_TRUE(isnan(param_value->double_array_value->values[2]));
+  EXPECT_TRUE(isnan(param_value->double_array_value->values[3]));
+  EXPECT_TRUE(isinf(param_value->double_array_value->values[4]));
+  EXPECT_TRUE(isinf(param_value->double_array_value->values[5]));
+
+  rcl_yaml_node_struct_print(params_hdl);
```
clalancette, line 543: Can we get rid of this print?  It looks like debugging output.

fujitatomoya (approved):
LGTM with @clalancette 's comments.

rcl_yaml_param_parser/src/parser.c
Ada-King: > I haven't heard that term before.

Me neither :). The pointer used to iterate the `value` a bit like a cursor or something, so i named it like this. Now, i prefer to name it `iter_ptr`, which i think it's more appropriate.

rcl_yaml_param_parser/test/test_parse_yaml.cpp
Ada-King: Ok

rcl_yaml_param_parser/test/test_parse_yaml.cpp
Ada-King: sure

Ada-King:
> Thanks for iterating on this! This is exactly what I was looking for.
> 
> I have a few more minor comments inline, and this also needs to be rebased. Once that happens, I'll run CI on this.

I think this is ready to be merged, please has a re-review. Thanks.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12030/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7056/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9806/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11963/)

clalancette:
So, Windows is failing because it doesn't like `strcasecmp`.  I'm not entirely sure how to solve that, but it needs further looking into.

fujitatomoya:
what about using `stricmp` with case insensitivity instead?

brawner:
It sounds like this PR can be merged as soon as it passes CI. #754 is not blocking anything at the moment, so I'm happy to rebase it on top of this PR.

clalancette:
> It sounds like this PR can be merged as soon as it passes CI. #754 is not blocking anything at the moment, so I'm happy to rebase it on top of this PR.

Sounds good, thanks @brawner

Ada-King:
@clalancette @fujitatomoya 
Maybe we can define a macro for the windows platform at the beginning of the `parser.c` file
```
#ifdef _WIN32
#define strcasecmp stricmp
#endif
```

rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h
```diff
@@ -32,6 +32,10 @@
 #include "rcl_yaml_param_parser/types.h"
 #include "rcl_yaml_param_parser/visibility_control.h"
 
+#ifdef _WIN32
+#define strcasecmp stricmp
+#endif
+
```
fujitatomoya, line 38: I think that there is no standard case insensitive string compare in C. but I am negative on this, this will cause complication and dependency especially for maintenance. (at least i believe that this is not the right place.) Instead of this, we would create `rcutils_strcasecmp` based on `strcasecmp` and `stricmp`? Any thoughts?

rcl_yaml_param_parser/src/parser.c
```diff
@@ -1105,7 +1106,22 @@ static void * get_value(
   {
     errno = 0;
     endptr = NULL;
-    dval = strtod(value, &endptr);
+    const char * iter_ptr = NULL;
+    if ((0 == strcasecmp(value, ".nan")) ||
```
iuhilnehc-ynos, line 1110: @Ada-King @clalancette @fujitatomoya 

Actually, I don't think we should support `strcasecmp`, because from the `yaml` spec([1.1](https://yaml.org/type/float.html) or [1.2](https://yaml.org/spec/1.2/spec.html)), 
it only supports
```
 [-+]? ( \.inf | \.Inf | \.INF )	 tag:yaml.org,2002:float (Infinity)
  \.nan | \.NaN | \.NAN       tag:yaml.org,2002:float (Not a number)
```

rcl_yaml_param_parser/test/test_parse_yaml.cpp
```diff
@@ -497,6 +498,49 @@ TEST(test_file_parser, maximum_number_parameters) {
...
+  EXPECT_STREQ(".nananan", param_value->string_array_value->data[1]);
+  EXPECT_STREQ(".infinf", param_value->string_array_value->data[2]);
+  param_value = rcl_yaml_node_struct_get(
+    "test_node", "nan_inf", params_hdl);
+  ASSERT_TRUE(NULL != param_value) << rcutils_get_error_string().str;
+  ASSERT_TRUE(NULL != param_value->double_array_value);
+  ASSERT_EQ(9U, param_value->double_array_value->size);
+  EXPECT_TRUE(std::isnan(param_value->double_array_value->values[2]));
+  EXPECT_TRUE(std::isnan(param_value->double_array_value->values[3]));
+  EXPECT_TRUE(std::isinf(param_value->double_array_value->values[4]));
```
iuhilnehc-ynos, line 540: @Ada-King 

About inf, there are positive infinity and negative infinity.
`isinf(x)      returns 1 if x is positive infinity, and -1 if x is negative infinity.`


Maybe it should be 
```
EXPECT_TRUE(std::isinf(param_value->double_array_value->values[4]) == 1);
EXPECT_TRUE(std::isinf(param_value->double_array_value->values[5]) == -1);
```

rcl_yaml_param_parser/src/parser.c
fujitatomoya: So you mean just using `strcmp` against those should be okay? Am i understanding correctly?

rcl_yaml_param_parser/test/test_parse_yaml.cpp
Ada-King: Ok, that make sense

rcl_yaml_param_parser/src/parser.c
iuhilnehc-ynos: @fujitatomoya 
Yes.

fujitatomoya: that makes sense to me. thanks

iuhilnehc-ynos: No idea if user will share the yaml configuration with other projects. So I think it's better to implement these features by following the yaml spec.

rcl_yaml_param_parser/test/test_parse_yaml.cpp
Ada-King: The [type](https://en.cppreference.com/w/cpp/numeric/math/isinf) of return value in c++ is `bool`, so the current test is sufficient.

Ada-King:
@fujitatomoya 
Although the [PR](https://github.com/ros2/rcutils/pull/280) is not used here, but i think it's still worth doing.

@clalancette friendly ping
I have modified the code based on above comments, Can you help to merge this :) ?

rcl_yaml_param_parser/src/parser.c
```diff
@@ -1105,7 +1106,27 @@ static void * get_value(
...
+      (0 == strcmp(value, "+.Inf")) ||
+      (0 == strcmp(value, "+.INF")) ||
+      (0 == strcmp(value, "-.inf")) ||
+      (0 == strcmp(value, "-.Inf")) ||
+      (0 == strcmp(value, "-.INF")))
+    {
+      for (iter_ptr = value; !isalpha(*iter_ptr); ) {
+        iter_ptr += 1;
+      }
+      dval = strtod(iter_ptr, &endptr);
```
iuhilnehc-ynos, line 1126: I think we should set a minus flag if it exists(-.inf, -.Inf and -.INF), and then set `dval=-dval` if the flag is true.
In my opinion, "-.inf" is different from "+.inf". (at least, ros2 param shows the different value, such as '-inf' and 'inf'.)

rcl_yaml_param_parser/test/test_parse_yaml.cpp
iuhilnehc-ynos: Sorry, It's my bad (not to check the c++ version).
I will use c version `isinf` to check the result accurately, anyway, it's up to you.

rcl_yaml_param_parser/src/parser.c
```diff
@@ -1105,7 +1106,30 @@ static void * get_value(
...
+      (0 == strcmp(value, ".NAN")) ||
+      (0 == strcmp(value, ".inf")) ||
+      (0 == strcmp(value, ".Inf")) ||
+      (0 == strcmp(value, ".INF")) ||
+      (0 == strcmp(value, "+.inf")) ||
+      (0 == strcmp(value, "+.Inf")) ||
+      (0 == strcmp(value, "+.INF")) ||
+      (0 == strcmp(value, "-.inf")) ||
+      (0 == strcmp(value, "-.Inf")) ||
+      (0 == strcmp(value, "-.INF")))
```
clalancette, line 1121: Once we get https://github.com/ros2/rcutils/pull/280 in, we can change this code to use `rcutils_strcasecmp`.

Ada-King: > Once we get [ros2/rcutils#280](https://github.com/ros2/rcutils/pull/280) in, we can change this code to use `rcutils_strcasecmp`.

@clalancette 
FYI (https://github.com/ros2/rcl/pull/741#discussion_r475394558)
The yaml spec only allows specific formats.

Besides, the existed build error doesn't seem to be caused by this PR.

Ada-King: I checked it, the build error result from [#727](https://github.com/ros2/rcl/pull/727), the [rcutils](https://github.com/ros2/rcutils/pull/264/files) dependency has been merged, but build system still based on ros-rolling-rcutils: 1.1.0, which is the problem.

Anyway, this has no effect on this merge process.

clalancette:

brawner:
The refactoring PR #754 was merged in order to make progress on adding in coverage unit tests. Rebasing this PR on master should be fairly straightforward. The modified logic in `parser.c` has been moved to `parse.c` and everything else can remain the same. Here is a draft PR with this PR's commits squashed and rebased into a single commit on top of master. Feel free to cherry-pick that commit, or just use it as an example. This logic also passes all the unit tests I'm adding in #771, #766, and #772.

https://github.com/ros2/rcl/pull/781

fujitatomoya:
@brawner 

thanks for opening and rebasing the PR, i will check https://github.com/ros2/rcl/pull/781 once again just in case.

clalancette:
I'm going to close this one out in favor of #781; we can continue the discussion over there.

PR #743, opened 2020-08-12
Fix bug error handling get_param_files

Blast545:
When the passed allocator fails, the function `rcl_arguments_get_param_files` fails to exit cleanly. The changed clause "break" caused memory not to be deallocated. The pointers changed caused segfaults.

I added a "bomb_allocator" required to reproduce the error. This can be used for other tests within `rcl`.

### DIFF
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -94,7 +94,7 @@ rcl_arguments_get_param_files(
       // deallocate allocated memory
       for (int r = i; r >= 0; --r) {
         if (NULL == (*parameter_files)[r]) {
-          break;
+          continue;
         }
         allocator.deallocate((*parameter_files)[r], allocator.state);
       }
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -93,9 +93,6 @@ rcl_arguments_get_param_files(
     if (NULL == (*parameter_files)[i]) {
       // deallocate allocated memory
       for (int r = i; r >= 0; --r) {
-        if (NULL == (*parameter_files)[r]) {
-          continue;
-        }
         allocator.deallocate((*parameter_files)[r], allocator.state);
       }
       allocator.deallocate((*parameter_files), allocator.state);
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -96,7 +96,7 @@ rcl_arguments_get_param_files(
         if (NULL == (*parameter_files)[r]) {
           break;
         }
-        allocator.deallocate((*parameter_files)[r]), allocator.state);
+        allocator.deallocate((*parameter_files)[r], allocator.state);
       }
       allocator.deallocate((*parameter_files), allocator.state);
       (*parameter_files) = NULL;
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -93,10 +93,10 @@ rcl_arguments_get_param_files(
     if (NULL == (*parameter_files)[i]) {
       // deallocate allocated memory
       for (int r = i; r >= 0; --r) {
-        if (NULL == (*parameter_files[r])) {
+        if (NULL == (*parameter_files)[r]) {
           break;
         }
-        allocator.deallocate((*parameter_files[r]), allocator.state);
+        allocator.deallocate((*parameter_files)[r]), allocator.state);
       }
       allocator.deallocate((*parameter_files), allocator.state);
       (*parameter_files) = NULL;
--- a/rcl/test/rcl/allocator_testing_utils.h
+++ b/rcl/test/rcl/allocator_testing_utils.h
@@ -87,6 +87,80 @@ set_failing_allocator_is_failing(rcutils_allocator_t & failing_allocator, bool s
   ((__failing_allocator_state *)failing_allocator.state)->is_failing = state;
 }
 
+typedef struct __time_bomb_allocator_state
+{
+  int count_until_failure;
+} __time_bomb_allocator_state;
+
+void * time_bomb_malloc(size_t size, void * state)
+{
+  __time_bomb_allocator_state * time_bomb_state = (__time_bomb_allocator_state *)state;
+  if (time_bomb_state->count_until_failure >= 0 &&
+      time_bomb_state->count_until_failure-- == 0)
+  {
+    return nullptr;
+  }
+  return rcutils_get_default_allocator().allocate(size, rcutils_get_default_allocator().state);
+}
+
+void *
+time_bomb_realloc(void * pointer, size_t size, void * state)
+{
+  __time_bomb_allocator_state * time_bomb_state = (__time_bomb_allocator_state *)state;
+  if (time_bomb_state->count_until_failure >= 0 &&
+      time_bomb_state->count_until_failure-- == 0)
+  {
+    return nullptr;
+  }
+  return rcutils_get_default_allocator().reallocate(
+    pointer, size, rcutils_get_default_allocator().state);
+}
+
+void
+time_bomb_free(void * pointer, size_t size, void * state)
+{
+  __time_bomb_allocator_state * time_bomb_state = (__time_bomb_allocator_state *)state;
+  if (time_bomb_state->count_until_failure >= 0 &&
+      time_bomb_state->count_until_failure-- == 0)
+  {
+    return;
+  }
+  rcutils_get_default_allocator().deallocate(pointer, rcutils_get_default_allocator().state);
+}
+
+void *
+time_bomb_calloc(size_t number_of_elements, size_t size_of_element, void * state)
+{
+  __time_bomb_allocator_state * time_bomb_state = (__time_bomb_allocator_state *)state;
+  if (time_bomb_state->count_until_failure >= 0 &&
+      time_bomb_state->count_until_failure-- == 0)
+  {
+    return nullptr;
+  }
+  return rcutils_get_default_allocator().zero_allocate(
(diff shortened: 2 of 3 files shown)


rcl/test/rcl/allocator_testing_utils.h
```diff
@@ -87,6 +87,80 @@ set_failing_allocator_is_failing(rcutils_allocator_t & failing_allocator, bool s
   ((__failing_allocator_state *)failing_allocator.state)->is_failing = state;
 }
 
+typedef struct __time_bomb_allocator_state
```
ivanpauno, line 90: Don't use `__` as a prefix (I can try to find the docs, but those symbols are reserved according to some c++ guideline/standard/etc).

rcl/src/rcl/arguments.c
```diff
@@ -93,10 +93,10 @@ rcl_arguments_get_param_files(
     if (NULL == (*parameter_files)[i]) {
       // deallocate allocated memory
       for (int r = i; r >= 0; --r) {
-        if (NULL == (*parameter_files[r])) {
-          break;
+        if (NULL == (*parameter_files)[r]) {
+          continue;
         }
```
ivanpauno, line 98: You can take advantage that deallocating `NULL` is a noop, that makes things easier.

```suggestion
```

rcl/test/rcl/allocator_testing_utils.h
```diff
@@ -87,6 +87,80 @@ set_failing_allocator_is_failing(rcutils_allocator_t & failing_allocator, bool s
...
+  }
+  return rcutils_get_default_allocator().zero_allocate(
+    number_of_elements, size_of_element, rcutils_get_default_allocator().state);
+}
+
+static inline rcutils_allocator_t
+get_time_bombed_allocator(void)
+{
+  static __time_bomb_allocator_state state;
+  state.count_until_failure = 1;
```
ivanpauno, line 148: why 1?
I understand it works now, but it's heavily relying on an implementation detail.

Tests that rely on an implementation detail are harmful because when you refactor the code they either stop being useful and continue passing without checking anything useful, or because they start failing and you don't understand even what's the intention of the test.

rcl/test/rcl/allocator_testing_utils.h
```diff
@@ -87,6 +87,80 @@ set_failing_allocator_is_failing(rcutils_allocator_t & failing_allocator, bool s
...
+    return nullptr;
+  }
+  return rcutils_get_default_allocator().zero_allocate(
+    number_of_elements, size_of_element, rcutils_get_default_allocator().state);
+}
+
+static inline rcutils_allocator_t
+get_time_bombed_allocator(void)
+{
+  static __time_bomb_allocator_state state;
```
ivanpauno, line 147: why static?

rcl/src/rcl/arguments.c
Blast545: Now that I think about it, the iteration stops on the first NULL, so actually this check is not needed because previous allocated elements are guaranteed to be not NULL.

rcl/test/rcl/allocator_testing_utils.h
Blast545: I didn't think much about the init value, the expected usage for this allocator involves making a call like `set_time_bombed_allocator_count(&alloc, count);`. Do you suggest me changing the default value to be 0 instead? (and fail in the first call).

ivanpauno: AFAIU, If you put zero it won't fail where you want, but I would fail here https://github.com/ros2/rcl/blob/632f0ac8c75888f82521587314fabaec3d940c9e/rcl/src/rcl/arguments.c#L86

rcl/test/rcl/allocator_testing_utils.h
Blast545: I can change it. Please note though, I did it like this to match the convention used for the other failing allocator in this file, should I open an issue to change that?

rcl/test/rcl/allocator_testing_utils.h
Blast545: See d1c4cb0

rcl/test/rcl/allocator_testing_utils.h
Blast545: With this PR it fails for the three `BAD_ALLOC` cases, please see:
https://github.com/ros2/rcl/blob/632f0ac8c75888f82521587314fabaec3d940c9e/rcl/test/rcl/test_arguments.cpp#L1054-L1066

And you can see I called the `set_time_bombed_allocator_count` each time.

rcl/test/rcl/allocator_testing_utils.h
Blast545: Removed with 9220f89

ivanpauno: >  I did it like this to match the convention used for the other failing allocator in this file, should I open an issue to change that?

If you want to fix that in another PR that's fine.

rcl/test/rcl/allocator_testing_utils.h
ivanpauno: ok, I think it's a bit fragile, but not too bad.

rcl/test/rcl/allocator_testing_utils.h
Blast545: Actually I put it back with a034d74. This is needed to keep the variable state that's modified with the other functions of the file.

Blast545:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11906/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6950/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9701/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11840/)

rcl/test/rcl/allocator_testing_utils.h
```diff
@@ -87,6 +87,80 @@ set_failing_allocator_is_failing(rcutils_allocator_t & failing_allocator, bool s
...
+get_time_bombed_allocator(void)
+{
+  time_bomb_allocator_state state;
+  state.count_until_failure = 1;
+  auto time_bombed_allocator = rcutils_get_default_allocator();
+  time_bombed_allocator.allocate = time_bomb_malloc;
+  time_bombed_allocator.deallocate = time_bomb_free;
+  time_bombed_allocator.reallocate = time_bomb_realloc;
+  time_bombed_allocator.zero_allocate = time_bomb_calloc;
+  time_bombed_allocator.state = &state;
```
ivanpauno, line 154: so, state goes out of scope here, and you're storing a pointer to it.
Then, you will have use the pointer when it's already invalid.

I see why you were using an static, but IMO that's a bit hacky.
I think that the best thing would be to allocate the state in the heap, and then deallocate it when we're done with the allocator

rcl/test/rcl/allocator_testing_utils.h
```diff
@@ -87,6 +87,80 @@ set_failing_allocator_is_failing(rcutils_allocator_t & failing_allocator, bool s
   ((__failing_allocator_state *)failing_allocator.state)->is_failing = state;
 }
 
+typedef struct time_bomb_allocator_state
+{
+  int count_until_failure;
+} time_bomb_allocator_state;
+
+void * time_bomb_malloc(size_t size, void * state)
```
ivanpauno, line 95: All the functions defined in a header should be marked `static` or `inline`.
In this case it doesn't matter much, as this only used in tests. `static` is fine

Blast545: See 1891ff0

rcl/test/rcl/allocator_testing_utils.h
Blast545: I agree it's a bit hacky, but as it is a testing util I personally don't think the extra heap management is required

Blast545:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11912/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6956/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9706/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11847/)

rcl/test/rcl/allocator_testing_utils.h
ivanpauno: > I agree it's a bit hacky, but as it is a testing util I personally don't think the extra heap management is required

Well, you can readd the `static` keyword in the state if you want.
But the current code is broken, because you're storing a pointer to an object that not longer exists.

Blast545: I added it again, the code CI was all red because I set up badly the job configuration

Blast545:

Blast545:
Tier 1 CI green + 1 review, merging! @ivanpauno Do you think this should be backported? as it fixes a bug that caused segfault on allocator error.

ivanpauno:
> Tier 1 CI green + 1 review, merging! @ivanpauno Do you think this should be backported? as it fixes a bug that caused segfault on allocator error.

Sure.
I'm curious on how this caused a segfault. We were definitely leaking, but I don't see how a segfault was caused.

Blast545:
It seems the `(*parameter_files[r])` accessed bad memory

ivanpauno:
> It seems the `(*parameter_files[r])` accessed bad memory

Oh god, I didn't realize of the precedence order error when reviewing.
I'm really concerned about people using `char ***` as an argument in public API, that's really a bad interface :confused: .

PR #748, opened 2020-08-14
Fix allocation arguments copy

Blast545:
When using `rcl_arguments_fini` on an interrupted copied parsed arguments, the call to 
`args->impl->allocator.deallocate(args->impl->parameter_files[p], args->impl->allocator.state);` 

Causes for memory that was not allocated, but it is not set to `NULL`, this solves this problem.

Added a test to verify this problem. Although these bomb allocator tests are somewhat fragile, they have proven to be useful to find this kind of errors, I wonder if the tests for bad allocation could be written to iterate until the first `RCL_RET_OK` is found instead of iterating until a fixed iteration value.

### DIFF
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -886,8 +886,8 @@ rcl_arguments_copy(
 
   // Copy parameter files
   if (args->impl->num_param_files_args) {
-    args_out->impl->parameter_files = allocator.allocate(
-      sizeof(char *) * args->impl->num_param_files_args, allocator.state);
+    args_out->impl->parameter_files = allocator.zero_allocate(
+      args->impl->num_param_files_args, sizeof(char *), allocator.state);
     if (NULL == args_out->impl->parameter_files) {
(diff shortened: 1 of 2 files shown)


Blast545:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11955/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6996/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9746/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11896/)

Blast545:
Rerunning CI before merging:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11984/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7016/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9768/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11923/)

PR #751, opened 2020-08-18
Complete rcl enclave validation API coverage.

hidmic:
Precisely what the title says. Bumps coverage above 98%.

### DIFF
--- a/rcl/test/rcl/test_validate_enclave_name.cpp
+++ b/rcl/test/rcl/test_validate_enclave_name.cpp
@@ -65,6 +65,8 @@ TEST(TestValidateEnclaveName, test_validate) {
         return RMW_RET_OK;
       });
 
+    // When applying RMW namespace validation rules, an enclave name may be too
+    // long for an RMW namespace but not necessarily for an enclave name.
     EXPECT_EQ(
       RCL_RET_OK,
       rcl_validate_enclave_name("/foo/baz", &validation_result, &invalid_index));
--- a/rcl/test/rcl/test_validate_enclave_name.cpp
+++ b/rcl/test/rcl/test_validate_enclave_name.cpp
@@ -17,10 +17,16 @@
 #include <string>
 #include <vector>
 
+#include "rcutils/snprintf.h"
+
 #include "rcl/rcl.h"
 #include "rcl/validate_enclave_name.h"
 
 #include "rcl/error_handling.h"
+
+#include "rmw/validate_namespace.h"
+
+#include "../mocking_utils/patch.hpp"
 
 TEST(TestValidateEnclaveName, test_validate) {
   int validation_result;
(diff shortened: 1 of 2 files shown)


rcl/test/rcl/test_validate_enclave_name.cpp
```diff
@@ -47,6 +53,57 @@ TEST(TestValidateEnclaveName, test_validate) {
...
+          *index = 0u;
+        }
+        *result = RMW_NAMESPACE_INVALID_TOO_LONG;
+        return RMW_RET_OK;
+      });
+
+    EXPECT_EQ(
+      RCL_RET_OK,
+      rcl_validate_enclave_name("/foo/baz", &validation_result, &invalid_index));
+    EXPECT_EQ(RCL_ENCLAVE_NAME_VALID, validation_result);
```
Blast545, line 71: Would be good to add
```suggestion
    EXPECT_EQ(RCL_ENCLAVE_NAME_VALID, validation_result);
    EXPECT_EQ(0u, invalid_index);
```

rcl/test/rcl/test_validate_enclave_name.cpp
```diff
@@ -47,6 +53,57 @@ TEST(TestValidateEnclaveName, test_validate) {
     RCL_RET_OK,
     rcl_validate_enclave_name("/foo/bar", &validation_result, &invalid_index));
   EXPECT_EQ(RCL_ENCLAVE_NAME_VALID, validation_result);
+
+  {
+    auto mock = mocking_utils::patch(
```
Blast545, line 58: Please add a comment describing that the `rmw` can return "TOO_LONG` and still be a valid case, otherwise the tests on its own looks odd later with the returned `RCL_ENCLAVE_NAME_VALID`

rcl/test/rcl/test_validate_enclave_name.cpp
```diff
@@ -47,6 +53,57 @@ TEST(TestValidateEnclaveName, test_validate) {
...
+    EXPECT_EQ(
+      RCL_RET_OK,
+      rcl_validate_enclave_name("/foo/baz", &validation_result, &invalid_index));
+    EXPECT_EQ(RCL_ENCLAVE_NAME_VALID, validation_result);
+  }
+}
+
+TEST(TestValidateEnclaveName, test_validate_on_internal_error) {
+  int validation_result;
+  size_t invalid_index;
```
Blast545, line 77: Should these be initialized?

rcl/test/rcl/test_validate_enclave_name.cpp
hidmic: Should we? Even when it's provided, Index is set if validation fails. API says nothing about its value if validation succeeds.

rcl/test/rcl/test_validate_enclave_name.cpp
hidmic: We could, but these are dummy variables. We have no use for their values.

rcl/test/rcl/test_validate_enclave_name.cpp
hidmic: Sounds good. See 39ef45a.

hidmic:
CI up to `rcl`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11996/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7025/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9776/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11933/)

PR #762, opened 2020-08-25
Cleanup rcl_get_secure_root() implementation.

hidmic:
Namely,
* Do not assume allocation will succeed.
* Do not leak allocated memory on failure.

### DIFF
--- a/rcl/src/rcl/security.c
+++ b/rcl/src/rcl/security.c
@@ -132,36 +132,35 @@ char * rcl_get_secure_root(
   const char * name,
   const rcl_allocator_t * allocator)
 {
-  bool ros_secure_enclave_override = true;
+  RCL_CHECK_ARGUMENT_FOR_NULL(name, NULL);
+  RCL_CHECK_ALLOCATOR_WITH_MSG(allocator, "allocator is invalid", return NULL);
 
   // find out if either of the configuration environment variables are set
   const char * env_buf = NULL;
-  if (NULL == name) {
-    return NULL;
-  }
-  const char * get_env_error_str = NULL;
+
   // check if enclave override environment variable is empty
-  get_env_error_str = rcutils_get_env(ROS_SECURITY_ENCLAVE_OVERRIDE, &env_buf);
+  const char * get_env_error_str = rcutils_get_env(ROS_SECURITY_ENCLAVE_OVERRIDE, &env_buf);
   if (NULL != get_env_error_str) {
-    RCUTILS_LOG_ERROR("rcutils_get_env failed: %s\n", get_env_error_str);
+    RCL_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "rcutils_get_env failed: %s", get_env_error_str);
     return NULL;
   }
-  if (!env_buf) {
-    return NULL;
+
+  char * ros_secure_enclave_override_env = NULL;
+  if (0 != strcmp("", env_buf)) {
+    ros_secure_enclave_override_env = rcutils_strdup(env_buf, *allocator);
+    if (NULL == ros_secure_enclave_override_env) {
+      RCL_SET_ERROR_MSG("failed to duplicate enclave override string");
+      return NULL;
+    }
   }
-  if (0 == strcmp("", env_buf)) {
-    ros_secure_enclave_override = false;
-  }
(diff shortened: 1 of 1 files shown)


rcl/src/rcl/security.c
```diff
@@ -132,36 +132,35 @@ char * rcl_get_secure_root(
...
-  const char * get_env_error_str = NULL;
+
   // check if enclave override environment variable is empty
-  get_env_error_str = rcutils_get_env(ROS_SECURITY_ENCLAVE_OVERRIDE, &env_buf);
+  const char * get_env_error_str = rcutils_get_env(ROS_SECURITY_ENCLAVE_OVERRIDE, &env_buf);
   if (NULL != get_env_error_str) {
-    RCUTILS_LOG_ERROR("rcutils_get_env failed: %s\n", get_env_error_str);
-    return NULL;
-  }
-  if (!env_buf) {
```
ivanpauno, line 149: why not checking `env_buf`?

rcl/src/rcl/security.c
```diff
@@ -132,36 +132,35 @@ char * rcl_get_secure_root(
...
 
   // check if keystore environment variable is empty
+  env_buf = NULL;
   get_env_error_str = rcutils_get_env(ROS_SECURITY_KEYSTORE_VAR_NAME, &env_buf);
   if (NULL != get_env_error_str) {
-    RCUTILS_LOG_ERROR("rcutils_get_env failed: %s\n", get_env_error_str);
-    allocator->deallocate(ros_secure_enclave_override_env, allocator->state);
-    return NULL;
-  }
-  if (!env_buf) {
```
ivanpauno, line 164: same

ivanpauno:
Adding a `cleanup` label and doing `goto cleanup` instead of returning may also improve readability (that will require defining some variables earlier).

rcl/src/rcl/security.c
hidmic: If `rcutils_get_env` succeeds (i.e. it returns `nullptr`), `env_buf` won't ever be `nullptr` (see [here](https://github.com/ros2/rcutils/blob/444fa85900f71b2755a52cac859dc3dd44e6ce82/include/rcutils/get_env.h#L26)). It does  no harm to check, but it creates an unnecessary branch. And I'd rather not test it by mocking `rcutils_get_env` as if it was broken.

rcl/src/rcl/security.c
hidmic: Same as above.

rcl/src/rcl/security.c
ivanpauno: > It does no harm to check, but it creates an unnecessary branch. And I'd rather not test it by mocking rcutils_get_env as if it was broken.

I prefer checking, specially because passing `NULL` to `rcutils_strdup` is UB.

hidmic: > It does no harm to check,

Actually, let me back a up a bit. I won't die on this hill, but I think I've had this discussion before (with @clalancette maybe). And the same rationale applies across the code base.

If `rcutils_get_env` returns `NULL` after success, that means it stopped working correctly (and tests did not caught it). I don't think we want to treat that as a possible error condition. Moreover, it's not only that passing a `NULL` pointer to `rcutils_strdup` is UB -- passing a pointer to anything that's not a valid  C-string to `rcutils_strdup` is UB. 

So, why guarding ourselves against a `NULL` pointer here? A correct `rcutils_get_env` implementation will never yield a `NULL` pointer and, unlike other invalid pointers, for `NULL` pointers the OS guarantees a program abort on dereference.

ivanpauno: ok, your point sounds reasonable.

hidmic:
@ivanpauno PTAL

hidmic:
CI up to `rcl`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12066/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7087/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9839/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11997/)

hidmic:
Alright, going in! Thanks for the reviews and suggestions!

PR #777, opened 2020-08-31
Bump rcl arguments' API test coverage.

hidmic:
Using fault injection. Line coverage for `arguments.c` goes up to 95.1%, global `rcl`'s goes up to 94.3%

PR #778, opened 2020-08-31
Fix rcl arguments' API memory leaks and bugs.

hidmic:
Precisely what the title says. All found while working on #777.

### DIFF
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -594,17 +594,20 @@ rcl_parse_arguments(
   }
 
   // Shrink remap_rules array to match number of successfully parsed rules
-  if (args_impl->num_remap_rules > 0) {
-    args_impl->remap_rules = rcutils_reallocf(
-      args_impl->remap_rules, sizeof(rcl_remap_t) * args_impl->num_remap_rules, &allocator);
-    if (NULL == args_impl->remap_rules) {
-      ret = RCL_RET_BAD_ALLOC;
-      goto fail;
-    }
-  } else {
+  if (0 == args_impl->num_remap_rules) {
     // No remap rules
     allocator.deallocate(args_impl->remap_rules, allocator.state);
     args_impl->remap_rules = NULL;
+  } else if (args_impl->num_remap_rules < argc) {
+    rcl_remap_t * new_remap_rules = allocator.reallocate(
+      args_impl->remap_rules,
+      sizeof(rcl_remap_t) * args_impl->num_remap_rules,
+      &allocator);
+    if (NULL == new_remap_rules) {
+      ret = RCL_RET_BAD_ALLOC;
+      goto fail;
+    }
+    args_impl->remap_rules = new_remap_rules;
   }
 
   // Shrink Parameter files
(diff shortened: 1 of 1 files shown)


hidmic:
CI up to `rcl`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12166/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7157/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9900/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12056/)

rcl/src/rcl/arguments.c
```diff
@@ -1800,25 +1805,31 @@ _rcl_parse_remap_rule(
...
     }
-  } else {
-    ret = rcl_lexer_lookahead2_fini(&lex_lookahead);
   }
+
+  // cleanup output rule but keep first error return code
+  if (RCL_RET_OK != rcl_remap_fini(output_rule)) {
+    RCUTILS_LOG_ERROR_NAMED(
+      ROS_PACKAGE_NAME, "Failed to fini remap rule after error occurred");
+  }
```
clalancette, line 1831: I'm not sure I agree with this logic, or, at least, I don't understand.

It looks to me with this change that we always cleanup output_rule, but that seems weird; we don't `remap_init` it in this method.  Shouldn't that be the responsibility of the caller?

In other words, it kind of looks to me like this method should be:

```
init_lexer();
if (failed_to_init_lexer)
  return fail;
parse_remap_begin_remap_rule(lexer, output_rule);
cleanup_lexer();
if (!ret) {
  if (!fini_ret)
    print_error();
  return ret;
}
```

And then the caller is responsible for `remap_fini` as appropriate.  Could you explain a little more what the idea is here?

hidmic: > It looks to me with this change that we always cleanup output_rule, but that seems weird

Only when something fails. 

> we don't remap_init it in this method.

We do, [right above](https://github.com/ros2/rcl/blob/6040298b429784ca9eb4d0404543d6211fd5f81d/rcl/src/rcl/arguments.c#L1792-L1800). Thing is, `remap_init` doesn't exist.

> Shouldn't that be the responsibility of the caller?

Right now it's not necessary. `_rcl_remap_parse_rule` takes a zero-initialized `rcl_remap_t` instance. On failure, it should stay zero-initialized.

It could be the case, but with the current `rcl_remap_t` structure and API implementation it doesn't add much value. The init/fini cycle would still be coupled to the parse API. However, if `rcl_remap_parse()` was capable of reusing `rcl_remap_t` structures (i.e. reuse past allocations), and if there was an `rcl_remap_move()` API to copy rules around cheaply, then relocating the init/fini cycle could save up quite a few unnecessary allocation/deallocation cycles. I didn't want to go that far.

clalancette: > We do, [right above](https://github.com/ros2/rcl/blob/6040298b429784ca9eb4d0404543d6211fd5f81d/rcl/src/rcl/arguments.c#L1792-L1800). Thing is, `remap_init` doesn't exist.

Oh, now I see.  OK, so that block of code is essentially `remap_init`, it just doesn't have that name.

> Right now it's not necessary. `_rcl_remap_parse_rule` takes a zero-initialized `rcl_remap_t` instance. On failure, it should stay zero-initialized.

I see.  So this PR is a cleanup of the allocated memory on failure further down.

I still find the way this is done weird.  The `rcl_remap_t` is currently an "inout" parameter, but it seems to me that it should just be an "out" parameter.  If we consider it more like that, then the caller can pass it in uninitialized.  This method would initialize it and add to it on success, and the caller would (eventually) need to free it.  On error, the parameter would still be considered "uninitialized".

The above just makes more sense to me.  That being said, I don't know how much you want to refactor this here.  So I'll leave it up to you whether you want to do that or just go with this code as-is.

hidmic: > I see. So this PR is a cleanup of the allocated memory on failure further down.

Exactly.

> I still find the way this is done weird. The rcl_remap_t is currently an "inout" parameter, but it seems to me that it should just be an "out" parameter. If we consider it more like that, then the caller can pass it in uninitialized. This method would initialize it and add to it on success, and the caller would (eventually) need to free it. On error, the parameter would still be considered "uninitialized".

That's almost exactly how it is handled right now, except that `_rcl_remap_parse_rule` does not explicitly zero-initialize the given `rcl_remap_t` structure, it assumes it is. So in a way it is an output parameter.

hidmic:
Ok, I'll move along. We can revisit arguments' and remaps implementation later on.

PR #788, opened 2020-09-03
Fix test_rcl_lifecycle

### LINKED ISSUE #787: test_rcl_lifecycle failing with cyclonedds
Example failure: http://build.ros2.org/view/Rci/job/Rci__nightly-cyclonedds_ubuntu_focal_amd64/

I think this error output is the problem:

> Not all nodes were finished before finishing the context

brawner:
This test had a few issues after digging through it when investigating #787. The root cause of #787 was that the node was not being shutdown during the maybe_fail test, but there was also a few other issues with finalization that led to memory leaks. Lastly to match the style in other tests, I reversed `EXPECT/ASSERT` so that the expected value was on the left-hand side.

If the last style fix is undesired, I can switch that back.

Fixes #787.

### DIFF
--- a/rcl_lifecycle/test/test_rcl_lifecycle.cpp
+++ b/rcl_lifecycle/test/test_rcl_lifecycle.cpp
@@ -43,53 +43,53 @@ static void * bad_realloc(void *, size_t, void *)
 
 TEST(TestRclLifecycle, lifecycle_state) {
   rcl_lifecycle_state_t state = rcl_lifecycle_get_zero_initialized_state();
-  EXPECT_EQ(state.id, 0u);
-  EXPECT_EQ(state.label, nullptr);
+  EXPECT_EQ(0u, state.id);
+  EXPECT_EQ(nullptr, state.label);
 
   rcl_allocator_t allocator = rcl_get_default_allocator();
   unsigned int expected_id = 42;
   const char expected_label[] = "label";
   rcl_ret_t ret = rcl_lifecycle_state_init(&state, expected_id, &expected_label[0], nullptr);
-  EXPECT_EQ(ret, RCL_RET_ERROR);
+  EXPECT_EQ(RCL_RET_ERROR, ret);
   rcutils_reset_error();
 
   ret = rcl_lifecycle_state_init(&state, expected_id, nullptr, &allocator);
-  EXPECT_EQ(ret, RCL_RET_ERROR);
+  EXPECT_EQ(RCL_RET_ERROR, ret);
   rcutils_reset_error();
 
   ret = rcl_lifecycle_state_init(nullptr, expected_id, &expected_label[0], &allocator);
-  EXPECT_EQ(ret, RCL_RET_ERROR);
+  EXPECT_EQ(RCL_RET_ERROR, ret);
   rcutils_reset_error();
 
   rcl_allocator_t bad_allocator = rcl_get_default_allocator();
   bad_allocator.allocate = bad_malloc;
   bad_allocator.reallocate = bad_realloc;
   ret = rcl_lifecycle_state_init(&state, expected_id, &expected_label[0], &bad_allocator);
-  EXPECT_EQ(ret, RCL_RET_ERROR);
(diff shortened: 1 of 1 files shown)


brawner:
Standard build, testing `--packages-select rcl_lifecycle`
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12213/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7196/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9939/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12104/)

cyclonedds only
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12214/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7197/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9940/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12105/)

rcl_lifecycle/test/test_rcl_lifecycle.cpp
```diff
@@ -228,148 +239,159 @@ TEST(TestRclLifecycle, state_machine) {
...
-  EXPECT_EQ(rcl_lifecycle_state_machine_is_initialized(&state_machine), RCL_RET_OK);
+  EXPECT_EQ(RCL_RET_OK, rcl_lifecycle_state_machine_is_initialized(&state_machine));
+
+  // get_state service is valid, but not change_state service
+  void * temp_function = state_machine.com_interface.srv_change_state.impl;
+  state_machine.com_interface.srv_change_state.impl = nullptr;
+  EXPECT_EQ(RCL_RET_ERROR, rcl_lifecycle_state_machine_is_initialized(&state_machine));
+  rcutils_reset_error();
+  state_machine.com_interface.srv_change_state.impl =
+    reinterpret_cast<rcl_service_impl_t *>(temp_function);
```
ivanpauno, line 338: how is this related to the PR?

rcl_lifecycle/test/test_rcl_lifecycle.cpp
```diff
@@ -228,148 +239,159 @@ TEST(TestRclLifecycle, state_machine) {
...
       allocator.state));
-  ASSERT_NE(state_machine.transition_map.transitions, nullptr);
+  ASSERT_NE(nullptr, state_machine.transition_map.transitions);
   state_machine.transition_map.transitions[0] = rcl_lifecycle_get_zero_initialized_transition();
 
-  EXPECT_EQ(rcl_lifecycle_state_machine_is_initialized(&state_machine), RCL_RET_OK);
+  EXPECT_EQ(RCL_RET_OK, rcl_lifecycle_state_machine_is_initialized(&state_machine));
+
+  // get_state service is valid, but not change_state service
+  void * temp_function = state_machine.com_interface.srv_change_state.impl;
```
ivanpauno, line 333: is this temporal a function?

ivanpauno (approved):
There are a lot of unrelated changes that make reviewing hard, but it lgtm.

rcl_lifecycle/test/test_rcl_lifecycle.cpp
brawner: This moves the check from line 302 to after `rcl_lifecycle_state_machine_is_initialized` succeeds once. That way, we know that state_machine is valid and the only thing that causes it to be invalid is by setting `srv_change_state.impl` to `NULL`. It`s not related to #787, but it is just a general fix for this unit test. I figured this PR could include several fixes at once, but if you'd like me to separate them out into multiple PRs, I'd be cool with that too.

rcl_lifecycle/test/test_rcl_lifecycle.cpp
brawner: Correct, `srv_change_state.impl` is a function pointer.

ivanpauno (approved):
lgtm

Issue #793, opened 2020-09-07
Created rmw subscription/publisher handle isn't freed while failing to get qos from rmw in rcl_subscription/publisher_init()


Barry-Xu-2018:
## Bug report

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - source
- Version or commit hash:
  - a80e55b
- DDS implementation:
  - fastrtps and cyclonedds
- Client library (if applicable):
  - rcl

#### Steps to reproduce issue

```bash
$ colcon build - --packages-up-to rcl --symlink-install -cmake-args -DCMAKE_BUILD_TYPE=Debug
$ source install/setup.bash
$ valgrind --leak-check=full ./build/rcl/test/test_subscription__rmw_fastrtps_cpp
$ valgrind --leak-check=full ./build/rcl/test/test_publisher__rmw_fastrtps_cpp
```

#### Expected behavior

No memory leak 

#### Actual behavior

```bash
$ valgrind --leak-check=full ./build/rcl/test/test_subscription__rmw_fastrtps_cpp
...
==247122== 483 (48 direct, 435 indirect) bytes in 1 blocks are definitely lost in loss record 44 of 47
==247122==    at 0x483F7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==247122==    by 0x5130F06: __default_allocate (allocator.c:37)
==247122==    by 0x5147597: rmw_allocate (allocators.c:29)
==247122==    by 0x514769F: rmw_subscription_allocate (allocators.c:76)
==247122==    by 0x50FC6B5: rmw_fastrtps_cpp::create_subscription(CustomParticipantInfo const*, rosidl_message_type_support_t const*, char const*, rmw_qos_profile_t const*, rmw_subscription_options_t const*, bool, bool) (subscription.cpp:170)
==247122==    by 0x50F9130: rmw_create_subscription (rmw_subscription.cpp:73)
==247122==    by 0x4875437: rcl_subscription_init (subscription.c:161)
==247122==    by 0x1258DC: TestSubscriptionFixture__rmw_fastrtps_cpp_test_subscription_bad_init_Test::TestBody() (test_subscription.cpp:238)
==247122==    by 0x18E689: void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (gtest.cc:2447)
[...]

Issue #740, opened 2020-08-11
Error handling often overwrites errors already set


hidmic:
## Bug report

**Required Info:**

- Operating System:
  - All Tier 1 platforms
- Installation type:
  - From source
- Version or commit hash:
  - Current [master](https://github.com/ros2/rcl/tree/f9a28d65cf2bdb6fb9373988c372caba3f6bfed5)

#### Steps to reproduce issue

Run `rcl` tests locally:
```sh
colcon test --event-handlers=console_direct+ --packages-select rcl
```

#### Expected behavior

All tests pass and no error handling error (an overwrite) is reported via `stderr`.

#### Actual behavior

All tests pass but plenty error handling errors (overwrites) are reported.

#### Additional information

These overwrites are the result of either ignoring errors set by `rcutils` or `rmw` APIs or attempting to propagate them, which isn't necessarily bad practice if it weren't for the fact that all our libraries share the same global error state.

hidmic:
IMO our main issue here is that there's no clear path forward. Even if we decide to rely on all libraries sharing the same error handling state, `rcutils` error handling API was not designed to ease error propagation, and suppressing portions of an error chain either decontextualizes an error message or obscures its origin.

hidmic:
FYI @Blast545 @brawner.

Blast545:
For future reference, discussion started here: https://github.com/ros2/rcl/pull/734#discussion_r468098092

I am not really sure what's the best way to tackle this. Thought a little on this, what if a "build" flag is set then the overwritten errors are print to stderr and if not set the error is overwritten without the warning?

hidmic:
> Thought a little on this, what if a "build" flag is set then the overwritten errors are print to stderr and if not set the error is overwritten without the warning?

Hmm, a possible outcome of this discussion is that we start setting error messages in every C function in a call chain. If we do that, then putting a message on `stderr` saying that there was an error with error handling and a previous message was overwritten will never be correct.

wjwwood:
What is supposed to happen is that when you encounter an error with an rcutils function, like `failed to duplicate string`, from within an rcl function is that you collected the rcutils error, add it to your error, clear the error state, and return with a new error state set, something like `failed to initialize publisher: failed to duplicate string`.

Any proposal to make this easier needs to (in my mind) consider two things:

- printing to stderr is often not desired (maybe the error is expected and will be ignored/suppressed)
[...]

PR #794, opened 2020-09-07
Fix memory leak in rcl_subscription_init()/rcl_publisher_init()

Barry-Xu-2018:
In rcl_subscription_init(), while rmw_subscription_get_actual_qos()
return failure, created rmw subscription handle isn't freed.
In rcl_publisher_init(), while rmw_publisher_get_actual_qos()
return failure, created rmw publisher handle isn't freed.

Signed-off-by: Barry Xu <barry.xu@sony.com>

### DIFF
--- a/rcl/src/rcl/publisher.c
+++ b/rcl/src/rcl/publisher.c
@@ -28,6 +28,7 @@ extern "C"
 #include "rcl/remap.h"
 #include "rcutils/logging_macros.h"
 #include "rcutils/macros.h"
+#include "rcutils/strdup.h"
 #include "rmw/error_handling.h"
 #include "rmw/validate_full_topic_name.h"
 #include "tracetools/tracetools.h"
@@ -205,6 +206,53 @@ rcl_publisher_init(
   goto cleanup;
 fail:
   if (publisher->impl) {
+    // Used for concatenating error messages in the fail: block.
+    // The cause or error which leads to jump to fail:
+    char * fail_error_message = NULL;
+    // The error happens in fail:
+    char * fini_error_message = NULL;
+    rcl_allocator_t default_allocator = rcl_get_default_allocator();
+
+    if (rcl_error_is_set()) {
+      fail_error_message = rcutils_strdup(rcl_get_error_string().str, default_allocator);
+      rcl_reset_error();
+    }
+
+    if (publisher->impl->rmw_handle) {
+      // If rmw_destroy_publisher fails, it will clobber the error string here.
+      // Concatenate the error strings if that happens
+      if (rmw_destroy_publisher(
+          rcl_node_get_rmw_handle(node),
+          publisher->impl->rmw_handle) == RMW_RET_ERROR)
+      {
+        if (rcl_error_is_set()) {
+          fini_error_message = rcutils_strdup(rcl_get_error_string().str, default_allocator);
+          rcl_reset_error();
+        }
+
+        RCL_SET_ERROR_MSG_WITH_FORMAT_STRING(
+          "rmw_destroy_publisher failed while handling a previous error."
+          "\nOriginal error:\n\t%s\nError encountered in rmw_destroy_publisher():\n\t%s\n",
+          fail_error_message != NULL ?
+          fail_error_message : "Failed to duplicate error while init publisher !",
+          fini_error_message != NULL ?
+          fini_error_message : "Failed to duplicate error while destory rmw publisher !");
+      }
+    }
+
+    if (!rcl_error_is_set()) {
+      RCL_SET_ERROR_MSG(
+        (fail_error_message != NULL) ?
+        fail_error_message : "Unspecified error in rcl_publisher_init() !");
+    }
+
+    if (fail_error_message != NULL) {
+      default_allocator.deallocate(fail_error_message, default_allocator.state);
+    }
+    if (fini_error_message != NULL) {
+      default_allocator.deallocate(fini_error_message, default_allocator.state);
+    }
+
     allocator->deallocate(publisher->impl, allocator->state);
     publisher->impl = NULL;
   }
--- a/rcl/src/rcl/publisher.c
+++ b/rcl/src/rcl/publisher.c
@@ -28,7 +28,6 @@ extern "C"
 #include "rcl/remap.h"
 #include "rcutils/logging_macros.h"
 #include "rcutils/macros.h"
-#include "rcutils/strdup.h"
 #include "rmw/error_handling.h"
 #include "rmw/validate_full_topic_name.h"
 #include "tracetools/tracetools.h"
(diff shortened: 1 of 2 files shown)


Barry-Xu-2018:
Fix #793

Barry-Xu-2018:
Since there is concatenating error messages, I refer to below codes

https://github.com/ros2/rcl/blob/e88b63ba3c153dec6f1e5f0cc76137a728d68e42/rcl_lifecycle/src/default_state_machine.c#L700-L735

fujitatomoya:
> Since there is concatenating error messages, I refer to below codes

I think there is discussion on that, https://github.com/ros2/rcl/issues/740.

fujitatomoya (approved):
lgtm

clalancette (changes requested):
I have to say that this code is really hard to follow now.  We were already basically at that point before, but now I think this is unmaintainable.  Also, this runs right into the problem we were discussing here: https://github.com/ros2/rcl/issues/740 (and it violates one of the principles we were trying to keep there, which is not allocating memory in the error paths).

My suggestion for going forward with this is to take a step back, and instead look at splitting this function into smaller, easier to read functions.  Each of those functions is responsible for doing one part of initialization; if it fails, that function is responsible for one part of the cleanup.

As far as chaining the error messages, it is nice, but I wouldn't actually do this until we come to some kind of agreement on https://github.com/ros2/rcl/issues/740 .

fujitatomoya:
@clalancette 

> but now I think this is unmaintainable.

at the time I think it would be better to avoid overwriting error message. but i do agree on this. it is better to keep it simple and focus on what we are fixing.

Barry-Xu-2018:
@clalancette  

Thanks for your comments.   
I agree `unmaintainable`. But according to current rcutils error handling implementation, I have no better idea to avoid overwriting error message.   

> My suggestion for going forward with this is to take a step back, and instead look at splitting this function into smaller, easier to read functions. Each of those functions is responsible for doing one part of initialization; if it fails, that function is responsible for one part of the cleanup.

Sorry. I not sure I understand your thought.    
Do you means rcl_subscription_init()/rcl_publisher_init() should be split into smaller function ?

> As far as chaining the error messages, it is nice, but I wouldn't actually do this until we come to some kind of agreement on #740

I will keep eye on #740.

clalancette:
> I agree `unmaintainable`. But according to current rcutils error handling implementation, I have no better idea to avoid overwriting error message.

Yeah, I agree.  Right now, there isn't a better way to do it.

> Sorry. I not sure I understand your thought.
> Do you means rcl_subscription_init()/rcl_publisher_init() should be split into smaller function ?

Yeah, that was what I was thinking.  It seems to me that if we split this up into smaller functions, it might be easier to read.

Another thing we can do here to move forward is to not worry about adding up error messages.  That is, if something failed, then in all likelihood the user is most interested in what caused the original error.  So if we just want to fix the memory leak, I'll suggest removing the custom code to handle the cascading errors, and just concentrate on the memory leak for now.  Then we can later think about additional errors.

Barry-Xu-2018:
@clalancette 

>> Sorry. I not sure I understand your thought.
>> Do you means rcl_subscription_init()/rcl_publisher_init() should be split into smaller function ?
>
> Yeah, that was what I was thinking. It seems to me that if we split this up into smaller functions, it might be easier to read.

Got it.

> Another thing we can do here to move forward is to not worry about adding up error messages. That is, if something failed, then in all likelihood the user is most interested in what caused the original error. So if we just want to fix the memory leak, I'll suggest removing the custom code to handle the cascading errors, and just concentrate on the memory leak for now. Then we can later think about additional errors.

Thanks for your suggestion.  
I will remove custom code on cascading errors.

rcl/src/rcl/subscription.c
```diff
@@ -193,6 +193,15 @@ rcl_subscription_init(
   goto cleanup;
 fail:
   if (subscription->impl) {
+    if (subscription->impl->rmw_handle) {
+      if (rmw_destroy_subscription(
+          rcl_node_get_rmw_handle(node),
+          subscription->impl->rmw_handle) != RMW_RET_OK)
+      {
+        RCL_SET_ERROR_MSG(rmw_get_error_string().str);
```
jacobperron, line 201: I'm pretty sure anytime we `goto fail`, an error is already set. This means we will always be overwriting the error.
Assuming the user probably cares more about the original error, I'd rather not overwrite the error here. @wjwwood @clalancette thoughts?

In any case, it's probably worth adding a TODO to handle this error cascade, referring to https://github.com/ros2/rcl/issues/740.

wjwwood: I think you're right @jacobperron, usually we just print to stderr if there's an error while cleaning up.

Alternatively you could build a new error message which contains both, like:

> Encountered error 'SECOND ERROR' while handling error 'FIRST ERROR'.

But then again you might have an error building that new error message, so you might just say `failed to destroy subscription during error handling in rcl_subscription_init()` or something similar.

rcl/src/rcl/publisher.c
```diff
@@ -205,6 +205,15 @@ rcl_publisher_init(
   goto cleanup;
 fail:
   if (publisher->impl) {
+    if (publisher->impl->rmw_handle) {
+      if (rmw_destroy_publisher(
+          rcl_node_get_rmw_handle(node),
+          publisher->impl->rmw_handle) != RMW_RET_OK)
+      {
```
wjwwood, line 212: I'd recommend this style instead:

```suggestion
      rmw_ret_t rmw_fail_ret = rmw_destroy_publisher(
        rcl_node_get_rmw_handle(node), publisher->impl->rmw_handle);
      if (RMW_RET_OK != rmw_fail_ret)
      {
```

rcl/src/rcl/subscription.c
wjwwood: I think this is what was being described above (in this issue) though.

jacobperron: +1 for just printing the rmw error string here. Same in publisher case.

clalancette (approved):
This looks much better to me, thanks.  I'm approving, but I'll wait from an additional approval from @jacobperron or @wjwwood before running CI.

rcl/src/rcl/subscription.c
wjwwood: This still is an issue for me. I'd prefer to see it print to stderr here, maybe with `RCUTILS_SAFE_FWRITE_TO_STDERR`:

https://github.com/ros2/rcutils/blob/060a68a9ef092f1ed909967ff0611a2bac5f9f18/include/rcutils/error_handling.h#L46

clalancette:
All right, I think that last commit should solve @wjwwood 's concerns.  Assuming that is the case, here is CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12738/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7698/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10451/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12684/)

jacobperron (approved):

jacobperron:
I've proposed backports for Foxy and Eloquent: https://github.com/ros2/rcl/pull/832 and https://github.com/ros2/rcl/pull/833

PR #795, opened 2020-09-09
Use valid clock in case of issue in rcl_timer_init

brawner:
This clock needs to be copied before rcl_timer_init in the case that `rcl_timer_init` fails. Previously, if it failed before it was copied, `rcl_timer_fini` would be called in `rcl_action_server_fini` that would try to access a pointer owned and already free'd by `rclcpp::Node`.
 
### DIFF
--- a/rcl_action/src/rcl_action/action_server.c
+++ b/rcl_action/src/rcl_action/action_server.c
@@ -175,8 +175,8 @@ rcl_action_server_init(
 
 // Initialize Timer
   ret = rcl_timer_init(
-    &action_server->impl->expire_timer, &action_server->impl->clock, node->context, options->result_timeout.nanoseconds,
-    NULL, allocator);
+    &action_server->impl->expire_timer, &action_server->impl->clock, node->context,
+    options->result_timeout.nanoseconds, NULL, allocator);
   if (RCL_RET_OK != ret) {
     goto fail;
   }
@@ -185,7 +185,7 @@ rcl_action_server_init(
   if (RCL_RET_OK != ret) {
     goto fail;
   }
-  
+
   // Copy action name
   action_server->impl->action_name = rcutils_strdup(action_name, allocator);
   if (NULL == action_server->impl->action_name) {
--- a/rcl_action/src/rcl_action/action_server.c
+++ b/rcl_action/src/rcl_action/action_server.c
@@ -170,9 +170,12 @@ rcl_action_server_init(
   PUBLISHER_INIT(feedback);
   PUBLISHER_INIT(status);
 
-  // Initialize Timer
+  // Copy clock
+  action_server->impl->clock = *clock;
+
+// Initialize Timer
   ret = rcl_timer_init(
-    &action_server->impl->expire_timer, clock, node->context, options->result_timeout.nanoseconds,
+    &action_server->impl->expire_timer, &action_server->impl->clock, node->context, options->result_timeout.nanoseconds,
     NULL, allocator);
   if (RCL_RET_OK != ret) {
     goto fail;
@@ -182,10 +185,7 @@ rcl_action_server_init(
   if (RCL_RET_OK != ret) {
     goto fail;
   }
-
-  // Copy clock
-  action_server->impl->clock = *clock;
-
+  
   // Copy action name
   action_server->impl->action_name = rcutils_strdup(action_name, allocator);
   if (NULL == action_server->impl->action_name) {


rcl_action/src/rcl_action/action_server.c
```diff
@@ -170,9 +170,12 @@ rcl_action_server_init(
   PUBLISHER_INIT(feedback);
   PUBLISHER_INIT(status);
 
-  // Initialize Timer
+  // Copy clock
+  action_server->impl->clock = *clock;
+
```
clalancette, line 175: This is a question to make sure I understand.  Is the purpose of copying the clock here so that if `rcl_timer_init` fails, `rcl_action_server_fini` in the `fail` block below has a valid pointer to cleanup?

clalancette (commented):
One question, but otherwise looks good to me.

rcl_action/src/rcl_action/action_server.c
brawner: Pretty much, basically. This one is a bit of a rabbit hole, so bare with me. In `rcl_timer_init`, it copies the pointer to the `clock` pointer passed to it. In `rcl_timer_fini` (called in `rcl_action_server_fini` line 235), it will try to access fields of that pointer by calling `rcutils_allocator_is_valid`, `rcl_clock_remove_jump_callback` and checking its type on the same clock pointer. Since ownership of the `clock` pointer is not supposed to be transferred in this function (rclcpp::Node still owns it), then it should be using the pointer to action_server`s clock instead during `rcl_timer_init`.

This was only found when running valgrind on rclcpp`s `test_server`. Basically, rclcpp::Node was destroying the clock before `rcl_timer_fini` was called in this action_server. Trackback below to one of the memory read errors without this fix.

```
==2946848== Invalid read of size 8
==2946848==    at 0x5896463: rcutils_allocator_is_valid (allocator.c:96)
==2946848==    by 0x57DB111: rcl_clock_remove_jump_callback (time.c:431)
==2946848==    by 0x57DCD41: rcl_timer_fini (timer.c:217)
==2946848==    by 0x49596D6: rcl_action_server_fini (action_server.c:235)
==2946848==    by 0x48E8666: rclcpp_action::ServerBase::ServerBase(std::shared_ptr<rclcpp::node_interfaces::NodeBaseInterface>, std::shared_ptr<rclcpp::node_interfaces::NodeClockInterface>, std::shared_ptr<rclcpp::node_interfaces::NodeLoggingInterface>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rosidl_action_type_support_t const*, rcl_action_server_options_t const&)::{lambda(rcl_action_server_t*)#1}::operator()(rcl_action_server_t*) const (server.cpp:90)
==2946848==    by 0x48F167F: std::_Sp_counted_deleter<rcl_action_server_t*, rclcpp_action::ServerBase::ServerBase(std::shared_ptr<rclcpp::node_interfaces::NodeBaseInterface>, std::shared_ptr<rclcpp::node_interfaces::NodeClockInterface>, std::shared_ptr<rclcpp::node_interfaces::NodeLoggingInterface>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rosidl_action_type_support_t const*, rcl_action_server_options_t const&)::{lambda(rcl_action_server_t*)#1}, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() (shared_ptr_base.h:471)
==2946848==    by 0x3432B5: std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release() (shared_ptr_base.h:155)
==2946848==    by 0x3334E2: std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count() (shared_ptr_base.h:730)
==2946848==    by 0x48F21CD: std::__shared_ptr<rcl_action_server_t, (__gnu_cxx::_Lock_policy)2>::~__shared_ptr() (shared_ptr_base.h:1169)
==2946848==    by 0x48F2211: std::shared_ptr<rcl_action_server_t>::~shared_ptr() (shared_ptr.h:103)
==2946848==    by 0x48F5CF3: rclcpp_action::ServerBaseImpl::~ServerBaseImpl() (server.cpp:35)
==2946848==    by 0x48F5D53: std::default_delete<rclcpp_action::ServerBaseImpl>::operator()(rclcpp_action::ServerBaseImpl*) const (unique_ptr.h:81)


==2946848==  Address 0xa0d7738 is 40 bytes inside a block of size 160 free'd
==2946848==    at 0x483F1CF: operator delete(void*, unsigned long) (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==2946848==    by 0x50F4571: std::_Sp_counted_ptr<rclcpp::Clock::Impl*, (__gnu_cxx::_Lock_policy)2>::_M_dispose() (shared_ptr_base.h:377)
==2946848==    by 0x3432B5: std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release() (shared_ptr_base.h:155)
==2946848==    by 0x3334E2: std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count() (shared_ptr_base.h:730)
==2946848==    by 0x50F343C: std::__shared_ptr<rclcpp::Clock::Impl, (__gnu_cxx::_Lock_policy)2>::~__shared_ptr() (shared_ptr_base.h:1169)
==2946848==    by 0x50F346E: std::shared_ptr<rclcpp::Clock::Impl>::~shared_ptr() (shared_ptr.h:103)
==2946848==    by 0x50F1212: rclcpp::Clock::~Clock() (clock.cpp:64)
==2946848==    by 0x44E865: void __gnu_cxx::new_allocator<rclcpp::Clock>::destroy<rclcpp::Clock>(rclcpp::Clock*) (new_allocator.h:153)
==2946848==    by 0x44637B: void std::allocator_traits<std::allocator<rclcpp::Clock> >::destroy<rclcpp::Clock>(std::allocator<rclcpp::Clock>&, rclcpp::Clock*) (alloc_traits.h:497)
==2946848==    by 0x4309C0: std::_Sp_counted_ptr_inplace<rclcpp::Clock, std::allocator<rclcpp::Clock>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() (shared_ptr_base.h:557)
==2946848==    by 0x3432B5: std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release() (shared_ptr_base.h:155)
==2946848==    by 0x3334E2: std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count() (shared_ptr_base.h:730)
```

brawner:
Cleaned up code to pass linters.

brawner:
Testing `--packages-select rcl_action rclcpp_action`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12249/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7226/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9969/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12142/)

rcl_action/src/rcl_action/action_server.c
clalancette: Thanks, great explanation.  That's really helpful.

clalancette (approved):

rcl_action/src/rcl_action/action_server.c
```diff
@@ -170,10 +170,13 @@ rcl_action_server_init(
   PUBLISHER_INIT(feedback);
   PUBLISHER_INIT(status);
 
-  // Initialize Timer
+  // Copy clock
+  action_server->impl->clock = *clock;
```
sloretz, line 174: @clalancette @brawner This appears to be the cause of the `tf2_ros` `test_buffer_server` test failing. The problem is `rcl_clock_t` can't be trivially copied because it stores time jump callbacks. Reverting would make the test failure go away, but I think a better solution would be to make the action server store a pointer to the clock, not the clock itself.

First the [`tf2_ros::Buffer` creates a time jump callback](https://github.com/ros2/geometry2/blob/d7276d572b65dae69e7597d4715352dae61b6aef/tf2_ros/src/buffer.cpp#L69). This causes [clock->jump_callbacks to be allocated](https://github.com/ros2/rcl/blob/fb190af758b7fae42050ddece6fa04cef91eda8f/rcl/src/rcl/time.c#L410-L417). Then the action server copies the clock (this line here). This means there are two `rcl_clock_t` instances pointing holding the same pointer in `clock->jump_callbacks` (let's call them `OriginalClock` and `ASClock`). Then the call to `rcl_timer_init()` below [adds a clock jump callback](https://github.com/ros2/rcl/blob/fb190af758b7fae42050ddece6fa04cef91eda8f/rcl/src/rcl/timer.c#L167) to `ASClock`. Since this is a realloc, the old pointer is free'd and a new pointer is returned. `ASClock->jump_callbacks` points to newly allocated memory while `OriginalClock->jump_callbacks` still points to the memory that has now been freed. The test creates a `tf2_ros::BufferServer` instance with `OriginalClock`, which [creates a timer](https://github.com/ros2/geometry2/blob/d7276d572b65dae69e7597d4715352dae61b6aef/tf2_ros/include/tf2_ros/buffer_server.h#L95) that again adds a jump callback. It tries `realloc` using `OriginalClock->jump_callbacks` but that is a use-after-free error.

brawner: @sloretz Thanks for bringing this to my attention and for the thorough writeup. I'll dig into it

sloretz: @brawner happy to. I have a branch locally that makes the action server use a pointer to the clock. Assuming tests pass locally, mind if I ping you later for a review?

PR #801, opened 2020-09-15
Fix memory leak on serialized message in test_publisher/subscription.cpp

Barry-Xu-2018:
This memory leak issue was found in #721.  

Related test:
- test_subscription__rmw_cyclonedds_cpp
- test_subscription__rmw_fastrtps_cpp 
- test_subscription__rmw_fastrtps_dynamic_cpp 
- test_publisher__rmw_cyclonedds_cpp
- test_publisher__rmw_fastrtps_cpp
- test_publisher__rmw_fastrtps_dynamic_cpp

### DIFF
--- a/rcl/test/rcl/test_publisher.cpp
+++ b/rcl/test/rcl/test_publisher.cpp
@@ -603,6 +603,9 @@ TEST_F(
   OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
   {
     test_msgs__msg__Strings__fini(&msg);
+    ASSERT_EQ(
+      RCUTILS_RET_OK,
+      rmw_serialized_message_fini(&serialized_msg)) << rcl_get_error_string().str;
   });
 
   ASSERT_TRUE(rosidl_runtime_c__String__assign(&msg.string_value, test_string));
--- a/rcl/test/rcl/test_publisher.cpp
+++ b/rcl/test/rcl/test_publisher.cpp
@@ -604,7 +604,7 @@ TEST_F(
   {
     test_msgs__msg__Strings__fini(&msg);
     ASSERT_EQ(
-      RCUTILS_RET_OK,
+      RMW_RET_OK,
       rmw_serialized_message_fini(&serialized_msg)) << rcl_get_error_string().str;
   });
 
--- a/rcl/test/rcl/test_subscription.cpp
+++ b/rcl/test/rcl/test_subscription.cpp
@@ -571,6 +571,10 @@ TEST_F(CLASSNAME(TestSubscriptionFixture, RMW_IMPLEMENTATION), test_subscription
   {
     ret = rcl_subscription_fini(&subscription, this->node_ptr);
     EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+
+    test_msgs__msg__Strings__fini(&msg);
+    ret = rmw_serialized_message_fini(&serialized_msg);
+    ASSERT_EQ(RCUTILS_RET_OK, ret) << rcl_get_error_string().str;
   });
   rcl_reset_error();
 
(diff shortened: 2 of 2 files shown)


fujitatomoya (approved):

rcl/test/rcl/test_publisher.cpp
```diff
@@ -603,6 +603,9 @@ TEST_F(
   OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
   {
     test_msgs__msg__Strings__fini(&msg);
+    ASSERT_EQ(
+      RCUTILS_RET_OK,
+      rmw_serialized_message_fini(&serialized_msg)) << rcl_get_error_string().str;
```
iuhilnehc-ynos, line 608:
```suggestion
    ASSERT_EQ(
      RMW_RET_OK,
      rmw_serialized_message_fini(&serialized_msg)) << rcl_get_error_string().str;
```

Same below

rcl/test/rcl/test_subscription.cpp
```diff
@@ -571,6 +571,10 @@ TEST_F(CLASSNAME(TestSubscriptionFixture, RMW_IMPLEMENTATION), test_subscription
   {
     ret = rcl_subscription_fini(&subscription, this->node_ptr);
     EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+
+    test_msgs__msg__Strings__fini(&msg);
+    ret = rmw_serialized_message_fini(&serialized_msg);
+    ASSERT_EQ(RMW_RET_OK, ret) << rcl_get_error_string().str;
```
clalancette, line 577: I'll suggest combining these two lines for one particular reason.  The types here are all messed up.  That is, as far as I can tell, `rmw_serialized_message_fini` is a macro that just calls `rcutils_uint8_array_fini`.  Thus, the fact that it is documented as returning `RMW_*` error codes is a lie; it actually returns `RCUTILS_*` error codes.  This all ends up working because they are all integers, but this should really all be sorted out.  You don't need to do that in this patch, but I don't want to make the situation worse by assigning an `RCUTILS_*` error code to an `rcl_ret_t`.  Let's just do:

```
ASSERT_EQ(RMW_RET_OK, rmw_serialized_message_fini(&serialized_msg)) << rcl_get_error_string().str;
```

And leave the other cleanup for another time.

rcl/test/rcl/test_subscription.cpp
```diff
@@ -595,6 +599,10 @@ TEST_F(CLASSNAME(TestSubscriptionFixture, RMW_IMPLEMENTATION), test_subscription
     ASSERT_EQ(RMW_RET_OK, ret);
     ASSERT_EQ(
       std::string(test_string), std::string(msg_rcv.string_value.data, msg_rcv.string_value.size));
+
+    test_msgs__msg__Strings__fini(&msg_rcv);
+    ret = rmw_serialized_message_fini(&serialized_msg_rcv);
+    ASSERT_EQ(RMW_RET_OK, ret) << rcl_get_error_string().str;
```
clalancette, line 605: Same comment as above; let's combine these two lines.

clalancette (changes requested):

rcl/test/rcl/test_subscription.cpp
Barry-Xu-2018: Thanks for your comments.  
Got it.

clalancette (approved):

rcl/test/rcl/test_subscription.cpp
clalancette: Thanks, appreciated.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12337/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7306/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10050/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12246/)

clalancette:
Also, I've filed https://github.com/ros2/rmw/issues/273 to follow-up with the error codes in the rmw layer.

PR #802, opened 2020-09-15
Remove MAX_NUM_PARAMS_PER_NODE and MAX_NUM_NODE_ENTRIES limitation.

### LINKED ISSUE #526: [YAML parser]: Remove arbitrary limits of MAX_NUM_PARAMS_PER_NODE and MAX_NUM_NODE_ENTRIES
In #524, we removed the maximum string limit in the yaml parser.  There are also two other hard-coded limits in there:

* [MAX_NUM_NODE_ENTRIES](https://github.com/ros2/rcl/blob/cdf1b4d17dd9058e19d62373c76c8d1ec70f97c0/rcl_yaml_param_parser/src/parser.c#L72) - This is the maximum number of nodes that can be in a single parameters file, and is currently set to 256
* [MAX_NUM_PARAMS_PER_NODE](https://github.com/ros2/rcl/blob/cdf1b4d17dd9058e19d62373c76c8d1ec70f97c0/rcl_yaml_param_parser/src/parser.c#L73) - This is the maximum number of parameters that a single node can have, and is currently set to 512

We should look at removing both of these limitations as well, as they are kind of arbitrary.  However, removing them is slightly tricky since they are used at initialization time to allocate the entire array of `char *` that can ever be used during parsing.  To remove them, we'll have to implement a scheme where we do reallocations as we are parsing to make room for the new entries.  The really naive, and potentially poorly-performing, way to do this is to reallocate on every new entry.  The slightly smarter way to do it is to reallocate in batches, where when we need to do a reallocation we reallocate 512 of them at a time or something.  This also requires tracking a capacity vs. used, so is a bit more complicated.

iuhilnehc-ynos:
Fixes #526


iuhilnehc-ynos:
- These init values can be 1. (Keep the original value here, I will update them after discussion/review).

https://github.com/iuhilnehc-ynos/rcl/blob/3ce867e4f02e59516dfa29e34b938de70ec42d89/rcl_yaml_param_parser/src/impl/types.h#L34

https://github.com/iuhilnehc-ynos/rcl/blob/3ce867e4f02e59516dfa29e34b938de70ec42d89/rcl_yaml_param_parser/src/parser.c#L38

- Keep test case `maximum_number_parameters` but update the result value ( And I am sure there is no memory leak.)

https://github.com/iuhilnehc-ynos/rcl/blob/3ce867e4f02e59516dfa29e34b938de70ec42d89/rcl_yaml_param_parser/test/test_parse_yaml.cpp#L487-L514

- No idea why double free for `max_num_params.yaml` in test_parse_file_with_bad_allocator

https://github.com/iuhilnehc-ynos/rcl/blob/3ce867e4f02e59516dfa29e34b938de70ec42d89/rcl_yaml_param_parser/test/test_parser.cpp#L324

(NOTE: use `allocator.reallocate` could get the same address value with before, maybe it's related the test framework, I'll investigate on it.)

iuhilnehc-ynos:
> No idea why double free for max_num_params.yaml in test_parse_file_with_bad_allocator

Already fixed with https://github.com/ros2/rcl/pull/802/commits/328d3a51bd37b25e1e64c9ad27071b138694bc00

@clalancette
Could you help me review this?

rcl_yaml_param_parser/src/node_params.c
```diff
@@ -19,25 +19,86 @@
...
-    MAX_NUM_PARAMS_PER_NODE, sizeof(rcl_variant_t), allocator.state);
+    capacity, sizeof(rcl_variant_t), allocator.state);
   if (NULL == node_params->parameter_values) {
     allocator.deallocate(node_params->parameter_names, allocator.state);
     node_params->parameter_names = NULL;
+    RCUTILS_SET_ERROR_MSG("allocate error");
+    return RCUTILS_RET_BAD_ALLOC;
+  }
+
+  node_params->capacity_params = capacity;
```
clalancette, line 55: Should we also set `node_params->num_params` to zero here?  I know it wasn't done before, but I feel like it should have been.

rcl_yaml_param_parser/src/node_params.c
```diff
@@ -19,25 +19,86 @@
...
   rcl_node_params_t * node_params,
   const rcutils_allocator_t allocator)
+{
+  return node_params_init_with_capacity(node_params, allocator, INIT_NUM_PARAMS_PER_NODE);
+}
+
+rcutils_ret_t node_params_init_with_capacity(
+  rcl_node_params_t * node_params,
+  const rcutils_allocator_t allocator,
+  size_t capacity)
```
clalancette, line 29: I feel like our typical usage is to have the `allocator` be the last argument to a function, so please rearrange these so that is the case.

(this applies elsewhere in this patch as well)

rcl_yaml_param_parser/src/node_params.c
```diff
@@ -19,25 +19,86 @@
...
+  if (capacity == 0) {
+    RCUTILS_SET_ERROR_MSG("capacity can't be zero");
+    return RCUTILS_RET_INVALID_ARGUMENT;
+  }
 
   node_params->parameter_names = allocator.zero_allocate(
-    MAX_NUM_PARAMS_PER_NODE, sizeof(char *), allocator.state);
+    capacity, sizeof(char *), allocator.state);
   if (NULL == node_params->parameter_names) {
+    RCUTILS_SET_ERROR_MSG("allocate error");
```
clalancette, line 42:
```suggestion
    RCUTILS_SET_ERROR_MSG("Failed to allocate memory for node parameter names");
```

rcl_yaml_param_parser/src/node_params.c
```diff
@@ -19,25 +19,86 @@
...
     return RCUTILS_RET_BAD_ALLOC;
   }
 
   node_params->parameter_values = allocator.zero_allocate(
-    MAX_NUM_PARAMS_PER_NODE, sizeof(rcl_variant_t), allocator.state);
+    capacity, sizeof(rcl_variant_t), allocator.state);
   if (NULL == node_params->parameter_values) {
     allocator.deallocate(node_params->parameter_names, allocator.state);
     node_params->parameter_names = NULL;
+    RCUTILS_SET_ERROR_MSG("allocate error");
```
clalancette, line 51:
```suggestion
    RCUTILS_SET_ERROR_MSG("Failed to allocate memory for node parameter values");
```

rcl_yaml_param_parser/src/node_params.c
```diff
@@ -19,25 +19,86 @@
...
+  if (new_capacity < node_params->num_params) {
+    RCUTILS_SET_ERROR_MSG("invalid capacity");
+    return RCUTILS_RET_INVALID_ARGUMENT;
+  }
+
+  void * parameter_names = allocator.reallocate(
+    node_params->parameter_names, new_capacity * sizeof(char *), allocator.state);
+  if (NULL == parameter_names) {
+    rcl_yaml_node_params_fini(node_params, allocator);
+    RCUTILS_SET_ERROR_MSG("reallocate error");
```
clalancette, line 77:
```suggestion
    RCUTILS_SET_ERROR_MSG("Failed to reallocate node parameter names");
```

rcl_yaml_param_parser/src/node_params.c
```diff
@@ -19,25 +19,86 @@
...
     return RCUTILS_RET_BAD_ALLOC;
   }
+  node_params->parameter_names = parameter_names;
 
+  void * parameter_values = allocator.reallocate(
+    node_params->parameter_values, new_capacity * sizeof(rcl_variant_t), allocator.state);
+  if (NULL == parameter_values) {
+    rcl_yaml_node_params_fini(node_params, allocator);
+    RCUTILS_SET_ERROR_MSG("reallocate error");
```
clalancette, line 86:
```suggestion
    RCUTILS_SET_ERROR_MSG("Failed to reallocate node parameter values");
```

rcl_yaml_param_parser/src/parse.c
```diff
@@ -809,18 +801,37 @@ rcutils_ret_t find_parameter(
...
+  if (node_param_st->num_params >= node_param_st->capacity_params) {
+    if (RCUTILS_RET_OK != node_params_reallocate(
+        node_param_st, allocator, node_param_st->capacity_params * 2))
+    {
+      ret = RCUTILS_RET_BAD_ALLOC;
+      goto clean;
+    }
+    reallocate = true;
+  }
+  // Parameter not found, add it.
```
clalancette, line 817: I think this comment belongs in the original place, that is, right after trying to find the parameter.

rcl_yaml_param_parser/src/parse.c
```diff
@@ -809,18 +801,37 @@ rcutils_ret_t find_parameter(
...
+    goto clean;
   }
   node_param_st->num_params++;
   return RCUTILS_RET_OK;
+
+clean:
+  if (reallocate) {
+    rcl_yaml_node_params_fini(node_param_st, allocator);
+  }
+  return ret;
```
clalancette, line 834: I think this is a bit strange.  That is, if we fail because we couldn't reallocate, we *don't* clean up the memory, but if we fail for another reason later on, we *do* clean up the memory.  How is a caller of this function supposed to react to a failure?  Call fini, or not call fini?  I don't think they can know.  I think we need to think about the post-conditions of this function in both success and failure cases, document it in the header file, and then implement whatever we decide there.

(this same comment applies to `find_node` below)

rcl_yaml_param_parser/src/parser.c
```diff
@@ -35,40 +35,103 @@
...
+  if (capacity == 0) {
+    RCUTILS_SET_ERROR_MSG("capacity can't be zero");
+    return NULL;
+  }
   rcl_params_t * params_st = allocator.zero_allocate(1, sizeof(rcl_params_t), allocator.state);
   if (NULL == params_st) {
-    RCUTILS_SAFE_FWRITE_TO_STDERR("Error allocating mem\n");
-    return NULL;
+    RCUTILS_SET_ERROR_MSG("Error allocating mem");
+    goto clean;
```
clalancette, line 61:
```suggestion
    return NULL;
```

(we haven't allocated anything at this point, so this is safe to short-circuit here)

rcl_yaml_param_parser/src/parser.c
```diff
@@ -35,40 +35,103 @@
...
   RCUTILS_CHECK_ALLOCATOR_WITH_MSG(&allocator, "invalid allocator", return NULL);
+  if (capacity == 0) {
+    RCUTILS_SET_ERROR_MSG("capacity can't be zero");
+    return NULL;
+  }
   rcl_params_t * params_st = allocator.zero_allocate(1, sizeof(rcl_params_t), allocator.state);
   if (NULL == params_st) {
-    RCUTILS_SAFE_FWRITE_TO_STDERR("Error allocating mem\n");
-    return NULL;
+    RCUTILS_SET_ERROR_MSG("Error allocating mem");
```
clalancette, line 60:
```suggestion
    RCUTILS_SET_ERROR_MSG("Failed to allocate memory for parameters");
```

rcl_yaml_param_parser/src/parser.c
```diff
@@ -35,40 +35,103 @@
...
   params_st->allocator = allocator;
 
   params_st->node_names = allocator.zero_allocate(
-    MAX_NUM_NODE_ENTRIES, sizeof(char *), allocator.state);
+    capacity, sizeof(char *), allocator.state);
   if (NULL == params_st->node_names) {
-    rcl_yaml_node_struct_fini(params_st);
-    RCUTILS_SAFE_FWRITE_TO_STDERR("Error allocating mem\n");
-    return NULL;
+    RCUTILS_SET_ERROR_MSG("Error allocating mem");
```
clalancette, line 69:
```suggestion
    RCUTILS_SET_ERROR_MSG("Failed to allocate memory for parameter node names");
```

rcl_yaml_param_parser/src/parser.c
```diff
@@ -35,40 +35,103 @@
...
   params_st->params = allocator.zero_allocate(
-    MAX_NUM_NODE_ENTRIES, sizeof(rcl_node_params_t), allocator.state);
+    capacity, sizeof(rcl_node_params_t), allocator.state);
   if (NULL == params_st->params) {
-    rcl_yaml_node_struct_fini(params_st);
-    RCUTILS_SAFE_FWRITE_TO_STDERR("Error allocating mem\n");
-    return NULL;
+    allocator.deallocate(params_st->node_names, allocator.state);
+    params_st->node_names = NULL;
+    RCUTILS_SET_ERROR_MSG("Error allocating mem");
```
clalancette, line 78:
```suggestion
    RCUTILS_SET_ERROR_MSG("Failed to allocate memory for parameter values");
```

clalancette (changes requested):
I've got a bunch of comments inline.

Also, I think we should add some additional tests here.  In particular, we need to ensure that the tests exercise both the allocate and reallocate code paths (so we probably need to add a yaml file with a lot of nodes in it).

iuhilnehc-ynos:
> I've got a bunch of comments inline.

Thanks for your comments.

> Also, I think we should add some additional tests here. In particular, we need to ensure that the tests exercise both the allocate and reallocate code paths (so we probably need to add a yaml file with a lot of nodes in it).

I totally agree with that.

rcl_yaml_param_parser/src/parse.c
iuhilnehc-ynos: > but if we fail for another reason later on, we do clean up the memory. How is a caller of this function supposed to react to a failure? Call fini, or not call fini?

Call `fini` is the responsibility inside the function. Actually, I think it's the same pattern as
```
function:
   a1 = allocate
   b1 = reallocate
   c1 = do_some
   if c1 is error, goto clean

   return true   
clean:
   deallocate(b1)
   deallocate(a1)
```

iuhilnehc-ynos: I found there is a problem,

https://github.com/ros2/rcl/blob/c1b3ff72d3d106e322661db3a2702b476eedfe27/rcl_yaml_param_parser/src/parser.c#L167

can only be called just once, and a function to call `rcl_yaml_node_struct_fini` must assign the `params_st` with NULL immediately, so I can't use this function in `clean` location because of some allocate failure.

iuhilnehc-ynos:
Please ignore about refactoring the structure of `rcl_yaml_param_parser` which I mentioned before if you see the information in notification emails, it's just my idea.

I didn't answer all of your comments but I think I have already update them, could you review it again?

rcl_yaml_param_parser/test/test_parser.cpp
```diff
@@ -321,7 +361,9 @@ TEST(RclYamlParamParser, test_parse_file_with_bad_allocator) {
     "correct_config.yaml",
     "empty_string.yaml",
     "indented_name_space.yaml",
-    "max_num_params.yaml",
+    "multiple_params.yaml",
+    "multiple_nodes.yaml",
+//    "multiple_nodes_parameters.yaml",  // Comment it because Timeout while running test.
```
iuhilnehc-ynos, line 366: Using 'colcon test' will be Timeout, but I ran successfully while run `./test_parser` directly.

clalancette: > Using 'colcon test' will be Timeout, but I ran successfully while running `./test_parser` directly.

It's possible to specify the TIMEOUT for tests (the default is 60 seconds) in the `ament_add_gtest` macro.

That being said, I'm concerned that this test is taking longer than 60 seconds.  I've got a few suggestions here:

1.  Can you do a bit of profiling and find out why it is taking so long?  512 nodes or parameters doesn't seem like that much, and thus I wouldn't expect this to take longer than 60 seconds.
1.  You could simplify this a bit by making one file which has > 512 parameters in a single node, and > 512 nodes in a single file.  That's a bit simpler than what you currently have.
1.  You could split these particular tests out of this file and put it in another test.
1.  Since we are removing the maximum limit now, you could reduce the initial number of parameters and nodes to a much smaller number.  In point of fact, we should probably do that anyway, since in most reasonable scenarios, people will only have a few dozen parameters and a similar number of nodes per file.  This will save users memory, and also make it so that the number of parameters/nodes you need to test "overflow" are smaller.  I'll suggest 128 for both the initial number of parameters and the initial number of nodes, but feel free to make different choices.

iuhilnehc-ynos: > Can you do a bit of profiling ? 

Yes, of course.

Uncommented `multiple_nodes_parameters.yaml`, and run google-pprof to get [test_parser](https://github.com/iuhilnehc-ynos/share_files/blob/master/test_parser.prof.pdf)

After removing `multiple_*.yaml` from `test_parser.cpp` and spliting the particular tests in new files `test_parser_multiple_nodes.cpp` and `test_parser_multiple_parameters.cpp`, and run google-pprof to get [test_parser_multiple_nodes](https://github.com/iuhilnehc-ynos/share_files/blob/master/test_parser_multiple_nodes.prof.pdf) and [test_parser_multiple_parameters](https://github.com/iuhilnehc-ynos/share_files/blob/master/test_parser_multiple_params.prof.pdf)


> and find out why it is taking so long

From the cpu profiling of [test_parser](https://github.com/iuhilnehc-ynos/share_files/blob/master/test_parser.prof.pdf) google-pprof , we can find the reason is mainly related to two functions, `find_node` to call `node_param_init`, `yaml_parser_parse`. It seems it's not related to the new added `*_reallocate` functions.


> 512 nodes or parameters doesn't seem like that much, and thus I wouldn't expect this to take longer than 60 seconds.

I supposed you knew it's a test case about bad allocator with macro `RCUTILS_FAULT_INJECTION_TEST`.

FYI:
```
[ RUN      ] test_file_parser.multiple_number_parameters
[       OK ] test_file_parser.multiple_number_parameters (2 ms)
[ RUN      ] test_file_parser.multiple_number_nodes
[       OK ] test_file_parser.multiple_number_nodes (4 ms)
[ RUN      ] test_file_parser.multiple_number_nodes_and_parameters
[       OK ] test_file_parser.multiple_number_nodes_and_parameters (4 ms)
```

> You could simplify this a bit by making one file which has > 512 parameters in a single node, and > 512 nodes in a single file. That's a bit simpler than what you currently have.

I had already provided these two files, please refer to "multiple_params.yaml" and "multiple_nodes.yaml".
OK, I'll only use these two yamls and remove the "multiple_nodes_parameters.yaml".

> You could split these particular tests out of this file and put it in another test.

After using 128 as initial value, `colcon test` can be passed, but I'd like to accept this suggestion to split these particular tests.


> I'll suggest 128 for both the initial number of parameters and the initial number of nodes

OK, I do like this number.

clalancette: > From the cpu profiling of [test_parser](https://github.com/iuhilnehc-ynos/share_files/blob/master/test_parser.prof.pdf) google-pprof , we can find the reason is mainly related to two functions, `find_node` to call `node_param_init`, `yaml_parser_parse`. It seems it's not related to the new added `*_reallocate` functions.

This is great work, thanks for taking the time to do this.  After looking through your profiling a bit, one obvious slowdown is that [find_parameter](https://github.com/ros2/rcl/blob/8bcada7b884693644d7ba2e3bd2c22fd9541eba7/rcl_yaml_param_parser/src/parse.c#L806) just iterates straight over the list, in `O(n)` time.  One thing we *could* do is to change this to a hash table with `O(1)` time for lookups.  That would improve the CPU usage quite a bit.  I'm not suggesting we do that in this PR, of course, but I would ask if you could open a follow-up issue to improve the performance there.

> I supposed you knew it's a test case about bad allocator with macro `RCUTILS_FAULT_INJECTION_TEST`.

I didn't know that, thanks for pointing it out.

> After using 128 as initial value, `colcon test` can be passed, but I'd like to accept this suggestion to split these particular tests.

Yeah, I think that makes more sense anyway.  I find it easier to run the tests individually anyway.

Thanks again for all of this work, I'll re-review the code.

rcl_yaml_param_parser/src/node_params.c
```diff
@@ -19,25 +19,97 @@
...
   }
+  node_params->parameter_values = parameter_values;
+  // zero initialization for the added memory
+  if (new_capacity > node_params->capacity_params) {
+    memset(
+      &node_params->parameter_values[node_params->capacity_params], 0,
+      (new_capacity - node_params->capacity_params) * sizeof(rcl_variant_t));
+  }
 
+  node_params->capacity_params = new_capacity;
```
clalancette, line 112: I'll make one other suggestion about this whole function.  I think it might be nicer if we guarantee that on success, everything has been reallocated properly, but on failure, nothing has been changed in the structure.  Then it is simple for the caller of this function to determine what to do in either case.  To do that, I'll suggest something like the following pseudo-code:

```
void *names = allocator.reallocate(node_params->parameter_names, new_capacity);
if (!names) {
  return ERROR;
}
zero_new_names_memory();
void *values = allocator.reallocate(node_params->parameter_values, new_capacity);
if (!values) {
  allocator.deallocate(names);
  return ERROR;
}
zero_new_values_memory();
node_params->capacity_params = new_capacity;
node_params->parameter_names = names;
node_params->parameter_values = values;
return OK;
```

Also, please document this post-condition in the header file above.

The same basic idea applies to `rcl_yaml_node_struct_reallocate` below.

rcl_yaml_param_parser/test/test_parse_yaml.cpp
```diff
@@ -495,7 +494,7 @@ TEST(test_file_parser, maximum_number_parameters) {
   {
     allocator.deallocate(test_path, allocator.state);
   });
-  char * path = rcutils_join_path(test_path, "max_num_params.yaml", allocator);
+  char * path = rcutils_join_path(test_path, "multiple_params.yaml", allocator);
```
clalancette, line 497: I'm a little confused here; you split this out to a separate file, but you are also testing it here.  Is that what you meant to do?  If so, why?

rcl_yaml_param_parser/test/test_parse_yaml.cpp
```diff
@@ -508,8 +507,59 @@ TEST(test_file_parser, maximum_number_parameters) {
...
+  rcutils_reset_error();
+  EXPECT_TRUE(rcutils_get_cwd(cur_dir, 1024)) << rcutils_get_error_string().str;
+  rcutils_allocator_t allocator = rcutils_get_default_allocator();
+  char * test_path = rcutils_join_path(cur_dir, "test", allocator);
+  ASSERT_TRUE(NULL != test_path) << rcutils_get_error_string().str;
+  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+  {
+    allocator.deallocate(test_path, allocator.state);
+  });
+  char * path = rcutils_join_path(test_path, "multiple_nodes.yaml", allocator);
```
clalancette, line 523: Same question here: do you mean to test this both here and in the separate file.  If so, why?

clalancette (changes requested):

rcl_yaml_param_parser/test/test_parser.cpp
iuhilnehc-ynos: > One thing we could do is to change this to a hash table with O(1) time for lookups. That would improve the CPU usage quite a bit. I'm not suggesting we do that in this PR, of course, but I would ask if you could open a follow-up issue to improve the performance there.

I'd like to do that.

rcl_yaml_param_parser/src/node_params.c
iuhilnehc-ynos: > but on failure, nothing has been changed in the structure

Thanks for your suggestion.

But we can't promise that(Actually, I did that in my first commit.), because `reallocate` could return address values same as or different from the old address.

```
void *names = allocator.reallocate(node_params->parameter_names, new_capacity);    
if (!names) {
  return ERROR;
}
zero_new_names_memory();
void *values = allocator.reallocate(node_params->parameter_values, new_capacity);
if (!values) {
  allocator.deallocate(names);      // `step A`, please refer to below explanation.
  return ERROR;
}
```

At `step A`
- if `names` address is different from `node_params->parameter_names`, the content in `node_params->parameter_names` address is not exist (copy the memory to the new address and free it by reallocate). so I need to put `node_params->parameter_names = names;` before `reallocate values`. 
- use `allocator.deallocate(names); ` is dangerous(memory leak that old name items not be destroyed) if `names` is same as `node_params->parameter_names` after `reallocate node_params->parameter_names`
- if failure happened, we could keep unchanged by using two method
  - 1. `reallocate node_params->parameter_names ` with the original node_params->capacity_params. (But the address of node_params->parameter_names might be changed.)
  - 2. not to deallocate or reallocate `node_params->parameter_names`, just leave node_params->parameter_names as it was.

What do you think about that ?

rcl_yaml_param_parser/test/test_parse_yaml.cpp
iuhilnehc-ynos: > I'm a little confused here; you split this out to a separate file, but you are also testing it here.

The particular tests I understand before are those that took a long time in `test_parser`.
OK, I'll move `test_file_parser, multiple_number_nodes` into test_parser_multiple_nodes.cpp, and deal with `test_file_parser, multiple_number_params` too.

iuhilnehc-ynos:
@clalancette 

I have pushed my latest commit to keep the content unchanged if a failure happened while reallocating, could you please review it again.

NOTE: There is unnecessary to deallocate the previous pointer which was reallocated before, because after returning failure, no matter what it continues to do, reallocating or finalizing the structure variable, the memory will not be leaked.
All users need to do is not use the old pointer after calling reallocate because the member pointer of this structure might be changed which was documented for the function.

iuhilnehc-ynos:
@clalancette 

If there is something I missed or make you confused, please point it out.

rcl_yaml_param_parser/src/node_params.c
clalancette: Thanks for the explanation; that makes sense.

In that case, I think we should just leave the code like it is; on failure, the structure will still contain all valid addresses.  The caller has to call `fini` on failure, but I think that is an OK constraint.

rcl_yaml_param_parser/test/test_node_params.cpp
```diff
@@ -26,15 +26,60 @@
...
   EXPECT_EQ(nullptr, node_params.parameter_values);
   EXPECT_EQ(0u, node_params.num_params);
+  EXPECT_EQ(0u, node_params.capacity_params);
+
+  // This function doesn't return anything, so just check it doesn't segfault on the second try
+  rcl_yaml_node_params_fini(&node_params, allocator);
+  rcl_yaml_node_params_fini(nullptr, allocator);
+}
+
+TEST(TestNodeParams, init_with_capability_fini) {
```
clalancette, line 46:
```suggestion
TEST(TestNodeParams, init_with_capacity_fini) {
```

(I think this is what you meant)

rcl_yaml_param_parser/test/test_node_params.cpp
```diff
@@ -26,15 +26,60 @@
...
+  EXPECT_EQ(nullptr, node_params.parameter_values);
+  EXPECT_EQ(0u, node_params.num_params);
+  EXPECT_EQ(0u, node_params.capacity_params);
+
+  // This function doesn't return anything, so just check it doesn't segfault on the second try
+  rcl_yaml_node_params_fini(&node_params, allocator);
+  rcl_yaml_node_params_fini(nullptr, allocator);
+}
+
+TEST(TestNodeParams, reallocate_with_capability_fini) {
```
clalancette, line 65:
```suggestion
TEST(TestNodeParams, reallocate_with_capacity_fini) {
```

clalancette (commented):
I've got two very small changes to make.  Otherwise, this looks good to me.  I'm still going to run CI on it because the two changes shouldn't make any difference to whether this passes CI or not.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12425/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7392/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10135/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12338/)

rcl_yaml_param_parser/src/node_params.c
```diff
@@ -19,25 +19,87 @@
...
+  rcl_node_params_t * node_params,
+  size_t new_capacity,
+  const rcutils_allocator_t allocator)
+{
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(node_params, RCUTILS_RET_INVALID_ARGUMENT);
+  RCUTILS_CHECK_ALLOCATOR_WITH_MSG(
+    &allocator, "invalid allocator", return RCUTILS_RET_INVALID_ARGUMENT);
+  // invalid if new_capacity is less than num_params
+  if (new_capacity < node_params->num_params) {
+    RCUTILS_SET_ERROR_MSG("invalid capacity");
```
fujitatomoya, line 70: it would be nice to notify what size is expected for `new_capacity`.

rcl_yaml_param_parser/src/impl/types.h
```diff
@@ -31,7 +31,7 @@ extern "C"
 #define NODE_NS_SEPERATOR "/"
 #define PARAMETER_NS_SEPERATOR "."
 
-#define MAX_NUM_PARAMS_PER_NODE 512U
+#define INIT_NUM_PARAMS_PER_NODE 128U
```
fujitatomoya, line 34: this should be move to `rcl_yaml_param_parser/src/node_params.c`?

rcl_yaml_param_parser/src/parser.c
```diff
@@ -35,40 +35,103 @@
...
+  rcl_params_t * params_st,
+  size_t new_capacity,
+  const rcutils_allocator_t allocator)
+{
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(params_st, RCUTILS_RET_INVALID_ARGUMENT);
+  RCUTILS_CHECK_ALLOCATOR_WITH_MSG(
+    &allocator, "invalid allocator", return RCUTILS_RET_INVALID_ARGUMENT);
+  // invalid if new_capacity is less than num_nodes
+  if (new_capacity < params_st->num_nodes) {
+    RCUTILS_SET_ERROR_MSG("invalid capacity");
```
fujitatomoya, line 101: same here.

fujitatomoya (approved):
lgtm with minor comments.

rcl_yaml_param_parser/src/node_params.c
iuhilnehc-ynos: Thank you.

clalancette:
CI looked generally good.  I'm actually going to run it one more time since there are enough changes here.  Assuming this passes, I'll go ahead and merge.  Thanks for the iterations.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12443/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7410/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10153/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12358/)

PR #807, opened 2020-09-21
Name remapping specific to a fully-qualified node's name

### LINKED ISSUE #296: Name remapping specific to a fully-qualified node's name
Remapping can be passed to a specific node with e.g. `talker:chatter:=my_chatter`.
Suppose I have multiple nodes named `chatter` in a process, but in different namespaces.
Is the ability to address the nodes individually implemented yet, so that I can do e.g. `/ns1/talker:chatter:=my_chatter`? I think not; am looking for confirmation. @sloretz

iuhilnehc-ynos:
Fixes #296


ivanpauno:
Can you add this to the remapping design doc? Similarly to what was done in https://github.com/ros2/design/pull/299 for https://github.com/ros2/rcl/pull/837.

Particularly, this change looks good to me (I haven't checked the implementation yet).
It enables a new use case (that was possible in ROS 1), without breaking API at all.

iuhilnehc-ynos:
@ivanpauno 

> Can you add this to the remapping design doc?

Yes, I'll create a new PR for the remapping design doc for the FQN after https://github.com/ros2/design/pull/299 is merged because some examples seem to be changed.

PR #809, opened 2020-09-22
add general check for yaml on rcl_yaml_param_parser

### LINKED ISSUE #301: rcl_yaml_param_parser allows two // in node name
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu Bionic
- Installation type:
  - source
- Version or commit hash:
  - master
- DDS implementation:
  - Fast-RTPS
  - N/A

#### Steps to reproduce issue

Create a yaml file called `empty_name_in_ns.yaml` in `rcl_yaml_param_parser/test` with the following

```yaml
lidar_ns//lidar_w:
  ros__parameters:
    id: 10
    name: front_lidar
    ports: [2438, 2439, 2440]
    driver1:
      dx: 4.56
      dy: 2.30
      fr_sensor_specs: [12, 3, 0, 7]
      bk_sensor_specs: [12.1, -2.3, 5.2, 9.0]
    is_front: true
    driver2:
      dx: 1.23
      dy: 0.45
```

Add this test block to `test_parse_yaml.cpp`

```c++
TEST(test_file_parser, empty_name_in_ns) {
  rcutils_reset_error();
  EXPECT_TRUE(rcutils_get_cwd(cur_dir, 1024));
  rcutils_allocator_t allocator = rcutils_get_default_allocator();
  char * test_path = rcutils_join_path(cur_dir, "test", allocator);
  char * path = rcutils_join_path(test_path, "empty_name_in_ns.yaml", allocator);
  fprintf(stderr, "cur_path: %s\n", path);
  EXPECT_TRUE(rcutils_exists(path));
  rcl_params_t * params_hdl = rcl_yaml_node_struct_init(allocator);
  EXPECT_FALSE(NULL == params_hdl);
  bool res = rcl_parse_yaml_file(path, params_hdl);
  fprintf(stderr, "%s\n", rcutils_get_error_string_safe());
  EXPECT_FALSE(res);
  rcl_yaml_node_struct_print(params_hdl);
  rcl_yaml_node_struct_fini(params_hdl);
  allocator.deallocate(test_path, allocator.state);
  allocator.deallocate(path, allocator.state);
}
```

#### Expected behavior

Parsing should fail to parse the yaml file.

#### Actual behavior

```
1:  Node Name                           Parameters
1: lidar_ns//lidar_w
1:                                                 id: 10
1:                                               name: front_lidar
1:                                              ports: 2438, 2439, 2440, 
1:                                         driver1.dx: 4.560000
1:                                         driver1.dy: 2.300000
1:                            driver1.fr_sensor_specs: 12, 3, 0, 7, 
1:                            driver1.bk_sensor_specs: 12.100000, -2.300000, 5.200000, 9.000000, 
1:                                           is_front: true
1:                                         driver2.dx: 1.230000
1:                                         driver2.dy: 0.450000
```

#### Additional information

See ros2/rcl#299 for another bug about `/` in the namespace names.

iuhilnehc-ynos:
Fixes #301

**Updated**: Besides fixing to not allow '//' in node name, adding the general check and some special rules about wildcards(including some new rules on https://github.com/ros2/design/pull/303)

### DIFF
--- a/rcl_yaml_param_parser/src/namespace.c
+++ b/rcl_yaml_param_parser/src/namespace.c
@@ -14,47 +14,7 @@
 
 #include "./impl/namespace.h"
 
-#include "rcutils/format_string.h"
 #include "rcutils/strdup.h"
-
-#include "rmw/error_handling.h"
-#include "rmw/validate_namespace.h"
-
-///
-/// Validate a name whether it is a valid namespace
-///
-static rcutils_ret_t
-__validate_namespace(const char * name, rcutils_allocator_t allocator)
-{
-  // rmw_validate_namespace will report RMW_NAMESPACE_INVALID_NOT_ABSOLUTE
-  // so add prefix '/' to name to make an absolute name if necessary
-  char * absolute_name = NULL;
-  if (name[0] != '/') {
-    absolute_name = rcutils_format_string(allocator, "/%s", name);
-    if (NULL == absolute_name) {
-      RCUTILS_SET_ERROR_MSG("failed to allocate memory for absolute name");
-      return RCUTILS_RET_BAD_ALLOC;
-    }
-  }
-  int validation_result = 0;
-  rmw_ret_t ret;
-  if (absolute_name) {
-    ret = rmw_validate_namespace(absolute_name, &validation_result, NULL);
-    allocator.deallocate(absolute_name, allocator.state);
-  } else {
-    ret = rmw_validate_namespace(name, &validation_result, NULL);
-  }
-  if (ret != RMW_RET_OK) {
-    RCUTILS_SET_ERROR_MSG(rmw_get_error_string().str);
-    return RCUTILS_RET_ERROR;
-  } else {
-    if (validation_result != RMW_NAMESPACE_VALID) {
-      RCUTILS_SET_ERROR_MSG(rmw_namespace_validation_result_string(validation_result));
-      return RCUTILS_RET_INVALID_ARGUMENT;
-    }
-  }
-  return RCUTILS_RET_OK;
-}
 
 rcutils_ret_t add_name_to_ns(
   namespace_tracker_t * ns_tracker,
@@ -88,12 +48,6 @@ rcutils_ret_t add_name_to_ns(
   /// Add a name to ns
   if (NULL == name) {
     return RCUTILS_RET_INVALID_ARGUMENT;
-  }
-  if (namespace_type == NS_TYPE_NODE) {
-    rcutils_ret_t ret = __validate_namespace(name, allocator);
-    if (RCUTILS_RET_OK != ret) {
-      return ret;
-    }
   }
   if (0U == *cur_count) {
     cur_ns = rcutils_strdup(name, allocator);
--- a/rcl_yaml_param_parser/src/namespace.c
+++ b/rcl_yaml_param_parser/src/namespace.c
@@ -14,7 +14,11 @@
 
 #include "./impl/namespace.h"
 
+#include "rcutils/format_string.h"
 #include "rcutils/strdup.h"
+
+#include "rmw/error_handling.h"
+#include "rmw/validate_namespace.h"
 
 ///
 /// Validate a name whether it is a valid namespace
@@ -22,10 +26,33 @@
 static rcutils_ret_t
 __validate_namespace(const char * name, rcutils_allocator_t allocator)
 {
-  if (strstr(name, "//") != NULL) {
-    return RCUTILS_RET_INVALID_ARGUMENT;
+  // rmw_validate_namespace will report RMW_NAMESPACE_INVALID_NOT_ABSOLUTE
+  // so add prefix '/' to name to make an absolute name if necessary
+  char * absolute_name = NULL;
+  if (name[0] != '/') {
+    absolute_name = rcutils_format_string(allocator, "/%s", name);
+    if (NULL == absolute_name) {
+      RCUTILS_SET_ERROR_MSG("failed to allocate memory for absolute name");
+      return RCUTILS_RET_BAD_ALLOC;
+    }
   }
-
+  int validation_result = 0;
+  rmw_ret_t ret;
+  if (absolute_name) {
+    ret = rmw_validate_namespace(absolute_name, &validation_result, NULL);
+    allocator.deallocate(absolute_name, allocator.state);
+  } else {
+    ret = rmw_validate_namespace(name, &validation_result, NULL);
+  }
+  if (ret != RMW_RET_OK) {
+    RCUTILS_SET_ERROR_MSG(rmw_get_error_string().str);
+    return RCUTILS_RET_ERROR;
+  } else {
+    if (validation_result != RMW_NAMESPACE_VALID) {
+      RCUTILS_SET_ERROR_MSG(rmw_namespace_validation_result_string(validation_result));
+      return RCUTILS_RET_INVALID_ARGUMENT;
+    }
+  }
   return RCUTILS_RET_OK;
 }
 
--- a/rcl_yaml_param_parser/src/namespace.c
+++ b/rcl_yaml_param_parser/src/namespace.c
@@ -14,7 +14,47 @@
 
 #include "./impl/namespace.h"
 
+#include "rcutils/format_string.h"
 #include "rcutils/strdup.h"
+
+#include "rmw/error_handling.h"
+#include "rmw/validate_namespace.h"
+
+///
+/// Validate a name whether it is a valid namespace
+///
+static rcutils_ret_t
+__validate_namespace(const char * name, rcutils_allocator_t allocator)
+{
+  // rmw_validate_namespace will report RMW_NAMESPACE_INVALID_NOT_ABSOLUTE
+  // so add prefix '/' to name to make an absolute name if necessary
+  char * absolute_name = NULL;
+  if (name[0] != '/') {
+    absolute_name = rcutils_format_string(allocator, "/%s", name);
+    if (NULL == absolute_name) {
+      RCUTILS_SET_ERROR_MSG("failed to allocate memory for absolute name");
+      return RCUTILS_RET_BAD_ALLOC;
+    }
+  }
+  int validation_result = 0;
+  rmw_ret_t ret;
+  if (absolute_name) {
+    ret = rmw_validate_namespace(absolute_name, &validation_result, NULL);
+    allocator.deallocate(absolute_name, allocator.state);
+  } else {
+    ret = rmw_validate_namespace(name, &validation_result, NULL);
+  }
+  if (ret != RMW_RET_OK) {
+    RCUTILS_SET_ERROR_MSG(rmw_get_error_string().str);
+    return RCUTILS_RET_ERROR;
+  } else {
+    if (validation_result != RMW_NAMESPACE_VALID) {
+      RCUTILS_SET_ERROR_MSG(rmw_namespace_validation_result_string(validation_result));
+      return RCUTILS_RET_INVALID_ARGUMENT;
+    }
+  }
+  return RCUTILS_RET_OK;
+}
 
 rcutils_ret_t add_name_to_ns(
   namespace_tracker_t * ns_tracker,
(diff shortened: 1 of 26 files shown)


rcl_yaml_param_parser/src/namespace.c
```diff
@@ -16,6 +16,19 @@
 
 #include "rcutils/strdup.h"
 
+///
+/// Validate a name whether it is a valid namespace
+///
+static rcutils_ret_t
+__validate_namespace(const char * name)
```
fujitatomoya, line 23: why it should not be dependent on `rmw`? could you elaborate a little bit? I think it would be better to use `rmw` to validate the namespace instead of having redundant code else where.

> I found that using rmw_validate_namespace has some limitations because there are some extra rules, such as "/**" (maybe more in the future), need to be parsed.

you mean https://github.com/ros2/rclcpp/issues/1265?

iuhilnehc-ynos: Thanks for your questions, @fujitatomoya 

> why it should not be dependent on rmw?

"/**" is a special rule at `rclcpp`, it is invalid checked by `rmw_validate_namespace`.
If using `rmw_validate_namespace`, we need to add more extra checking for these special rules inside `__validate_namespace`, which seems not a requirement of this issue.

>> maybe more in the future

> you mean ros2/rclcpp#1265?

Yes.

jacobperron: Though this solves the reported issue, I think we should do a more general check if the node namespace (and node name) are valid. I would do this using `rmw_validate_namespace` and `rmw_validate_node_name`, and handle the special cases for wildcards `*` and `**`.

I suggest implementing this function as follows:

1. Make a copy of the name to validate
2. Substitute substrings `*` and `**`  with arbitrary valid strings (e.g. `WILDCARD`)
3. Pass the modified name to `rmw_validate_namespace`

By using the rmw functions we are consistent with what it means to be a valid name.

What do you think?

jacobperron (commented):
I think we should also validate the node name. If you don't want to add it to this PR, we do it in a follow-up.

rcl_yaml_param_parser/src/namespace.c
iuhilnehc-ynos: @jacobperron

> I suggest implementing this function as follows

Thanks for your suggestion. I'll update it later.

iuhilnehc-ynos:
@jacobperron

After rebasing from `master`, I need to force push these commits to this branch.
Could you please help review it again?

rcl_yaml_param_parser/src/parse.c
```diff
@@ -404,6 +409,118 @@ rcutils_ret_t parse_value(
...
+}
+
+///
+/// Check a name (namespace/node_name) whether it is valid
+///
+static rcutils_ret_t
+__validate_name(const char * name, rcutils_allocator_t allocator)
+{
+  // special rules
+  if (strcmp(name, "/**") == 0) {
```
jacobperron, line 461: I think we could add additional rules for wildcards now (even if rclcpp and rclpy don't support them yet).
If you don't want to, I think it's worth adding a TODO here.

jacobperron (commented):
LGTM.

Could you also add a test for the `/**` case? Thanks!

rcl_yaml_param_parser/src/parse.c
iuhilnehc-ynos: > If you don't want to, I think it's worth adding a TODO here.

I have added a TODO there.

iuhilnehc-ynos: @jacobperron

I found you had created a PR for wildcards design on https://github.com/ros2/design/pull/303.
so I'll add additional rules for wildcards later.

iuhilnehc-ynos:
@jacobperron 

> Could you also add a test for the /** case? Thanks!

I have added the test and added additional rules for wildcards based on https://github.com/ros2/design/pull/303.
Could you review it again?

rcl_yaml_param_parser/src/parse.c
```diff
@@ -404,6 +409,128 @@ rcutils_ret_t parse_value(
...
+  } else {
+    node_name = rcutils_strdup(separator_pos + 1, allocator);
+    if (NULL == node_name) {
+      ret = RCUTILS_RET_BAD_ALLOC;
+      goto clean;
+    }
+
+    char * namespace = NULL;
+    // Besides "*", node name also support /*.
+    if (0 == strcmp(node_name, "*") && '/' == name[separator_pos - name]) {
```
jacobperron, line 484: I think the colcon (`:`) following a namespace should implicitly act like a slash separate (`/`).
For example:

```yaml
/foo:
  /*:
    ros__parameters:
      ...
```

Should result in the fully-qualified name `/foo//*`, which is an invalid name.

Therefore, I think this case can be simplified to:

```c
char * namespace = rcutils_strndup(name, separator_pos - name, allocator);
```

As for the test, a valid parameters yaml file would be:

```yaml
/foo/*:
    ros__parameters:
      ...
```

iuhilnehc-ynos: > Should result in the fully-qualified name /foo//*, which is an invalid name.

okay.

> As for the test, a valid parameters yaml file would be:
> ```
> /foo/*:
>     ros__parameters:
>       ...
> ```

Thank you.

iuhilnehc-ynos: @jacobperron

I have a question, do we need to support the following setting?
```
/foo/**/bar:
  some_node:
     ros__parameters:
        ...
```

jacobperron: I would say that example is valid following the matching rules on the design document. That being said, I don't know if it will be supported in the client libraries any time soon so we can probably revisit this case as a follow-up. WDYT?

rcl_yaml_param_parser/src/parse.c
```diff
@@ -404,6 +409,147 @@ rcutils_ret_t parse_value(
...
+            goto clean;
+          }
+        }
+        allocator.deallocate(namespace, allocator.state);
+        i += len;
+      } while (NULL != (separator_pos = index(absolute_namespace + i + 1, '/')));
+    }
+  }
+
+  if (0 != strcmp(node_name, "*") && 0 != strcmp(node_name, "**")) {
```
jacobperron, line 536: I don't think a node name set to `**` should be valid, since it implies it has zero or more parts (eg. `""` or `"/foo/bar"`).

rcl_yaml_param_parser/src/parse.c
```diff
@@ -404,6 +409,147 @@ rcutils_ret_t parse_value(
...
+  if (absolute_namespace) {
+    size_t i = 0;
+    separator_pos = index(absolute_namespace + i + 1, '/');
+    if (NULL == separator_pos) {
+      ret = __validate_namespace(absolute_namespace);
+      if (RCUTILS_RET_OK != ret) {
+        goto clean;
+      }
+    } else {
+      do {
```
jacobperron, line 509: I think we can refactor the following logic so it is easier to read and maintain. Instead of making repeated allocations in the loop, what if we just did a find-and-replace of all `*` characters in the string with a random valid character (e.g. `*` -> `W`)? Then pass the result to `__validate_namespace`. It seems like it would be a lot simpler to me. The only caveat is that we risk running over the limit `RMW_NAMESPACE_MAX_LENGTH`, but I'm not sure that is a case we should worry about handling.

rcl_yaml_param_parser/src/parse.c
jacobperron: Nevermind, I see you added logic to support this. See my other comment https://github.com/ros2/rcl/pull/809#discussion_r504865010.

rcl_yaml_param_parser/src/parse.c
iuhilnehc-ynos: Actually, I thought that before. If it's partial matches for wildcards, such as '/foo/**part/node', this policy failed.

rcl_yaml_param_parser/src/parse.c
iuhilnehc-ynos: This case was for '/foo/bar/**'. OK, I'll add more codes to not allow node name to be '**'.

rcl_yaml_param_parser/src/parse.c
jacobperron: Ah, yeah that makes sense. Nevermind then.

rcl_yaml_param_parser/src/parse.c
jacobperron: I see. Yeah, this is tricky case. Maybe it's fine to leave the code as-is.

iuhilnehc-ynos: okay.

jacobperron (approved):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12770/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7729/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10483/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12714/)

iuhilnehc-ynos:
The `rcl_yaml_param_parser.cppcheck.error: syntaxError (src/parse.c:478)` on [macOS](http://ci.ros2.org/job/ci_osx/10483/) confused me, I'll rebuild it on my local environment tommorow because I have no environment now.

iuhilnehc-ynos:

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12779/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7738/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10492/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12725/)

rcl_yaml_param_parser/src/parse.c
```diff
@@ -18,14 +18,54 @@
...
 #include "rcl_yaml_param_parser/parser.h"
 
+///
+/// Check a name space whether it is valid
+///
+/// \param[in] namespace the namespace to check
+/// \return RCUTILS_RET_OK if namespace is valid, or
+/// \return RCUTILS_RET_INVALID_ARGUMENT if namespace is not valid, or
+/// \return RCUTILS_RET_ERROR if an unspecified error occurred.
+RCL_YAML_PARAM_PARSER_LOCAL
```
jacobperron, line 41: We should include `rcl_yaml_param_parser/visibility_control.h`.

rcl_yaml_param_parser/src/parse.c
```diff
@@ -18,14 +18,54 @@
...
+rcutils_ret_t
+__validate_namespace(const char * namespace);
```
jacobperron, line 43: I think we should use a single underscore. Double underscores are reserved identifiers.
```suggestion
_validate_namespace(const char * namespace);
```

Same for other functions names below.

jacobperron (commented):
I'm not sure about the cppcheck failure on macOS or the Windows build failure :thinking:

rcl_yaml_param_parser/src/parse.c
iuhilnehc-ynos: OK. I'll add it.
(Actually, I thought rcl_yaml_param_parser/parser.h already includes rcl_yaml_param_parser/visibility_control.h, so I ignore it before.)

rcl_yaml_param_parser/src/parse.c
iuhilnehc-ynos: That's good to know.

iuhilnehc-ynos:
> cppcheck failure on macOS

the version of `cppcheck` on macOS is 1.89, so I should not use 'namespace' as a variable name which is a keyword in C++.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12787/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7746/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10501/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12734/)

rcl_yaml_param_parser/src/parse.c
jacobperron: Maybe it's redundant, but I think it's generally good to directly include whatever is being used in a file.

jacobperron (approved):
LGTM, thanks for iterating!

PR #811, opened 2020-09-27
Improve trigger test for graph guard condition

Barry-Xu-2018:


fujitatomoya:
fix https://github.com/ros2/rcl/issues/574 (to keep the cross-reference)

Barry-Xu-2018:
@fujitatomoya 

Thanks for adding linker.  
I plan to add linker after internal checking. So not directly add linker to issue :)

fujitatomoya (changes requested):
@Barry-Xu-2018 
CC: @ivanpauno 

the current test code is to see if graph guard condition is triggered, but not checking the `graph side-effects` such as interfaces with https://github.com/ros2/rcl/blob/master/rcl/include/rcl/graph.h. I think that requirement is to check the `graph side-effects` too. (comes from https://github.com/ros2/rcl/issues/574#issuecomment-697687462)

Barry-Xu-2018:
@fujitatomoya 

> the current test code is to see if graph guard condition is triggered, but not checking the graph side-effects such as interfaces with https://github.com/ros2/rcl/blob/master/rcl/include/rcl/graph.h. I think that requirement is to check the graph side-effects too. (comes from #574 (comment))

Do you means I should call function in `graph.h` such as rcl_count_publishers() to check ? 

> With a "graph side-effect", I mean anything that can change the output of all the functions defined here and publisher_count/subscription_count functions.

My understanding is   
All `actions` affect output of functions list in [here](https://github.com/ros2/rcl/blob/master/rcl/include/rcl/graph.h). This effect is `graph side-effects`. So we should test these `actions`.

fujitatomoya:
> Do you means I should call function in graph.h such as rcl_count_publishers() to check ?

yes, i was thinking in that way. but i may be wrong since there are already tests for each method for https://github.com/ros2/rcl/blob/master/rcl/include/rcl/graph.h.

Barry-Xu-2018:
Yes. 
I check the remaining test in test_graph.cpp.  API listed in `graph.h` are checked.   
We can hear ivanpauno' opinion.

rcl/test/rcl/test_graph.cpp
```diff
@@ -1224,75 +1225,209 @@ TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_graph_query_functio
...
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+  {
+    EXPECT_EQ(RCL_RET_OK, rcl_shutdown(&context)) << rcl_get_error_string().str;
+    EXPECT_EQ(RCL_RET_OK, rcl_context_fini(&context)) << rcl_get_error_string().str;
+  });
+
+  rcl_node_t node = rcl_get_zero_initialized_node();
+  rcl_node_options_t node_options = rcl_node_get_default_options();
+  ret = rcl_node_init(&node, "test_graph", "", &context, &node_options);
```
ivanpauno, line 1263:
```suggestion
  ret = rcl_node_init(&node, "test_graph", "/", &context, &node_options);
```

rcl/test/rcl/test_graph.cpp
```diff
@@ -1224,75 +1225,209 @@ TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_graph_query_functio
...
-  ASSERT_GE(graph_changes_count, 4ul);
+    rcl_node_get_graph_guard_condition(&node);
+
+  // Graph change since first node created
+  ret = rcl_wait_set_clear(&wait_set);
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  ret = rcl_wait_set_add_guard_condition(&wait_set, graph_guard_condition, NULL);
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  ret = rcl_wait(&wait_set, time_to_sleep.count());
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
```
ivanpauno, line 1289: this is waiting to what graph change?

rcl/test/rcl/test_graph.cpp
```diff
@@ -1224,75 +1225,209 @@ TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_graph_query_functio
...
+    &pub, &node, ROSIDL_GET_MSG_TYPE_SUPPORT(test_msgs, msg, BasicTypes),
+    "/chatter_test_graph_guard_condition_topics", &pub_ops);
+  EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+
+  ret = rcl_wait_set_clear(&wait_set);
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  ret = rcl_wait_set_add_guard_condition(&wait_set, graph_guard_condition, NULL);
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  ret = rcl_wait(&wait_set, time_to_sleep.count());
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
```
ivanpauno, line 1304: can you add a macro/function doing this?

rcl/test/rcl/test_graph.cpp
```diff
@@ -1224,75 +1225,209 @@ TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_graph_query_functio
...
+
+  // Should not get graph change if no change
+  ret = rcl_wait_set_clear(&wait_set);
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  ret = rcl_wait_set_add_guard_condition(&wait_set, graph_guard_condition, NULL);
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  ret = rcl_wait(&wait_set, time_to_sleep.count());
+  ASSERT_EQ(RCL_RET_TIMEOUT, ret) << rcl_get_error_string().str;
```
ivanpauno, line 1430: mmm, I wouldn't add this. It will be a source of flaky fails.

ivanpauno:
> Do you means I should call function in graph.h such as rcl_count_publishers() to check ?

Calling `rcl_count_publishers()` doesn't trigger the graph guard condition, what should trigger it is the creation of a `publisher` that matches a `subscription` (and the opposite too).

The issue is that detecting a new publisher/subscription also triggers the graph guard condition, so checking that sounds impossible.

rcl/test/rcl/test_graph.cpp
```diff
@@ -35,6 +35,7 @@
 
 #include "rcutils/logging_macros.h"
 #include "rcutils/logging.h"
+#include "rcutils/env.h"
```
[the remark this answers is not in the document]
Barry-Xu-2018: Not necessary, Remove it

rcl/test/rcl/test_graph.cpp
```diff
@@ -1224,75 +1225,209 @@ TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_graph_query_functio
...
+  rcl_init_options_t init_options = rcl_get_zero_initialized_init_options();
+  ret = rcl_init_options_init(&init_options, rcl_get_default_allocator());
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+  {
+    EXPECT_EQ(RCL_RET_OK, rcl_init_options_fini(&init_options)) << rcl_get_error_string().str;
+  });
+
+  // Test in new ROS domain
+  ret = rcl_init_options_set_domain_id(&init_options, 66);
```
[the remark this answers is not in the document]
Barry-Xu-2018: > AFAIU, some of the CI machines have an assigned domain id to avoid collisions with other machines in the network, and changing the number can cause random failures.

Get it.  
But for this test, we need new domain to do test. How to choose suitable domain ID?  
After checking other test codes, I find domain ID "42" are used in below test codes. 
Whether can I use it ?

https://github.com/ros2/rcl/blob/3a9711180a240e07578ce9a9522a9bf25d6ac445/rcl/test/rcl/test_node.cpp#L119-L121

https://github.com/ros2/rcl/blob/3a9711180a240e07578ce9a9522a9bf25d6ac445/rcl/test/rcl/test_domain_id.cpp#L26-L27

rcl/test/rcl/test_graph.cpp
Barry-Xu-2018: It is related above first node creating.

rcl/test/rcl/test_graph.cpp
Barry-Xu-2018: I considered adding macro.  
But while check  failure happens, it is hard to address which step is failure by using macro.

rcl/test/rcl/test_graph.cpp
Barry-Xu-2018: Do you have any advice on how to check it ?

rcl/test/rcl/test_graph.cpp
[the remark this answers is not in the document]
ivanpauno: > But for this test, we need new domain to do test. How to choose suitable domain ID?

Why a different domain id is needed?

rcl/test/rcl/test_graph.cpp
ivanpauno: but it doesn't have to trigger the guard condition in this case, because you first created the node and then the guard condition.
You can expect the guard condition to be triggered in later not creations, but not in this one.

rcl/test/rcl/test_graph.cpp
ivanpauno: You can use a function and [scoped traces](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#adding-traces-to-assertions), but I would be great to not repeat the same 5 lines of code everywhere.

rcl/test/rcl/test_graph.cpp
ivanpauno: > Do you have any advice on how to check it ?

I don't, but I don't think it's possible to test that in a reliable way.
Maybe we can left this as-is, and I can try a repeated test in CI to see if is reliable or not.

rcl/test/rcl/test_graph.cpp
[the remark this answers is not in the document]
Barry-Xu-2018: If use old domain, previous test will affect graph change.  So use new domain for clear test environment.  
Is there any way which can clear existing graph change ?    
If yse, we can use existing domain.

rcl/test/rcl/test_graph.cpp
Barry-Xu-2018: Thanks. I will use scoped traces.

rcl/test/rcl/test_graph.cpp
Barry-Xu-2018: > but it doesn't have to trigger the guard condition in this case, because you first created the node and then the guard condition.

Yes. I also think so.   
But I find there is one graph change here actually.  So I have to add above codes to read this graph change.

rcl/test/rcl/test_graph.cpp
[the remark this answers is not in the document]
ivanpauno: > Is there any way which can clear existing graph change ?

You could wait repeatedly with a timeout until the graph guard condition is not more ready, that should "clear" previous graph changes.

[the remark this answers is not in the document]
Barry-Xu-2018: > You could wait repeatedly with a timeout until the graph guard condition is not more ready, that should "clear" previous graph changes.

Okay. I will try it.

Barry-Xu-2018:
@ivanpauno 

> You could wait repeatedly with a timeout until the graph guard condition is not more ready, that should "clear" previous graph changes.

Yeah. I have confirmed it and updated codes. 
Please help review again.

rcl/test/rcl/test_graph.cpp
```diff
@@ -1223,75 +1223,175 @@ TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_graph_query_functio
...
-  std::shared_future<bool> future = topic_changes_promise.get_future();
-  size_t graph_changes_count = 0;
-  // while the topic thread is not done, wait and count the graph changes
-  while (future.wait_for(std::chrono::seconds(0)) != std::future_status::ready) {
-    ret = rcl_wait_set_clear(this->wait_set_ptr);
+    rcl_node_get_graph_guard_condition(node_ptr);
+
+  // Wait for no graph change condition
+  int idx = 0;
+  while (idx < 100) {
```
ivanpauno, line 1260: where is `++idx`?

maybe a `for` loop is more clear

ivanpauno (commented):
LGTM with the only pending comment addressed

rcl/test/rcl/test_graph.cpp
Barry-Xu-2018: Sorry. It's a mistake. will fix it.

Barry-Xu-2018:
@ivanpauno 

I fix it and change to `for` loop.  
Please check again.

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13317/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8247/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11047/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13357/)

ivanpauno:
Thanks for the contribution @Barry-Xu-2018 !

clalancette:
I think this PR is causing the `test_graph__rmw_connext_cpp` failures we are seeing on all platforms: https://ci.ros2.org/view/nightly/job/nightly_linux_debug/1802/testReport/junit/projectroot/test/test_graph__rmw_connext_cpp/ .  If I revert this PR, then that test starts succeeding for me locally.

It's weird that the CI above was green, but it is failing in the nightlies.  Maybe there is something else at work here.  But for now I'm going to open a revert PR, and we can follow-up with another fix later.  @Barry-Xu-2018 @ivanpauno FYI.

ivanpauno:
> It's weird that the CI above was green

That one was only fastrtps, I think it makes sense to revert for the moment.

clalancette:
> That one was only fastrtps, I think it makes sense to revert for the moment.

Ah, I missed that.  That explains it, thanks.

PR #812, opened 2020-09-28
Use hash map for rcl_params_t and rcl_node_params_s

### LINKED ISSUE #808: Use hash map for rcl_params_t and rcl_node_params_s
#### Feature description
Currently, `find_node` and `find_parameter` use a loop find with an array to get an item which costs O(n), it's better to use a hash map instead of the array to store data so that it will costs O(1).

```
typedef struct rcl_node_params_s
{
  rcutils_hash_map_t params;      /// char* : rcl_variant_t*
} rcl_node_params_s;

typedef struct rcl_params_s
{
  rcutils_hash_map_t node_params; /// char* : rcl_node_params_t*
  rcutils_allocator_t allocator;  ///< Allocator used
} rcl_params_t;

```

iuhilnehc-ynos:
Fixes #808

### DIFF
--- a/rcl_yaml_param_parser/src/node_params.c
+++ b/rcl_yaml_param_parser/src/node_params.c
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "rcutils/strdup.h"
+
 #include "./impl/node_params.h"
 #include "./impl/types.h"
 #include "./impl/yaml_variant.h"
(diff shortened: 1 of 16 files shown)


iuhilnehc-ynos:
@clalancette 

Before continuing to modify the relevant `rclcpp` code, I think we need to reach a consensus on the modification of `rcl(rcl_yaml_param_parser)`. This PR updates a lot of things, could you please review it in your spare time? I'd like to hear your suggestions about it.

iuhilnehc-ynos:
Actually, there is no performance improvement by using `hash map` for `rcl_yaml_param_parser`.
(NOTE: I think it's because `rcutils_hash_map_t` need to copy all old items while `hash_map_check_and_grow_map`)

I checked the following test (already comment assert data after `rcl_parse_yaml_file`) ,

| test case | original | use hash map |
|---|---|---|
|./test_parser_multiple_nodes --gtest_filter=*multiple_number_nodes|6-7ms|7-9ms|
|./test_parser_multiple_params --gtest_filter=*multiple_params|2ms|4ms|

and `rclcpp` just copy `rcl_params_t` data into its `std::map`, I wonder if we should use hash map for `rcl_yaml_param_parser` now?

clalancette:
> Actually, there is no performance improvement by using `hash map` for `rcl_yaml_param_parser`.
> (NOTE: I think it's because `rcutils_hash_map_t` need to copy all old items while `hash_map_check_and_grow_map`)

Thanks for doing that analysis.  There is one underlying reason for using a hashmap; to make lookups faster.  That should benefit anyone calling `find_parameter`, which includes the initial parser.  I'm not sure why that is no better.

> and `rclcpp` just copy `rcl_params_t` data into its `std::map`,

That's unfortunate.  The whole point of the `rcl` layer is so we can share this stuff between client libraries, but I guess we are not using it here.

> I wonder if we should use hash map for `rcl_yaml_param_parser` now?

Yeah, I'm not sure.  I still think there is some value here if we figure out why it is slower, and can fix that.  But if you don't think that is worth the time, I can't argue with you.  I'll leave it up to you whether to pursue this further or close this PR out (both are fine with me).

rcl_yaml_param_parser/src/node_params.c
```diff
@@ -38,74 +40,89 @@ rcutils_ret_t node_params_init_with_capacity(
...
-  node_params->parameter_names = parameter_names;
-  // zero initialization for the added memory
-  if (new_capacity > node_params->capacity_params) {
-    memset(
-      node_params->parameter_names + node_params->capacity_params, 0,
-      (new_capacity - node_params->capacity_params) * sizeof(char *));
+  char * parameter_name = NULL;
+  rcl_variant_t * parameter_value = NULL;
+  char * param_name = NULL;
+  rcl_variant_t * param_value = NULL;
```
fujitatomoya, line 69: nitpicks:

```suggestion
  char * src_param_name = NULL;
  rcl_variant_t * src_param_value = NULL;
  char * dest_param_name = NULL;
  rcl_variant_t * dest_param_value = NULL;
```

rcl_yaml_param_parser/src/node_params.c
```diff
@@ -116,30 +133,33 @@ void rcl_yaml_node_params_fini(
...
-  if (NULL != node_params_st->parameter_values) {
-    for (size_t parameter_idx = 0U; parameter_idx < node_params_st->num_params;
-      parameter_idx++)
-    {
-      rcl_yaml_variant_fini(&(node_params_st->parameter_values[parameter_idx]), allocator);
+  char * param_name = NULL;
+  rcl_variant_t * param_value = NULL;
+  rcutils_ret_t ret = rcutils_hash_map_get_next_key_and_data(
+    &node_params_st->node_params_map, NULL, &param_name, &param_value);
+  while (RCUTILS_RET_OK == ret) {
```
fujitatomoya, line 140: I would use do while statement here.

rcl_yaml_param_parser/src/node_params.c
```diff
@@ -116,30 +133,33 @@ void rcl_yaml_node_params_fini(
...
-    node_params_st->parameter_values = NULL;
+    allocator.deallocate(param_name, allocator.state);
+    param_name = NULL;
+    allocator.deallocate(param_value, allocator.state);
+    param_value = NULL;
+    ret = rcutils_hash_map_get_next_key_and_data(
+      &node_params_st->node_params_map, NULL, &param_name, &param_value);
+  }
+  if (RCUTILS_RET_HASH_MAP_NO_MORE_ENTRIES != ret) {
+    RCUTILS_SET_ERROR_MSG("Failed to get next item for node parameters");
```
fujitatomoya, line 157: error message is not exactly correct.

```suggestion
    RCUTILS_SET_ERROR_MSG("Failed to remove all entries in the hash map");
```

fujitatomoya (commented):
a few minor comments for now.

fujitatomoya:
@iuhilnehc-ynos thanks for the PR!

> I wonder if we should use hash map for rcl_yaml_param_parser now?

this modifies a lot of current code, so I would say maybe it is not right time unless we have something better.

clalancette:
> this modifies a lot of current code, so I would say maybe it is not right time unless we have something better.

Yep, that is totally fine with me.  @iuhilnehc-ynos feel free to close this if you agree.

iuhilnehc-ynos:
Thanks for reviewing this, @fujitatomoya.

> I would say maybe it is not right time unless we have something better.

I agree with you.

@clalancette, sorry to take your time, I'll close this PR and the relevant issue.

PR #819, opened 2020-10-02
[foxy] Fix bug error handling get_param_files (#743)

jacobperron:
Backport #743 to Foxy.

### DIFF
--- a/rcl/QUALITY_DECLARATION.md
+++ b/rcl/QUALITY_DECLARATION.md
@@ -2,7 +2,7 @@ This document is a declaration of software quality for the `rcl` package, based 
 
 # `rcl` Quality Declaration
 
-The package `rcl` claims to be in the **Quality Level 4** category.
+The package `rcl` claims to be in the **Quality Level 2** category.
 
 Below are the rationales, notes, and caveats for this claim, organized by each requirement listed in the [Package Quality Categories in REP-2004](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#package-quality-categories).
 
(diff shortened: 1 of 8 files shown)


jacobperron:
Looks like there may be a prerequisite I've missed in order to backport this change.

ivanpauno:
`rcl/test/rcl/test_arguments.cpp` is missing an include to `rcl/test/rcl/allocator_testing_utils.h`, that was added in a PR before.

IMO, backporting the PRs adding those tests isn't needed.
I would only backport the fix without backporting any of the new tests.

brawner:
I would recommend backporting test coverage PRs to foxy. @chapulina and @ahcorde can also weigh in on that

ivanpauno:
> I would recommend backporting test coverage PRs to foxy. @chapulina and @ahcorde can also weigh in on that

Backporting tests is ok, but that shouldn't block the backport of a (one line) fix.
If the test utilities haven't been backported yet, the test cases depending on them can be backported in the future.

brawner:
@ivanpauno that's reasonable. Tests can be backported in a separate PR

jacobperron:

jacobperron:
There's a missing header include that was added in https://github.com/ros2/rcl/pull/703, I'm not sure if there are plans to backport it. If not, then I can cherry-pick the missing include line.

ivanpauno:
> There's a missing header include that was added in #703, I'm not sure if there are plans to backport it. 

I think that most of the test coverage that was added to Rolling is being backported to Foxy, but I'm not sure about that.
@brawner probably knows.

> If not, then I can cherry-pick the missing include line.

:+1:
You can also only backport the changes in `rcl/src/rcl/arguments.c` without backporting the tests, I'm ok with what you prefer doing.

jacobperron:
I asked around and it sounds like there are plan to backport #703, so I'll just wait.

jacobperron:
I don't have a reference, but this was backported to Foxy elsewhere.

brawner:
I think in https://github.com/ros2/rcl/pull/848

PR #830, opened 2020-10-13
[rcl_lifecycle] Add warnings

audrow:
This PR enables `-Wformat=2`, `-Wconversion`, `-Wshadow`, and `-Wsign-conversion` in `rcl_lifecycle`. This PR relies on using gtest v1.10.0, see https://github.com/ament/googletest/pull/8.

### DIFF
--- a/rcl_lifecycle/src/com_interface.c
+++ b/rcl_lifecycle/src/com_interface.c
@@ -250,9 +250,9 @@ rcl_lifecycle_com_interface_publish_notification(
   rcl_lifecycle_com_interface_t * com_interface,
   const rcl_lifecycle_state_t * start, const rcl_lifecycle_state_t * goal)
 {
-  msg.start_state.id = start->id;
+  msg.start_state.id = (uint8_t) start->id;
   rosidl_runtime_c__String__assign(&msg.start_state.label, start->label);
-  msg.goal_state.id = goal->id;
+  msg.goal_state.id = (uint8_t) goal->id;
   rosidl_runtime_c__String__assign(&msg.goal_state.label, goal->label);
 
   return rcl_publish(&com_interface->pub_transition_event, &msg, NULL);
--- a/rcl_lifecycle/src/com_interface.c
+++ b/rcl_lifecycle/src/com_interface.c
@@ -250,9 +250,9 @@ rcl_lifecycle_com_interface_publish_notification(
   rcl_lifecycle_com_interface_t * com_interface,
   const rcl_lifecycle_state_t * start, const rcl_lifecycle_state_t * goal)
 {
-  msg.start_state.id = (uint8_t) start->id;
+  msg.start_state.id = start->id;
   rosidl_runtime_c__String__assign(&msg.start_state.label, start->label);
-  msg.goal_state.id = (uint8_t) goal->id;
+  msg.goal_state.id = goal->id;
   rosidl_runtime_c__String__assign(&msg.goal_state.label, goal->label);
 
   return rcl_publish(&com_interface->pub_transition_event, &msg, NULL);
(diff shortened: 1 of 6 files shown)


audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12720/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7681/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10434/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12667/)

fujitatomoya (approved):
would you share why you came up with these options and plan to add these options in other packages as well?

audrow:
Hi @fujitatomoya, for the next ROS2 release, we want to add several warnings to all of the maintained C and C++ packages. The thinking is that this will help maintain code quality and catch bugs (it has already caught a few in other packages). I made a proposal and worked with the team to find a set of warnings to include. I'll ask if I can make the proposal public.

Below is the full set of warnings. This PR doesn't do all of them because some of the warnings require solving additional problems (so far, mostly to do with vendor packages, like googletest).
```
-Wall -Wextra -Wpedantic -Wformat=2 -Wundef -Wconversion -Woverloaded-virtual -Wswitch-default -Wshadow -Wsign-conversion -Wnon-virtual-dtor -Wold-style-cast -Wcast-qual
```

fujitatomoya:
@audrow 

sounds great 👍  thanks 😃

rcl_lifecycle/src/com_interface.c
```diff
@@ -250,9 +250,9 @@ rcl_lifecycle_com_interface_publish_notification(
   rcl_lifecycle_com_interface_t * com_interface,
   const rcl_lifecycle_state_t * start, const rcl_lifecycle_state_t * goal)
 {
-  msg.start_state.id = start->id;
+  msg.start_state.id = (uint8_t) start->id;
```
clalancette, line 253: So, turning on the warning here has shown a potential bug (which is a good thing!).

The problem is that rcl_lifecycle_state_t has an `id` member of type `unsigned int`: https://github.com/ros2/rcl/blob/0ca8c6162136594e490ba69fbf288de1b5f7073d/rcl_lifecycle/include/rcl_lifecycle/data_types.h#L35

But the message we are trying to stuff it into (State message) has an `id` member of type `uint8`: https://github.com/ros2/rcl_interfaces/blob/7a6ebaa2064151a48dd189cd514fc7db5d260047/lifecycle_msgs/msg/State.msg#L54

That means that it is conceivable that we could generate an ID internally that we cannot represent in the message.  Just think of the case where we have 256 as the `start->id`; this will wrap back to 0 and publish a bogus state id.

There are 2 ways we can fix this:
1.  Increase the size of the State.msg field to accommodate the possible ID.
1.  Make the internal `rcl_lifecycle_state_t` member `id` be a uint8.

I would go for the second one.  Given that these are meant to be the IDs of states, then 255 really should be enough.  If it turns out it isn't in the future, we'll have to do work on both the internal structure and the external message to expand them.

clalancette (changes requested):

audrow:
I also tested above packages to see if changing the type of `rcl_lifecycle_state_t`'s member `id` caused any warnings.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12815/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7772/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10529/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12760/)

clalancette (approved):

PR #832, opened 2020-10-14
[foxy] Fix memory leak in rcl_subscription_init()/rcl_publisher_init() (#794)

jacobperron:
Backport #794 to Foxy.

### DIFF
--- a/rcl/src/rcl/publisher.c
+++ b/rcl/src/rcl/publisher.c
@@ -201,8 +201,7 @@ fail:
       rmw_ret_t rmw_fail_ret = rmw_destroy_publisher(
         rcl_node_get_rmw_handle(node), publisher->impl->rmw_handle);
       if (RMW_RET_OK != rmw_fail_ret) {
-        RCUTILS_SAFE_FWRITE_TO_STDERR_WITH_FORMAT_STRING(
-          "%s, at %s:%d\n", rmw_get_error_string().str, __FILE__, __LINE__);
+        RCUTILS_SAFE_FWRITE_TO_STDERR(rmw_get_error_string().str);
       }
     }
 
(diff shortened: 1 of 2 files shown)


wjwwood (approved):

fujitatomoya (approved):

jacobperron:
The rcutils macro being used isn't available in Foxy, so I've opted to not use a format string to print the file and line number (4bf6989).

rcl/src/rcl/publisher.c
```diff
@@ -197,6 +197,14 @@ rcl_publisher_init(
   goto cleanup;
 fail:
   if (publisher->impl) {
+    if (publisher->impl->rmw_handle) {
+      rmw_ret_t rmw_fail_ret = rmw_destroy_publisher(
+        rcl_node_get_rmw_handle(node), publisher->impl->rmw_handle);
+      if (RMW_RET_OK != rmw_fail_ret) {
+        RCUTILS_SAFE_FWRITE_TO_STDERR(rmw_get_error_string().str);
```
wjwwood, line 204: The formatting wasn't required, see:

https://github.com/ros2/rcutils/blob/0c2cc271c3d3a357d16f15037606c15e6f7b0978/include/rcutils/error_handling.h#L284-L297

So if the other pr had that it may have been redundant?

wjwwood, line 204: You also need to have a `\n`, since this write doesn't add one, I'd just add another statement like this:

```suggestion
        RCUTILS_SAFE_FWRITE_TO_STDERR(rmw_get_error_string().str);
        RCUTILS_SAFE_FWRITE_TO_STDERR("\n");
```

jacobperron: Yes, it's redundant. Good catch. Fixed in https://github.com/ros2/rcl/pull/834

jacobperron: Applied in 62db41b

wjwwood (approved):

PR #836, opened 2020-10-16
Add function to convert qos policy values to and from strings

ivanpauno:
I'm not sure if the correct place would be here or in `rmw` (moving the impl is easy), maybe some rmw implementations can take advantage of these functions.

Required by https://github.com/ros2/rclcpp/pull/1408, the idea is to avoid custom conversions in different clients (rclcpp, rclpy).

### DIFF
--- a/rcl/CMakeLists.txt
+++ b/rcl/CMakeLists.txt
@@ -52,6 +52,7 @@ set(${PROJECT_NAME}_sources
   src/rcl/node.c
   src/rcl/node_options.c
   src/rcl/publisher.c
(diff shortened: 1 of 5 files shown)


jacobperron (commented):
Considering the functions operate only on RMW types, I'm inclined to say `rmw` is the best place for the implementation to live. No strong feelings about it though. @wjwwood Thoughts?

wjwwood:
I don't have a strong opinion either, so I guess the small benefit of having them lower in the stack in `rmw` might be the way to go.

PR #844, opened 2020-10-28
Fix for external log segfault after SIGINT

iuhilnehc-ynos:
Fixes #https://github.com/ros2/rclcpp/issues/1418

### DIFF
--- a/rcl/src/rcl/logging.c
+++ b/rcl/src/rcl/logging.c
@@ -126,6 +126,12 @@ rcl_ret_t rcl_logging_fini()
 {
   rcl_ret_t status = RCL_RET_OK;
   rcutils_logging_set_output_handler(rcutils_logging_console_output_handler);
+  // In order to output log message to `rcutils_logging_console_output_handler`
+  // and `rcl_logging_ext_lib_output_handler` is not called after `rcl_logging_fini`,
+  // in addition to calling `rcutils_logging_set_output_handler`,
+  // the `g_rcl_logging_num_out_handlers` and `g_rcl_logging_out_handlers` must be updated.
+  g_rcl_logging_num_out_handlers = 1;
+  g_rcl_logging_out_handlers[0] = rcutils_logging_console_output_handler;
 
   if (g_rcl_logging_rosout_enabled) {
     status = rcl_logging_rosout_fini();


fujitatomoya (approved):

rcl/src/rcl/logging.c
```diff
@@ -126,6 +126,12 @@ rcl_ret_t rcl_logging_fini()
 {
   rcl_ret_t status = RCL_RET_OK;
   rcutils_logging_set_output_handler(rcutils_logging_console_output_handler);
+  // In order to output log message to `rcutils_logging_console_output_handler`
+  // and `rcl_logging_ext_lib_output_handler` is not called after `rcl_logging_fini`,
+  // in addition to calling `rcutils_logging_set_output_handler`,
+  // the `g_rcl_logging_num_out_handlers` and `g_rcl_logging_out_handlers` must be updated.
+  g_rcl_logging_num_out_handlers = 1;
+  g_rcl_logging_out_handlers[0] = rcutils_logging_console_output_handler;
```
ivanpauno, line 134: My first impression was that this is correct, but checking it again I don't get why these lines are needed.

L128 is uninstalling the previous output handler and reinstalling the console output handler, that should be enough to make sure that `g_rcl_logging_out_handlers` elements aren't used anymore (that's always used through `rcl_logging_ext_lib_output_handler`).

There's clearly a synchronization problem I'm not seeing here.

iuhilnehc-ynos: > L128 is uninstalling the previous output handler and reinstalling the console output handler, that should be enough to make sure that g_rcl_logging_out_handlers elements aren't used anymore

The log function in rcutils(`rcutils_logging_set_output_handler` and `rcutils_log`) is not thread-safe.
If the function `rcutils_log` already enter into `rclcpp_logging_output_handler`, but not to call `rcl_logging_multiple_output_handler` immediately, at this time, `rcl_logging_fini` call `rcutils_logging_set_output_handler` to uninstall the previous output handler and reinstall the console output handler, even if this case, the function `rclcpp_logging_output_handler` is still going to execute `rcl_logging_multiple_output_handler` that will continue to call `rcl_logging_ext_lib_output_handler` by using `g_rcl_logging_out_handlers`.

BTW: Did you already know how the issue happened? Do these two comments (https://github.com/ros2/rclcpp/issues/1418#issuecomment-714879097 and https://github.com/ros2/rclcpp/issues/1418#issuecomment-717299537) still make you confused?

iuhilnehc-ynos: Sorry, actually I don't know what kind of information I should provide for the tag `more-information-needed`, I just repeat the issue again and again.

ivanpauno: > The log function in rcutils(rcutils_logging_set_output_handler and rcutils_log) is not thread-safe.
> If the function rcutils_log already enter into rclcpp_logging_output_handler, but not to call rcl_logging_multiple_output_handler immediately, at this time, rcl_logging_fini call rcutils_logging_set_output_handler to uninstall the previous output handler and reinstall the console output handler, even if this case, the function rclcpp_logging_output_handler is still going to execute rcl_logging_multiple_output_handler that will continue to call rcl_logging_ext_lib_output_handler by using g_rcl_logging_out_handlers.

I'm confused because the [rclcpp output handler](https://github.com/ros2/rclcpp/blob/aa159a5e8f6d3bf50c4c0c40dbfb656f7e8c6f97/rclcpp/src/rclcpp/context.cpp#L134) and [logging fini](https://github.com/ros2/rclcpp/blob/aa159a5e8f6d3bf50c4c0c40dbfb656f7e8c6f97/rclcpp/src/rclcpp/context.cpp#L329) are mutually exclusive.
i.e. all calls to `rcl_logging_multiple_output_handler` and `rcl_logging_fini` should be happening in a mutually exclusive fashion (because rclcpp is [installing a different output handler](https://github.com/ros2/rclcpp/blob/aa159a5e8f6d3bf50c4c0c40dbfb656f7e8c6f97/rclcpp/src/rclcpp/context.cpp#L219-L222)).
If this is a bug, it means that there are calls to `rcl_logging_multiple_output_handler` coming from another place, or the synchronization mentioned above is somewhat wrong.

I also don't see how this patch could be fixing anything, it's writing to some non atomic variables without any synchronization, so the problem can still happen (though maybe this patch is making the error less likely).

Maybe making the global rcutils log output handler an atomic with sequential consistency would be a good idea: https://github.com/ros2/rcutils/blob/4c534768869bdec1bf380832729e1eb2f2bed896/src/logging.c#L83, but it would be good to really understand why the synchronization mentioned above is not solving the problem.

> BTW: Did you already know how the issue happened? Do these two comments (https://github.com/ros2/rclcpp/issues/1418#issuecomment-714879097 and https://github.com/ros2/rclcpp/issues/1418#issuecomment-717299537) still make you confused?

Yes, I read those comments and the issue isn't still clear to me.
e.g.: [here](https://github.com/ros2/rclcpp/issues/1418#issuecomment-714879097) interleaving step 4 and 5 between step 3 and 6 shouldn't be possible by the mutually exclusion mechanism commented above, so why is that happening?

> Sorry, actually I don't know what kind of information I should provide for the tag more-information-needed, I just repeat the issue again and again.

The tag means that the reviewer needs more information to understand the issue and feel comfortable to merge the fix.
If you need to repeat something to make things more clear that's ok, people sometimes overlook or fail to understand previous comments.
This kind of comment will not make the review happen faster and doesn't add any value to the discussion.

iuhilnehc-ynos: > e.g.: here interleaving step 4 and 5 between step 3 and 6 shouldn't be possible by the mutually exclusion mechanism commented above, so why is that happening?

Please see the new added "<updated>" for lock order, it will prove the steps 4 and 5 and can happen between step 3 and 6. (notice that lock 2 is not before calling `rcutils_log`)
```
                rcutils_log
                    rclcpp_logging_output_handler                  step 3    ready to log by rcl_logging_external_log
                        <updated>lock 2
                        rcl_logging_multiple_output_handler
                            rcl_logging_ext_lib_output_handler
                                g_root_logger->log                 step 6    crash here if using the shutdowned log
                         <updated>unlock 2

        // shutdown logger
        <updated> lock 1
        rcl_logging_fini()
            rcutils_logging_set_output_handler                     step 4    setting handler is too late for this case
            rcl_logging_external_shutdown                          step 5    shutdown the external log
        <updated> unlock 1
```

ivanpauno: ah, I see. Ok, this should work then.

ivanpauno (approved):

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12921/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7874/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10637/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12870/)

ivanpauno:
The macOS failure seems to be unrelated, a similar failure can be found in nightlies: https://ci.ros2.org/view/nightly/job/nightly_osx_repeated/2141/testReport/junit/(root)/rcl/test_node__rmw_fastrtps_cpp_gtest_missing_result/.

Thanks for the contribution @iuhilnehc-ynos !

Issue #851, opened 2020-11-02
👩‍🌾 test_node_info_publishers flaky on Windows Release


chapulina:
It seems to be failing about 1/20 builds:

https://ci.ros2.org/view/nightly/job/nightly_win_rel/1740/testReport/rcl/NodeGraphMultiNodeFixture/test_node_info_publishers/history/

The last 2 failures have been the same:

```
C:\ci\ws\src\ros2\rcl\rcl\test\rcl\test_graph.cpp:966
Expected: (attempts) <= (max_attempts), actual: 5 vs 4
Unable to attain all required nodes
```

PR #854, opened 2020-11-03
Make WaitForAllNodesAlive in test_graph.cpp more reliable

ivanpauno:
\+ and fix leaks

Fixes https://github.com/ros2/rcl/issues/851.

### DIFF
--- a/rcl/test/rcl/test_graph.cpp
+++ b/rcl/test/rcl/test_graph.cpp
@@ -946,25 +946,23 @@ public:
 
   void WaitForAllNodesAlive()
   {
-    rcl_ret_t ret;
-    rcutils_string_array_t node_names = rcutils_get_zero_initialized_string_array();
-    rcutils_string_array_t node_namespaces = rcutils_get_zero_initialized_string_array();
-    OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
-    {
-      ret = rcutils_string_array_fini(&node_names);
-      ASSERT_EQ(RCUTILS_RET_OK, ret);
-      ret = rcutils_string_array_fini(&node_namespaces);
-      ASSERT_EQ(RCUTILS_RET_OK, ret);
-    });
     // wait for all 3 nodes to be discovered: remote_node, old_node, node
     size_t attempts = 0;
-    size_t max_attempts = 4;
-    while (node_names.size < 3) {
+    size_t max_attempts = 10;
+    rcutils_string_array_t node_names = rcutils_get_zero_initialized_string_array();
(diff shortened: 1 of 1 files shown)


rcl/test/rcl/test_graph.cpp
```diff
@@ -946,25 +946,23 @@ class NodeGraphMultiNodeFixture : public CLASSNAME(TestGraphFixture, RMW_IMPLEME
...
       std::this_thread::sleep_for(std::chrono::seconds(1));
-      ret = rcl_get_node_names(this->remote_node_ptr, allocator, &node_names, &node_namespaces);
+      node_names = rcutils_get_zero_initialized_string_array();
+      node_namespaces = rcutils_get_zero_initialized_string_array();
+      ASSERT_EQ(
+        RCL_RET_OK,
+        rcl_get_node_names(this->remote_node_ptr, allocator, &node_names, &node_namespaces));
       attempts++;
+      ASSERT_EQ(RCUTILS_RET_OK, rcutils_string_array_fini(&node_names));
+      ASSERT_EQ(RCUTILS_RET_OK, rcutils_string_array_fini(&node_namespaces));
```
wjwwood, line 963: Why not leave this in a scope exit?

ivanpauno: I don't see a big benefit of using that here.

Before we were doing so, and the test was still leaking because of passing the same string array `node_names` and `node_namespaces` multiple times to `rcl_get_node_names()`.

Now we're calling `rcutils_string_array_fini()` immediately after the function that makes the allocation, so the scope exit macro does not seem to add much value.

wjwwood: Oh I guess there's no reason to not get to these lines if `rcl_get_node_names` succeeds, I was thinking like if someone added an assertion between lines 959 and 962, it would leak if that assertion failed.

But as-is it is fine.

wjwwood (approved):

wjwwood:
Also these method names are styled incorrectly, but I guess that slipped past a review.

ivanpauno:
> Also these method names are styled incorrectly, but I guess that slipped past a review.

Fixed in https://github.com/ros2/rcl/pull/854/commits/135125a0e2cbc6315c82d500831770beb8af19eb.

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12963/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7912/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10681/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12912/)

ivanpauno:

PR #860, opened 2020-11-06
Allow forward slashes within a parameter name rule in argument parsing

ivanpauno:
While implementing a demo for https://github.com/ros2/design/pull/300, https://github.com/ros2/rclcpp/pull/1408, I found that though parameter names containing forward slashes are valid and you can set them from a parameter file, you cannot set them with a command line parameter rule.

e.g.:
```
ros2 run quality_of_service_demo_cpp qos_overrides_listener --ros-args -p qos_overrides./qos_overrides_chatter.subscription.depth:=11
```
doesn't work without this PR while
```yaml
qos_overrides_listener:  # node name
  ros__parameters:  # All parameters are nested in this key
    qos_overrides:  # Parameter group for all qos overrides
      /qos_overrides_chatter:  # Parameter group for the topic
        subscription:  # Profile for subscriptions in the topic
        # subscription_custom_identifier:  # Subscription with override id="custom_identifier"
          reliability: reliable
          depth: 11
```

does work.

### DIFF
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -1248,6 +1248,45 @@ _rcl_parse_resource_match(
   return RCL_RET_OK;
 }
 
+RCL_LOCAL
+rcl_ret_t
+_rcl_parse_param_name_token(rcl_lexer_lookahead2_t * lex_lookahead)
+{
+  rcl_ret_t ret;
+  rcl_lexeme_t lexeme;
+
+  // Check arguments sanity
+  assert(NULL != lex_lookahead);
+
+  ret = rcl_lexer_lookahead2_peek(lex_lookahead, &lexeme);
+  if (RCL_RET_OK != ret) {
+    return ret;
+  }
+  if (RCL_LEXEME_TOKEN != lexeme && RCL_LEXEME_FORWARD_SLASH != lexeme) {
+    if (RCL_LEXEME_WILD_ONE == lexeme) {
+      RCL_SET_ERROR_MSG("Wildcard '*' is not implemented");
+      return RCL_RET_ERROR;
+    } else if (RCL_LEXEME_WILD_MULTI == lexeme) {
+      RCL_SET_ERROR_MSG("Wildcard '**' is not implemented");
+      return RCL_RET_ERROR;
+    } else {
+      RCL_SET_ERROR_MSG("Expecting token or wildcard");
+      return RCL_RET_WRONG_LEXEME;
+    }
+  }
+  do {
+    ret = rcl_lexer_lookahead2_accept(lex_lookahead, NULL, NULL);
+    if (RCL_RET_OK != ret) {
+      return ret;
+    }
+    ret = rcl_lexer_lookahead2_peek(lex_lookahead, &lexeme);
+    if (RCL_RET_OK != ret) {
+      return ret;
+    }
+  } while (RCL_LEXEME_TOKEN == lexeme || RCL_LEXEME_FORWARD_SLASH == lexeme);
+  return RCL_RET_OK;
+}
+
 /// Parse a parameter name in a parameter override rule (ex: `foo.bar`)
 /**
  * \sa _rcl_parse_param_rule()
@@ -1277,7 +1316,7 @@ _rcl_parse_param_name(
   }
 
   // token ( '.' token )*
-  ret = _rcl_parse_resource_match_token(lex_lookahead);
+  ret = _rcl_parse_param_name_token(lex_lookahead);
   if (RCL_RET_OK != ret) {
     return ret;
   }
@@ -1290,7 +1329,7 @@ _rcl_parse_param_name(
     if (RCL_RET_WRONG_LEXEME == ret) {
       return RCL_RET_INVALID_REMAP_RULE;
     }
-    ret = _rcl_parse_resource_match_token(lex_lookahead);
+    ret = _rcl_parse_param_name_token(lex_lookahead);
     if (RCL_RET_OK != ret) {
       return ret;
     }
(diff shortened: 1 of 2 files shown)


hidmic (commented):
Hmm, is it a good idea to allow slashes? I mean, there's nothing fundamentally wrong about it, but it can be confusing. Moreover, there's a [design proposal](https://github.com/ros2/design/pull/241) (true, it's collecting dust) that suggests we stop using periods `.` and use slashes instead.

ivanpauno:
> Hmm, is it a good idea to allow slashes? I mean, there's nothing fundamentally wrong about it and your patch is correct, but it can be confusing. Moreover, there's a [design proposal](https://github.com/ros2/design/pull/241) (true, it's collecting dust) that suggests we stop using periods `.` and use slashes instead.

In the case of the parameter file, I think that it wouldn't be a problem if we change the addressing scheme in the future:

```yaml
/my/node/name:
  ros__parameters:
    qos_overrides:
      /my/topic/name:
```

is now

```
ros2_param://my/node/name/qos_overrides./my/topic/name
```

and in the new addressing scheme would be:

```
ros2_param://my.node.name/qos_overrides/my/topic/name
```

(using `/` is equivalent to creating a nested dict in the last example)

---

In the case of parameters passed through the command line, if we ever change the addressing scheme we will already be breaking how to override `param_group.param_name`: from `-p param_group.param_name:=value` to `-p param_group/param_name:=value`.

It doesn't seem that not allowing forward slashes in parameter names now will help to make forward compatibility with the mentioned design proposal easier.
The only place where we're not accepting slashes in parameter names is in command line rules, everywhere else you can already use forward slashes.

clalancette:
> Moreover, there's a [design proposal](https://github.com/ros2/design/pull/241) (true, it's collecting dust) that suggests we stop using periods `.` and use slashes instead.

Honestly, I think that ship has sailed.  While it would be nice to change to using slashes for parameters to support real URIs, it would cause a *huge* amount of downstream churn.  I personally don't think the benefits outweigh the costs there.  I don't necessarily want to get into a debate about that here (we can talk about it more on ros2/design#241), but given that assumption, I think we need to decide on whether this PR makes sense on its own.

So there are 2 ways I can see us going here:
1.  Make the command-line tools support slashes in parameter names to match the rest of the code (this PR).
1.  Declare that slashes aren't allowed in parameter names.

The benefit to the first one is that it keeps existing behavior and just expands the command-line parsing to support it.  The downside is that slashes in parameter names seems somewhat confusing to me (they do seem like a delimiter, even though we aren't using them as such right now).

The benefit to the second one is that we disallow things that look like delimiters in the parameter names.  The big downside is that we potentially break users who are already using `/` in their parameter names.  I expect that the vast majority of parameter names don't currently use slashes, so the breakage would be constrained.  But we would still probably have to do a deprecation cycle for it.

On balance, my preference is to go with option 1, and take this PR (or something like it).  No breakage of downstream users, and we make things in the ecosystem more consistent.  But I'm still interested in other opinions.  @ros2/team any thoughts?

hidmic:
> I think that it wouldn't be a problem if we change the addressing scheme in the future

If we allow the delimiter character as a valid token character, we end up with ambiguous URIs or parameter aliasing e.g. with

```yaml
/my/node/name:
  ros__parameters:
    qos_overrides:
      /my:
          topic/name: foo
      /my/topic/name: bar
```

either `ros2_param://my.node.name/qos_overrides/my/topic/name` URI refers to both `foo` and `bar`, or we have two ways to specify the same parameter (actually more). The latter may be regarded as a feature but definitely not something you can add w/o silently breaking downstream applications. Same if we were to allow periods `.` in parameter names.

> Honestly, I think that ship has sailed.

Considering slashes are allowed everywhere else, it probably has. FYI @wjwwood. 

I will also say that the current state of things precludes unambiguous `rosparam` URIs entirely. Does

```
rosparam://my/node/name/qos_overrides./my/topic/name
```

refer to `/my/node`'s `name/qos_overrides./my/topic/name` parameter or `/my/node/name`'s `qos_overrides./my/topic/name` parameter?

ivanpauno:
> Same if we were to allow periods . in parameter names.

IIRC, we currently allow `.` in parameter files key names :

```yaml
/my/node/name:
  ros__parameters:
    param_group:
      nested_param_group:
        param_name: value
```

seems to behave as:

```yaml
/my/node/name:
  ros__parameters:
    param_group.nested_param_group.param_name: value
```

hidmic:
Interesting. So we do have aliasing already. I wonder if that was intentional or an unintended consequence of `rcl_yaml_param_parser` implementation.

ivanpauno:
> I will also say that the current state of things precludes unambiguous rosparam URIs entirely

That's fair.
We're not currently using parameters URI's anywhere AFAIK.

We could make them unambiguous by not allowing slashes in the "outer" parameter group name, but that sounds too bizarre.

---

Unrelated question: for the qos parameteres overrides proposal, what could be the alternative of using forward slashes in the parameter name?

hidmic:
> for the qos parameteres overrides proposal, what could be the alternative of using forward slashes in the parameter name

Hmm, considering parameter limitations, perhaps having profiles that can match one or more publishers and/or subscriptions would do. But it's less straightforward.

But hey, it looks like we already have (support?) parameter aliasing! When/if we switch from periods to slashes, parameter URIs ambiguity would go away. It'd force that change if we ever need parameter URIs though.

wjwwood:
> > Honestly, I think that ship has sailed.

> Considering slashes are allowed everywhere else, it probably has. FYI @wjwwood.

I think we should keep an open mind about these kinds of changes, slow though they may need to be rolled out. Conceptually it is just broken, not to mention breaks with ROS 1's style for parameter namespacing.

hidmic:
We never made a decision here, which means QoS overrides still can't be passed through the command line :D 

We won't be changing parameter addressing for Galactic, but we can merge and backport this.

jacobperron:
I'm okay with merging this. Should we add a test or is this already covered?

jacobperron:
This is still broken. I think we should merge this "fix" and backport it. It sounds like any change to parameter namespaces would be disruptive with or without this change anyways.

clalancette:
> This is still broken. I think we should merge this "fix" and backport it. It sounds like any change to parameter namespaces would be disruptive with or without this change anyways.

Yes, I agree.  I think we should take this fix, and then deal with the larger question separately.  @ivanpauno do you mind rebasing this, and adding a test for this case?

fujitatomoya (approved):
lgtm

ivanpauno:
Rpr failure is unrelated and seems to be fine according to [this comment](https://github.com/ros2/rcl/pull/951#issuecomment-1075514115).

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16417/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11025/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16788/)

PR #863, opened 2020-11-19
Fix another failing test on CentOS 7.

clalancette:
Replace the use of inject_on_return with patch_and_return.

### DIFF
--- a/rcl/test/CMakeLists.txt
+++ b/rcl/test/CMakeLists.txt
@@ -37,7 +37,6 @@ if("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
       if("${key}" STREQUAL "id")
         string(REGEX REPLACE "^.*=(.*)" "\\1" DISTRIBUTION "${line}")
         string(TOLOWER "${DISTRIBUTION}" DISTRIBUTION)
-        message(STATUS "CHRIS: ${DISTRIBUTION}")
         break()
       endif()
     endforeach()
@@ -183,6 +182,9 @@ function(test_target_function)
     AMENT_DEPENDENCIES ${rmw_implementation}
   )
 
+  # CentOS does not support mimick's inject_on_return functionality, which
+  # causes test_node to segfault when it runs.  Since CentOS is not a Tier 1
+  # platform, just disable the test since we are covering it elsewhere.
   if(NOT "${DISTRIBUTION}" STREQUAL "\"centos\"")
     rcl_add_custom_gtest(test_node${target_suffix}
       SRCS rcl/test_node.cpp
--- a/rcl/test/CMakeLists.txt
+++ b/rcl/test/CMakeLists.txt
@@ -23,6 +23,27 @@ set(extra_lib_dirs "${rcl_lib_dir}")
 set(test_resources_dir_name "resources")
 add_definitions(-DTEST_RESOURCES_DIRECTORY="${CMAKE_CURRENT_BINARY_DIR}/${test_resources_dir_name}")
 
+set(DISTRIBUTION "Unknown")
+if("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
+  # If we are on Linux, look for /etc/os-release, which has "key=value" items.
+  # Parse those items, looking for a key named "id" (case-insensitive), and
+  # if we find it, set DISTRIBUTION to that value.  That gives us some idea
+  # of which Linux distribution we are on.
+  if(EXISTS "/etc/os-release")
+    file(STRINGS "/etc/os-release" OS_RELEASE)
+    foreach(line ${OS_RELEASE})
+      string(REGEX REPLACE "^(.*)=.*" "\\1" key "${line}")
+      string(TOLOWER "${key}" key)
+      if("${key}" STREQUAL "id")
+        string(REGEX REPLACE "^.*=(.*)" "\\1" DISTRIBUTION "${line}")
+        string(TOLOWER "${DISTRIBUTION}" DISTRIBUTION)
+        message(STATUS "CHRIS: ${DISTRIBUTION}")
+        break()
+      endif()
+    endforeach()
+  endif()
+endif()
+
 # finding gtest once in the highest scope
 # prevents finding it repeatedly in each local scope
 ament_find_gtest()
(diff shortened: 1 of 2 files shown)


rcl/test/rcl/test_node.cpp
```diff
@@ -511,15 +511,21 @@ TEST_F(CLASSNAME(TestNodeFixture, RMW_IMPLEMENTATION), test_rcl_node_init_with_i
...
   }
+
   // Try normal init but force an internal error on fini.
   {
     ret = rcl_node_init(&node, name, namespace_, &context, &options);
     EXPECT_EQ(RCL_RET_OK, ret);
-    auto mock = mocking_utils::inject_on_return("lib:rcl", rmw_destroy_node, RMW_RET_ERROR);
-    ret = rcl_node_fini(&node);
-    EXPECT_EQ(RCL_RET_ERROR, ret);
+    auto mock = mocking_utils::patch_and_return(
```
brawner, line 519: Does this match the coverage? Seems like rcl_node_fini would exit elsewhere with this new patch?

clalancette: That's a good question.

I suspect that the coverage will remain the same; we weren't previously covering the failure path when `rcl_logging_rosout_fini_publisher_for_node` failed, so we should be swapping one error path for another.  But I'll run a local coverage build to check for sure.

clalancette: I ran coverage before and after this change, and it is as I thought; we are trading coverage on one line for another.  In both cases, just running the unit tests, we are covering 221/233 lines in src/rcl/node.c

clalancette:

brawner (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13134/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8065/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10843/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13128/)

rcl/test/rcl/test_node.cpp
```diff
@@ -511,15 +511,21 @@ TEST_F(CLASSNAME(TestNodeFixture, RMW_IMPLEMENTATION), test_rcl_node_init_with_i
...
+      "lib:rcl", rcl_logging_rosout_fini_publisher_for_node, RCL_RET_ERROR);
```
clalancette, line 520: Unfortunately, this does not work on macOS; according to @hidmic, the problem is that intra-library calls use relative calls, and thus we cannot patch it this way.

Alternately, I also tried using the RCUTILS_FAULT_INJECTION framework.  But in brief try here, I also couldn't get that to work properly.

At this point, I may end up just disabling this test on CentOS.  I actually don't know quite how to do *that* either, but at least we keep the test as-is for the Tier 1 platforms and just disable it there.

hidmic: Hmm, perhaps it's not the cleanest and prettiest solution, but what if we add a macro or a function to check if `inject_on_return()` is supported so we can `GTEST_SKIP()` the entire test if not?

clalancette: > Hmm, perhaps it's not the cleanest and prettiest solution, but what if we add a macro or a function to check if `inject_on_return()` is supported so we can `GTEST_SKIP()` the entire test if not?

I did something slightly more coarse in f7577f5 .  There, I go and detect the Linux distribution, and if it is "centos", I just skip the test.  On everything that is not Linux, or does not have `/etc/os-release`, I just assume that we should compile and run it.  What do you think about that approach?

hidmic: It looks good. What I had in mind was a bit more general, but I don't really know if I can extrapolate what we know about Mimick to any platform.

rcl/test/CMakeLists.txt
```diff
@@ -23,6 +23,27 @@ set(extra_lib_dirs "${rcl_lib_dir}")
...
+  # of which Linux distribution we are on.
+  if(EXISTS "/etc/os-release")
+    file(STRINGS "/etc/os-release" OS_RELEASE)
+    foreach(line ${OS_RELEASE})
+      string(REGEX REPLACE "^(.*)=.*" "\\1" key "${line}")
+      string(TOLOWER "${key}" key)
+      if("${key}" STREQUAL "id")
+        string(REGEX REPLACE "^.*=(.*)" "\\1" DISTRIBUTION "${line}")
+        string(TOLOWER "${DISTRIBUTION}" DISTRIBUTION)
+        message(STATUS "CHRIS: ${DISTRIBUTION}")
```
hidmic, line 40: @clalancette leftover?

rcl/test/CMakeLists.txt
```diff
@@ -23,6 +23,27 @@ set(extra_lib_dirs "${rcl_lib_dir}")
...
+  # Parse those items, looking for a key named "id" (case-insensitive), and
+  # if we find it, set DISTRIBUTION to that value.  That gives us some idea
+  # of which Linux distribution we are on.
+  if(EXISTS "/etc/os-release")
+    file(STRINGS "/etc/os-release" OS_RELEASE)
+    foreach(line ${OS_RELEASE})
+      string(REGEX REPLACE "^(.*)=.*" "\\1" key "${line}")
+      string(TOLOWER "${key}" key)
+      if("${key}" STREQUAL "id")
+        string(REGEX REPLACE "^.*=(.*)" "\\1" DISTRIBUTION "${line}")
```
hidmic, line 38: @clalancette nit: if we do
```suggestion
        string(REGEX REPLACE "^.*=\"(.*)\"" "\\1" DISTRIBUTION "${line}")
```
we can forget about quotes later on.

rcl/test/CMakeLists.txt
```diff
@@ -162,14 +183,16 @@ function(test_target_function)
...
 
-  rcl_add_custom_gtest(test_node${target_suffix}
-    SRCS rcl/test_node.cpp
-    ENV ${rmw_implementation_env_var} ${memory_tools_ld_preload_env_var}
-    APPEND_LIBRARY_DIRS ${extra_lib_dirs}
-    LIBRARIES ${PROJECT_NAME} mimick osrf_testing_tools_cpp::memory_tools
-    AMENT_DEPENDENCIES ${rmw_implementation} "osrf_testing_tools_cpp"
-    TIMEOUT 240  # Large timeout to wait for fault injection tests
-  )
+  if(NOT "${DISTRIBUTION}" STREQUAL "\"centos\"")
```
hidmic, line 186: @clalancette nit: a small note explaining this would be nice.

rcl/test/CMakeLists.txt
clalancette: I wanted my name in LIGHTS! :smiley: 

It's a leftover, I'll remove it.

rcl/test/CMakeLists.txt
clalancette: Sigh.  Unfortunately, on Ubuntu, the quotes are *not* there, so the regex wouldn't match at all.  I guess we could take that as a sign that this is not CentOS, but I think what is there right now is a little more generic (even though we have to worry about the quotes).

rcl/test/CMakeLists.txt
clalancette: Good point, will add a note.

clalancette: Added a note in 111cb41

rcl/test/CMakeLists.txt
clalancette: Removed in 111cb41

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13202/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8133/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10918/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13219/)
* Linux CentOS [Build Status](http://ci.ros2.org/job/ci_linux-centos/90/)

rcl/test/CMakeLists.txt
hidmic: Interesting. Nevermind then.

hidmic (approved):
LGTM ~once CI goes green~ !

clalancette:
The only warning on Windows is coming from Fast-CDR, and is a CMake warning (so unrelated to this).  Given that everything else is green, I'm going to go ahead and merge.  Thanks for the review.

PR #864, opened 2020-11-20
Address various clang static analysis fixes

brawner:
This addresses several clang static analysis fixes, most of which were actual issues. After this PR, rcl_action will build clean with `scan-build`

### DIFF
--- a/rcl_action/test/rcl_action/test_action_server.cpp
+++ b/rcl_action/test/rcl_action/test_action_server.cpp
@@ -727,7 +727,13 @@ TEST_F(TestActionServer, test_action_server_get_goal_status_array)
   EXPECT_NE(status_array.msg.status_list.data, nullptr);
   EXPECT_EQ(status_array.msg.status_list.size, 1u);
   rcl_action_goal_info_t * goal_info_out = &status_array.msg.status_list.data[0].goal_info;
-  EXPECT_TRUE(uuidcmp(goal_info_out->goal_id.uuid, goal_info_in.goal_id.uuid));
+
+  // Clang static analysis has trouble following macros here, so it flags this pointer as
+  // potentially null.
+  const auto * goal_info_out_uuid = goal_info_out->goal_id.uuid;
+  assert(nullptr != goal_info_out_uuid);
+
+  EXPECT_TRUE(uuidcmp(goal_info_out_uuid, goal_info_in.goal_id.uuid));
   ret = rcl_action_goal_status_array_fini(&status_array);
   EXPECT_EQ(ret, RCL_RET_OK);
 
@@ -870,7 +876,12 @@ TEST_F(TestActionServerCancelPolicy, test_action_process_cancel_request_single_g
     ASSERT_EQ(cancel_response.msg.goals_canceling.size, 1u);
     EXPECT_EQ(cancel_response.msg.return_code, action_msgs__srv__CancelGoal_Response__ERROR_NONE);
     rcl_action_goal_info_t * goal_info = &cancel_response.msg.goals_canceling.data[0];
-    EXPECT_TRUE(uuidcmp(goal_info->goal_id.uuid, cancel_request.goal_info.goal_id.uuid));
+
+    // Clang static analysis has trouble following macros here, so it flags this pointer as
+    // potentially null.
+    auto * goal_info_uuid = goal_info->goal_id.uuid;
+    assert(nullptr != goal_info_uuid);
+    EXPECT_TRUE(uuidcmp(goal_info_uuid, cancel_request.goal_info.goal_id.uuid));
     EXPECT_EQ(RCL_RET_OK, rcl_action_cancel_response_fini(&cancel_response));
   }
   {
--- a/rcl_action/src/rcl_action/action_server.c
+++ b/rcl_action/src/rcl_action/action_server.c
@@ -656,12 +656,16 @@ rcl_action_expire_goals(
       }
     }
   }
-  ret_final = _recalculate_expire_timer(
+  rcl_ret_t expire_timer_ret = _recalculate_expire_timer(
     &action_server->impl->expire_timer,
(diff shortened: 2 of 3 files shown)


rcl_action/test/rcl_action/test_action_server.cpp
```diff
@@ -727,7 +727,13 @@ TEST_F(TestActionServer, test_action_server_get_goal_status_array)
   EXPECT_NE(status_array.msg.status_list.data, nullptr);
   EXPECT_EQ(status_array.msg.status_list.size, 1u);
   rcl_action_goal_info_t * goal_info_out = &status_array.msg.status_list.data[0].goal_info;
-  EXPECT_TRUE(uuidcmp(goal_info_out->goal_id.uuid, goal_info_in.goal_id.uuid));
+
+  // Clang static analysis has trouble following macros here, so it flags this pointer as
+  // potentially null.
+  const auto * goal_info_out_uuid = goal_info_out->goal_id.uuid;
+  assert(nullptr != goal_info_out_uuid);
```
fujitatomoya, line 734: Is there any reason that we cannot use `ASSERT_NE(nullptr, goal_info_out_uuid)`?

rcl_action/test/rcl_action/test_action_server.cpp
```diff
@@ -870,7 +876,12 @@ TEST_F(TestActionServerCancelPolicy, test_action_process_cancel_request_single_g
     ASSERT_EQ(cancel_response.msg.goals_canceling.size, 1u);
     EXPECT_EQ(cancel_response.msg.return_code, action_msgs__srv__CancelGoal_Response__ERROR_NONE);
     rcl_action_goal_info_t * goal_info = &cancel_response.msg.goals_canceling.data[0];
-    EXPECT_TRUE(uuidcmp(goal_info->goal_id.uuid, cancel_request.goal_info.goal_id.uuid));
+
+    // Clang static analysis has trouble following macros here, so it flags this pointer as
+    // potentially null.
+    auto * goal_info_uuid = goal_info->goal_id.uuid;
+    assert(nullptr != goal_info_uuid);
```
fujitatomoya, line 883: same here.

rcl_action/test/rcl_action/test_action_server.cpp
clalancette: Yeah, I think either that or:

```
ASSERT(nullptr != goal_info_out_uuid);
```

would be preferred.

brawner: Yah, I wanted to get some more ideas about this. I can't just use `ASSERT_NE` as clang static analysis doesn't follow gtest macros, so it doesn't realize that `ASSERT_NE` will prevent goal_info_out_uuid from being `null` in the code below. I added `ASSERT_NE` on top of `assert` though so `ASSERT_NE` will fail before `assert` does.

It's already impossible for it to be null since it's an array, not a pointer so if you use `ASSERT_NE` directly on `goal_info_out->goal_id.uuid` gtest will warn you that the compiler won't let `uuid` be null.


For reference, this is the warning from scan-build if you don't check it with assert statements.
```
home/brawner/workspace/ros2_master/src/ros2/rcl/rcl_action/test/rcl_action/test_action_server.cpp:886:17: warning: Null pointer passed to 1st parameter expecting 'nonnull'
llptr != goal_info_uuid);
    ~~~~~~~~~~~~^~~~~~~~~
/home/brawner/workspace/ros2_master/src/ros2/rcl/rcl_action/include/rcl_action/types.h:58:37: note: expanded from macro 'fo_uuid'
#define uuidcmp(uuid0, uuid1) (0 == memcmp(uuid0, uuid1, UUID_SIZE))
                                    ^      ~~~~~
/home/brawner/workspace/ros2_master/install/gtest_vendor/src/gtest_vendor/include/gtest/gtest.h:1969:23: note: expanded from macro 'r'
  GTEST_TEST_BOOLEAN_(condition, #condition, false, true, \
  ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/brawner/workspace/ros2_master/install/gtest_vendor/src/gtest_vendor/include/gtest/internal/gtest-internal.h:1325:34: note: expanded from macro 'GTEST_TEST_BOOLEAN_'
      ::testing::AssertionResult(expression)) \
```

clalancette: I think I'm seeing different results than you.  If I take this PR, remove the `assert`, and switch the `ASSERT_NE` to:

```
ASSERT_TRUE(nullptr != goal_info_out_uuid);
```

Then I don't see any warnings from scan-build.  What version of scan-build are you using?

brawner: I couldn't find a way to get it to display the exact version, but llvm 10. I found that passing the array by reference would work fine and would not need confusing null checks, so I updated here and in line 885.

rcl_action/test/rcl_action/test_action_server.cpp
brawner: Changed to passing by reference

brawner:
Testing `--packages-select rcl_action`
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13183/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8114/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10899/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13197/)

brawner:
Windows has cmake warnings due to cmake version, which are being addressed elsewhere.

rcl_action windows test failures also occur on nightlies: https://ci.ros2.org/view/nightly/job/nightly_win_rel/1767/testReport/rcl_action/

clalancette (approved):
Looks good to me!  Thanks for the iterations.

PR #868, opened 2020-11-26
Update build.ros2.org links

brawner:
I think I made this mistake in a couple of places. Checking over things

@nuclearsandwich Without build.ros2.org, can you confirm?


chapulina:
Ouch, judging by the URL patterns of build.ros.org, I think you're right that the repository name is used. For example:

http://build.ros.org/view/Mpr/job/Mpr__gazebo_ros_pkgs__ubuntu_bionic_amd64/lastCompletedBuild/testReport/

jacobperron (approved):

PR #712, opened 2020-07-10
Removed doxygen warnings

ahcorde:
Improved API doc. Removed doxygen warnings

Signed-off-by: ahcorde <ahcorde@gmail.com>

PR #873, opened 2020-12-03
Enable EXTRACT_ALL in Doxyfile to display documentation

ThijsRay:
Currently, the documentation on https://docs.ros2.org/latest/api/rcl/index.html does not display anything besides the contents of the header file. This change was introduced in https://github.com/ros2/rcl/pull/712. This makes it difficult to use the documentation at all, because all information regarding how a function should be used is not displayed.

This change enables the previously removed EXTRACT_ALL to restore the documentation to its intended state.

### DIFF
--- a/rcl/Doxyfile
+++ b/rcl/Doxyfile
@@ -17,6 +17,7 @@ MACRO_EXPANSION        = YES
 EXPAND_ONLY_PREDEF     = YES
 PREDEFINED             += RCL_PUBLIC=
 PREDEFINED             += RCL_WARN_UNUSED=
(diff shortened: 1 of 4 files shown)


clalancette:
I guess the question is why EXTRACT_ALL was removed in the first place, since it seems deliberate.  @ahcorde @Blast545 (maybe @chapulina) can you comment?

clalancette:
After looking at this, I don't think this is the way we want to go.  Instead we want to do something more like #896, where we explicitly enable documentation of the files as necessary.  Because of that, I'm going to close this out.

Issue #878, opened 2020-12-18
Potential null ptr passed to strcmp in remap.c


nnmm:
## Bug report

**Required Info:**

- Operating System: Ubuntu 20.04
- Clang-tidy version: 10.0.0
- Version or commit hash: 30464d12e73d6350b641f73e665d21a5cfa99651 

#### Steps to reproduce issue
```
# Produce compilation database while building ROS 2 rolling
colcon build --symlink-install --merge-install --cmake-args -DCMAKE_EXPORT_COMPILE_COMMANDS=1
clang-tidy -p build/compile_commands.json src/ros2/rcl/rcl/src/rcl/*.c
```

#### Expected behavior
The arguments to `strcmp` are never null.

#### Actual behavior
```
~/ros2_rolling/src/ros2/rcl/rcl/src/rcl/remap.c:145:23: warning: Null pointer passed to 2nd parameter expecting 'nonnull' [clang-analyzer-core.NonNullParamChecker]
      matched = (0 == strcmp(expanded_match, name));
                      ^
~/ros2_rolling/src/ros2/rcl/rcl/src/rcl/remap.c:316:3: note: Assuming the condition is false
  RCUTILS_CAN_SET_MSG_AND_RETURN_WITH_ERROR_OF(RCL_RET_INVALID_ARGUMENT);
  ^

... many more notes to the effect that none of the early returns are taken ...

~/ros2_rolling/src/ros2/rcl/rcl/src/rcl/remap.c:324:61: note: Passing null pointer value via 4th parameter 'name'
    local_arguments, global_arguments, RCL_NAMESPACE_REMAP, NULL, node_name, NULL, NULL,
                                                            ^
/usr/lib/llvm-10/lib/clang/10.0.0/include/stddef.h:89:16: note: expanded from macro 'NULL'
#  define NULL ((void*)0)
               ^
~/ros2_rolling/src/ros2/rcl/rcl/src/rcl/remap.c:323:10: note: Calling 'rcl_remap_name'
  return rcl_remap_name(
         ^
...

~/ros2_rolling/src/ros2/rcl/rcl/src/rcl/remap.c:193:7: note: Passing null pointer value via 4th parameter 'name'
      name, node_name, node_namespace, substitutions, allocator, &rule);
      ^
~/ros2_rolling/src/ros2/rcl/rcl/src/rcl/remap.c:191:21: note: Calling 'rcl_remap_first_match'
    rcl_ret_t ret = rcl_remap_first_match(
                    ^

[...]

PR #879, opened 2020-12-18
Add NULL check in remap.c

nnmm:
Followup to https://github.com/ros2/rcl/issues/878

### DIFF
--- a/rcl/src/rcl/remap.c
+++ b/rcl/src/rcl/remap.c
@@ -142,9 +142,7 @@ rcl_remap_first_match(
         }
         continue;
       }
-      if (NULL != name) {
-        matched = (0 == strcmp(expanded_match, name));
-      }
+      matched = (0 == strcmp(expanded_match, name));
       allocator.deallocate(expanded_match, allocator.state);
     } else {
       // nodename and namespace replacement apply if the type and node name prefix checks passed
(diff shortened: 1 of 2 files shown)


nnmm:
I took a second look – calling `rcl_remap_name` with a nullptr in `name` and the `type_bitmask` equal to either `RCL_TOPIC_REMAP` or `RCL_SERVICE_REMAP` is incorrect, right? So that function should return an error instead when it is called like that, which would hopefully also satisfy clang-tidy.

clalancette (changes requested):
Ah, this is a much better fix.  You are right, this is a precondition that should hold, so I like this better.

Could you also add a test that violates this precondition?  Thanks.

nnmm:
Sorry, unfortunately clang-tidy doesn't get it when I add the check to `rcl_remap_name`. But that function is private, so you were right – it can never be called with the invalid combination. I'll do what you initially suggested and also add an explanatory comment. And a test.

Should I add a test that   EXPECT_EQ(RCL_RET_INVALID_ARGUMENT, rcl_remap_copy(&remap_empty, &remap_dst));

clalancette (approved):
Looks good to me, so I'll approve and run CI.  @ros2/team I'd appreciate another review, since I essentially suggested the code here.

clalancette:
CI (build: `--packages-up-to rcl`, test: `--packages-select rcl`; this seems safe since we are adding a check that shouldn't trigger):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13295/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8228/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11028/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13336/)

sloretz (approved):

clalancette:
Thanks for the additional approval!  Merging.

PR #882, opened 2021-01-06
make rcl_lifecycle_com_interface optional in lifecycle nodes

Karsten1987:
connects to ros2/rclcpp#1506

### DIFF
--- a/rcl_lifecycle/include/rcl_lifecycle/rcl_lifecycle.h
+++ b/rcl_lifecycle/include/rcl_lifecycle/rcl_lifecycle.h
@@ -194,6 +194,11 @@ rcl_lifecycle_transition_fini(
   rcl_lifecycle_transition_t * transition,
   const rcl_allocator_t * allocator);
 
+/// Return a default initialized state machine options struct.
+RCL_LIFECYCLE_PUBLIC
+rcl_lifecycle_state_machine_options_t
+rcl_lifecycle_get_default_state_machine_options();
+
 /// Return a rcl_lifecycle_state_machine_t struct with members set to `NULL` or 0.
 /**
  * Should be called to get a null rcl_lifecycle_state_machine_t before passing to
@@ -228,11 +233,7 @@ rcl_lifecycle_get_zero_initialized_state_machine();
  *    available transitions
  * \param[in] ts_srv_get_transition_graph pointer to the service that allows to get transitions from
  *    the graph
- * \param[in] enable_com_interface if `false` the services listed above are not initialized,
- *    if true all services are available and active
- * \param[in] default_states if `true` a new default state machine is initialized, otherwise
- *    the state_machine pointer is only used to initialize the interfaces
- * \param[in] allocator a valid allocator used to initialized the state machine
+ * \param[in] state_machine_optios collection of config options for initializing the state machine
  * \return `RCL_RET_OK` if the state machine was initialized successfully, or
  * \return `RCL_RET_INVALID_ARGUMENT` if input params is NULL, or
  * \return `RCL_RET_ERROR` if an unspecified error occurs.
@@ -249,9 +250,7 @@ rcl_lifecycle_state_machine_init(
   const rosidl_service_type_support_t * ts_srv_get_available_states,
   const rosidl_service_type_support_t * ts_srv_get_available_transitions,
   const rosidl_service_type_support_t * ts_srv_get_transition_graph,
-  bool enable_com_interface,
-  bool default_states,
-  const rcl_allocator_t * allocator);
+  const rcl_lifecycle_state_machine_options_t * state_machine_options);
 
 /// Finalize a rcl_lifecycle_state_machine_t.
 /**
--- a/rcl_lifecycle/include/rcl_lifecycle/rcl_lifecycle.h
+++ b/rcl_lifecycle/include/rcl_lifecycle/rcl_lifecycle.h
@@ -228,6 +228,8 @@ rcl_lifecycle_get_zero_initialized_state_machine();
  *    available transitions
  * \param[in] ts_srv_get_transition_graph pointer to the service that allows to get transitions from
  *    the graph
+ * \param[in] enable_com_interface if `false` the services listed above are not initialized,
+ *    if true all services are available and active
  * \param[in] default_states if `true` a new default state machine is initialized, otherwise
  *    the state_machine pointer is only used to initialize the interfaces
  * \param[in] allocator a valid allocator used to initialized the state machine
@@ -247,6 +249,7 @@ rcl_lifecycle_state_machine_init(
   const rosidl_service_type_support_t * ts_srv_get_available_states,
   const rosidl_service_type_support_t * ts_srv_get_available_transitions,
   const rosidl_service_type_support_t * ts_srv_get_transition_graph,
+  bool enable_com_interface,
   bool default_states,
   const rcl_allocator_t * allocator);
 
--- a/rcl_lifecycle/include/rcl_lifecycle/rcl_lifecycle.h
+++ b/rcl_lifecycle/include/rcl_lifecycle/rcl_lifecycle.h
@@ -268,7 +268,6 @@ rcl_lifecycle_state_machine_init(
  *
  * \param[inout] state_machine struct to be finalized
  * \param[in] node_handle valid (not finalized) handle to the node
- * \param[in] allocator a valid allocator used to finalize the state machine
  * \return `RCL_RET_OK` if the state was finalized successfully, or
  * \return `RCL_RET_INVALID_ARGUMENT` if any arguments are invalid, or
  * \return `RCL_RET_ERROR` if an unspecified error occurs.
@@ -278,8 +277,7 @@ RCL_WARN_UNUSED
 rcl_ret_t
 rcl_lifecycle_state_machine_fini(
   rcl_lifecycle_state_machine_t * state_machine,
-  rcl_node_t * node_handle,
-  const rcl_allocator_t * allocator);
+  rcl_node_t * node_handle);
 
 /// Check if a state machine is active.
 /**
--- a/rcl_lifecycle/include/rcl_lifecycle/rcl_lifecycle.h
+++ b/rcl_lifecycle/include/rcl_lifecycle/rcl_lifecycle.h
@@ -233,7 +233,7 @@ rcl_lifecycle_get_zero_initialized_state_machine();
  *    available transitions
  * \param[in] ts_srv_get_transition_graph pointer to the service that allows to get transitions from
  *    the graph
- * \param[in] state_machine_optios collection of config options for initializing the state machine
+ * \param[in] state_machine_options collection of config options for initializing the state machine
  * \return `RCL_RET_OK` if the state machine was initialized successfully, or
  * \return `RCL_RET_INVALID_ARGUMENT` if input params is NULL, or
  * \return `RCL_RET_ERROR` if an unspecified error occurs.
--- a/rcl_lifecycle/src/rcl_lifecycle.c
+++ b/rcl_lifecycle/src/rcl_lifecycle.c
@@ -237,8 +237,7 @@ rcl_lifecycle_state_machine_init(
     if (ret != RCL_RET_OK) {
       // init default state machine might have allocated memory,
       // so we have to call fini
-      ret = rcl_lifecycle_state_machine_fini(
-        state_machine, node_handle, &state_machine->options.allocator);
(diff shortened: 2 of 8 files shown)


rcl_lifecycle/include/rcl_lifecycle/rcl_lifecycle.h
```diff
@@ -228,6 +228,8 @@ rcl_lifecycle_get_zero_initialized_state_machine();
  *    available transitions
  * \param[in] ts_srv_get_transition_graph pointer to the service that allows to get transitions from
  *    the graph
+ * \param[in] enable_com_interface if `false` the services listed above are not initialized,
```
fujitatomoya, line 231: this tells me publication will be available on topic name `/<node_name>/transition_event`, but actually this publication will not be initialized with `rcl_lifecycle_com_interface_init` if enable_com_interface is false.

rcl_lifecycle/test/test_rcl_lifecycle.cpp
```diff
@@ -239,72 +238,82 @@ TEST(TestRclLifecycle, state_machine) {
...
   rcutils_reset_error();
   EXPECT_EQ(RCL_RET_INVALID_ARGUMENT, rcl_lifecycle_state_machine_is_initialized(&state_machine));
   rcutils_reset_error();
 
+  // Com interface not enabled
+  ret = rcl_lifecycle_state_machine_init(
+    &state_machine, &node, pn, cs, gs, gas, gat, gtg, false, false, &allocator);
+  EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  EXPECT_NE(nullptr, &state_machine.com_interface);
+  EXPECT_EQ(nullptr, state_machine.com_interface.srv_change_state.impl);
```
fujitatomoya, line 308: it would be nice to check `pub_transition_event.impl` is also NULL.

Karsten1987:
@fujitatomoya I've opted for introducing two more sets of `init/fini`, namely `..publisher_init`, `..publisher_fini`, `..services_init`, `..services_fini`. I've incorporated these appropriately in the `goto fail:` sections in order to deduplicate code.

As discussed in the rclcpp ticket, that allows us to explicitly initialize the transition event publisher without initializing the services. Let me know what you think.

fujitatomoya:
> As discussed in the rclcpp ticket, that allows us to explicitly initialize the transition event publisher without initializing the services.

it looks okay. i am okay to go with this approach.

but I'd like to get back to discussion if `lifecycle_com_interface` includes `/<node_name>/transition_event`. as far as i see from the current implementation, it does. in that case, publication should be disabled along with services? this can be done, disabling all communication path, and we can check `state_machine_` see if publisher is initialized and then update `publish_update`.

CC: @ivanpauno @clalancette @wjwwood

Karsten1987:
My/Our use case here is actually pretty straight forward. We'd like to disable external access to state changes. That is, we'd like to block all incoming communication which modifies the state machine. That however does not prevent any outgoing traffic, such as a publication on the transition event topics. In fact, I believe it should stay active in order to let other nodes in the network know about the triggered transitions.

Now, granted, with this definition we could also split the services further, where each individual service could be activated or not. That is, we only disable the `change_state` service, but leave the `get_state` et. al. services alive.

clalancette:
@fujitatomoya Since you had some initial comments, I assigned this one to you to take a look at.  If you don't have time for this, please feel free to unassign yourself.

fujitatomoya:
@Karsten1987 

overall, i am good to go with this. there are three `unused_warning` https://build.ros2.org/job/Rpr__rcl__ubuntu_focal_amd64/471/gcc/new/, probably we could change where to log the error messages? what do you think?

Karsten1987:
@fujitatomoya I think I've addressed the unused warnings. I tried to mention it in the line comments, however I do believe we don't have to log twice and can ignore the return value of the `fini` calls in the `fail` section. The called function itself logs the reason of failure.

I did however have to add another flag to the `com_interface` for bookkeeping whether the com interface was initialized or not. That was necessary to make the call to `is_initialized` successful in both cases.

I'll link CI to the overall issue ticket once more: https://github.com/ros2/rclcpp/issues/1506#issuecomment-788537922

fujitatomoya (approved):
looks good to me.

fujitatomoya:
@clalancette just in case, could you also take a look when you got time?

rcl_lifecycle/include/rcl_lifecycle/rcl_lifecycle.h
```diff
@@ -247,6 +249,7 @@ rcl_lifecycle_state_machine_init(
   const rosidl_service_type_support_t * ts_srv_get_available_states,
   const rosidl_service_type_support_t * ts_srv_get_available_transitions,
   const rosidl_service_type_support_t * ts_srv_get_transition_graph,
+  bool enable_com_interface,
```
clalancette, line 252: So I guess my only real concern here is this change, which is an API break.  Looking at the development history of rcl/rcl_lifecycle over the last year, there are only 2 other API breaks that I see:

* #830, which switched an unsigned int -> uint8_t (so while technically an API break, unlikely to be noticed by most users)
* #676 , which moved the logging header file from `rcl` into `rcl_logging_interface` (so any users of the API would have to include it from somewhere else.  Its unlikely to be a problem anyway because it is really an internal API)

So this would be the first real API break we have.  We could avoid it by leaving this API as-is, marking it deprecated, and adding in `rcl_lifecycle_state_machine_init2`, or whatever.

@jacobperron @wjwwood @ivanpauno How do you feel about an API break here?

Karsten1987: What is wrong with an API change here? It is a new ROS distro and should be eligible for an API break in my mind. It's a bit unfortunate that we don't have function overloading in C, but renaming it to `..init2` seems pretty complicated to me as people then have to change it back to the original signature for H-Turtle.

ivanpauno: The churn of breaking API is that non-official client libraries have to catch up, e.g. rcljava.
But I haven't seem the lifecycle node interface wrapped in other client library that's not rclcpp.

This would also be a compile error, which is easy to detect and correct.

IMHO, this API break is reasonable.

---

We could also define a `rcl_lifecycle_state_machine_options_t`, so code will look like:

```c
rcl_lifecycle_state_machine_options_t options = rcl_lifecycle_get_default_state_machine_options();
rcl_ret_t ret = rcl_lifecycle_state_machine_init(
  state_machine,
  node_handle,
  <lots of typesupports>,
  options);
  ```

and next time a new option is added you only break ABI.
If a different client library doesn't care about exposing the new option they don't have to modify code, only re-compile.

Comment (Karsten1987) [reply rcl_lifecycle/include/rcl_lifecycle/rcl_lifecycle.h:252]: I like the idea of having a dedicated options struct. We could technically even turn this up a notch by introducing a `com_interface_options` struct which entails all the typesupports as well. That way, the function signature doesn't have millions of arguments ;-)

Comment (clalancette) [reply rcl_lifecycle/include/rcl_lifecycle/rcl_lifecycle.h:252]: > What is wrong with an API change here?

If this was a breaking change in rclcpp, we would most likely do the deprecation and adding of a new API.  We do try to do the tick-tock method of deprecation.

> The churn of breaking API is that non-official client libraries have to catch up, e.g. rcljava.

But given that this is in rcl (so isn't a user-facing API), we can more easily get away with an API break.  That being said, I know for sure that we will break rclnodejs: https://github.com/RobotWebTools/rclnodejs/blob/8490c2f8fa26019b8b284bcb17dc2001363d9195/src/rcl_lifecycle_bindings.cpp#L78

Still, if the maintainers of rcl think that is OK, it is OK with me.  If we go ahead with the API break, I will request a release note at https://docs.ros.org/en/rolling/Releases/Release-Galactic-Geochelone.html

Comment (wjwwood) [reply rcl_lifecycle/include/rcl_lifecycle/rcl_lifecycle.h:252]: Breaking API has to happen sometimes, and in C it's less friendly, unfortunately, but I don't think we should ever avoid a change because it breaks API (assuming there's isn't a better way to do it). Without C++'s overloading, doing a deprecation cycle is annoying and not worth it much of the time. We have the versioning scheme that allows downstream stuff to `#ifdef` their way around it, e.g.:

https://github.com/ros2/rmw/blob/ace271650ad39a656d81239050c9639c4190f077/rmw/CMakeLists.txt#L81

I don't think we're doing that for `rcl` or `rcl_lifecycle` but we could.

Comment (ivanpauno) [reply rcl_lifecycle/include/rcl_lifecycle/rcl_lifecycle.h:252]: >Generic rcl_lifecycle_state_machine_options_t struct with a single parameter "enable_com_interface"

I was thinking in including the `default_states` and `allocator` arguments in the struct too, which have reasonable defaults.

```c
typedef struct rcl_lifecycle_state_machine_options_t {
  bool enable_com_interface;
  bool default_states;
  rcl_allocator_t allocator;
} rcl_lifecycle_state_machine_options_t;
```

> or more specific rcl_lifecycle_com_interface_options_t which has the parameter as well as the typesupports?

that also makes sense to me, the `rcl_lifecycle_state_machine_options_t` struct could also have the typesupports and `rcl_lifecycle_get_default_state_machine_options()` could provide the default ones (like [here](https://github.com/ros2/rcl/blob/5ef4b2dfca54c1a5b2e8f70be3f414d33e529975/rcl_lifecycle/test/test_rcl_lifecycle.cpp#L227-L238), `rcl` already depends on `rcl_interfaces`).

`rcl_lifecycle_state_machine_options_t` sounds better to me so you can include no-com related options there, though having them grouped separetly isn't a bad idea either ... I don't mind strongly.

jacobperron: +1 for breaking API and switching to an options struct.

Karsten1987: So this turns out to become a bigger change than anticipated. 

While looking at the allocator, I think it's generally good practice to store the allocator used for `..._init` locally and reuse that allocator instance within the `..._fini`. However, the lifecycle state machine seems to receive an external allocator in its `fini`.
I'll address that accordingly as I won't be able to cleanly separate this PR from a potential follow-up PR which addresses the allocator.

wjwwood: +1 for not taking an allocator in the fini.

Karsten1987:
This is good for another review. CI can be found on the top level issue: https://github.com/ros2/rclcpp/issues/1506#issuecomment-804440165

rcl_lifecycle/include/rcl_lifecycle/rcl_lifecycle.h
```diff
@@ -228,9 +233,7 @@ rcl_lifecycle_get_zero_initialized_state_machine();
  *    available transitions
  * \param[in] ts_srv_get_transition_graph pointer to the service that allows to get transitions from
  *    the graph
- * \param[in] default_states if `true` a new default state machine is initialized, otherwise
- *    the state_machine pointer is only used to initialize the interfaces
- * \param[in] allocator a valid allocator used to initialized the state machine
+ * \param[in] state_machine_optios collection of config options for initializing the state machine
```
clalancette, line 236:
```suggestion
 * \param[in] state_machine_options collection of config options for initializing the state machine
```

clalancette (approved):
One nit inline, but otherwise looks pretty good to me.

rcl_lifecycle/src/rcl_lifecycle.c
```diff
@@ -234,11 +257,12 @@ rcl_lifecycle_state_machine_init(
...
   rcl_lifecycle_state_machine_t * state_machine,
-  rcl_node_t * node_handle,
-  const rcl_allocator_t * allocator)
+  rcl_node_t * node_handle)
 {
   RCL_CHECK_ALLOCATOR_WITH_MSG(
-    allocator, "can't free state machine, no allocator given\n", return RCL_RET_INVALID_ARGUMENT);
+    &state_machine->options.allocator,
+    "Can't cleanup state machine. No allocator associated with the state machine.\n",
+    return RCL_RET_ERROR);
```
ivanpauno, line 265: checking the allocator doesn't seem necessary here, as the `init()` function already did that

ivanpauno (approved):
lgtm

Karsten1987:
CI is green (https://github.com/ros2/rclcpp/issues/1506#issuecomment-805036539), merging.
Thanks for the review.

PR #883, opened 2021-01-07
Revert "Improve trigger test for graph guard condition (#811)"

clalancette:
This reverts commit a6cacef15839deb249433f3bc0dfa621ac60095d.

PR #811 seems to be causing a regression on the buildfarm; https://ci.ros2.org/view/nightly/job/nightly_linux_debug/1802/testReport/junit/projectroot/test/test_graph__rmw_connext_cpp/ is one example of it.  I'm not quite sure why it is failing, but this reverts that commit and fixes the test (at least, it fixed it for me locally).

### DIFF
--- a/rcl/test/rcl/test_graph.cpp
+++ b/rcl/test/rcl/test_graph.cpp
@@ -1223,175 +1223,75 @@ TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_graph_query_functio
     9);  // number of retries
 }
 
-/* Test the graph guard condition notices below changes.
- * publisher create/destroy, subscription create/destroy
- * service create/destroy, client create/destroy
- * Other node added/removed
+/* Test the graph guard condition notices topic changes.
  *
  * Note: this test could be impacted by other communications on the same ROS Domain.
  */
-TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_graph_guard_condition_trigger_check) {
-#define CHECK_GUARD_CONDITION_CHANGE(EXPECTED_RESULT)   do { \
-    ret = rcl_wait_set_clear(&wait_set); \
-    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str; \
-    ret = rcl_wait_set_add_guard_condition(&wait_set, graph_guard_condition, NULL); \
-    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str; \
-    ret = rcl_wait(&wait_set, time_to_sleep.count()); \
-    ASSERT_EQ(EXPECTED_RESULT, ret) << rcl_get_error_string().str; \
-} while (0)
-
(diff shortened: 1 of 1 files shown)


clalancette:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13368/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8295/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11097/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13421/)

ivanpauno (approved):

hidmic:
Shall we attempt to figure this out? It fails differently because the test itself is different. The former test is not checking for graph updates on local node creation. The test is leaking entities when an assertion fails too. FYI @Barry-Xu-2018.

clalancette:
> Shall we attempt to figure this out? It fails differently because the test itself is different. The former test is not checking for graph updates on local node creation. The test is leaking entities when an assertion fails too. FYI @Barry-Xu-2018.

I agree that there is a problem with the old code (which I'm restoring here).  But this is currently causing all builds to be yellow, which affects all other PRs.  My suggestion is that we put this revert PR in, and then we can figure out the real problem afterwards.

Barry-Xu-2018:
@clalancette @hidmic @ivanpauno 

I will investigate this and fix it.

Barry-Xu-2018:
@clalancette @hidmic @ivanpauno

After checking, the cause is  
After a new node is created, timeout occurs while waiting the change of graph guard condition.  

Currently, timeout is set to 400ms. This value is no problem for fastdds and cyclonedds. But for connext, it spends about 1s in my test environment.  

We should set a suitable timeout for different test environments(e.g. arm platform).    
I update codes to set default timeout to 1s. And for new node creation case, the timeout is set to 3s.   
What do you think ?   
Updated codes like https://github.com/Barry-Xu-2018/rcl/commit/767a61b7f37d5ee2cd069bab109c74d09f7217e1   

@hidmic 
> The former test is not checking for graph updates on local node creation. The test is leaking entities when an assertion fails too.

Sorry. I not clear your concern. Could you explain more about your concern ? 
Before executing tests, below codes make sure existing graph change doesn't affect following graph tests. 
https://github.com/Barry-Xu-2018/rcl/blob/767a61b7f37d5ee2cd069bab109c74d09f7217e1/rcl/test/rcl/test_graph.cpp#L1259-L1275

clalancette:
> We should set a suitable timeout for different test environments(e.g. arm platform).
> I update codes to set default timeout to 1s. And for new node creation case, the timeout is set to 3s.
> What do you think ?
> Updated codes like [Barry-Xu-2018@767a61b](https://github.com/Barry-Xu-2018/rcl/commit/767a61b7f37d5ee2cd069bab109c74d09f7217e1)

That sounds reasonable to me; we've noticed that Connext is much slower in the past as well.  Please feel free to open a new pull request with your new code, and we'll do a full round of CI on it.  Thanks for looking into it!

Barry-Xu-2018:
@clalancette @ivanpauno 

Sorry for my late response.  
I created #884 and please help to do the full round of CI on this commit and review it.

PR #891, opened 2021-02-18
Change destruction behaviour

jparisu:
Change RCL Entities destruction behavior in failure.

RCL Entities do not stop the destruction when a failure occurs (it could happen because they destroy lower entities during the process).
With this approach the destruction will stop before deallocate the entity memory and return an error in case the lower entity (rmw) fails in destruction.

NOTE: This is a first approach to show a possible design.

Signed-off-by: jparisu <javierparis@eprosima.com>

### DIFF
--- a/rcl/src/rcl/publisher.c
+++ b/rcl/src/rcl/publisher.c
@@ -183,7 +183,7 @@ rcl_publisher_fini(rcl_publisher_t * publisher, rcl_node_t * node)
       rmw_destroy_publisher(rmw_node, publisher->impl->rmw_handle);
     if (ret != RMW_RET_OK) {
       RCL_SET_ERROR_MSG(rmw_get_error_string().str);
-      result = RCL_RET_ERROR;
+      return RCL_RET_ERROR;
     }
     allocator.deallocate(publisher->impl, allocator.state);
     publisher->impl = NULL;
--- a/rcl/test/rcl/test_publisher.cpp
+++ b/rcl/test/rcl/test_publisher.cpp
@@ -820,8 +820,13 @@ TEST_F(CLASSNAME(TestPublisherFixture, RMW_IMPLEMENTATION), test_mock_publisher_
     &publisher, this->node_ptr, ts, topic_name, &publisher_options);
   ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
 
-  // Internal rmw failure destroying publisher
-  auto mock = mocking_utils::patch_and_return("lib:rcl", rmw_destroy_publisher, RMW_RET_ERROR);
+  {
+    // Internal rmw failure destroying publisher
+    auto mock = mocking_utils::patch_and_return("lib:rcl", rmw_destroy_publisher, RMW_RET_ERROR);
+    ret = rcl_publisher_fini(&publisher, this->node_ptr);
+    EXPECT_EQ(RCL_RET_ERROR, ret) << rcl_get_error_string().str;
+  }
+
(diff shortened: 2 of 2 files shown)


rcl/src/rcl/publisher.c
```diff
@@ -183,7 +183,7 @@ rcl_publisher_fini(rcl_publisher_t * publisher, rcl_node_t * node)
       rmw_destroy_publisher(rmw_node, publisher->impl->rmw_handle);
     if (ret != RMW_RET_OK) {
       RCL_SET_ERROR_MSG(rmw_get_error_string().str);
-      result = RCL_RET_ERROR;
+      return RCL_RET_ERROR;
```
fujitatomoya, line 186: i understand this works in the rcl test. what i was thinking is user application with [rclcpp](https://github.com/ros2/rclcpp) (and i think [rclpy](https://github.com/ros2/rclpy) is the same). with this fix, user application needs to be aware of failure and reason, and then retry. unless, it will be more memory leaks cz `publisher->impl` will never be deallocated.

https://github.com/ros2/rclcpp/blob/ab743392e4daf546eb0de704936b5f14ef6e0417/rclcpp/src/rclcpp/publisher_base.cpp#L50-L63

if i am not mistaken, you are suggesting we should change the rclcpp interfaces accordingly? i am not sure about that, i'd like to hear from others.

jparisu: It is true that this is most likely a design choice than an implementation one.
I had an error in that specific test case in `rcl` and that is why I tried to find a solution, but there is not a good one after all. 

In my opinion the correct behaviour should be to not deallocate anything that has subentities, and this is the pattern that DDS specification follows.

But it is understood that ROS2 wanted to follow other way, and it is coherent within different layers.

So yes, to your question, I suggest to change every interface higher or lower (rclcpp, rmw).
But probably it is not realistic such change in a complex environment right now.

I will close the issue. And I would like to be communicated if any choice regarding this path is taken or if there are any other opinions and suggestions.

fujitatomoya: > In my opinion the correct behaviour should be to not deallocate anything that has subentities, and this is the pattern that DDS specification follows.

i cannot disagree on this. probably we can hear from others for their opinion 😄

jparisu: It is true that this is most likely a design choice than an implementation one.
I had an error in that specific test case in `rcl` and that is why I tried to find a solution, but there is not a good one after all. 

In my opinion the correct behaviour should be to not deallocate anything that has subentities, and this is the pattern that DDS specification follows.

But it is understood that ROS2 wanted to follow other way, and it is coherent within different layers.

So yes, to your question, I suggest to change every interface higher or lower (rclcpp, rmw).
But probably it is not realistic such change in a complex environment right now.

I will close the issue. And I would like to be communicated if any choice regarding this path is taken or if there are any other opinions and suggestions.

jacobperron: I don't know the history as to why we deal with destruction this way, but it seems reasonable as we're trying to cleanup as best we can.

Perhaps it makes sense to do something similar to [`rmw_destroy_publisher`](https://github.com/ros2/rmw/blob/ace271650ad39a656d81239050c9639c4190f077/rmw/include/rmw/rmw.h#L351-L352): do not change state if the return value is `RMW_RET_INVALID_ARGUMENT` or `RMW_RET_INCORRECT_RMW_IMPLEMENTATION`. Still, I'm not sure how we would handle this in rclcpp or rclpy.

PR #896, opened 2021-03-08
Cleanup documentation for doxygen.

clalancette:
In particular:

1. Add in @file on all of the files we want indexed (otherwise
   Doxygen just ignores them).
2. Make sure to document all things that are missing (as pointed
   out by Doxygen).
3. Switch to #ENUM to reference enums (this makes Doxygen put
   links in properly).
4. Switch from `my_func` or `my_func()` to my_func() everywhere
   (Doxygen supports all 3 for making links, but for consistency
   we just stick with one style).
5. Hide a few things from Doxygen.
6. Switch all \returns to \return (Doxygen supports both, but
   we should be consistent).


rcl/include/rcl/remap.h
```diff
@@ -53,12 +55,12 @@ rcl_get_zero_initialized_remap(void);
...
  * return RCL_RET_INVALID_ARGUMENT.
  *
- * `global_arguments` is usually the arguments passed to `rcl_init()`.
+ * `global_arguments` is usually the arguments passed to rcl_init().
  * \sa rcl_init()
  * \sa rcl_get_global_arguments()
  *
  * Remap rules are checked in the order they were given.
- * For rules passed to `rcl_init` this usually is the order they were passed on the command line.
+ * For rules passed to rcl_ini()` this usually is the order they were passed on the command line.
```
ivanpauno, line 63:
```suggestion
 * For rules passed to rcl_init()` this usually is the order they were passed on the command line.
```

?

rcl/include/rcl/service.h
```diff
@@ -252,7 +256,22 @@ rcl_take_request_with_info(
   rmw_service_info_t * request_header,
   void * ros_request);
 
-/// backwards compatibility version that takes a request_id only
+/// Backwards compatibility function to ake a pending ROS request using a rcl service.
```
ivanpauno, line 259:
```suggestion
/// Backwards compatibility function to take a pending ROS request using a rcl service.
```

ivanpauno: is this function only used for backwards compatibility?
It seems to me that we're actually using it ...

ivanpauno (approved):
except for a few typos, lgtm!

rcl/include/rcl/allocator.h
```diff
@@ -22,15 +24,35 @@ extern "C"
 
 #include "rcutils/allocator.h"
 
+/// Encapsulation of an allocator.
+/**
+ * \sa rcutils_allocator_t
+ */
 typedef rcutils_allocator_t rcl_allocator_t;
 
+/// Return a properly initialized rcutils_allocator_t with default values.
```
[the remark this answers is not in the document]
clalancette: Fixed in 2206ac8

rcl/include/rcl/remap.h
clalancette: Fixed in 2206ac8

rcl/include/rcl/service.h
clalancette: Fixed the typo in 2206ac8 .

wjwwood:
> Add in @file on all of the files we want indexed (otherwise Doxygen just ignores them).

I'm curious why this is the case? Is it because it's C (`.c`)?

Also, should this have not been `\file`?

clalancette:
> I'm curious why this is the case? Is it because it's C (`.c`)?

As far as I know, Doxygen only looks at files that are marked with `@file` for generating documentation (unless you have EXTRACT_ALL set to YES).  So that's why I added it to the files we wanted documented.

As far the `@` vs. `\`; yeah, in retrospect, it should have been `\` to be more consistent.

wjwwood:
Should we be using `EXTRACT_ALL`? Just seems odd to add this to every file. It's not that important, I just want to understand before we roll it out to every package with doxygen docs.

clalancette:
> Should we be using `EXTRACT_ALL`?

The problem I ran into with EXTRACT_ALL is that it goes and sweeps everything, and Doxygen can get upset about parsing certain things.  It is possible to use EXTRACT_ALL in combination with EXCLUDE (and/or PREPROCESSOR) to make it work, but I feel like using `@file` to use a whitelist is more straightforward.

PR #907, opened 2021-04-01
Add functions for waiting for publishers and subscribers

jacobperron:
* Add two new functions for waiting for a count of publishers (or subscribers) b1dbca9
* Refactor existing tests to use these new functions 466ee6e

---

I plan to add equivalent rclcpp and rclpy APIs.

I'd like to add similar API for services, perhaps in a separate PR.


rcl/include/rcl/graph.h
```diff
@@ -581,6 +582,95 @@ rcl_count_subscribers(
...
+/**
+ * The `node` parameter must point to a valid node.
+ *
+ * The `allocator` parameter must point to a valid allocator.
+ *
+ * The `topic_name` parameter must not be `NULL`, and must not be an empty string.
+ * It should also follow the topic name rules.
+ *
+ * This function blocks and will return when the number of publishers for `topic_name`
+ * is greater than or equal to the `count` parameter, or the specified `timeout` is reached.
```
wjwwood, line 595: I don't know where, but it might be worth mentioning here, that the node's graph guard condition cannot be used concurrently by any other wait sets, as there is only one and guard conditions can be used with at most one wait set at a time.

rcl/include/rcl/graph.h
```diff
@@ -581,6 +582,95 @@ rcl_count_subscribers(
...
+ *
+ * The `timeout` parameter is in nanoseconds.
+ * A negative value disables the timeout (i.e. this function blocks until the number of
+ * publishers is greater than or equals to `count`).
```
wjwwood, line 599: might want to mention the timeout is based on steady time elapsed, rather than system time.

rcl/src/rcl/graph.c
```diff
@@ -452,6 +456,158 @@ rcl_count_subscribers(
...
+  // Add it to the wait set
+  ret = rcl_wait_set_add_guard_condition(&wait_set, guard_condition, NULL);
+  if (ret != RCL_RET_OK) {
+    // Error message already set
+    return ret;
+  }
+
+  // Get current time
+  rcutils_time_point_value_t start;
+  rcutils_ret_t time_ret = rcutils_steady_time_now(&start);
```
wjwwood, line 519: I think a kind of meta issue we have to resolve in rcl (not here, and it has come up elsewhere), is that we don't know really what kind of clock the wait set uses to figure its timeout. I think it says system atm, and some do that and other don't? I don't remember off the top of my head and don't have time to dig atm. Anyway I mention that because here we're using steady time and the wait might be using system time, which may give some odd results.

But since this isn't an issue unless you're majorly shifting the date while running this function, it seems low risk atm.

rcl/src/rcl/graph.c
```diff
@@ -452,6 +456,158 @@ rcl_count_subscribers(
...
+
+    // Check count
+    count_ret = count_entities_func(node, topic_name, &count);
+    if (count_ret != RCL_RET_OK) {
+      // Error already set
+      return count_ret;
+    }
+    if (expected_count <= count) {
+      *success = true;
+      break;
```
wjwwood, line 543: Could return directly from here?

wjwwood (approved):
I pointed out a few things to consider, but generally, lgtm.

rcl/src/rcl/graph.c
jacobperron: I can't find docs in rcl about the clock, but rmw is documented to use system time:

https://github.com/ros2/rmw/blob/ace271650ad39a656d81239050c9639c4190f077/rmw/include/rmw/rmw.h#L2371-L2374

Maybe it's better to use system time here instead; then we're consistent with `rcl_wait` (assuming the RMW is respecting the docs).

rcl/src/rcl/graph.c
```diff
@@ -452,6 +456,158 @@ rcl_count_subscribers(
...
+    return count_ret;
+  }
+  if (expected_count <= count) {
+    *success = true;
+    return RCL_RET_OK;
+  }
+
+  // Create a wait set and add the node graph guard condition to it
+  rcl_wait_set_t wait_set = rcl_get_zero_initialized_wait_set();
+  rcl_ret_t ret = rcl_wait_set_init(
```
jacobperron, line 497: oops, I should finalize the wait set structure before leaving this function.

rcl/src/rcl/graph.c
wjwwood: Yeah may be safer.

Somewhere I expressed the desire that rmw_wait would use steady time, but ultimately we're at the mercy of the underlying middleware.

rcl/src/rcl/graph.c
wjwwood: Oops, I missed that too.

jacobperron: Fixed in 51b3235

rcl/src/rcl/graph.c
jacobperron: Leaving this `break` to ensure we do the cleanup (51b3235)

rcl/include/rcl/graph.h
jacobperron: Added note about it: 0b94ee5

rcl/include/rcl/graph.h
jacobperron: Done: 0b94ee5

rcl/src/rcl/graph.c
jacobperron: Switched to system time: af14fc4

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14194/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8992/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11883/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14297/)

wjwwood (approved):

Karsten1987 (approved):
lgtm with green ci

rcl/test/rcl/test_graph.cpp
```diff
@@ -706,16 +706,104 @@ TEST_F(
...
+TEST_F(
+  CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION),
+  test_rcl_wait_for_publishers
+) {
+  rcl_ret_t ret;
+  rcl_node_t zero_node = rcl_get_zero_initialized_node();
+  rcl_allocator_t zero_allocator = static_cast<rcl_allocator_t>(
+    rcutils_get_zero_initialized_allocator());
+  rcl_allocator_t allocator = rcl_get_default_allocator();
+  const char * topic_name = "/topic_test_rcl_wait_for_publishers";
```
Karsten1987, line 720: I tried to use this API and I get a `invalid topic name` return code if the topic isn't absolute.

PR #913, opened 2021-04-21
Add rcl_publisher_wait_for_all_acked support

Barry-Xu-2018:
Related to ros2/rmw#295

For test timeout return value, the test case is as the below
1. Create 1 publisher and 3 subscribers 
2. The publisher send big message (1MB) 500 times continuously
3. Call rcl_publisher_wait_for_all_acked with small timeout value (500ms)

Refer to test file `test_publisher_wait_all_ack.cpp`

- Cyclonedds
  It uses sync publish. So the above test method is unsuitable. So not do test
- Fastdds
  By default, fastdds use intraprocess mode in current test scenario. But this leads to high-speed data transmission. Timeout cannot occur. So I have to use fastdds profile file to disable this mode (refer to the content of `disable_intraprocess.xml`). In order to avoid other test case, I have to create a new test class. But I find that, in a gtest program, fastdds seems to be loaded once. So I have to move new test class and test codes to create a new test file.  
- Connextdds
  No problem.
  
 I am not sure whether it is acceptable that a new test file is added only for testing one case.
 I would like to hear your opinions.

### DIFF
--- a/rcl/include/rcl/publisher.h
+++ b/rcl/include/rcl/publisher.h
@@ -439,7 +439,7 @@ rcl_publisher_assert_liveliness(const rcl_publisher_t * publisher);
 /**
  * This function waits until all published message data were acknowledged by peer node or timeout.
  *
- * The unit of timeout is nanoseconds.
+ * The timeout unit is nanoseconds.
  * If the timeout is negative then this function will block indefinitely until all published message
  * data were acknowledged.
  * If the timeout is 0 then this function will be non-blocking; checking all published message data
@@ -449,7 +449,7 @@ rcl_publisher_assert_liveliness(const rcl_publisher_t * publisher);
  * elapsed (return RCL_RET_TIMEOUT) or all published message data were acknowledged (return
  * RCL_RET_OK).
  *
- * This function only works effectively while QOS profile of publisher is set to RELIABLE.
+ * This function only waits for acknowledgments if the publisher's QOS profile is RELIABLE.
  * Otherwise this function will immediately return RCL_RET_OK.
  *
  * <hr>
@@ -464,7 +464,7 @@ rcl_publisher_assert_liveliness(const rcl_publisher_t * publisher);
  * \param[in] timeout the duration to wait for all published message data were acknowledged, in
  *   nanoseconds.
  * \return #RCL_RET_OK if successful, or
- * \return #RCL_RET_TIMEOUT if wait timed out, or
+ * \return #RCL_RET_TIMEOUT if timed out, or
  * \return #RCL_RET_PUBLISHER_INVALID if publisher is invalid, or
  * \return #RCL_RET_ERROR if an unspecified error occurs, or
  * \return #RCL_RET_UNSUPPORTED if the middleware does not support that feature.
--- a/rcl/include/rcl/publisher.h
+++ b/rcl/include/rcl/publisher.h
@@ -27,6 +27,7 @@ extern "C"
 #include "rcl/macros.h"
 #include "rcl/node.h"
 #include "rcl/visibility_control.h"
+#include "rcl/time.h"
 
 /// Internal rcl publisher implementation struct.
 struct rcl_publisher_impl_t;
@@ -442,9 +443,11 @@ rcl_publisher_assert_liveliness(const rcl_publisher_t * publisher);
  * If the timeout is negative then this function will block indefinitely until all published message
  * data were acknowledged.
  * If the timeout is 0 then this function will be non-blocking; checking all published message data
- * were acknowledged, but not waiting.
+ * were acknowledged (If acknowledged, return RCL_RET_OK. Otherwise, return RCL_RET_TIMEOUT), but
+ * not waiting.
  * If the timeout is greater than 0 then this function will return after that period of time has
- * elapsed or all published message data were acknowledged.
+ * elapsed (return RCL_RET_TIMEOUT) or all published message data were acknowledged (return
+ * RCL_RET_OK).
  *
  * This function only works effectively while QOS profile of publisher is set to RELIABLE.
  * Otherwise this function will immediately return RCL_RET_OK.
@@ -471,7 +474,7 @@ RMW_WARN_UNUSED
 rcl_ret_t
 rcl_publisher_wait_for_all_acked(
   const rcl_publisher_t * publisher,
-  int64_t timeout);
+  rcl_duration_value_t timeout);
 
 /// Get the topic name for the publisher.
 /**
--- a/rcl/include/rcl/publisher.h
+++ b/rcl/include/rcl/publisher.h
@@ -469,8 +469,8 @@ rcl_publisher_assert_liveliness(const rcl_publisher_t * publisher);
  * \return #RCL_RET_ERROR if an unspecified error occurs, or
  * \return #RCL_RET_UNSUPPORTED if the middleware does not support that feature.
  */
-RMW_PUBLIC
-RMW_WARN_UNUSED
+RCL_PUBLIC
+RCL_WARN_UNUSED
 rcl_ret_t
 rcl_publisher_wait_for_all_acked(
   const rcl_publisher_t * publisher,
--- a/rcl/include/rcl/publisher.h
+++ b/rcl/include/rcl/publisher.h
@@ -434,6 +434,45 @@ RCL_WARN_UNUSED
 rcl_ret_t
 rcl_publisher_assert_liveliness(const rcl_publisher_t * publisher);
 
+/// Wait until all published message data is acknowledged or until the specified timeout elapses.
+/**
+ * This function waits until all published message data were acknowledged by peer node or timeout.
+ *
+ * The unit of timeout is nanoseconds.
+ * If the timeout is negative then this function will block indefinitely until all published message
+ * data were acknowledged.
+ * If the timeout is 0 then this function will be non-blocking; checking all published message data
+ * were acknowledged, but not waiting.
+ * If the timeout is greater than 0 then this function will return after that period of time has
+ * elapsed or all published message data were acknowledged.
+ *
+ * This function only works effectively while QOS profile of publisher is set to RELIABLE.
+ * Otherwise this function will immediately return RCL_RET_OK.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | Yes
+ * Uses Atomics       | No
+ * Lock-Free          | No
+ *
+ * \param[in] publisher handle to the publisher that needs to wait for all acked.
+ * \param[in] timeout the duration to wait for all published message data were acknowledged, in
+ *   nanoseconds.
+ * \return #RCL_RET_OK if successful, or
+ * \return #RCL_RET_TIMEOUT if wait timed out, or
+ * \return #RCL_RET_PUBLISHER_INVALID if publisher is invalid, or
+ * \return #RCL_RET_ERROR if an unspecified error occurs, or
+ * \return #RCL_RET_UNSUPPORTED if the middleware does not support that feature.
+ */
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rcl_ret_t
+rcl_publisher_wait_for_all_acked(
+  const rcl_publisher_t * publisher,
+  int64_t timeout);
+
 /// Get the topic name for the publisher.
 /**
  * This function returns the publisher's internal topic name string.
--- a/rcl/src/rcl/publisher.c
+++ b/rcl/src/rcl/publisher.c
@@ -307,7 +307,7 @@ rcl_publisher_assert_liveliness(const rcl_publisher_t * publisher)
 }
 
 rcl_ret_t
-rcl_publisher_wait_for_all_acked(const rcl_publisher_t * publisher, int64_t timeout)
+rcl_publisher_wait_for_all_acked(const rcl_publisher_t * publisher, rcl_duration_value_t timeout)
 {
   if (!rcl_publisher_is_valid(publisher)) {
     return RCL_RET_PUBLISHER_INVALID;  // error already set
@@ -316,7 +316,7 @@ rcl_publisher_wait_for_all_acked(const rcl_publisher_t * publisher, int64_t time
   rmw_time_t rmw_timeout;
   if (timeout > 0) {
     rmw_timeout.sec = RCL_NS_TO_S(timeout);
-    rmw_timeout.nsec = timeout;
+    rmw_timeout.nsec = timeout % 1000000000;
   } else if (timeout < 0) {
     rmw_time_t infinite = RMW_DURATION_INFINITE;
     rmw_timeout = infinite;
--- a/rcl/src/rcl/publisher.c
+++ b/rcl/src/rcl/publisher.c
@@ -327,10 +327,11 @@ rcl_publisher_wait_for_all_acked(const rcl_publisher_t * publisher, rcl_duration
 
   rmw_ret_t ret = rmw_publisher_wait_for_all_acked(publisher->impl->rmw_handle, rmw_timeout);
   if (ret != RMW_RET_OK) {
-    RCL_SET_ERROR_MSG(rmw_get_error_string().str);
     if (ret == RMW_RET_TIMEOUT) {
       return RCL_RET_TIMEOUT;
-    } else if (ret == RMW_RET_UNSUPPORTED) {
+    }
+    RCL_SET_ERROR_MSG(rmw_get_error_string().str);
+    if (ret == RMW_RET_UNSUPPORTED) {
       return RCL_RET_UNSUPPORTED;
     } else {
       return RCL_RET_ERROR;
--- a/rcl/src/rcl/publisher.c
+++ b/rcl/src/rcl/publisher.c
@@ -27,6 +27,8 @@ extern "C"
 #include "rcl/node.h"
 #include "rcutils/logging_macros.h"
 #include "rcutils/macros.h"
+#include "rcl/time.h"
+#include "rmw/time.h"
 #include "rmw/error_handling.h"
 #include "tracetools/tracetools.h"
 
@@ -301,6 +303,40 @@ rcl_publisher_assert_liveliness(const rcl_publisher_t * publisher)
     RCL_SET_ERROR_MSG(rmw_get_error_string().str);
     return RCL_RET_ERROR;
   }
+  return RCL_RET_OK;
+}
+
+rcl_ret_t
+rcl_publisher_wait_for_all_acked(const rcl_publisher_t * publisher, int64_t timeout)
+{
+  if (!rcl_publisher_is_valid(publisher)) {
+    return RCL_RET_PUBLISHER_INVALID;  // error already set
+  }
+
+  rmw_time_t rmw_timeout;
+  if (timeout > 0) {
+    rmw_timeout.sec = RCL_NS_TO_S(timeout);
+    rmw_timeout.nsec = timeout;
+  } else if (timeout < 0) {
+    rmw_time_t infinite = RMW_DURATION_INFINITE;
+    rmw_timeout = infinite;
+  } else {
+    rmw_time_t zero = RMW_DURATION_UNSPECIFIED;
+    rmw_timeout = zero;
+  }
+
+  rmw_ret_t ret = rmw_publisher_wait_for_all_acked(publisher->impl->rmw_handle, rmw_timeout);
+  if (ret != RMW_RET_OK) {
+    RCL_SET_ERROR_MSG(rmw_get_error_string().str);
+    if (ret == RMW_RET_TIMEOUT) {
+      return RCL_RET_TIMEOUT;
+    } else if (ret == RMW_RET_UNSUPPORTED) {
+      return RCL_RET_UNSUPPORTED;
+    } else {
+      return RCL_RET_ERROR;
+    }
+  }
+
   return RCL_RET_OK;
 }
 
--- a/rcl/test/rcl/test_publisher_wait_all_ack.cpp
+++ b/rcl/test/rcl/test_publisher_wait_all_ack.cpp
@@ -18,6 +18,7 @@
 #include <chrono>
 #include <string>
 
+#include "rcl/allocator.h"
 #include "rcl/publisher.h"
 #include "rcl/subscription.h"
 #include "rcpputils/filesystem_helper.hpp"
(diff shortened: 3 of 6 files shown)


fujitatomoya (changes requested):

rcl/include/rcl/publisher.h
```diff
@@ -434,6 +434,45 @@ RCL_WARN_UNUSED
...
+ * \return #RCL_RET_PUBLISHER_INVALID if publisher is invalid, or
+ * \return #RCL_RET_ERROR if an unspecified error occurs, or
+ * \return #RCL_RET_UNSUPPORTED if the middleware does not support that feature.
+ */
+RMW_PUBLIC
+RMW_WARN_UNUSED
+rcl_ret_t
+rcl_publisher_wait_for_all_acked(
+  const rcl_publisher_t * publisher,
+  int64_t timeout);
```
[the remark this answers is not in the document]
Barry-Xu-2018: I refer to the below definition. So I define int64_t.

https://github.com/ros2/rcl/blob/1aa50bc14fc8a46acb7f19717c5e5d18ac7e479d/rcl/include/rcl/wait.h#L470-L473

Yeah. It is better to change it. 

https://github.com/ros2/rcl/blob/1aa50bc14fc8a46acb7f19717c5e5d18ac7e479d/rcl/include/rcl/time.h#L48

I think I can use `rcl_duration_value_t` instead.

rcl/include/rcl/publisher.h
```diff
@@ -434,6 +434,45 @@ RCL_WARN_UNUSED
...
 
+/// Wait until all published message data is acknowledged or until the specified timeout elapses.
+/**
+ * This function waits until all published message data were acknowledged by peer node or timeout.
+ *
+ * The unit of timeout is nanoseconds.
+ * If the timeout is negative then this function will block indefinitely until all published message
+ * data were acknowledged.
+ * If the timeout is 0 then this function will be non-blocking; checking all published message data
+ * were acknowledged, but not waiting.
```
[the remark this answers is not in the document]
Barry-Xu-2018: Yeah. I will supplement the description.

rcl/src/rcl/publisher.c
```diff
@@ -304,6 +306,40 @@ rcl_publisher_assert_liveliness(const rcl_publisher_t * publisher)
   return RCL_RET_OK;
 }
 
+rcl_ret_t
+rcl_publisher_wait_for_all_acked(const rcl_publisher_t * publisher, int64_t timeout)
```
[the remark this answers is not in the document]
Barry-Xu-2018: Will update

rcl/src/rcl/publisher.c
```diff
@@ -304,6 +306,40 @@ rcl_publisher_assert_liveliness(const rcl_publisher_t * publisher)
...
+{
+  if (!rcl_publisher_is_valid(publisher)) {
+    return RCL_RET_PUBLISHER_INVALID;  // error already set
+  }
+
+  rmw_time_t rmw_timeout;
+  if (timeout > 0) {
+    rmw_timeout.sec = RCL_NS_TO_S(timeout);
+    rmw_timeout.nsec = timeout;
```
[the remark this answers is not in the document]
Barry-Xu-2018: `timeout` is nanoseconds.   So here need not be converted.

rcl/test/rcl/test_publisher_wait_all_ack.cpp
```diff
@@ -0,0 +1,192 @@
...
+    is_connextdds = (std::string(rmw_get_implementation_identifier()).find("rmw_connextdds") == 0);
+
+    if (is_fastdds) {
+      // By default, fastdds use intraprocess mode in this scenario. But this leads to high-speed
+      // data transmission. test_wait_for_all_acked need low data transmission. So disable this
+      // mode via fastdds profile file.
+      rcpputils::fs::path fastdds_profile(TEST_RESOURCES_DIRECTORY);
+      fastdds_profile /= "test_profile/disable_intraprocess.xml";
+#ifdef _WIN32
+      _putenv(env_var.c_str());
```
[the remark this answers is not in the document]
Barry-Xu-2018: Yeah. Miss some codes. I will fix it.

rcl/test/rcl/test_publisher_wait_all_ack.cpp
```diff
@@ -0,0 +1,192 @@
...
+    is_cyclonedds = (std::string(rmw_get_implementation_identifier()).find("rmw_cyclonedds") == 0);
+    is_connextdds = (std::string(rmw_get_implementation_identifier()).find("rmw_connextdds") == 0);
+
+    if (is_fastdds) {
+      // By default, fastdds use intraprocess mode in this scenario. But this leads to high-speed
+      // data transmission. test_wait_for_all_acked need low data transmission. So disable this
+      // mode via fastdds profile file.
+      rcpputils::fs::path fastdds_profile(TEST_RESOURCES_DIRECTORY);
+      fastdds_profile /= "test_profile/disable_intraprocess.xml";
+#ifdef _WIN32
```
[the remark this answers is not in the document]
Barry-Xu-2018: Yeah.

rcl/src/rcl/publisher.c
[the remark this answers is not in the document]
fujitatomoya: i think it does, what if timeout is 100,123,456,789 nano seconds?

in that case, rmw_timeout.nsec will be 100,123,456,789. i think it should be 123,456,789 instead.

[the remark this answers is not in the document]
Barry-Xu-2018: Yeah. It's a mistake. I will fix it.

Barry-Xu-2018:
@fujitatomoya 

Please review again.

rcl/test/rcl/test_publisher_wait_all_ack.cpp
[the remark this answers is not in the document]
fujitatomoya: just checking, we are not gonna use `rcutils_set_env`? i would prefer to use that since that is use case ❓

fujitatomoya:
@Barry-Xu-2018 

i think overall it seems to be okay. thanks

rcl/test/rcl/test_publisher_wait_all_ack.cpp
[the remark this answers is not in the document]
Barry-Xu-2018: Sorry. Miss it. It should be changed.   
I update codes again.

Barry-Xu-2018:
@fujitatomoya 

Update again for using `rcutils_set_env`.   
Please check again.

rcl/include/rcl/publisher.h
```diff
@@ -434,6 +435,47 @@ RCL_WARN_UNUSED
...
+ * If the timeout is negative then this function will block indefinitely until all published message
+ * data were acknowledged.
+ * If the timeout is 0 then this function will be non-blocking; checking all published message data
+ * were acknowledged (If acknowledged, return RCL_RET_OK. Otherwise, return RCL_RET_TIMEOUT), but
+ * not waiting.
+ * If the timeout is greater than 0 then this function will return after that period of time has
+ * elapsed (return RCL_RET_TIMEOUT) or all published message data were acknowledged (return
+ * RCL_RET_OK).
+ *
+ * This function only works effectively while QOS profile of publisher is set to RELIABLE.
```
ivanpauno, line 452:
```suggestion
 * This function only waits for acknowledgments if the publisher's QOS profile is RELIABLE.
```

rcl/include/rcl/publisher.h
```diff
@@ -434,6 +435,47 @@ RCL_WARN_UNUSED
...
+ * Allocates Memory   | No
+ * Thread-Safe        | Yes
+ * Uses Atomics       | No
+ * Lock-Free          | No
+ *
+ * \param[in] publisher handle to the publisher that needs to wait for all acked.
+ * \param[in] timeout the duration to wait for all published message data were acknowledged, in
+ *   nanoseconds.
+ * \return #RCL_RET_OK if successful, or
+ * \return #RCL_RET_TIMEOUT if wait timed out, or
```
ivanpauno, line 467:
```suggestion
 * \return #RCL_RET_TIMEOUT if timed out, or
```

rcl/src/rcl/publisher.c
```diff
@@ -304,6 +306,40 @@ rcl_publisher_assert_liveliness(const rcl_publisher_t * publisher)
...
+    return RCL_RET_PUBLISHER_INVALID;  // error already set
+  }
+
+  rmw_time_t rmw_timeout;
+  if (timeout > 0) {
+    rmw_timeout.sec = RCL_NS_TO_S(timeout);
+    rmw_timeout.nsec = timeout % 1000000000;
+  } else if (timeout < 0) {
+    rmw_time_t infinite = RMW_DURATION_INFINITE;
+    rmw_timeout = infinite;
```
ivanpauno, line 322:
```suggestion
    rmw_timeout = RMW_DURATION_INFINITE;
```

rcl/src/rcl/publisher.c
```diff
@@ -304,6 +306,40 @@ rcl_publisher_assert_liveliness(const rcl_publisher_t * publisher)
...
+  } else {
+    rmw_time_t zero = RMW_DURATION_UNSPECIFIED;
+    rmw_timeout = zero;
+  }
+
+  rmw_ret_t ret = rmw_publisher_wait_for_all_acked(publisher->impl->rmw_handle, rmw_timeout);
+  if (ret != RMW_RET_OK) {
+    RCL_SET_ERROR_MSG(rmw_get_error_string().str);
```
ivanpauno, line 330: Should this be called when `ret == RMW_RET_TIMEOUT`?
I don't remember what we did in the rmw PRs, but `RMW_RET_TIMEOUT` is a non-error return code, so we shouldn't be setting any error neither calling RCL_SET_ERROR_MSG() here.

rcl/test/rcl/test_publisher_wait_all_ack.cpp
```diff
@@ -0,0 +1,268 @@
...
+    this->node_ptr, \
+    ts, \
+    topic_name, \
+    &subscription_options); \
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str; \
+  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT( \
+  { \
+    ret = rcl_subscription_fini(&subscription ## idx, this->node_ptr); \
+    EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str; \
+  });
```
ivanpauno, line 145: can we move the defines out of the function definition?

rcl/test/rcl/test_publisher_wait_all_ack.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+  INIT_SUBSCRIPTION(1)
+  INIT_SUBSCRIPTION(2)
+  INIT_SUBSCRIPTION(3)
+
+  ASSERT_TRUE(wait_for_established_subscription(&publisher, 10, 100));
+
+  // Prepare 1MB message
+#define ONE_MEGABYTE (1024 * 1024)
```
ivanpauno, line 154: same

rcl/test/rcl/test_publisher_wait_all_ack.cpp
```diff
@@ -0,0 +1,268 @@
...
+  ret = rcl_publisher_wait_for_all_acked(
+    &publisher,
+    RCL_MS_TO_NS(500));
+
+  if (is_cyclonedds) {
+    // cyclonedds use sync publish, so above scenario cannot lead to timeout.
+    EXPECT_EQ(RCL_RET_OK, ret);
+  } else {
+    EXPECT_EQ(RCL_RET_TIMEOUT, ret);
+  }
```
ivanpauno, line 188: can we do something like

```cpp
  if (ret != RCL_RET_OK && ret != RCL_RET_TIMOEUT) {
    ASSERT_FAILURE()
  }
```

?

Other rmw implementations can also be configured to use synchronous publishing, so better to do the test independent of the rmw implementation.

rcl/test/rcl/test_publisher_wait_all_ack.cpp
```diff
@@ -0,0 +1,268 @@
...
+  ret = rcl_publisher_wait_for_all_acked(&publisher, -1);
+  EXPECT_EQ(RCL_RET_OK, ret);
+}
+
+TEST_F(
+  CLASSNAME(TestPublisherFixtureSpecial, RMW_IMPLEMENTATION),
+  test_wait_for_all_acked_with_best_effort)
```
ivanpauno, line 195: the test seems to be too complex to only test that calling `rcl_publisher_wait_for_all_acked()` in a best effort publisher does nothing.

I would just create a publisher, no subscription, don't publish anything, and call `rcl_publisher_wait_for_all_acked()` expecting it to return OK.

ivanpauno (approved):
LGTM with comments addressed

rcl/src/rcl/publisher.c
Barry-Xu-2018:
```c++
#define RMW_DURATION_INFINITE {9223372036LL, 854775807LL}
```
So `RMW_DURATION_INFINITE` cannot be assigned to rmw_timeout.

rcl/src/rcl/publisher.c
Barry-Xu-2018: Yes. Return `RMW_RET_TIMEOUT` isn't an error.

rcl/test/rcl/test_publisher_wait_all_ack.cpp
Barry-Xu-2018: If modify as you said, rcl test with cyclonedds always report failure in CI.  
Developers have to check what is failed for rcl test.  
Does this failure bring inconvenience for other developer ？

rcl/test/rcl/test_publisher_wait_all_ack.cpp
Barry-Xu-2018: > I would just create a publisher, no subscription, don't publish anything, and call rcl_publisher_wait_for_all_acked() expecting it to return OK.

Even if a reliable publisher is used, the above steps also can return OK.   
So I have to make a complex case to test it.

rcl/test/rcl/test_publisher_wait_all_ack.cpp
ivanpauno: > If modify as you said, rcl test with cyclonedds always report failure in CI.

Sorry, but I think that's not the case.
For cyclonedds `ret` will be `RCL_RET_OK`, so `ret != RCL_RET_OK && ret != RCL_RET_TIMOEUT` is `false` and we don't enter the `if()`.
i.e. I want to modify the test so it passes when the function either returns RMW_RET_OK or RMW_RET_TIMEOUT, without listing what rmw implementation is supposed to return which value.

rcl/test/rcl/test_publisher_wait_all_ack.cpp
ivanpauno: > Even if a reliable publisher is used, the above steps also can return OK.

That could also happen in the current test case if a reliable publisher would be able to send all those messages in 500ms.
The test doesn't seem to be adding much value really.

Barry-Xu-2018: > That could also happen in the current test case if a reliable publisher would be able to send all those messages in 500ms.  

Yes. It is possible.  
But this scenario also leads to the failure of  test case `test_wait_for_all_acked`.  
```
  ret = rcl_publisher_wait_for_all_acked(
    &publisher,
    RCL_MS_TO_NS(500));

  if (is_cyclonedds) {
    // cyclonedds use sync publish, so above scenario cannot lead to timeout.
    EXPECT_EQ(RCL_RET_OK, ret);
  } else {
    EXPECT_EQ(RCL_RET_TIMEOUT, ret);  <== We expect to get RCL_RET_TIMEOUT
  }
```

Anyway, I will simplify this test case.

rcl/test/rcl/test_publisher_wait_all_ack.cpp
Barry-Xu-2018: Okay. I understand.

Barry-Xu-2018:
@ivanpauno 

Address your comments and do rebase.  
Please review again.

Barry-Xu-2018:
@fujitatomoya 

One comment (https://github.com/ros2/rcl/pull/913#discussion_r620220921) has been handled (https://github.com/ros2/rcl/pull/913#issuecomment-827270099). 
Please confirm it.

rcl/src/rcl/publisher.c
ivanpauno: got it, that's unfortunate.

rcl/test/rcl/test_publisher_wait_all_ack.cpp
```diff
@@ -0,0 +1,204 @@
...
+  constexpr char topic_name[] = "test_wait_for_all_acked_with_best_effort";
+  rcl_publisher_options_t publisher_options = rcl_publisher_get_default_options();
+  publisher_options.qos.reliability = RMW_QOS_POLICY_RELIABILITY_BEST_EFFORT;
+  publisher_options.qos.depth = 10000;
+  ret = rcl_publisher_init(&publisher, this->node_ptr, ts, topic_name, &publisher_options);
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+  {
+    rcl_ret_t ret = rcl_publisher_fini(&publisher, this->node_ptr);
+    EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
```
ivanpauno, line 197: EXPECT_EQ() inside a OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(), doesn't work, because it's just returning from the lambda (see [here](https://github.com/google/googletest/blob/master/docs/advanced.md#propagating-fatal-failures)).

You can use the [FAIL()](https://github.com/google/googletest/blob/master/docs/reference/assertions.md#fail-fail) macro instead.

rcl/test/rcl/test_publisher_wait_all_ack.cpp
```diff
@@ -0,0 +1,204 @@
...
+  constexpr char topic_name[] = "test_wait_for_all_acked";
+  rcl_publisher_options_t publisher_options = rcl_publisher_get_default_options();
+  publisher_options.qos.reliability = RMW_QOS_POLICY_RELIABILITY_RELIABLE;
+  publisher_options.qos.depth = 10000;
+  ret = rcl_publisher_init(&publisher, this->node_ptr, ts, topic_name, &publisher_options);
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+  {
+    rcl_ret_t ret = rcl_publisher_fini(&publisher, this->node_ptr);
+    EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
```
ivanpauno, line 134: same issue here

ivanpauno (approved):

Barry-Xu-2018:
@ivanpauno 

I check build error (https://build.ros2.org/job/Rpr__rcl__ubuntu_focal_amd64/570/console).  
```
10:00:45 [ 31%] Building CXX object test/CMakeFiles/test_log_level.dir/rcl/test_log_level.cpp.o
10:00:45 [ 32%] Linking CXX executable test_log_level
10:00:45 /usr/bin/ld: ../librcl.so: undefined reference to `rmw_publisher_wait_for_all_acked'
10:00:45 collect2: error: ld returned 1 exit status
```
`ros-rolling-rmw-implementation: 2.5.0-2focal.20210729.202045` used in CI doesn't include merged commit ros2/rmw_implementation#188.

fujitatomoya (approved):

ivanpauno:
> I check build error (https://build.ros2.org/job/Rpr__rcl__ubuntu_focal_amd64/570/console).

I will release a new version of `rmw_implementation` in rolling, that should fix the issue.

ivanpauno:
I have finished a new rolling release of rmw_implementation: https://github.com/ros/rosdistro/pull/30437.
In some hours we can rerun the Rpr checker.

ivanpauno:
@Barry-Xu-2018 could you check the windows' failures

Barry-Xu-2018:
> @Barry-Xu-2018 could you check the windows' failures?

Sorry，I am on vacation. So I am not at the computer. After I go back to home，I check it at once.

rcl/include/rcl/publisher.h
```diff
@@ -434,6 +435,47 @@ RCL_WARN_UNUSED
...
+ * \param[in] publisher handle to the publisher that needs to wait for all acked.
+ * \param[in] timeout the duration to wait for all published message data were acknowledged, in
+ *   nanoseconds.
+ * \return #RCL_RET_OK if successful, or
+ * \return #RCL_RET_TIMEOUT if timed out, or
+ * \return #RCL_RET_PUBLISHER_INVALID if publisher is invalid, or
+ * \return #RCL_RET_ERROR if an unspecified error occurs, or
+ * \return #RCL_RET_UNSUPPORTED if the middleware does not support that feature.
+ */
+RMW_PUBLIC
```
sloretz, line 472: Please change `RMW_PUBLIC` to `RCL_PUBLIC`. This is probably the reason for the failure to build on windows.

sloretz (changes requested):

rcl/include/rcl/publisher.h
Barry-Xu-2018: According to build log in CI, the failure is 
```
01:25:02 C:\ci\ws\src\ros2\rcl\rcl\test\rcl\test_publisher_wait_all_ack.cpp(91,5): error C3861: 'unsetenv': identifier not found [C:\ci\ws\build\rcl\test\test_publisher_wait_all_ack__rmw_cyclonedds_cpp.vcxproj]
```
`unsetenv` isn't supported on window.  
`rcutils_set_env()` support removing environment variable on window. So I use it to replace  `unsetenv`.

Barry-Xu-2018:
@sloretz  @ivanpauno 

I have fixed code.   
Please confirm it and help to re-execute CI.

ivanpauno (approved):
LGTM!

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15093/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9825/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12780/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15260/)

Barry-Xu-2018:
The failure on Windows is
```
[ RUN      ] TestPublisherFixtureSpecial__rmw_cyclonedds_cpp.test_wait_for_all_acked
unknown file: error: SEH exception with code 0xc00000fd thrown in the test body.
```
The exception code `0xc00000fd` is STATUS_STACK_OVERFLOW.   
This should be related  to below codes.  
```c++
char test_string[ONE_MEGABYTE];
```
So I will update code to fix this problem.

ivanpauno:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15265/)

Barry-Xu-2018:
@sloretz 

Please confirm my change.

PR #918, opened 2021-05-02
Set rmw log severity

### LINKED ISSUE #915: Make `rcl_logging_configure` call `rmw_set_log_severity`
#### Feature description

`rmw_set_log_severity` was added in:

* https://github.com/ros2/rmw/pull/124
* https://github.com/ros2/rmw_implementation/pull/30

It's implemented by the rmw implementation, but it doesn't really seem to be used anywhere/by anything.

It would make sense to have `rcl_logging_configure()` call `rmw_set_log_severity()`.

TBD

christophebedard:
Closes #915

This makes `rcl` call `rmw_set_log_severity` if the default level is set, e.g. with `--ros-args --log-level debug`.

Signed-off-by: Christophe Bedard <bedard.christophe@gmail.com>

### DIFF
--- a/rcl/src/rcl/logging.c
+++ b/rcl/src/rcl/logging.c
@@ -31,6 +31,7 @@ extern "C"
 #include "rcl/macros.h"
 #include "rcutils/logging.h"
 #include "rcutils/time.h"
+#include "rmw/rmw.h"
 
 #define RCL_LOGGING_MAX_OUTPUT_FUNCS (4)
 
@@ -76,6 +77,12 @@ rcl_logging_configure_with_output_handler(
   if (log_levels) {
     default_level = (int)log_levels->default_logger_level;
     rcutils_logging_set_default_logger_level(default_level);
+    if (RCUTILS_LOG_SEVERITY_UNSET != default_level) {
+      rmw_ret_t rmw_status = rmw_set_log_severity((rmw_log_severity_t)default_level);
+      if (RMW_RET_OK != rmw_status) {
+        return RCL_RET_ERROR;
+      }
+    }
 
     for (size_t i = 0; i < log_levels->num_logger_settings; ++i) {
       rcutils_ret_t rcutils_status = rcutils_logging_set_logger_level(


christophebedard:
Not sure if this is the best place to put it, but it feels like the simplest solution, and it's right next to `rcutils_logging_set_default_logger_level` so it does make sense.

rcl/src/rcl/logging.c
```diff
@@ -76,6 +77,12 @@ rcl_logging_configure_with_output_handler(
   if (log_levels) {
     default_level = (int)log_levels->default_logger_level;
     rcutils_logging_set_default_logger_level(default_level);
+    if (RCUTILS_LOG_SEVERITY_UNSET != default_level) {
+      rmw_ret_t rmw_status = rmw_set_log_severity((rmw_log_severity_t)default_level);
+      if (RMW_RET_OK != rmw_status) {
+        return RCL_RET_ERROR;
```
fujitatomoya, line 83: if implementation does not support `rmw_set_log_severity`, this will generate exception `RCLError` to user application to fail initialization. I think it is up to implementation to support `rmw_set_log_severity` or not. so how about print error to show `it cannot enable rmw implementation logger` instead of returning error?

christophebedard: Good point, done in e8a34876bde0621555ba8a4b73836c1ce10e2b35

christophebedard:
I've also added a test: 90ed0db3e87f868b7bbe811e865585de20d9b742

rcl/src/rcl/logging.c
```diff
@@ -76,6 +77,14 @@ rcl_logging_configure_with_output_handler(
   if (log_levels) {
     default_level = (int)log_levels->default_logger_level;
     rcutils_logging_set_default_logger_level(default_level);
+    if (RCUTILS_LOG_SEVERITY_UNSET != default_level) {
+      rmw_ret_t rmw_status = rmw_set_log_severity((rmw_log_severity_t)default_level);
```
clalancette, line 81: What you are doing here makes a certain sense; if we are changing the default logging level, we should change it everywhere.

On the other hand, I'm not sure what semantic we really want `--log-level` to have.  As it stands right now, the semantic is more-or-less "core ROS 2 code".  If we add this code in, it becomes "core ROS 2 code + RMW backend code".  I'm not entirely sure that's what we want to do.

I think this would be helped by spelling out what the log-levels mean in a more rigorous way, probably as part of https://github.com/ros2/design/issues/314 .

christophebedard: I understand. The fact that rmw implementations don't support the "unset" logging level/state definitely hightlights this.

Are there any plans to tackle that soon-ish after Galactic? I can't commit to doing ros2/design#314, but I can probably help, and I can certainly implement whatever is in the design document for this.

clalancette: > Are there any plans to tackle that soon-ish after Galactic?

The honest answer is that I don't know right now.  We've been concentrating on Galactic, so we don't have a roadmap for H-Turtle yet.  I do hope it is something that the ROS 2 community gets to for H-Turtle, as I think it is important.

christophebedard: Got it. I'll see what I can do (in a few weeks) then.

I'll leave this draft here for now.

Yadunund:
I'd be interested in resurrecting the effort here, especially to address https://github.com/ros2/rmw_zenoh/issues/423.

I'm in favor of having `--log-level` define the logging level at the RMW level as well since a lot of users are familiar with this arg and in fact, other tier-1 middlewares which rely on `RCUTILS_LOG_XX` macros seem to honor the `--log-level` arg. However in `rmw_zenoh`, we have custom logging macros and having `rcl` call `rmw_set_log_severity` would be a convenient way of changing the logging level of the custom logger.

PR #919, opened 2021-05-04
Wait for action server in rcl_action comm tests.

hidmic:
Similar to https://github.com/ros2/rmw_implementation/pull/191. Current tests assume action server and client in the same process discover each other immediately. That doesn't seem to be true for `rmw_connextdds` and thus [the flakes in amd64 Linux repeated CI](https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/lastCompletedBuild/testReport/rcl_action/TestActionCommunication__rmw_connextdds/test_valid_cancel_comm/), which can be reproduced locally (FYI @asorbini). 

This patch amends this. 

Repeated CI up to `rcl_action`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14557/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9314/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12223/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14686/)

### DIFF
--- a/rcl_action/test/rcl_action/test_action_communication.cpp
+++ b/rcl_action/test/rcl_action/test_action_communication.cpp
@@ -13,6 +13,9 @@
 // limitations under the License.
 #include <gtest/gtest.h>
 
+#include <chrono>
+#include <thread>
+
 #include "osrf_testing_tools_cpp/scope_exit.hpp"
 
 #include "rcl_action/action_client.h"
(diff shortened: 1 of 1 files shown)


clalancette (approved):
Seems reasonable to me.  There is no guarantee that the action server is immediately available, so I feel like this should have always been in place.

ivanpauno (approved):

PR #921, opened 2021-06-28
use parent logger

iuhilnehc-ynos:
to fix https://github.com/ros2/rclcpp/issues/1694


rcl/src/rcl/logging_rosout.c
```diff
@@ -236,10 +237,31 @@ void rcl_logging_rosout_output_handler(
...
   rcutils_ret_t rcutils_ret = rcutils_hash_map_get(&__logger_map, &name, &entry);
+  // if logger name not found, continue to check if its parent logger name exists
+  if (RCUTILS_RET_OK != rcutils_ret) {
+    name_dup = rcutils_strdup(name, allocator);
+    if (NULL == name_dup) {
+      RCUTILS_SAFE_FWRITE_TO_STDERR("unable to allocate memory\n");
+      return;
+    }
+    do {
+      char * name_sep = strrchr(name_dup, RCUTILS_LOGGING_SEPARATOR_STRING[0]);
```
clalancette, line 254: This is making an assumption that the `RCUTILS_LOGGING_SEPARATOR_STRING` is only a single character.  While that is true (and is unlikely to change anytime soon), it does seem weird that it is called "string", but then we are using it as a character.  Instead of using `strrchr`, can we search backwards in the string for a substr?

rcl/src/rcl/logging_rosout.c
```diff
@@ -236,10 +237,31 @@ void rcl_logging_rosout_output_handler(
...
+      if (NULL == name_sep) {
+        break;
+      }
+      name_dup[(size_t)(name_sep - name_dup)] = '\0';
+      if (!rcutils_logging_logger_is_enabled_for(name_dup, severity)) {
+        goto clean;
```
clalancette, line 260: Do we even need this goto label?  This function is setup to always run through to the end, so I think if we just `break` here, we'll clean up.

rcl/src/rcl/logging_rosout.c
```diff
@@ -290,6 +312,9 @@ void rcl_logging_rosout_output_handler(
       RCUTILS_SAFE_FWRITE_TO_STDERR("\n");
     }
   }
+
+clean:
+  allocator.deallocate(name_dup, allocator.state);
```
clalancette, line 317: You could move this deallocation directly after the loop you added above; we never use `name_dup` after that.

rcl/src/rcl/logging_rosout.c
```diff
@@ -236,10 +237,31 @@ void rcl_logging_rosout_output_handler(
...
+      RCUTILS_SAFE_FWRITE_TO_STDERR("unable to allocate memory\n");
+      return;
+    }
+    do {
+      char * name_sep = strrchr(name_dup, RCUTILS_LOGGING_SEPARATOR_STRING[0]);
+      if (NULL == name_sep) {
+        break;
+      }
+      name_dup[(size_t)(name_sep - name_dup)] = '\0';
+      if (!rcutils_logging_logger_is_enabled_for(name_dup, severity)) {
```
clalancette, line 259: This kind of seems like the wrong layer of the logging to do this in.  Everywhere else, this is done at the `rcutils` layer, while here we are introducing this at the `rcl` layer.

clalancette (changes requested):
I've got a couple of changes suggested inline.

Separately, I'm slightly concerned about the run-time performance hit of this.  That is, this function is called for *every* log message that goes out, and we are now making this function even more expensive for sub-loggers.  I'm kind of wondering if we should instead add all sub-loggers to the `__logger_map` instead, which means we would use more memory, but keep this function faster.  Thoughts?

iuhilnehc-ynos:
> I've got a couple of changes suggested inline.

Thanks for your review.

> Separately, I'm slightly concerned about the run-time performance hit of this. That is, this function is called for _every_ log message that goes out, and we are now making this function even more expensive for sub-loggers. I'm kind of wondering if we should instead add all sub-loggers to the `__logger_map` instead, which means we would use more memory, but keep this function faster. Thoughts?

You're right. It definitely costs more consumption each time while using the sub-loggers not existing. I think we could add all sub-loggers to the `__logger_map`. I'll reconsider it.

iuhilnehc-ynos:
@clalancette 

I think a node is unnecessary to have multiple publishers named '/rosout', so I will have the publisher shared by the sub-logger. 

The following is my thought.
- update the [rosout_map_entry_t structure](https://github.com/ros2/rcl/blob/80efa3fa0861da45c1f81622763d06fc1f76f129/rcl/src/rcl/logging_rosout.c#L77-L81)
```
typedef struct rosout_map_entry_t
{
  rcl_node_t * node;
  rcl_publisher_t publisher;
  // identicate whether the publisher is shared from parent node
  bool share_from_parent;
  // store the child logger names to make sure clean them if current node is finalized
  rcutils_array_list_t child_logger_names;
} rosout_map_entry_t;
```

- rcl_logging_rosout_output_handler will create a new map item (with subloggername and its parent entry, the entry need to be updated) for a coming sublogger if necessary
- also update rcl_logging_rosout_init_publisher_for_node and rcl_logging_rosout_fini_publisher_for_node

Do you have any suggestions?

clalancette:
I think I understand your proposal; basically, whenever we see a new sublogger in `rcl_logging_rosout_output_handler`, we'll create a new entry in the map for it, and then subsequent calls to `rcl_logging_rosout_output_handler` can reuse that map entry.  The rest of the changes are bookkeeping for cleaning up subloggers later on.  Is that correct?

That will work, though it makes `rcl_logging_rosout_output_handler` more complicated and more jittery; it will "sometimes" have to do memory allocations in the fast path.

Before we go that route, let me ask a question: is the rosout logging layer notified when a new sublogger is created from a parent logger?  If it isn't, can we make it so?  Being able to do all of this setup work at sublogger creation time would be better, in my opinion.

iuhilnehc-ynos:
> I think I understand your proposal; basically, whenever we see a new sublogger in `rcl_logging_rosout_output_handler`, we'll create a new entry in the map for it, and then subsequent calls to `rcl_logging_rosout_output_handler` can reuse that map entry. The rest of the changes are bookkeeping for cleaning up subloggers later on. Is that correct?

Yes.

> That will work, though it makes `rcl_logging_rosout_output_handler` more complicated and more jittery; it will "sometimes" have to do memory allocations in the fast path.

Sorry for not finding out a better idea based on the current design. Maybe in the future, we could make an implicit node to create a publisher on the topic '/rosout'. (I don't know why ros2 needs to create multiple publishers on topic '/rosout' if there are multiple nodes.)

> Before we go that route, let me ask a question: is the rosout logging layer notified when a new sublogger is created from a parent logger? If it isn't, can we make it so? Being able to do all of this setup work at sublogger creation time would be better, in my opinion.

Could you give me more explanations about `is the rosout logging layer notified when a new sublogger is created from a parent logger`? I don't know the purpose of this notification.

clalancette:
> Could you give me more explanations about `is the rosout logging layer notified when a new sublogger is created from a parent logger`? I don't know the purpose of this notification.

Going back to the original problem, it seems like the problem comes about in a line like this:

```
    RCLCPP_INFO(get_logger().get_child("child"), "Child logger");
```

Reading the rclcpp code a bit, it looks like `get_logger()` returns a [Logger instance](https://github.com/ros2/rclcpp/blob/e9e398d2d49de2ecc3148d425f8659d9db74918c/rclcpp/include/rclcpp/logger.hpp#L63-L65).  And `get_child()` is a [method on the Logger instance](https://github.com/ros2/rclcpp/blob/e9e398d2d49de2ecc3148d425f8659d9db74918c/rclcpp/include/rclcpp/logger.hpp#L158-L166).

My suggestion is that we add a new function to https://github.com/ros2/rcl/blob/master/rcl/include/rcl/logging_rosout.h called something like `rcl_add_sublogger()`, which would allocate and add a sublogger to the `__logger_map` if it doesn't already exist.  And then we would call that anytime we construct a new child logger (maybe just anytime we construct a Logger at all, I'm not sure).  That way, we do the initialization at child logger construction time, and keep `rcl_logging_rosout_output_handler` simpler.

iuhilnehc-ynos:
> Going back to the original problem, it seems like the problem comes about in a line like this:
> 
> ```
>     RCLCPP_INFO(get_logger().get_child("child"), "Child logger");
> ```
> 
> Reading the rclcpp code a bit, it looks like `get_logger()` returns a [Logger instance](https://github.com/ros2/rclcpp/blob/e9e398d2d49de2ecc3148d425f8659d9db74918c/rclcpp/include/rclcpp/logger.hpp#L63-L65). And `get_child()` is a [method on the Logger instance](https://github.com/ros2/rclcpp/blob/e9e398d2d49de2ecc3148d425f8659d9db74918c/rclcpp/include/rclcpp/logger.hpp#L158-L166).
> 
> My suggestion is that we add a new function to https://github.com/ros2/rcl/blob/master/rcl/include/rcl/logging_rosout.h called something like `rcl_add_sublogger()`, which would allocate and add a sublogger to the `__logger_map` if it doesn't already exist. And then we would call that anytime we construct a new child logger (maybe just anytime we construct a Logger at all, I'm not sure). That way, we do the initialization at child logger construction time, and keep `rcl_logging_rosout_output_handler` simpler.

Thank you for your suggestion. I'd really appreciate that.
Do you mean to use the `sublogger` must through get_child()? If so, I think I thought of that implementation before, but it seems there is unexpected behaviour, if we can accept that, I'll continue to implement it.

there is a node named 'node',
1. normal case
```cpp
RCLCPP_INFO(get_logger().get_child("child"), "Child logger");
```
`get_child()` call `rcl_add_sublogger` to add a sublogger, so it will use the real sublogger to log message. 

2. expected case
```cpp
RCLCPP_INFO(get_logger("node.child"), "not to publish message on rosout");
```
Because rcl_add_sublogger was not called before and there is no node named `/node/child`, it can't publish the message on `rosout`

3. unexpected case

if `get_child()` was called before, such as
```cpp
RCLCPP_INFO(get_logger().get_child("child"), "Child logger");
```
and then,
```cpp
RCLCPP_INFO(get_logger("node.child"), "it will use the child logger");
```

Do you think the cases(2, 3) will make users confused?

fujitatomoya:
adding dynamic memory allocation and hashmap operation would be issue for logging as mentioned before. in production, logging is really important to make sure that we can have sufficient information w/o overhead.

> Because rcl_add_sublogger was not called before

i think we can call internally based on `"node.child"`? if i am not mistaken.

iuhilnehc-ynos:
> adding dynamic memory allocation and hashmap operation would be issue for logging as mentioned before. in production, logging is really important to make sure that we can have sufficient information w/o overhead.
> 
> > Because rcl_add_sublogger was not called before
> 
> i think we can call internally based on `"node.child"`? if i am not mistaken.

Thank @fujitatomoya, if so, I think the added function `rcl_add_sublogger` only needs to be called in `rcl` instead of `rclcpp`. (In other words, `rcl_add_sublogger` can be a static function inside `rcl`, which will be called at `rcl_logging_rosout_output_handler`)

clalancette:
@iuhilnehc-ynos Hm, good point about the `"node.child"` issue; I hadn't thought about that one.

In that case, I think I would propose:
1.  Create a function in rcl which is `rcl_add_sublogger`.  You'll probably also want an `rcl_remove_sublogger` for destruction time.
2. Call `rcl_add_sublogger` when a `Logger` instance is constructed (and `rcl_remove_sublogger` when it is destroyed).
3. Also call `rcl_add_sublogger` from `rcl_logger_rosout_output_handler` if the logger isn't in the `__logger_map`.

That way, you can avoid allocations at runtime by doing:

```
auto sublogger = get_logger().get_child("child");
RCLCPP_INFO(sublogger, "foo");
```

But if you forget to do that and just do:

```
RCLCPP_INFO(get_logger("node.child"), "foo");
```

things will still work.  What do you think about that?

iuhilnehc-ynos:
@clalancette

Thanks for your proposal. There is something that needs to be confirmed.
1. Currently, the `Logger` of `rcpcpp` does not use `rcl_node_t` inside, which means we can't use the following interface in `Logger`,
```
rcl_ret_t
rcl_logging_rosout_add_sublogger(
  rcl_node_t * node, const char * sublogger_name)
{
  char * logger_name = NULL;

  RCL_LOGGING_ROSOUT_VERIFY_INITIALIZED
  RCL_CHECK_ARGUMENT_FOR_NULL(node, RCL_RET_NODE_INVALID);
  if (NULL == sublogger_name) {
    RCL_SET_ERROR_MSG("Sub-logger name was null.");
    return RCL_RET_ERROR;
  }

  logger_name = _rcl_logging_rosout_get_sublogger_name(node, sublogger_name);
  if (NULL == logger_name) {
    // Error already set
    return RCL_RET_ERROR;
  }

  status = _rcl_logging_rosout_add_logger(node, logger_name);
  // TODO: save logger_name into a list,
  // it can't be freed immediately because the hashmap need it

  return status;
}
```
so the `rcl_logging_rosout_add_sublogger` seems to be implemented as,
```
rcl_ret_t
rcl_logging_rosout_add_sublogger(
  const char * sublogger_name)
{
  1. to search from `__logger_map` by substring the `sublogger_name` to locate an existing rcl_node_t
  2. create a new string `sublogger_name_dup` from sublogger_name
  3. use the rcl_node_t to create a new publisher, and then add the item(sublogger_name_dup, new_entry) into `__logger_map`
  4. put the new `sublogger_name_dup` into a list
}
```

I don't think it's what we want because this kind of API (without `rcl_node_t *` argument) can't be the same group inside `logging_rosout.c`, I need to wait for your confirmation.

clalancette:
>     1. Currently, the `Logger` of `rcpcpp` does not use `rcl_node_t` inside, which means we can't use the following interface in `Logger`,

Good point.

> ```
> rcl_ret_t
> rcl_logging_rosout_add_sublogger(
>   const char * sublogger_name)
> {
>   1. to search from `__logger_map` by substring the `sublogger_name` to locate an existing rcl_node_t
>   2. create a new string `sublogger_name_dup` from sublogger_name
>   3. use the rcl_node_t to create a new publisher, and then add the item(sublogger_name_dup, new_entry) into `__logger_map`
>   4. put the new `sublogger_name_dup` into a list
> }
> ```

OK, I understand this pseudo code.

> I don't think it's what we want because this kind of API (without `rcl_node_t *` argument) can't be the same group inside `logging_rosout.c`, I need to wait for your confirmation.

I don't quite understand this part.  Can you rephrase?

iuhilnehc-ynos:
> > I don't think it's what we want because this kind of API (without `rcl_node_t *` argument) can't be the same group inside `logging_rosout.c`, I need to wait for your confirmation.
> 
> I don't quite understand this part. Can you rephrase?

Sorry to make you confused. I thought it seems like an indirect API in RCL because this API need to search a rcl_node_t inside. If only considering the implementation level of `RCL`, I'd prefer to encapsulate an API like the first one(`rcl_logging_rosout_add_sublogger(rcl_node_t *, const char *)`).

Anyway, I'll implement this API as
```
rcl_ret_t
rcl_logging_rosout_add_sublogger(
  const char * logger_name, const char * sublogger_name)
```

rcl/src/rcl/logging_rosout.c
```diff
@@ -292,6 +343,122 @@ void rcl_logging_rosout_output_handler(
...
+
+  RCL_LOGGING_ROSOUT_VERIFY_INITIALIZED
+  if (NULL == logger_name) {
+    RCL_SET_ERROR_MSG("Logger name was null.");
+    return RCL_RET_INVALID_ARGUMENT;
+  }
+  if (NULL == sublogger_name) {
+    RCL_SET_ERROR_MSG("Sub-logger name was null.");
+    return RCL_RET_INVALID_ARGUMENT;
+  }
```
fujitatomoya, line 378: these are redundant for add/remove API, this could be checked in `_rcl_logging_rosout_get_sublogger_name`?

rcl/src/rcl/logging_rosout.c
```diff
@@ -292,6 +343,122 @@ void rcl_logging_rosout_output_handler(
...
+
+  rcutils_ret_t rcutils_ret = rcutils_hash_map_get(&__logger_map, &logger_name, &entry);
```
fujitatomoya, line 380: hash map management would be called after we can get the `full_sublogger_name` w/o error.

rcl/test/rcl/test_logging_rosout.cpp
```diff
@@ -67,43 +67,39 @@ std::ostream & operator<<(
 
 class CLASSNAME (TestLogRosoutFixtureNotParam, RMW_IMPLEMENTATION) : public ::testing::Test {};
 
-class TEST_FIXTURE_P_RMW (TestLoggingRosoutFixture)
-  : public ::testing::TestWithParam<TestParameters>
+class TestLoggingRosout : public ::testing::Test
```
fujitatomoya, line 70: just confirmation for my understanding. introducing this class as generic, so that this can be used in rosout parameter test and new test. correct?

fujitatomoya (changes requested):

rcl/src/rcl/logging_rosout.c
iuhilnehc-ynos: OK, I'll update it because we could call https://github.com/ros2/rcl/blob/108d4fa793d69933d8d8f298f6f868c5dbb2f8ba/rcl/src/rcl/logging_rosout.c#L380
after getting the `full_sublogger_name`.

rcl/test/rcl/test_logging_rosout.cpp
iuhilnehc-ynos: Yes.

rcl/src/rcl/logging_rosout.c
```diff
@@ -292,6 +343,110 @@ void rcl_logging_rosout_output_handler(
...
+    __rosout_allocator.deallocate(full_sublogger_name, __rosout_allocator.state);
+    return RCL_RET_ERROR;
+  }
+
+  if (rcutils_hash_map_key_exists(&__logger_map, &full_sublogger_name)) {
+    __rosout_allocator.deallocate(full_sublogger_name, __rosout_allocator.state);
+    return RCL_RET_SUBLOGGER_ALREADY_EXIST;
+  }
+
+  status = _rcl_logging_rosout_add_logger(entry.node, full_sublogger_name);
```
iuhilnehc-ynos, line 393: Use `rcutils_hash_map_set(&__logger_map, &full_sublogger_name, &entry) instead of `_rcl_logging_rosout_add_logger` that adding a new publisher.

iuhilnehc-ynos:
@clalancette 

Could you help review it in your spare time? Thanks in advance.

fujitatomoya:
@clalancette friendly ping.

fujitatomoya (approved):

fujitatomoya (approved):
lgtm

fujitatomoya:
@clalancette friendly ping.

rcl/src/rcl/logging_rosout.c
```diff
@@ -87,13 +89,52 @@ rcl_ret_t rcl_logging_rosout_init(const rcl_allocator_t * allocator)
...
+    return status;
+  }
+
+  __sublogger_names = rcutils_get_zero_initialized_array_list();
+  status = rcl_ret_from_rcutils_ret(
+    rcutils_array_list_init(
+      &__sublogger_names, 2, sizeof(const char **), allocator));
+  if (RCL_RET_OK != status) {
+    status = rcl_ret_from_rcutils_ret(rcutils_hash_map_fini(&__logger_map));
+    return status;
```
clalancette, line 102: I don't think we want to return the status from `rcutils_hash_map_fini` here.  Instead, we should return the status from above.

rcl/src/rcl/logging_rosout.c
```diff
@@ -292,3 +355,133 @@ void rcl_logging_rosout_output_handler(
...
+  if (RCL_RET_OK != status) {
+    RCL_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "Failed to add publisher to map for logger '%s'.", full_sublogger_name);
+    goto cleanup;
+  }
+
+  status = rcl_ret_from_rcutils_ret(
+    rcutils_array_list_add(&__sublogger_names, &full_sublogger_name));
+  if (RCL_RET_OK != status) {
+    goto cleanup;
```
clalancette, line 424: If we fail here, we actually should go back and remove the sublogger from the __logger_map.

rcl/src/rcl/logging_rosout.c
```diff
@@ -292,3 +355,133 @@ void rcl_logging_rosout_output_handler(
...
+    _rcl_logging_rosout_get_sublogger_name(logger_name, sublogger_name, &full_sublogger_name);
+  if (RCL_RET_OK != status) {
+    // Error already set
+    return status;
+  }
+
+  // remove the entry from the map
+  if (!rcutils_hash_map_key_exists(&__logger_map, &full_sublogger_name)) {
+    RCL_SET_ERROR_MSG_WITH_FORMAT_STRING("Sub-logger '%s' not exist.", full_sublogger_name);
+    __rosout_allocator.deallocate(full_sublogger_name, __rosout_allocator.state);
```
clalancette, line 456: Like `rcl_logging_rosout_add_sublogger`, we should have a `cleanup` label and go there when we fail.

This will also allow us to unravel the cleanup below, so we can `goto cleanup` during failures.

rcl/src/rcl/logging_rosout.c
```diff
@@ -74,6 +75,7 @@ typedef struct rosout_map_entry_t
 static rcutils_hash_map_t __logger_map;
 static bool __is_initialized = false;
 static rcl_allocator_t __rosout_allocator;
+static rcutils_array_list_t __sublogger_names;
```
clalancette, line 78: I'm wondering whether this should be a hash_map, rather than a list.  That way when removing them, we don't have to look through all subloggers to find it, which should improve performance.  Thoughts?

rcl/test/rcl/test_logging_rosout.cpp
```diff
@@ -134,12 +130,49 @@ class TEST_FIXTURE_P_RMW (TestLoggingRosoutFixture)
...
+  virtual void call_rcl_init()
+  {
+    rcl_ret_t ret = rcl_init(0, NULL, &init_options_, this->context_ptr);
+    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  }
+
+  virtual void update_node_option() {}
+
+  rcl_init_options_t init_options_;
+  rcl_node_options_t node_options_;
```
clalancette, line 142: I'll suggest to remove the trailing `_`, just to be consistent with the rest.  We aren't totally consistent in tests, but we may as well be consistent within the same class.

clalancette (changes requested):
Mostly small comments, though there is one question about the underlying data structure.

iuhilnehc-ynos:
Thanks for your reviews. @clalancette

iuhilnehc-ynos:
try CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18065/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12593/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18709/)

iuhilnehc-ynos:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18069/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12596/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18714/)

rcl/src/rcl/logging_rosout.c
```diff
@@ -87,48 +98,138 @@ rcl_ret_t rcl_logging_rosout_init(const rcl_allocator_t * allocator)
...
+    return RCL_RET_OK;
+  }
+  rcl_ret_t status = RCL_RET_OK;
+  rosout_map_entry_t entry;
+  rosout_sublogger_entry_t sublogger_entry;
+
+  status = _rcl_logging_rosout_clear_hashmap(
+    &__logger_map, _rcl_logging_rosout_clear_logger_map_item, &entry);
+  if (RCL_RET_OK != status) {
+    goto exit;
```
clalancette, line 219: This is purely stylistic, but I think that we can just `return status` here.  That is, the extra `goto` here doesn't really buy us much.  Same goes a few other times below.

rcl/src/rcl/logging_rosout.c
```diff
@@ -146,6 +247,10 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(rcl_node_t * node)
   RCL_CHECK_ARGUMENT_FOR_NULL(node, RCL_RET_NODE_INVALID);
   logger_name = rcl_node_get_logger_name(node);
   if (NULL == logger_name) {
+    // Prevent from printing an extra error information in some tests
```
clalancette, line 250: Can you please comment on what tests were printing this extra error information?

clalancette (commented):
I left two thoughts inline.  Other than those, this is looking pretty good to me.

rcl/src/rcl/logging_rosout.c
iuhilnehc-ynos: When I run the test with `./test_logging_rosout__rmw_fastrtps_cpp --gtest_filter="*.test_bad_params_init_fini_node_publisher"`, the test outputs as follows,

```
chenlh test $ ./test_logging_rosout__rmw_fastrtps_cpp --gtest_filter="*.test_bad_params_init_fini_node_publisher"
Running main() from /home/chenlh/Projects/ROS2/ros2-master/local_install_test/gtest_vendor/src/gtest_vendor/src/gtest_main.cc
Note: Google Test filter = *.test_bad_params_init_fini_node_publisher
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from TestLogRosoutFixtureNotParam__rmw_fastrtps_cpp
[ RUN      ] TestLogRosoutFixtureNotParam__rmw_fastrtps_cpp.test_bad_params_init_fini_node_publisher

  'rcl node implementation is invalid, at /home/chenlh/Projects/ROS2/ros2-master/src/ros2/rcl/rcl/src/rcl/node.c:414'

with this new error message:

  'Logger name was null., at /home/chenlh/Projects/ROS2/ros2-master/src/ros2/rcl/rcl/src/rcl/logging_rosout.c:249'

<<<
[       OK ] TestLogRosoutFixtureNotParam__rmw_fastrtps_cpp.test_bad_params_init_fini_node_publisher (0 ms)
[----------] 1 test from TestLogRosoutFixtureNotParam__rmw_fastrtps_cpp (0 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (0 ms total)
[  PASSED  ] 1 test.
```
I expected a clean output with only gtest information.

Considering that it's not what this PR should do, I'll remove this fix, because I think this issue exists most places.

iuhilnehc-ynos:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18077/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12605/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18721/)

clalancette (approved):
Thanks for the patience and all of the iterations here.

PR #933, opened 2021-09-08
Include rmw_event_t instead of forward declaring it

hidmic:
Connected to https://github.com/ros2/rmw/pull/313.

### DIFF
--- a/rcl/include/rcl/event.h
+++ b/rcl/include/rcl/event.h
@@ -47,7 +47,7 @@ typedef enum rcl_subscription_event_type_t
 } rcl_subscription_event_type_t;
 
 /// rmw struct.
-typedef struct rmw_event_t rmw_event_t;
+typedef struct rmw_event_s rmw_event_t;
 
 /// Internal rcl implementation struct.
 struct rcl_event_impl_t;
--- a/rcl/include/rcl/event.h
+++ b/rcl/include/rcl/event.h
@@ -22,6 +22,8 @@ extern "C"
 {
 #endif
 
+#include <rmw/event.h>
+
 #include "rcl/client.h"
 #include "rcl/macros.h"
 #include "rcl/publisher.h"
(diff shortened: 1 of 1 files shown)


rcl/include/rcl/event.h
```diff
@@ -47,7 +47,7 @@ typedef enum rcl_subscription_event_type_t
 } rcl_subscription_event_type_t;
 
 /// rmw struct.
-typedef struct rmw_event_t rmw_event_t;
+typedef struct rmw_event_s rmw_event_t;
```
clalancette, line 50: I know you didn't add this code, but this line actually feels wrong.  We should never be redefining `rmw_*` structures in this file, just using them.

Elsewhere in this package, we have lines like:

```
typedef rmw_serialized_message_t rcl_serialized_message_t;
```

I *suspect* that was what was originally intended, but it got crossed up somehow.

Regardless, I think we should either remove this line completely (and just use `rmw_event_t` everywhere), or change this to:

```
typedef rmw_event_t rcl_event_t;
```

And change the function signatures to match.  Thoughts?

hidmic: I can only guess, but I *think* it was intentional. `rcl/event.h` does not include `rmw/event.h`, so this would be a forward declaration. Of course, we can drop this and simply include the corresponding header.

clalancette: > Of course, we can drop this and simply include the corresponding header.

Yeah, that seems preferable to me.

hidmic: See https://github.com/ros2/rcl/pull/933/commits/af75126a45ffd600dd62d8c61746dc5f9e16e65b.

clalancette (approved):
Looks good to me.  I'll suggest a full CI on this just to be sure there are no surprises in downstream consumers.

hidmic:
Full CI (from https://github.com/ros2/rmw/pull/313):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15218/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9938/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12898/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15400/)

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2021-9-16/22372/1

Issue #934, opened 2021-09-14
Question about conditional statements in graph.c file


YoungJin-gurum:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - from source
- Version or commit hash:
  - 3.2.0(master) 
- DDS implementation:
  - GurumDDS(after 2.7.2774), rmw_gurumdds_cpp(rolling, galactic)
- Client library (if applicable):
  - rcl 

#### Steps to reproduce issue

```
Execute test of rcl (colcon test) with gurumdds, rmw_gurumdds_cpp.
```

#### Expected behavior
rcl test_graph_query_functions
At check_graph_state after subscription fini
expected_in_tnat result passed (expect topic isn't in tnat)
#### Actual behavior
failed
#### Additional information
I have a question about the code at line 491 and 548 of <graph.c>
```
  if (expected_count <= count) {
    *success = true;
    return RCL_RET_OK;
  }
```
Variable count is result of count_entities_func.
After entity fini, if it has not yet received signal of deleted entity, count will be 1.
Then the condition is satisfied (expected_count 0 <= count 1).
So it will return OK immediately, then rcl_get_topic_names_and_types gets topic_cahce.
As I think, condition should be changed to
```
 expected_count == count
```
I think that makes waitset operate normally until the graph changes.
If there is a valid reason for the inequality sign in the conditional statement, please let me know.

aprotyas:
For context, directing you to #907 where this change was merged.

YoungJin-gurum:
> For context, directing you to #907 where this change was merged.

Thank you for your reply. I'm already read #907, but the reason for the conditional statement is not explained.

I want to understand why success is when count is greater than expected_count.

aprotyas:
> I want to understand why success is when count is greater than expected_count.

I don't have much context, but I'm quoting a section from the `rcl_wait_for_publishers` function header https://github.com/ros2/rcl/blob/7b899dae4a803575394588397c3e12e4be214abb/rcl/include/rcl/graph.h#L596-L597

Note that `rcl_wait_for_publishers` calls `_rcl_wait_for_entities` (which contains the code you originally referenced) https://github.com/ros2/rcl/blob/7b899dae4a803575394588397c3e12e4be214abb/rcl/src/rcl/graph.c#L594-L611

With it passing in the `expected_count`, inferring from the expected behavior specified in the function header, I would expect the wait function to be successful when `count >= expected_count`.

[...]

PR #935, opened 2021-09-14
test should check specified number of entities.

fujitatomoya:
address https://github.com/ros2/rcl/issues/934

### DIFF
--- a/rcl/test/rcl/test_graph.cpp
+++ b/rcl/test/rcl/test_graph.cpp
@@ -797,62 +797,6 @@ TEST_F(
 }
 
 void
-check_graph_state(
-  const rcl_node_t * node_ptr,
-  std::string & topic_name,
-  size_t expected_publisher_count,
-  size_t expected_subscriber_count,
-  bool expected_in_tnat,
-  const std::chrono::nanoseconds & timeout)
-{
-  RCUTILS_LOG_DEBUG_NAMED(
-    ROS_PACKAGE_NAME,
-    "Expecting %zu publishers, %zu subscribers, and that the topic is%s in the graph.",
-    expected_publisher_count,
-    expected_subscriber_count,
-    expected_in_tnat ? "" : " not"
-  );
-  bool is_in_tnat = false;
-  rcl_names_and_types_t tnat {};
-  rcl_ret_t ret;
-  rcl_allocator_t allocator = rcl_get_default_allocator();
-
-  // Wait for expected number of publishers
-  bool success = false;
-  ret = rcl_wait_for_publishers(
-    node_ptr, &allocator, topic_name.c_str(), expected_publisher_count, timeout.count(), &success);
-  ASSERT_EQ(ret, RCL_RET_OK);
-  EXPECT_TRUE(success);
-  // Wait for expected number of subscribers
-  success = false;
-  ret = rcl_wait_for_subscribers(
-    node_ptr, &allocator, topic_name.c_str(), expected_subscriber_count, timeout.count(), &success);
-  ASSERT_EQ(ret, RCL_RET_OK);
-  EXPECT_TRUE(success);
-
-  tnat = rcl_get_zero_initialized_names_and_types();
-  ret = rcl_get_topic_names_and_types(node_ptr, &allocator, false, &tnat);
-  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
-  is_in_tnat = false;
-  for (size_t i = 0; RCL_RET_OK == ret && i < tnat.names.size; ++i) {
-    if (topic_name == std::string(tnat.names.data[i])) {
-      ASSERT_FALSE(is_in_tnat) << "duplicates in the tnat";  // Found it more than once!
-      is_in_tnat = true;
-    }
-  }
-  if (RCL_RET_OK == ret) {
-    ret = rcl_names_and_types_fini(&tnat);
-    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
-  }
-
-  if (expected_in_tnat) {
-    EXPECT_TRUE(is_in_tnat);
-  } else {
-    EXPECT_FALSE(is_in_tnat);
-  }
-}
-
-void
 check_entity_count(
   const rcl_node_t * node_ptr,
   std::string & topic_name,
--- a/rcl/test/rcl/test_graph.cpp
+++ b/rcl/test/rcl/test_graph.cpp
@@ -853,7 +853,7 @@ check_graph_state(
 }
 
 void
-check_entity_number(
+check_entity_count(
   const rcl_node_t * node_ptr,
   std::string & topic_name,
   size_t expected_publisher_count,
@@ -1256,7 +1256,7 @@ TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_graph_query_functio
   RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME, "Using topic name: %s", topic_name.c_str());
   rcl_ret_t ret;
   // First assert the "topic_name" is not in use.
-  check_entity_number(
+  check_entity_count(
     this->node_ptr,
     topic_name,
     0,    // expected publishers on topic
@@ -1271,7 +1271,7 @@ TEST_F(CLASSNAME(TestGraphFixture, RMW_IMPLEMENTATION), test_graph_query_functio
   EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
   rcl_reset_error();
   // Check the graph.
-  check_entity_number(
+  check_entity_count(
     this->node_ptr,
     topic_name,
     1,  // expected publishers on topic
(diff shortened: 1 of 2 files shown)


fujitatomoya:
@aprotyas your review will be appreciated. CC: @ivanpauno @clalancette

rcl/test/rcl/test_graph.cpp
```diff
@@ -852,6 +852,67 @@ check_graph_state(
   }
 }
 
+void
+check_entity_number(
```
aprotyas, line 856: Nit: Consider this renaming? Just to be very explicit about the point of the check.
```suggestion
check_entity_count(
```

rcl/test/rcl/test_graph.cpp
```diff
@@ -852,6 +852,67 @@ check_graph_state(
...
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+  } while (std::chrono::system_clock::now() - start_time < timeout);
+  EXPECT_EQ(expected_publisher_count, pub_count);
+  EXPECT_EQ(expected_subscriber_count, sub_count);
+
+  tnat = rcl_get_zero_initialized_names_and_types();
+  ret = rcl_get_topic_names_and_types(node_ptr, &allocator, false, &tnat);
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+  is_in_tnat = false;
+  for (size_t i = 0; i < tnat.names.size; ++i) {
```
aprotyas, line 898: We can leave it like how it was in `check_graph_state` for early termination? Or was there a specific reason you removed the `ret` check?
```suggestion
  for (size_t i = 0; RCL_RET_OK == ret && i < tnat.names.size; ++i) {
```

aprotyas (commented):
### On the changes

Left some minor comments inline.

Overall, the changes introduced look fine to me. Also, shouldn't the `check_graph_state` function be removed?

---

### On the need for changes

I think I understand the necessity behind the new test function, but I'm curious: is this a known flaky test? If not, does this happen only (and/or consistently) with GurumDDS/`rmw_gurumdds_cpp` - which is the DDS implementation reported in #934? 

Again, I understand that if entity deletion signals are not received soon enough following the invocation of entity fini methods, there might be sort of a race condition with the entity count returned by `rcl_count_[x]` being outdated. However, that wouldn't explain why this test hasn't failed with other DDS implementations.

I could be completely off in asking that question. If so, my bad - I don't have a deep understanding here.

YoungJin-gurum:
> ### On the need for changes
> I think I understand the necessity behind the new test function, but I'm curious: is this a known flaky test? If not, does this happen only (and/or consistently) with GurumDDS/`rmw_gurumdds_cpp` - which is the DDS implementation reported in #934?

GurumDDS does not always fail, depending on the test environment, it may or may not succeed.
 GurumDDS receives EDP over the network via socket. The local network is fast and the test sometimes succeeds by handling the EDP before calling the rcl_get_topic_names_and_types function.
 However, sometimes the test fails if a bottleneck on the network causes delays in sending and receiving EDP.

 In my opinion, the test checks how far this delay can be tolerated through timeout. If the EDP arrives later than the timeout, as intended by the test, it is correct to fail.
 However, the problem is that timeout is not used properly in the existing test code. It fails because it does not properly check the parts that can go beyond the allowable range.
 Other vendors also believe that testing is likely to fail in certain environments.

aprotyas:
> I think other vendors also could fail to test in certain environments.

👍 I won't disagree with this.

---

> The local network is fast and the test sometimes succeeds by handling the EDP before calling the `rcl_get_topic_names_and_types` function.

It's interesting that you bring up this function. I assumed the test failed in the assertion before the invocation of this method - check below. Is that correct?https://github.com/ros2/rcl/blob/fe67cfd9c003a1487e0ceea2a766e6f6d1940558/rcl/test/rcl/test_graph.cpp#L820-L831

---

>In my opinion, the test checks how far this delay can be tolerated through timeout.
If the EDP arrives later than the timeout, as intended by the test, it is correct to fail.
However, the problem is that timeout is not used properly in the existing test code.

Do you mind a small snippet/write-up of what you're envisioning as the use of the timeout? I understand the concern that the timeout seems to only add a tolerance bound for the delay, but how would the timeout be used properly?

---

> It fails because it does not properly check the parts that can go beyond the allowable range.

By the "allowable range", did you mean for time instances after the timeout? Or are you speaking of a different range? @fujitatomoya is there any design document that specifies ranges for these facilities?

YoungJin-gurum:
> It's interesting that you bring up this function. I assumed the test failed in the assertion before the invocation of this method - check below. Is that correct?

No, it's not. As I mentioned at issue, the point of failure of the test is to check expected_in_tnat.
ret from `rcl_wait_for_[x]` is  `RCL_RET_OK`.
Because `rcl_wait_for_[x]` calls `_rcl_wait_for_entities`, and it return `RCL_RET_OK`.
when expected_count is 0 and count is 1.
https://github.com/ros2/rcl/blob/7b899dae4a803575394588397c3e12e4be214abb/rcl/src/rcl/graph.c#L484-L494

The variable `count`(which is the result of count_entities_func) is set to 1
And it makes true for variable `is_in_tnat` after `rcl_get_topic_names_and_types`.
Even after the fini function is called.
https://github.com/ros2/rcl/blob/7b899dae4a803575394588397c3e12e4be214abb/rcl/test/rcl/test_graph.cpp#L833-L842

> Do you mind a small snippet/write-up of what you're envisioning as the use of the timeout? I understand the concern that the timeout seems to only add a tolerance bound for the delay, but how would the timeout be used properly?

> By the "allowable range", did you mean for time instances after the timeout? Or are you speaking of a different range? @fujitatomoya is there any design document that specifies ranges for these facilities?

What I said is 4 seconds, arguments of check_graph_state function.
https://github.com/ros2/rcl/blob/7b899dae4a803575394588397c3e12e4be214abb/rcl/test/rcl/test_graph.cpp#L1251-L1257
"allowable range" means within 4 seconds, as presented in the test. There is no other special meaning.
'The timeout be used properly', means if they are not received signal of entity deleted, have to wait until the timeout range is reached for graph_state change.
This is because the purpose of the test is to verify that it has been properly deleted.
Because the existing test code returns RCL_RET_OK before `count` becomes `0`, it does not wait for timeout.
I hope my answer solves your question.

fujitatomoya:
@aprotyas 

> but I'm curious: is this a known flaky test?

AFAIK, i do not think so.

> If not, does this happen only (and/or consistently) with GurumDDS

i think this just depends on implantation such as sync/async operation or cache/bypass graph update if in the process space or participant(if that is DDS). i think what important to test here is check if it is equal to the specified number of entities as expected. (NOT less/greater than specific number)

@YoungJin-gurum 

> If the EDP arrives later than the timeout, as intended by the test, it is correct to fail.

test should be agnostic from DDS. i think it expects that graph should get updated within timeout. if not, fails.

> However, the problem is that timeout is not used properly in the existing test code.

i am not clear on this, could you elaborate on this?

fujitatomoya:
> Also, shouldn't the check_graph_state function be removed?

i thought that too, but i would keep it for future test cases. (i do not have strong opinion here.)

rcl/test/rcl/test_graph.cpp
fujitatomoya: 👍  thanks! i will change it as you suggested.

rcl/test/rcl/test_graph.cpp
fujitatomoya: `ret` is checked a line before with `ASSERT_EQ`, this guarantees `ret` is `RCL_RET_OK`, otherwise test assertion fails and will never get to be this `for` loop.

aprotyas (approved):

aprotyas:
About the timeout itself, I don't think that should be the focus of the conversation. We're not testing in a loaded network or within a loaded process - as long as the graph is updated within a (generous) timeout, the test should pass.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15232/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9952/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12912/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15415/)

fujitatomoya:
@ivanpauno @clalancette minor test enhancement, could you review when you got time?

rcl/test/rcl/test_graph.cpp
```diff
@@ -852,6 +852,67 @@ check_graph_state(
   }
 }
 
+void
+check_entity_count(
```
clalancette, line 856: It looks to me like `check_entity_count` is completely replacing `check_graph_state`.  Given that `check_graph_state` isn't called anymore, we should remove it.

clalancette (changes requested):
One thing to fix inline, then I think this is OK to go.

rcl/test/rcl/test_graph.cpp
fujitatomoya: > It looks to me like check_entity_count is completely replacing check_graph_state

almost..., but they are not exactly the same. 'check_entity_count' checks if number of entity is equal to expected number. `check_graph_state` checks if number of entity is equal to or greater than expected number.
I am okay to delete `check_graph_state`, but probably there would be some use cases for `check_graph_state` in the future.

clalancette: Well, it's completely unused as of this PR.  So I'd say we either add a test that uses it, or we delete it.  If we delete it, it will still be in the git history so we can always pull it back out if we need it in the future.

fujitatomoya: sounds good to me!

fujitatomoya:
@clalancette your comments are addressed, could you take a look?

clalancette (approved):

PR #940, opened 2021-09-22
Increase Windows timeout 15 -> 25 ms

sloretz:
TOLERANCE may be too short for windows. @hidmic said the [Windows timeout was chosen because of the window's scheduler's time slice](https://github.com/ros2/rcl/pull/683#discussion_r441592518) , and and the timeout tolerance is currently 15 ms, but [this page on the microsoft documentation says it might be about 20ms](https://docs.microsoft.com/en-us/windows/win32/procthread/multitasking).

The goal is to fix the flaky test failures seen on the Windows repeated job: https://ci.ros2.org/job/nightly_win_rep/2375

```
rcl.WaitSetTestFixture__rmw_connextdds.finite_timeout
projectroot.test.test_wait__rmw_connextdds
rcl.WaitSetTestFixture__rmw_connextdds.canceled_timer
```

```
29: C:\ci\ws\src\ros2\rcl\rcl\test\rcl\test_wait.cpp(149): error: Expected: (diff) <= (timeout + ((15) * (1000LL * 1000LL))), actual: 25364600 vs 25000000
```

```
29: C:\ci\ws\src\ros2\rcl\rcl\test\rcl\test_wait.cpp(387): error: Expected: (diff) <= (((10) * (1000LL * 1000LL)) + ((15) * (1000LL * 1000LL))), actual: 25437100 vs 25000000
```

I increased the timeout by 10 ms, but the test failures this time were only over by 0.4ish ms, so maybe this PR is increasing it more than necessary. Since I'm uncertain about what the window's time scheduler actually is, a larger increase makes sense to me.

### DIFF
--- a/rcl/test/rcl/test_wait.cpp
+++ b/rcl/test/rcl/test_wait.cpp
@@ -42,7 +42,7 @@
 #ifndef _WIN32
 #define TOLERANCE RCL_MS_TO_NS(6)
 #else
-#define TOLERANCE RCL_MS_TO_NS(15)
+#define TOLERANCE RCL_MS_TO_NS(25)
 #endif
 
 class CLASSNAME (WaitSetTestFixture, RMW_IMPLEMENTATION) : public ::testing::Test


rcl/test/rcl/test_wait.cpp
```diff
@@ -42,7 +42,7 @@
 #ifndef _WIN32
 #define TOLERANCE RCL_MS_TO_NS(6)
 #else
-#define TOLERANCE RCL_MS_TO_NS(15)
+#define TOLERANCE RCL_MS_TO_NS(25)
```
clalancette, line 45: Honestly, I wonder if we should just get rid of this TOLERANCE and just switch any place that uses it to `EXPECT_GT`.  Since we aren't dealing with real-time OSs or threads here, there is no guarantee that any of these tests come back in a reasonable amount of time; the OS could choose not to schedule this process for long periods of time.

We'll lose checking if the timeout is similar to what the user asked for, but I think that is relatively low risk.

hidmic: That's fair, and I agree that time-sensitive tests are inherently flaky for the OSes we support, but I'll point out that validating that timeout is the sole purpose of several of these tests. We may as well drop them if we get rid of this TOLERANCE.

hidmic (commented):
Barring the high-level discussion about time-sensitive tests, this makes sense to me. Pending green CI.

rcl/test/rcl/test_wait.cpp
clalancette: I don't feel super strongly about it.  If updating the tolerance to 25ms here makes the tests less flaky, that's at least a step in the right direction.

sloretz:
CI (build: ` --packages-up-to rcl` test: `--packages-select rcl`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15298/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10013/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12979/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15490/)

hidmic (approved):

PR #943, opened 2021-10-06
Only change the default logger level if default_logger_level is set

JafarAbdi:
Required for https://github.com/ros2/rcutils/pull/347 to set the default logger value from the config file, otherwise, https://github.com/ros2/rcutils/blob/master/src/logging.c#L378 will set the default logger value back to RCUTILS_DEFAULT_LOGGER_DEFAULT_LEVEL after changing it in the linked PR

### DIFF
--- a/rcl/src/rcl/logging.c
+++ b/rcl/src/rcl/logging.c
@@ -74,8 +74,10 @@ rcl_logging_configure_with_output_handler(
   g_rcl_logging_num_out_handlers = 0;
 
   if (log_levels) {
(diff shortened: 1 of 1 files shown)


aprotyas (approved):

aprotyas:
Not sure I follow what happened in the `rclpy` tests, but at any rate those tests don't seem related. Running windows CI again regardless:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15588/)

PR #951, opened 2021-11-05
suppress false positive from clang-tidy

wjwwood:
I believe this issue identified from clang-tidy is a false positive, but we should have more eyes on it.

It's similar to, but perhaps not the same, issue as in this clang-tidy bug: https://bugs.llvm.org/show_bug.cgi?id=41311

Either way, looking at the code I believe it to be safe, and clang-tidy simply cannot tell that the args are always initialized elsewhere. Again, more eyes would be good here.

This pull request suppresses these warnings with this in mind.

closes https://github.com/space-ros/space-ros/issues/4

### DIFF
--- a/rcl/src/rcl/logging.c
+++ b/rcl/src/rcl/logging.c
@@ -165,8 +165,11 @@ static
 void
 rcl_logging_ext_lib_output_handler(
   const rcutils_log_location_t * location,
-  int severity, const char * name, rcutils_time_point_value_t timestamp,
-  const char * format, va_list * args)
+  int severity,
+  const char * name,
+  rcutils_time_point_value_t timestamp,
+  const char * format,
+  va_list * args)
 {
   rcl_ret_t status;
   char msg_buf[1024] = "";
@@ -188,7 +191,9 @@ rcl_logging_ext_lib_output_handler(
   };
 
   va_list args_clone;
-  va_copy(args_clone, *args);
+  // The args are initialized, but clang-tidy cannot tell.
+  // It may be related to this bug: https://bugs.llvm.org/show_bug.cgi?id=41311
+  va_copy(args_clone, *args);  // NOLINT(clang-analyzer-valist.Uninitialized)
   status = rcutils_char_array_vsprintf(&msg_array, format, args_clone);
   va_end(args_clone);
 
--- a/rcl/src/rcl/logging_rosout.c
+++ b/rcl/src/rcl/logging_rosout.c
@@ -231,8 +231,11 @@ rcl_ret_t rcl_logging_rosout_fini_publisher_for_node(
 
 void rcl_logging_rosout_output_handler(
   const rcutils_log_location_t * location,
-  int severity, const char * name, rcutils_time_point_value_t timestamp,
-  const char * format, va_list * args)
+  int severity,
+  const char * name,
+  rcutils_time_point_value_t timestamp,
+  const char * format,
+  va_list * args)
 {
   rosout_map_entry_t entry;
   rcl_ret_t status = RCL_RET_OK;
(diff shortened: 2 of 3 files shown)


rcl/src/rcl/logging.c
```diff
@@ -188,7 +191,9 @@ rcl_logging_ext_lib_output_handler(
   };
 
   va_list args_clone;
-  va_copy(args_clone, *args);
+  // The args are initialized, but clang-tidy cannot tell.
+  // It may be related to this bug: https://bugs.llvm.org/show_bug.cgi?id=41311
+  va_copy(args_clone, *args);  // NOLINT(clang-analyzer-valist.Uninitialized)
```
jacobperron, line 196: It's not obvious to me that `args` is initialized. IIUC, they come from this public function and there's no guard for a null value:

https://github.com/ros2/rcl/blob/3a07bef352ff82c2502b1f0502930041d756c064/rcl/src/rcl/logging.c#L155

I think we're primarily relying on `rcutils` to call the function [here](https://github.com/ros2/rcutils/blob/67095577eaf39977f30f7db1ae92f7b0b2ec48d0/src/logging.c#L558) (and `args` is not null), but there's nothing stopping another user from calling `rcl_logging_multiple_output_handler` with a null value.

rcl/src/rcl/logging_rosout.c
```diff
@@ -251,7 +254,9 @@ void rcl_logging_rosout_output_handler(
     };
 
     va_list args_clone;
-    va_copy(args_clone, *args);
+    // The args are initialized, but clang-tidy cannot tell.
+    // It may be related to this bug: https://bugs.llvm.org/show_bug.cgi?id=41311
+    va_copy(args_clone, *args);  // NOLINT(clang-analyzer-valist.Uninitialized)
```
jacobperron, line 259: Same here. How do I know `args` is not null? I guess we should add a guard to avoid accessing null.

rcl/src/rcl/logging.c
wjwwood: True, but the place that this function is being used, we ensure the args are not-null and initialized (different things).

We can check for null, but that doesn't guard against uninitialized, which would be something like this:

```c
va_list args;  // uninitialized
rcl_logging_multiple_output_handler(..., &args);
```

rcl/src/rcl/logging_rosout.c
wjwwood: Same thing, I'll check for null, but we cannot check for initialized unfortunately, we can only document that it needs to be initialized.

rcl/src/rcl/logging.c
jacobperron: Understood. I agree that `args` is initialized the way we are calling this function.

jacobperron (approved):

rcl/src/rcl/logging.c
wjwwood: Yeah, so for this function, there's nothing useful for me to do if `args` is `nullptr` since it returns `void` and there's no access to any error handler here. This function is local the `.c` file so I think we just ensure it's always called with non-nullptr. If it were C++ I would say we should use a reference, but because it's not I'd say we just document it and leave it as is.

rcl/src/rcl/logging_rosout.c
wjwwood: Again, this one can't do anything useful if it is nullptr, so I'd say that we just document how it is supposed to be used: 26059dd

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16327/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10947/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16699/)

fujitatomoya (approved):

fujitatomoya:
CI is failing with the same problem issued here, https://github.com/cpplint/cpplint/issues/184

wjwwood:
Yeah, I think it's addressed for us by https://github.com/ament/ament_lint/pull/339. I'll re-run ci:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16384/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10993/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16756/)

clalancette:

wjwwood:
The Rpr job will continue to fail until `ament_cpplint` is released.

wjwwood:
The Windows test failures are unrelated.

PR #955, opened 2021-12-08
Make rcl_difference_times args const

sloretz:
This allows passing `const rcl_time_point_t *` variables into this function.

### DIFF
--- a/rcl/include/rcl/time.h
+++ b/rcl/include/rcl/time.h
@@ -473,7 +473,7 @@ RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_ret_t
 rcl_difference_times(
-  rcl_time_point_t * start, rcl_time_point_t * finish, rcl_duration_t * delta);
+  const rcl_time_point_t * start, const rcl_time_point_t * finish, rcl_duration_t * delta);
 
 /// Fill the time point value with the current value of the associated clock.
 /**
--- a/rcl/src/rcl/time.c
+++ b/rcl/src/rcl/time.c
@@ -231,7 +231,7 @@ rcl_system_clock_fini(
 
 rcl_ret_t
 rcl_difference_times(
-  rcl_time_point_t * start, rcl_time_point_t * finish, rcl_duration_t * delta)
+  const rcl_time_point_t * start, const rcl_time_point_t * finish, rcl_duration_t * delta)
 {
   RCL_CHECK_ARGUMENT_FOR_NULL(start, RCL_RET_INVALID_ARGUMENT);
   RCL_CHECK_ARGUMENT_FOR_NULL(finish, RCL_RET_INVALID_ARGUMENT);


sloretz:
CI (build: `--packages-up-to rcl` test: `--packages-select rcl`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15742/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10424/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13407/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15989/)

fujitatomoya (approved):

j-rivero:
Changes looks good to me. They are going to break API/ABI for the consumers of rcl. Would make sense to run a wider CI run to be able to detect most of the problems?

sloretz:
> They are going to break API/ABI for the consumers of rcl. Would make sense to run a wider CI run to be able to detect most of the problems?

I think the API and ABI aren't broken since this is C and going from non-`const` to `const`. Here's one linux job building everything above `rcl` to make sure: [Build Status](https://ci.ros2.org/job/ci_linux/15747/)

Here are a few things I played with

<details><summary>If rcl had a C++ API then const-ness would matter here</summary>

```c++
#include <iostream>

bool
func(int * abc)
{
  std::cout << "non-const Func " << *abc << "\n";
  return true;
}

bool
func(const int * abc)
{
  std::cout << "Const Func " << *abc << "\n";
  return true;
}

int main()
{
  std::cout << "Hello world\n";
  int foobar = 42;
  const int foobar2 = 64;
  func(&foobar);
  func(&foobar2);
  return 0;
};
```

```console
$ g++ asdf.cpp -o const_test && ./const_test
Hello world
non-const Func 42
Const Func 64
```

The `const`-ness of the arguments is included in the ABI

```console
nm const_test | grep func
00000000000009fe t _GLOBAL__sub_I__Z4funcPi
00000000000008ba T _Z4funcPi
0000000000000902 T _Z4funcPKi
```

</details>


<details><summary>Since rcl has a C API, const-ness doesn't matter here</summary>

```c
#include <stdio.h>

int
func(int * abc)
{
  printf("non-const Func %d\n", *abc);
  return 64;
}

// int
// func(const int * abc)
// {
//   printf("const Func %d\n", *abc);
//   return 128;
// }

int main()
{
  printf("Hello C world\n");
  int foobar = 42;
  func(&foobar);
  return 0;
};
```

```console
$ gcc asdf.c -o c_const && ./c_const
Hello C world
const Func 42
```

```console
$ gcc asdf.c -o c_non_const && ./c_non_const
Hello C world
non-const Func 42
```

The `const`-ness of the arguments is not part of the ABI. That's why one function has to be commented out.

```console
$ nm c_const | grep func
00000000000006fa T func
$ nm c_non_const | grep func
00000000000006fa T func
```

</details>

clalancette (approved):

j-rivero:
> Here are a few things I played with

wow, I did know about it, thanks so much for the time and experiments.

PR #959, opened 2022-01-20
Install includes to include/${PROJECT_NAME}

sloretz:
Part of ros2/ros2#1150 - This eliminates issues with the include directory search order when overriding these packages by installing their includes to `include/${PROJECT_NAME}`

### DIFF
--- a/rcl/CMakeLists.txt
+++ b/rcl/CMakeLists.txt
@@ -69,7 +69,7 @@ set(${PROJECT_NAME}_sources
 add_library(${PROJECT_NAME} ${${PROJECT_NAME}_sources})
 target_include_directories(${PROJECT_NAME} PUBLIC
   "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
-  "$<INSTALL_INTERFACE:include>")
+  "$<INSTALL_INTERFACE:include/${PROJECT_NAME}>")
 # specific order: dependents before dependencies
 ament_target_dependencies(${PROJECT_NAME}
   "rcl_interfaces"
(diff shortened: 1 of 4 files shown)


sloretz:
CI (build: `--packages-up-to rclpy rclcpp rclcpp_action rclcpp_lifecycle` test: `--packages-select  rclpy rclcpp rclcpp_action rclcpp_lifecycle rcl_lifecycle rcl rcl_action rcl_yaml_param_parser`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15975/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10646/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16297/)

clalancette (approved):
Looks good with green CI.

pablogs9:
Hello @clalancette and @sloretz, does this change mean that from now on we will need to include rcl headers using `#include <rcl/rcl/allocator.h>` instead of `#include <rcl/allocator.h>`?

clalancette:
> Hello @clalancette and @sloretz, does this change mean that from now on we will need to include rcl headers using `#include <rcl/rcl/allocator.h>` instead of `#include <rcl/allocator.h>`?

No, the CMake changes take care of all of that.  All downstream consumers can still use `#include <rcl/allocator.h>`.

pablogs9:
At least in micro-ROS, we are having problems with this patch. It only happens in rolling and it complains about not having `#include <rcl/rcl.h>`. 

I found that it is solved using `#include <rcl/rcl/rcl.h>` but as it implies changes along a huge amount of packages I have to revert the patch in our micro-ROS fork.

Any idea on why we need to use those kinds of includes when we are not using the merge install option?

clalancette:
> Any idea on why we need to use those kinds of includes when we are not using the merge install option?

Not off-hand.  Can you open a new issue about this, giving instructions on how to reproduce?  We can then take a look at it.

russkel:
I agree with @pablogs9 . As I am trying to integrate various bits and pieces into non-cmake projects, I have to write some additional build logic to check for nested directories like `install/rcl/include/rcl/rcl`. This doesn't seem right. I am surprised this was the suggested solution.

clalancette:
> This doesn't seem right. I am surprised this was the accepted solution.

We agree as well, this shouldn't be necessary.  The existing logic should take care of it.  Can you please open an issue with details about what problems you are running into?

russkel:
Sure.

PR #960, opened 2022-02-14
Remove fastrtps customization on test_events

MiguelCompany:
This PR removes the custom code path on test_events.cpp that checks certain events return `RCL_RET_UNSUPPORTED` on rmw_fastrtps.


clalancette:
CI for this is in https://github.com/ros2/rmw_fastrtps/pull/583#issuecomment-1045169781.

PR #965, opened 2022-03-09
Micro-optimizations in rcl

clalancette:
While doing some profiling of the rcl code, I noticed a few different places where we were unnecessarily doing work. This PR has three commits:

1. The first commit gets rid of duplicate checking of whether `wait_set` is NULL, since we just did that in the previous line.  Here I've chosen to remove the duplication by not calling `rcl_wait_set_is_valid` and just checking the `impl` pointer directly.  Another way to fix this is to remove the `RCL_CHECK_ARGUMENT_FOR_NULL` call on the previous line, and rely on `rcl_wait_set_is_valid` for validating both pointers.  I don't have a strong opinion about which way to go.
2. The second commit refactors a couple of the `RCUTILS_DEBUG` statements to try to reduce their overhead on the hot path.
3. The third commit somewhat controversially removes the `RCUTILS_DEBUG` statements from within `rcl_wait` completely.  Since this is a hot-path, the overhead of continually checking whether a debug statement should be output or not actually shows up in profiles.  Also, my feeling is that in general, these debug statements are not useful; if you are actually debugging something here, you need to use something more targeted.

With these statements removed, and in combination with https://github.com/ros2/rclcpp/pull/1896 , I saw about a 20% reduction in overhead in the application I was benchmarking.

### DIFF
--- a/rcl/src/rcl/wait.c
+++ b/rcl/src/rcl/wait.c
@@ -217,7 +217,7 @@ rcl_wait_set_get_allocator(const rcl_wait_set_t * wait_set, rcl_allocator_t * al
 
 #define SET_ADD(Type) \
   RCL_CHECK_ARGUMENT_FOR_NULL(wait_set, RCL_RET_INVALID_ARGUMENT); \
-  if (!rcl_wait_set_is_valid(wait_set)) { \
+  if (!wait_set->impl) { \
     RCL_SET_ERROR_MSG("wait set is invalid"); \
     return RCL_RET_WAIT_SET_INVALID; \
   } \
--- a/rcl/src/rcl/wait.c
+++ b/rcl/src/rcl/wait.c
@@ -586,6 +586,10 @@ rcl_wait(rcl_wait_set_t * wait_set, int64_t timeout)
     temporary_timeout_storage.sec = 0;
     temporary_timeout_storage.nsec = 0;
     timeout_argument = &temporary_timeout_storage;
+    RCUTILS_LOG_DEBUG_NAMED(
+      ROS_PACKAGE_NAME,
+      "Waiting with timeout: %" PRIu64 "s + %" PRIu64 "ns, based on next scheduled timer: %d",
+      temporary_timeout_storage.sec, temporary_timeout_storage.nsec, is_timer_timeout);
   } else if (timeout > 0 || is_timer_timeout) {
     // If min_timeout was negative, we need to wake up immediately.
     if (min_timeout < 0) {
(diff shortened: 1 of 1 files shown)


fujitatomoya (approved):
this looks good to me.

btw, do we have ability to disable logging macros into complete NOP with specific build configuration? if so, we can keep the debug / info messages, and also turn them into complete NOP with that compile option for the performance.

clalancette:
> btw, do we have ability to disable logging macros into complete NOP with specific build configuration? if so, we can keep the debug / info messages, and also turn them into complete NOP with that compile option for the performance.

We do have that ability, by setting `RCUTILS_LOG_MIN_SEVERITY` to a level higher than debug, which turns these into a noop.  The problem is that you have to do that at compile time, which means that in order to take advantage of it a user would have to recompile `rcl`.  While that is certainly possible to do for users who want the absolute best performance, it also is a step that I think most users aren't going to take.  I'd still rather remove debug statements from hot paths.

fujitatomoya:
> While that is certainly possible to do for users who want the absolute best performance, it also is a step that I think most users aren't going to take. I'd still rather remove debug statements from hot paths.

that is true. we have our own CI system to build the target distribution from the source code. so that was not really under my radar. thanks for the comment!

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16307/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10927/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16675/)

clalancette:
There are two additional failures in Windows that I don't see in the nightlies.

The first is test_publisher_subscriber__rclcpp__rclpy__rmw_fastrtps_cpp, and the output shows:

```
test_processes_finished_gracefully[Defaults] (test_communication.TestPublisherSubscriberAfterShutdown)
Test that both executables finished gracefully. ... 
Starting test run test_communication.test_publisher_subscriber__rclcpp__rclpy__rmw_fastrtps_cpp_Release.launch_tests[Empty]
[INFO] [launch]: All log files can be found below C:\Users\ContainerAdministrator\.ros\log\2022-03-10-00-09-18-399086-69e0b0403c7a-10872
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [daemon-stop-22]: process started with pid [8796]
[INFO] [daemon-stop-22]: process has finished cleanly [pid 8796]
FAIL

======================================================================
FAIL: test_processes_finished_gracefully[Defaults] (test_communication.TestPublisherSubscriberAfterShutdown)
Test that both executables finished gracefully.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\ci\ws\build\test_communication\test_publisher_subscriber__rclcpp__rclpy__rmw_fastrtps_cpp_Release.py", line 77, in test_processes_finished_gracefully
    launch_testing.asserts.assertExitCodes(
  File "C:\ci\ws\install\Lib\site-packages\launch_testing\asserts\assert_exit_codes.py", line 62, in assertExitCodes
    assert info.returncode in allowable_exit_codes, 'Proc {} exited with code {}'.format(
AssertionError: Proc test_subscriber-21 exited with code 3221225477
```

The second is test_launch_ros.test.test_launch_ros.actions.test_load_composable_nodes.test_load_node_with_param_file, and the output shows:

```
test\test_launch_ros\actions\test_load_composable_nodes.py:350: in test_load_node_with_param_file
    assert request.node_name == 'node_1'
E   AssertionError: assert 'node_2' == 'node_1'
E     - node_1
E     ?      ^
E     + node_2
E     ?      ^
```

I don't see how either of them could be related to this PR, so I'm going to go ahead and merge this.  Thanks @fujitatomoya !

PR #966, opened 2022-03-14
Add rcl guard condition on_trigger_callback

mauropasse:
This PR adds the `on_trigger_callback` callback to rcl guard conditions.

If the `on_trigger_callback` is set to a guard condition, it will be called every time the guard condition is triggered. 
If not, a counter will count the times the guard condiiton was triggered, and if eventually a callback is set it will use the counter as argument.

This PR is aligned to the recently added listener callbacks to the RMW entities ([subscriptions](https://github.com/ros2/rcl/blob/master/rcl/src/rcl/subscription.c#L449), [clients](rcl_client_set_on_new_response_callback), [services](https://github.com/ros2/rcl/blob/master/rcl/src/rcl/service.c#L350)).

The motivation behind this PR, be able to wake up events driven executors (not relying on a wait set) when guard conditions are triggered. For example, a timer is reset after being cancelled, so it's inner guard condition is triggered and we can take actions to update the next trigger time for the timer.

Signed-off-by: Mauro Passerino <mpasserino@irobot.com>

### DIFF
--- a/rcl/include/rcl/guard_condition.h
+++ b/rcl/include/rcl/guard_condition.h
@@ -24,6 +24,7 @@ extern "C"
 
 #include "rcl/allocator.h"
 #include "rcl/context.h"
+#include "rcl/event_callback.h"
 #include "rcl/macros.h"
 #include "rcl/types.h"
 #include "rcl/visibility_control.h"
@@ -47,6 +48,14 @@ typedef struct rcl_guard_condition_options_s
   /// Custom allocator for the guard condition, used for internal allocations.
   rcl_allocator_t allocator;
 } rcl_guard_condition_options_t;
+
+/// On trigger callback data
+typedef struct rcl_guard_condition_callback_data_s
+{
+  rcl_event_callback_t on_trigger_callback;
+  const void * user_data;
+  size_t trigger_count;
+} rcl_guard_condition_callback_data_t;
 
 /// Return a rcl_guard_condition_t struct with members set to `NULL`.
 RCL_PUBLIC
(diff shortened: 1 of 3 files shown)


rcl/include/rcl/guard_condition.h
```diff
@@ -48,6 +49,14 @@ typedef struct rcl_guard_condition_options_s
   rcl_allocator_t allocator;
 } rcl_guard_condition_options_t;
 
+/// On trigger callback data
+typedef struct rcl_guard_condition_callback_data_s
+{
+  rcl_event_callback_t on_trigger_callback;
+  const void * user_data;
+  size_t trigger_count;
```
fujitatomoya, line 57: @mauropasse implementation looks good, but could you explain the use case of `trigger_count`?

rcl/include/rcl/guard_condition.h
```diff
@@ -262,6 +271,34 @@ RCL_WARN_UNUSED
 rmw_guard_condition_t *
 rcl_guard_condition_get_rmw_handle(const rcl_guard_condition_t * guard_condition);
 
+/// Set the on trigger callback function for the guard_condition.
+/**
+ * This API sets the callback function to be called whenever the
+ * guard_condition is triggered.
+ *
```
fujitatomoya, line 278: I think it would be nice to add explanation about how `trigger_count` works internally.

fujitatomoya:
I think i should have looked at https://github.com/ros2/rcl/pull/995 1st.

mauropasse:
Hi @fujitatomoya, sorry I should have closed this PR when opened https://github.com/ros2/rcl/pull/995.

The goal was having a way to generate an event when a timer was reset. Since a guard condition is triggered when a timer is reset, the first idea was adding a callback directly to a guard condition (`on_trigger_callback`). But now in the new PR, the callback belongs directly to the timer (`on_reset_callback`). 

The last case in which I'd still like a guard condition to have a callback, is the situation where a new entity (pub/sub/etc) is created from a node. When this happens, the node's guard condition is triggered. If we have a callback there, we can rebuild the waitset looking only for new entities added/removed from the node, instead of iterating through all the nodes of the system (which is quite expensive). But I guess we could have there an `on_updated_node` callback or something like that.

PR #969, opened 2022-03-15
Fix missing terminating \0 in rcl_context_impl_t.argv

yushijinhun:


rcl/src/rcl/init.c
```diff
@@ -113,7 +113,7 @@ rcl_init(
       fail_ret = RCL_RET_BAD_ALLOC; goto fail);
     int64_t i;
     for (i = 0; i < argc; ++i) {
-      size_t argv_i_length = strlen(argv[i]);
+      size_t argv_i_length = strlen(argv[i]) + 1;
```
clalancette, line 116: I agree that this is a bug, but this actually won't fix it properly.

The problem is that this allocates the extra byte for the `\0`, but it never puts the `\0` into place.  We may get lucky and have the allocated memory be `\0` by default, but we can't depend on that.

To fix that part of it, we need to explicitly set the last character to `\0`.  I'll suggest adding a line right before the `memcpy` below that looks like:

```
context->impl->argv[i][argv_i_length - 1] = '\0';
```

(we could also change the `memcpy` line to copy one additional byte, but I think the above is more explicit)

clalancette (changes requested):

rcl/src/rcl/init.c
yushijinhun: `memcpy` will also copy the extra `\0` byte of the original string, so this is not a problem.

clalancette: Oh, you are right,  I totally got that confused.  Sorry for the noise!

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16322/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10942/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16692/)

fujitatomoya (approved):

clalancette:
All of the failures on CI are well-known.  With that, and approval, I'm going to merge.  Thanks @yushijinhun !

Issue #983, opened 2022-05-13
RCL overwriting error state after rmw_create_node


allsey87:
## Bug report

**Required Info:**

- Operating System: WebAssembly
- Installation type: Source
- Version or commit hash: Galactic
- DDS implementation: Custom
- Client library (if applicable): rclcpp

#### Expected behavior
RCL should fail gracefully when `rmw_create_node` returns `NULL` and sets the error state via `RMW_SET_ERROR_MSG`

#### Actual behavior
RCL ignores the fact the `rmw_create_node` set an error and returned `NULL` and then proceeds to overwrite the error state as follows:
```
failed to initialize rcl node: rcl node's rmw handle is invalid, at /home/developer/workspace/src/ros2/rcl/rcl/src/rcl/node.c:414
>>> [rcutils|error_handling.c:108] rcutils_set_error_state()
This error state is being overwritten:
'RMW_ERROR_MESSAGE, at RMW_FILENAME:0, at /home/developer/workspace/src/ros2/rcl/rcl/src/rcl/node.c:262'
with this new error message:
'rcl node's rmw handle is invalid, at /home/developer/workspace/src/ros2/rcl/rcl/src/rcl/node.c:414'
rcutils_reset_error() should be called after error handling to avoid this.
<<<

[ERROR] [0000000003.694284999] [rcl]: Failed to fini publisher for node: 1
```

fujitatomoya:
i think this is bug.

https://github.com/ros2/rcl/blob/493ebf3a5d8fe2d861a93006a7d6908fd4e592d9/rcl/src/rcl/node.c#L251-L261

`logger_name` is assigned before `rmw_create_node` above. and if `rmw_create_node` fails, it goes to

https://github.com/ros2/rcl/blob/493ebf3a5d8fe2d861a93006a7d6908fd4e592d9/rcl/src/rcl/node.c#L305-L311

and call `rcl_logging_rosout_fini_publisher_for_node` w/o calling `rcl_logging_rosout_init_publisher_for_node`.

this leads to call `rcl_node_is_valid_except_context` to overwrite the error message.

fujitatomoya:
https://github.com/ros2/rcl/pull/985 can fix the problem, but according to https://github.com/ros2/rcl/pull/985#pullrequestreview-975525676, it would be nice to refactor the `rcl_node_init` and `rcl_node_fini` process to fix the complication and possible problem in the future.

PR #985, opened 2022-05-16
node logger_name should be assigned after rmw_create_node.

fujitatomoya:
address https://github.com/ros2/rcl/issues/983


clalancette (commented):
While this addresses the immediate concern, I think this bug report points out a deeper problem; `rcl_node_init` is far too complicated as-is.  I think the right thing to do here is to refactor this function completely so that it calls out to functions to do most of the work for it, which will make it much easier to follow.  Also, a lot of the work that is done in the `fail` section could be shared with `rcl_node_fini`.

@fujitatomoya I won't outright block on this, but I think this whole thing will be much easier to reason about and maintain if we do that refactoring.  What do you think?

Issue #986, opened 2022-05-19
Do not require the node argument in publisher and subscription fini functions


nnmm:
## Feature request
Do not require the node argument in publisher and subscription fini functions.

#### Feature description
Passing in the node to `rcl_subscription_fini` and `rcl_publisher_fini` allows passing in a node that's not the same node as the one used in the init function.

The publisher and subscription could instead store the node pointer given to the init function and use it later in the fini function. This would also make the `rclrs` client library slightly more elegant.

#### Implementation considerations
I assume changing the signatures of the fini functions would cause too much breakage (although I'm not really sure about that, there probably aren't many users of the `rcl` API in the wild). So we could add a `rcl_subscription_fini_2`/`rcl_publisher_fini_2` function.

nnmm:
cc @sloretz, you mentioned this yesterday. I can work on it, but can you remind me of what you think regarding changing signatures?

wjwwood:
The reason for having the node argument passed via this fini function is to make it clear (by contract) that the node needs to be finalized after the subscription/publisher.

I would not remove this argument, but I do support storing and checking that the correct node was used in fini.

fujitatomoya:
This issue has no activity more than 6 months, i will go ahead to close this.

PR #987, opened 2022-05-20
Store the node pointer in publisher and subscription

nnmm:
The `rcl_publisher_fini()` and `rcl_subscription_fini()` function now check that the correct node was given to it.

Part 1 of #986.

Signed-off-by: Nikolai Morin <nikolai.morin@apex.ai>

### DIFF
--- a/rcl/include/rcl/publisher.h
+++ b/rcl/include/rcl/publisher.h
@@ -179,7 +179,8 @@ rcl_publisher_init(
  * \return #RCL_RET_OK if publisher was finalized successfully, or
  * \return #RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
  * \return #RCL_RET_PUBLISHER_INVALID if the publisher is invalid, or
- * \return #RCL_RET_NODE_INVALID if the node is not the node used to create the publisher, or
+ * \return #RCL_RET_NODE_INVALID if the node is invalid, or
+ * \return #RCL_RET_INCORRECT_NODE if the node is not the node used to create the publisher, or
  * \return #RCL_RET_ERROR if an unspecified error occurs.
  */
 RCL_PUBLIC
--- a/rcl/include/rcl/publisher.h
+++ b/rcl/include/rcl/publisher.h
@@ -179,7 +179,7 @@ rcl_publisher_init(
  * \return #RCL_RET_OK if publisher was finalized successfully, or
  * \return #RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
  * \return #RCL_RET_PUBLISHER_INVALID if the publisher is invalid, or
- * \return #RCL_RET_NODE_INVALID if the node is invalid, or
+ * \return #RCL_RET_NODE_INVALID if the node is not the node used to create the publisher, or
  * \return #RCL_RET_ERROR if an unspecified error occurs.
  */
 RCL_PUBLIC
--- a/rcl/include/rcl/subscription.h
+++ b/rcl/include/rcl/subscription.h
@@ -175,7 +175,7 @@ rcl_subscription_init(
  *
  * After calling, calls to rcl_wait and rcl_take will fail when using this
  * subscription.
- * Additioanlly rcl_wait will be interrupted if currently blocking.
+ * Additionally, rcl_wait will be interrupted if currently blocking.
  * However, the given node handle is still valid.
  *
  * <hr>
@@ -191,7 +191,7 @@ rcl_subscription_init(
  * \return #RCL_RET_OK if subscription was deinitialized successfully, or
  * \return #RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
  * \return #RCL_RET_SUBSCRIPTION_INVALID if the subscription is invalid, or
- * \return #RCL_RET_NODE_INVALID if the node is invalid, or
+ * \return #RCL_RET_NODE_INVALID if the node is not the node used to create the subscription, or
  * \return #RCL_RET_ERROR if an unspecified error occurs.
  */
 RCL_PUBLIC
--- a/rcl/include/rcl/subscription.h
+++ b/rcl/include/rcl/subscription.h
@@ -191,7 +191,8 @@ rcl_subscription_init(
  * \return #RCL_RET_OK if subscription was deinitialized successfully, or
  * \return #RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
  * \return #RCL_RET_SUBSCRIPTION_INVALID if the subscription is invalid, or
- * \return #RCL_RET_NODE_INVALID if the node is not the node used to create the subscription, or
+ * \return #RCL_RET_NODE_INVALID if the node is invalid, or
+ * \return #RCL_RET_INCORRECT_NODE if the node is not the node used to create the subscription, or
  * \return #RCL_RET_ERROR if an unspecified error occurs.
  */
 RCL_PUBLIC
--- a/rcl/src/rcl/publisher.c
+++ b/rcl/src/rcl/publisher.c
@@ -180,7 +180,7 @@ rcl_publisher_fini(rcl_publisher_t * publisher, rcl_node_t * node)
   if (publisher->impl) {
     if (node != publisher->impl->node) {
       RCL_SET_ERROR_MSG("fini called with incorrect node");
-      return RCL_RET_INVALID_ARGUMENT;
+      return RCL_RET_INCORRECT_NODE;
     }
     rcl_allocator_t allocator = publisher->impl->options.allocator;
     rmw_node_t * rmw_node = rcl_node_get_rmw_handle(node);
--- a/rcl/src/rcl/publisher.c
+++ b/rcl/src/rcl/publisher.c
@@ -105,6 +105,8 @@ rcl_publisher_init(
     publisher->impl, "allocating memory failed", ret = RCL_RET_BAD_ALLOC; goto cleanup);
 
   // Fill out implementation struct.
+  // node
+  publisher->impl->node = node;
   // rmw handle (create rmw publisher)
   // TODO(wjwwood): pass along the allocator to rmw when it supports it
   publisher->impl->rmw_handle = rmw_create_publisher(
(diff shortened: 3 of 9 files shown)


rcl/include/rcl/publisher.h
```diff
@@ -179,7 +179,7 @@ rcl_publisher_init(
  * \return #RCL_RET_OK if publisher was finalized successfully, or
  * \return #RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
  * \return #RCL_RET_PUBLISHER_INVALID if the publisher is invalid, or
- * \return #RCL_RET_NODE_INVALID if the node is invalid, or
+ * \return #RCL_RET_NODE_INVALID if the node is not the node used to create the publisher, or
```
fujitatomoya, line 182: IMO, `invalid` can include `not the node used to create the publisher` case. and there is a case that generally `node` is just invalid verified on `rcl_node_is_valid_except_context`? so probably current doc is good enough?

rcl/include/rcl/subscription.h
```diff
@@ -191,7 +191,7 @@ rcl_subscription_init(
  * \return #RCL_RET_OK if subscription was deinitialized successfully, or
  * \return #RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
  * \return #RCL_RET_SUBSCRIPTION_INVALID if the subscription is invalid, or
- * \return #RCL_RET_NODE_INVALID if the node is invalid, or
+ * \return #RCL_RET_NODE_INVALID if the node is not the node used to create the subscription, or
```
fujitatomoya, line 194: same comment with `rcl_publisher_fini`.

rcl/src/rcl/publisher.c
```diff
@@ -176,6 +178,10 @@ rcl_publisher_fini(rcl_publisher_t * publisher, rcl_node_t * node)
 
   RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME, "Finalizing publisher");
   if (publisher->impl) {
+    if (node != publisher->impl->node) {
+      RCL_SET_ERROR_MSG("fini called with incorrect node");
+      return RCL_RET_INVALID_ARGUMENT;
```
fujitatomoya, line 183:
```suggestion
      return RCL_RET_NODE_INVALID;
```

rcl/src/rcl/subscription.c
```diff
@@ -172,6 +174,10 @@ rcl_subscription_fini(rcl_subscription_t * subscription, rcl_node_t * node)
     return RCL_RET_NODE_INVALID;  // error already set
   }
   if (subscription->impl) {
+    if (node != subscription->impl->node) {
+      RCL_SET_ERROR_MSG("fini called with incorrect node");
+      return RCL_RET_INVALID_ARGUMENT;
```
fujitatomoya, line 179:
```suggestion
      return RCL_RET_NODE_INVALID ;
```

rcl/test/rcl/test_publisher.cpp
```diff
@@ -241,20 +253,32 @@ TEST_F(CLASSNAME(TestPublisherFixture, RMW_IMPLEMENTATION), test_publisher_init_
...
-  rcl_reset_error();
 
   // Pass invalid node to fini
   ret = rcl_publisher_fini(&publisher, nullptr);
   EXPECT_EQ(RCL_RET_NODE_INVALID, ret) << rcl_get_error_string().str;
   rcl_reset_error();
 
+  // Pass a different node to fini
+  ret = rcl_publisher_fini(&publisher, this->different_node_ptr);
+  EXPECT_EQ(RCL_RET_INVALID_ARGUMENT, ret) << rcl_get_error_string().str;
```
fujitatomoya, line 264:
```suggestion
  EXPECT_EQ(RCL_RET_NODE_INVALID, ret) << rcl_get_error_string().str;
```

rcl/test/rcl/test_subscription.cpp
```diff
@@ -256,6 +268,11 @@ TEST_F(CLASSNAME(TestSubscriptionFixture, RMW_IMPLEMENTATION), test_subscription
   rcl_reset_error();
   EXPECT_EQ(RCL_RET_NODE_INVALID, rcl_subscription_fini(&subscription, &invalid_node));
   rcl_reset_error();
+  EXPECT_EQ(
+    RCL_RET_INVALID_ARGUMENT, rcl_subscription_fini(
```
fujitatomoya, line 272:
```suggestion
    RCL_RET_NODE_INVALID, rcl_subscription_fini(
```

fujitatomoya (changes requested):
this makes sense to me. some minor fixes need to be addressed.

nnmm:
Actually, @fujitatomoya, would it be okay to ignore the node argument to the `fini()` functions entirely (and stating that it is unused in the docs)?

nnmm:
> I think we should keep it. since it explicitly requires that application is responsible to keep the valid node handle object until finalizing publisher(s) or subscription(s).

Some background: In the Rust client library, I'd like to implement the destructor (drop function) for `rcl_publisher_t`. This function only has access to the type it's implemented on, so I can't provide a `rcl_node_t` there. There is a workaround which we have in place, but it would be nice to get rid of it. When I mentioned this in the client library WG, @sloretz remarked that the node argument was unnecessary in the first place and should instead be stored. Then we could create something like `rcl_publisher_fini2(rcl_publisher_t *)`, without the node argument, and this PR was going to be the first step towards that.

I realized we could avoid this ugly "overloaded" function by instead modifying the existing fini functions to accept a `nullptr` in the second argument, hence my suggestion.

As for the extra argument expressing a lifetime requirement, I think it would be better to just clearly document this requirement. Other types also don't express lifetime requirements this way, e.g. `rcl_wait_set_fini()` and `rcl_node_fini()` don't take a context argument.

So, to summarize: ignoring the node argument would be useful, and, in my opiniion, make the fini functions more consistent. Let me know what you think :)

rcl/include/rcl/publisher.h
wjwwood: I would actually recommend a new error code. This is conflating two things, first that the node is invalid and second that it is valid (perhaps) but not the same node that created the publisher.

wjwwood (changes requested):
I think a new error code is order, so I recommended that.

I think this change is good, but I wouldn't go further and remove it, as I said here: https://github.com/ros2/rcl/issues/986#issuecomment-1142509827

nnmm:
Okay! Below I'm quote-replying to @wjwwood about some general points, but if nothing new comes up I'm fine with leaving it like it is now (modulo the new error type).

> You should use rust to ensure that the node is never dropped before the publisher/subscription.

Definitely, the publisher and subscription have shared pointers of the node. And the wait set and node have shared pointers of the context.

> So I think the tension you are running into should not be worked around. So actually, this argument has achieved exactly what I intended it to, it made us discuss relative lifetimes of these objects.

Fair point, though I shouldn't have called it a workaround: we're not avoiding doing things properly, it's just that the code is a bit more complicated than what we'd do if `fini()` only took one argument.

Also, this pattern isn't used consistently across `rcl`, so if a "parent" object is not an input to `fini()`, I'm not sure if it doesn't need to be kept alive, or the function doesn't follow this pattern.

Code-as-documentation also not applicable everywhere. For instance, how long does a subscription that was added to a wait set need to live? It wouldn't be practical to add all the subscriptions as extra parameters to `rcl_wait_set_fini()` or `rcl_wait_set_clear()`. Similarly, does the node name string passed to `rcl_node_init()` get stored or copied?

This is why I think lifetime requirements should be explained via the API docs. In my dream world, lifetimes would be clarified by a combination of a central API conventions doc that states things like "`char*` are always copied, never stored",  and a note about the lifetime of each function parameter not covered by these conventions, for each function.

> You need to use the lifetime feature to ensure this. I know that can be really difficult in dependent types, but its the "right thing to do" imo.

If you're suggesting the publisher should hold a reference to the node – we're using shared pointers because the lifetime parameters that result from references are infectious and often result in bad ergonomics. Maybe we'll do an experiment to validate whether it's workable, but my guess is it's not. Though that topic is entirely orthogonal to the rest of the discussion, since the lifetime requirements are the same either way. Sorry for going down a tangent, I guess I just can't leave anything uncommented, especially Rust things :)

nnmm:
@fujitatomoya @wjwwood I added a new return code and made the fini functions return it, as suggested.

fujitatomoya (approved):
lgtm

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16791/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11372/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17201/)

nnmm:
@wjwwood Do the latest changes match what you requested?

nnmm:
Friendly ping @wjwwood

nnmm:
@wjwwood It's been a few weeks, could you take another look and check if your "changes requested" still applies?

nnmm:
@wjwwood ping

nnmm:
I'm closing this and opening a new PR since the "changes requested" is blocking this PR.

clalancette:
> I'm closing this and opening a new PR since the "changes requested" is blocking this PR.

In general, this isn't the way to proceed.  Opening new PRs just means all context from the previous one is lost, and it doesn't necessarily mean you'll get a faster response.  I'll suggest in the future that you just ping on the current PR rather than opening a new one.

fujitatomoya:
@nnmm can you reopen this one instead of https://github.com/ros2/rcl/pull/1015?

i think https://github.com/ros2/rcl/pull/987#discussion_r886006603 has been addressed by https://github.com/ros2/rcl/pull/987/commits/0e4dce8afaaee2723c31276886936707a885abb5, so we just need one more maintainer review.

nnmm:
Okay, sorry if that was bad style. Isn't it the case that one "changes requested" unconditionally blocks the MR from being merged? I thought that was the case, and I had little hope of reaching William after asking for a review four times over a span of more than four months.

clalancette:
> Okay, sorry if that was bad style. Isn't it the case that one "changes requested" unconditionally blocks the MR from being merged?

Not necessarily.  The default GitHub UI doesn't enforce this, though you can add rules to make this the case.  We haven't done this on the ROS repositories.

Regardless, we still need to have whatever it is resolved.  If we have enough consensus, we can move forward without @wjwwood 's  approval (or we can eventually get it).

nnmm:
Thanks for clarifying! I assumed that it acted as a blocker, I wouldn't have closed this otherwise.

fujitatomoya:
@nnmm can you reopen this?

nnmm:
It doesn't look like I can, the button is greye bout.

clalancette:
Yes, same here.  I'm guessing the source branch was deleted or something like that.

At this point, please just go ahead and open a new one.  In the future we can just leave them open.

nnmm:
It's probably that I squashed the commit history for the new PR.

PR #990, opened 2022-06-29
Service Introspection

ihasdapie:
This PR is a prototype for the service introspection REP (https://github.com/ros-infrastructure/rep/pull/360) implementing the [proposed serialized service event approach](https://github.com/ros-infrastructure/rep/pull/360#discussion_r902987531). 

Few items of note
- This would introduce new dependencies for `rcl` on `rosidl_typesupport_introspection_c` and `rosidl_typesupport_c` 
- This implements a _rather_ hacky way of loading the respective Request/Response typesupports given a Service typesupport by inspecting the service typesupport's symbol names to find the service name, then using that to build the name of the Request/Response typesupport libaries and loading them in.   
https://github.com/ros2/rcl/blob/9c0b42edace34ff9783274e4eb18ceeaacc87015/rcl/src/rcl/introspection.c#L57-L119
- The `client_id` is not populated on client send request events due to the `writer_guid` not being accessible from the `client_send_request` scope. PRs will be made to the `rmw` implementations to make this accessible.
- There are no tests/docs/minimal error handling. It also _definitely_ doesn't pass linting and there are superfluous #includes. I will fix this later :)

For instructions on how to build and run this see the meta-ticket https://github.com/ros2/ros2/issues/1285 

Resolves https://github.com/ros2/ros2/issues/1285


clalancette:
> * This would introduce new dependencies for `rcl` on `rosidl_typesupport_introspection_c` and `rosidl_typesupport_c`

I just took a quick look, and it looks like there is already an (build) indirect dependency between `rcl` and `rosidl_typesupport_introspection_c`/`rosidl_typesupport_c`:

`rcl` -> `rcl_interfaces` -> `rosidl_default_generators` -> `rosidl_typesupport_introspection_c` (same goes for `rosidl_typesupport_c`).  Thus I think turning this into a direct dependency is fine.

rcl/include/rcl/introspection.h
```diff
@@ -0,0 +1,162 @@
...
+  rcl_clock_t * clock;
+  rcl_publisher_t * publisher;
+  rosidl_message_type_support_t * request_type_support;
+  rosidl_message_type_support_t * response_type_support;
+  char * service_name;
+  char * service_type_name;
+  char * service_event_topic_name;
+  bool _enabled;
+  bool _content_enabled;
+} rcl_service_introspection_utils_t;
```
ivanpauno, line 41: IMHO, the name is not pretty clear.
Maybe `rcl_service_events_publisher` or something like that.

rcl/include/rcl/introspection.h
```diff
@@ -0,0 +1,162 @@
...
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_service_introspection_utils_t
+rcl_get_zero_initialized_introspection_utils();
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_service_introspection_init(
```
ivanpauno, line 51: whatever the name of the `struct` ends up being, this should be named `{name_of_the_struct_without_trailing"_t"}_init`.
Based on the current struct name, it should be `rcl_service_introspection_utils_init` (though see my other comment about the struct name as well).

Same applies to all the functions

rcl/include/rcl/introspection.h
```diff
@@ -0,0 +1,162 @@
...
+  const rcl_node_t * node,
+  rcl_allocator_t * allocator);
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_service_introspection_fini(
+  rcl_service_introspection_utils_t * introspection_utils,
+  rcl_allocator_t * allocator,
+  rcl_node_t *  node);
```
ivanpauno, line 64: nit:

```suggestion
  rcl_node_t * node);
```

rcl/include/rcl/introspection.h
```diff
@@ -0,0 +1,162 @@
...
+ *
+ */
+
+RCL_LOCAL
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_service_introspection_disable(
+    rcl_service_introspection_utils_t * introspection_utils,
+    rcl_node_t * node,
+    const rcl_allocator_t * allocator);
```
ivanpauno, line 107: I would expect to be able to enable/disable service introspection (or service events) in `rcl_service_t`, and not here.
If service events are disabled, this struct should not be inited.

rcl/include/rcl/introspection.h
```diff
@@ -0,0 +1,162 @@
...
+ *
+ *
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_service_introspection_configure_service_events(
+    rcl_service_t * service,
+    rcl_node_t * node,
+    bool enable);
```
ivanpauno, line 125: it seems that this function should be in `service.h`, if it affects `rcl_service_t` and not `rcl_service_introspection_utils_t`.

rcl/include/rcl/introspection.h
```diff
@@ -0,0 +1,162 @@
...
+
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_service_introspection_configure_client_events(rcl_client_t * client,
+    rcl_node_t * node,
+    bool enable);
```
ivanpauno, line 132: same here

rcl/include/rcl/introspection.h
```diff
@@ -0,0 +1,162 @@
...
+/*
+ *
+ */
+RCL_PUBLIC
+void
+rcl_service_introspection_configure_client_content(rcl_client_t * client, bool enable);
+
+RCL_PUBLIC
+void
+rcl_service_introspection_configure_service_content(rcl_service_t * service, bool enable);
```
ivanpauno, line 144: Same for this two, but I also don't understand what they are for....

rcl/include/rcl/service.h
```diff
@@ -361,6 +361,16 @@ RCL_WARN_UNUSED
...
 
+/// Get the service type name for the service
+/**
+ 
+ * TODO(ihasdapie): document this function
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+const char *
+rcl_service_get_service_type_name(const rosidl_service_type_support_t * service_type_support);
```
ivanpauno, line 372: This function should be added [here](https://github.com/ros2/rosidl/blob/rolling/rosidl_runtime_c/include/rosidl_runtime_c/service_type_support_struct.h), and not in this file.

rcl/include/rcl/introspection.h
```diff
@@ -0,0 +1,162 @@
...
+
+#include "rcl/publisher.h"
+#include "rcl/service.h"
+#include "rcl/time.h"
+#include "rmw/rmw.h"
+#include "rcl/client.h"
+
+#define RCL_SERVICE_INTROSPECTION_TOPIC_POSTFIX "/_service_event"
+
+typedef struct rcl_service_introspection_utils_s {
```
ivanpauno, line 31: if this is only usef from rcl itself and not from rclcpp/rclpy/etc, the we should move the header file to the src file (i.e. not make it public).
In that case, ignore my other comment about using pimpl.

ivanpauno (commented):
partial review

rcl/src/rcl/introspection.c
```diff
@@ -0,0 +1,409 @@
...
+#include "rcl_interfaces/msg/service_event_type.h"
+#include "rcutils/logging_macros.h"
+#include "rcutils/macros.h"
+#include "rcutils/shared_library.h"
+#include "rmw/error_handling.h"
+#include "rosidl_runtime_c/primitives_sequence_functions.h"
+#include "rosidl_runtime_c/string_functions.h"
+#include "rosidl_typesupport_c/type_support_map.h"
+#include "rosidl_typesupport_introspection_c/identifier.h"
+#include "rosidl_typesupport_introspection_c/service_introspection.h"
```
ivanpauno, line 39: do you need this header files for anything?

rcl/package.xml
```diff
@@ -24,6 +24,9 @@
   <depend>rosidl_runtime_c</depend>
   <depend>tracetools</depend>
 
+  <depend>rosidl_typesupport_introspection_c</depend>
```
ivanpauno, line 27: why do you need this dependency?
It doesn't seem that you're using this anywhere.

ihasdapie: There _was_ a dependency on `rosidl_typesupport_introspection_c` but that has since been removed as I iterated on this PR. I will remove the dependency.

rcl/include/rcl/introspection.h
ihasdapie: Good point, I can reoganize these functions accordingly

rcl/include/rcl/introspection.h
ihasdapie: These two functions are for configuring whether or not to include the Request/Response event message in the introspected output. For example someone might only want to introspect the timing of Request/Response events and so does not want to incur the added overhead of publishing out (potentially large) service event messages.

rcl/include/rcl/introspection.h
ihasdapie: Parts of this header file is used from the client libraries. See: https://github.com/ros2/rclcpp/tree/service_introspection-serialized (WIP)

rcl/include/rcl/introspection.h
ivanpauno: > These two functions are for configuring whether or not to include the Request/Response event message in the introspected output. For example someone might only want to introspect the timing of Request/Response events and so does not want to incur the added overhead of publishing out (potentially large) service event messages.

Ah, I see.
I think it makes more sense to have only one `rcl_service_events_options_t`, which could be a member of [rcl_service_options_t](https://github.com/ros2/rcl/blob/4eccc3cbe65f5b58981f22efaf612a65731cb2f0/rcl/include/rcl/service.h#L43-L50), instead of having many functions to enable disable the events and to enable disable sending the messages contents.

We should also improve naming, something like "message payload" would be more clear than "service content".

rcl/include/rcl/introspection.h
ivanpauno: If you move all the functions that act on a `rcl_service_t` directly to `service.h` that's probably not more the case already.
i.e. the functions that take `rcl_service_introspection_utils_t` as an argument are probably not being used from other client libraries (if not currently the case, I think it could be avoided).
Though I haven't checked your `rclcpp` WIP yet.

rcl/include/rcl/introspection.h
jacobperron: +1 for `message_payload`, or maybe the more verbose `request_response_payload`

jacobperron (commented):
The approach looks good to me. Besides the string manipulation related to service typesupport, the code looks straight-forward :+1:  I also agree with Ivan's comments.

I realize you haven't focused on error handling yet, but when you do you should look out for the following (which might help uncover bugs downstream):

1. Always check the return values of functions (this isn't happening in many places)
2. Make sure to do any necessary memory deallocation before returning from functions (pay attention to places where you are returning early due to an error).
3. Check arguments to functions before using them (e.g. if they are null or invalid).

rcl/src/rcl/introspection.c
```diff
@@ -0,0 +1,383 @@
...
+    sizeof(char) * ((strlen(typesupport_library_fmt) - 2) + strlen(map->package_name) + 1),
+    allocator->state);
+  char * request_message_symbol = allocator->allocate(
+    sizeof(char) * ((strlen(service_message_fmt) - 6) + strlen(map->package_name) +
+                    strlen(service_type_name) + strlen("Request") + 1),
+    allocator->state);
+  char * response_message_symbol = allocator->allocate(
+    sizeof(char) * ((strlen(service_message_fmt) - 6) + strlen(map->package_name) +
+                    strlen(service_type_name) + strlen("Request") + 1),
+    allocator->state);
```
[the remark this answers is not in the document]
jacobperron: We can also use `snprintf` to guard against buffer overflow. We should use the portable [rcutils implementation](https://github.com/ros2/rcutils/blob/0fa7d5c318fa357244ecba45a1a9b0764ea734d1/include/rcutils/snprintf.h#L59), which works on Windows.

ihasdapie:
Closing in lieu of https://github.com/ros2/rcl/pull/997 to keep all service introspection work under a single branch name

rcl/src/rcl/introspection.c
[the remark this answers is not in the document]
ihasdapie: We no longer need this function (See https://github.com/ros2/rcl/pull/997)

rcl/src/rcl/introspection.c
```diff
@@ -0,0 +1,383 @@
...
+}
+
+rcl_ret_t rcl_introspection_send_message(
+  const rcl_service_introspection_utils_t * introspection_utils, const uint8_t event_type,
+  const void * ros_response_request, const int64_t sequence_number, const uint8_t uuid[16],
+  const rcl_allocator_t * allocator)
+{
+  // Early exit of service introspection if it isn't enabled
+  // TODO(ihasdapie): Different return code?
+  if (!introspection_utils->_enabled) return RCL_RET_OK;
```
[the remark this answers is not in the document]
ihasdapie: I'm using this check to turn service introspection on/off during runtime, so it wouldn't make sense to have it be an error code. I think RCL_RET_OK is the right choice here because an early return without doing anything is expected.

rcl/src/rcl/introspection.c
```diff
@@ -0,0 +1,383 @@
...
+{
+  rcl_ret_t ret;
+  ret = rcl_publisher_fini(introspection_utils->publisher, node);
+  if (RCL_RET_OK != ret) {
+    RCL_SET_ERROR_MSG(rmw_get_error_string().str);
+    return RCL_RET_ERROR;
+  }
+
+  allocator->deallocate(introspection_utils->publisher, allocator->state);
+  introspection_utils->publisher = NULL;
```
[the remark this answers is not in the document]
ihasdapie: Yup we can. Resolved in the other PR.

rcl/src/rcl/introspection.c
```diff
@@ -0,0 +1,383 @@
...
+
+  introspection_utils->publisher = allocator->allocate(sizeof(rcl_publisher_t), allocator->state);
+  *introspection_utils->publisher = rcl_get_zero_initialized_publisher();
+  const rosidl_message_type_support_t * service_event_typesupport =
+    ROSIDL_GET_MSG_TYPE_SUPPORT(rcl_interfaces, msg, ServiceEvent);
+
+  rcl_publisher_options_t publisher_options = rcl_publisher_get_default_options();
+  ret = rcl_publisher_init(
+    introspection_utils->publisher, node, service_event_typesupport,
+    introspection_utils->service_event_topic_name, &publisher_options);
```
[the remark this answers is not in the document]
ihasdapie: I don't think it's worth consolidating into a reusable function because it is only performing the pretty standard-but-verbose-in-rcl task creating a publisher. In prior implementations I checked if it were enabled/disabled already in the server/client-specific functions I created but it does make more sense to put this logic into `rcl_service_introspection_enable`/disable. 
This has been done in https://github.com/ros2/rcl/pull/997/commits/e1a2273929921002d255de2244898befecdc35d4

rcl/src/rcl/introspection.c
```diff
@@ -0,0 +1,383 @@
...
+    allocator->allocate(strlen(service_event_topic_name) + 1, allocator->state);
+
+  strcpy(introspection_utils->service_event_topic_name, service_event_topic_name);
+
+  introspection_utils->publisher = allocator->allocate(sizeof(rcl_publisher_t), allocator->state);
+  *introspection_utils->publisher = rcl_get_zero_initialized_publisher();
+  const rosidl_message_type_support_t * service_event_typesupport =
+    ROSIDL_GET_MSG_TYPE_SUPPORT(rcl_interfaces, msg, ServiceEvent);
+
+  rcl_publisher_options_t publisher_options = rcl_publisher_get_default_options();
```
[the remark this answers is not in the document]
ihasdapie: Resolved with
- https://github.com/ros2/rcl/pull/997/commits/caac9e1ac45a2e8f1dcb0e632d4ec0370afd0631
- https://github.com/ros2/rclcpp/pull/1985/commits/0b83cb3ed7d0188df3aa124098af16cbb6bd957f
- https://github.com/ros2/rclpy/pull/988/commits/e6a3f9f6af00172592ab984a6446cfc2380e3135

ivanpauno:
@ihasdapie I think you meant to close this PR, so I'm closing it now.
Please, reopen if that was not the case.

ihasdapie:
Yup, thanks! I had forgotten.

PR #992, opened 2022-07-09
Make loaned messages const on the subscription side

nnmm:
It seems to me that it is an error for the receiver of a loaned message to modify it – other subscriptions could be reading the same message at the same time. Therefore, this PR makes the API stricter by handing out only const pointers.

I have also made corresponding changes in rclcpp, which I'll open a PR for as well and link here.

### DIFF
--- a/rcl/src/rcl/subscription.c
+++ b/rcl/src/rcl/subscription.c
@@ -610,7 +610,7 @@ rcl_take_serialized_message(
 rcl_ret_t
 rcl_take_loaned_message(
   const rcl_subscription_t * subscription,
-  void ** loaned_message,
+  const void ** loaned_message,
   rmw_message_info_t * message_info,
   rmw_subscription_allocation_t * allocation)
 {
@@ -646,7 +646,7 @@ rcl_take_loaned_message(
 rcl_ret_t
 rcl_return_loaned_message_from_subscription(
   const rcl_subscription_t * subscription,
-  void * loaned_message)
+  const void * loaned_message)
 {
   RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME, "Subscription releasing loaned message");
   if (!rcl_subscription_is_valid(subscription)) {
--- a/rcl/include/rcl/subscription.h
+++ b/rcl/include/rcl/subscription.h
@@ -630,7 +630,7 @@ RCL_WARN_UNUSED
 rcl_ret_t
 rcl_take_loaned_message(
   const rcl_subscription_t * subscription,
-  void ** loaned_message,
+  const void ** loaned_message,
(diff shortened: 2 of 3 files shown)


fujitatomoya (approved):
I think this makes sense.

CC: @MiguelCompany @ivanpauno @wjwwood

nnmm:
Can somebody help me understand why the  Rpr__rcl__ubuntu_jammy_amd64 check is red? I'm not familiar with Jenkins, and in the console log it says "0 tests failed".

clalancette:
> Can somebody help me understand why the Rpr__rcl__ubuntu_jammy_amd64 check is red? I'm not familiar with Jenkins, and in the console log it says "0 tests failed".

It's because there are compiler warnings: https://build.ros2.org/job/Rpr__rcl__ubuntu_jammy_amd64/52/gcc/new/

nnmm:
> It's because there are compiler warnings: https://build.ros2.org/job/Rpr__rcl__ubuntu_jammy_amd64/52/gcc/new/

Thanks. I'll get on it.

ivanpauno:
>  Should I also do this change in the rmw layer?

Yes, and after that it would be nice to cleanup the casts that will become unnecessary here.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16948/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11513/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17347/)

ivanpauno:
> @ivanpauno I'm not sure how this can be tested, but it needs the changes to https://github.com/ros2/rclcpp/pull/1971 as well.

ah sorry, my bad.
I didn't see the other PR.

rcl/src/rcl/subscription.c
```diff
@@ -630,7 +630,8 @@ rcl_take_loaned_message(
   // Call rmw_take_with_info.
   bool taken = false;
   rmw_ret_t ret = rmw_take_loaned_message_with_info(
-    subscription->impl->rmw_handle, loaned_message, &taken, message_info_local, allocation);
+    subscription->impl->rmw_handle, (void **) loaned_message, &taken, message_info_local,
```
clalancette, line 633: It seems to me that rather than doing this cast here, we should first change `rmw_take_loaned_message_with_info` to take a `const void ** loaned_message`.  At that point, everything will be const and no casts will be needed.

Same goes below for `rmw_return_loaned_message_from_subscription`.

clalancette (changes requested):
Sorry, I know this has been sitting a long time.  I just took a look and suggested a way to make this better if you are still interested in making this happen.

PR #995, opened 2022-07-29
Add timer on reset callback

mauropasse:
@wjwwood this PR is created after discussions originated from this PR: https://github.com/ros2/rcl/pull/966
Instead of adding an `on_trigger` callback to the rcl guard conditions, we add here an `on_reset` callback to the rcl timers.

The issue which originated this need, is wake up an events based executor when a timer is reset (which could also be caused by a time jump).

FYI @alsora

### DIFF
--- a/rcl/include/rcl/timer.h
+++ b/rcl/include/rcl/timer.h
@@ -26,6 +26,7 @@ extern "C"
 
 #include "rcl/allocator.h"
 #include "rcl/context.h"
+#include "rcl/event_callback.h"
 #include "rcl/guard_condition.h"
 #include "rcl/macros.h"
 #include "rcl/time.h"
@@ -40,6 +41,14 @@ typedef struct rcl_timer_s
   /// Private implementation pointer.
   rcl_timer_impl_t * impl;
 } rcl_timer_t;
+
+/// Structure which encapsulates the on reset callback data
+typedef struct rcl_timer_on_reset_callback_data_s
+{
+  rcl_event_callback_t on_reset_callback;
+  const void * user_data;
+  size_t reset_counter;
+} rcl_timer_on_reset_callback_data_t;
 
 /// User callback signature for timers.
 /**
@@ -589,6 +598,33 @@ RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_guard_condition_t *
 rcl_timer_get_guard_condition(const rcl_timer_t * timer);
+
+/// Set the on reset callback function for the timer.
+/**
+ * This API sets the callback function to be called whenever the
+ * timer is reset.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | No
+ *
+ * \param[in] timer The handle to the timer on which to set the callback
+ * \param[in] on_reset_callback The callback to be called when timer is reset
+ * \param[in] user_data Given to the callback when called later, may be NULL
+ * \return `RCL_RET_OK` if successful, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if `timer` is NULL
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_timer_set_on_reset_callback(
+  const rcl_timer_t * timer,
+  rcl_event_callback_t on_reset_callback,
+  const void * user_data);
 
 #ifdef __cplusplus
 }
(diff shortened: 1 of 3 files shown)


rcl/include/rcl/timer.h
```diff
@@ -41,6 +42,14 @@ typedef struct rcl_timer_s
   rcl_timer_impl_t * impl;
 } rcl_timer_t;
 
+/// Structure which encapsulates the on reset callback data
+typedef struct rcl_timer_on_reset_callback_data_s
+{
+  rcl_event_callback_t on_reset_callback;
+  const void * user_data;
+  size_t reset_counter;
```
fujitatomoya, line 50: just a question. this needs to be `size_t`? why cannot be `bool`? I do not see how application makes difference, could you elaborate the actual use case?

rcl/include/rcl/timer.h
```diff
@@ -590,6 +599,33 @@ RCL_WARN_UNUSED
 rcl_guard_condition_t *
 rcl_timer_get_guard_condition(const rcl_timer_t * timer);
 
+/// Set the on reset callback function for the timer.
+/**
+ * This API sets the callback function to be called whenever the
+ * timer is reset.
+ *
```
fujitatomoya, line 606: i think it would be better to add doc that callback in argument will be issue via this API if timer has been already reset.

rcl/include/rcl/timer.h
mauropasse: I would have used just a `bool` for this. But all the other entities callbacks have this `size_t` counter, so just in case I left it here.
If I have to be creative and find a use case, could be, I want to know how many times a timer had a backward time jump. If it has happened N times, I'd have this callback be called with N times as argument.

fujitatomoya: @mauropasse thanks for the comments.

rcl/include/rcl/timer.h
mauropasse: Done in 9f5c847

fujitatomoya (approved):

timonegk:
@wjwwood @clalancette It would be very nice to merge this and https://github.com/ros2/rcl/pull/995 soon. We are trying to use https://github.com/irobot-ros/events-executor/, but it requires these pull requests, and manually rebuilding rcl and rclcpp with the pull requests merged is not sufficient because the rclcpp pull request includes ABI changes that require a rebuild of all packages using `TimerBase`. It also seems as if the comments on both pull requests have been addressed.

Flova:
Another ping on this one due to the breaking ABI changes and the need to build nearly everything from source as a consequence if the events executor is used in a single node.

Also I think you mean https://github.com/ros2/rclcpp/pull/1979 @timonegk right?

Flova:
As the maintainers were updated recently also a frindly ping at @audrow and @ivanpauno

ivanpauno:
> The issue which originated this need, is wake up an events based executor when a timer is reset (which could also be caused by a time jump).

Could you explain that in more details?

mauropasse:
> Could you explain that in more detail?

There are two situations in which a wait-set based executor can be awaken when something changes related to the timer, by means of triggering its guard condition:
1. Timer is reset: https://github.com/ros2/rcl/pull/995/files#diff-a66cf81f81551a776d4b6595ae523ce53c747a7027df404d6c098c836b8ee984R446
2. Timer detects a time jump: https://github.com/ros2/rcl/pull/995/files#diff-a66cf81f81551a776d4b6595ae523ce53c747a7027df404d6c098c836b8ee984R114

Before these same situations, the `on_reset_callback` is called.
Maybe a new `on_time_jump_callback` could be created, along the `on_reset_callback`. 
In the current state, the `on_reset_callback` is used for both purposes.

ivanpauno:
> Timer is reset:

I think we trigger the guard condition because the timeout used in `rcl_wait()` wasn't calculated considering that timer.
A callback isn't trigger immediately when a timer is reset.
Based on that, I'm not sure if you would need to trigger a callback in this case for the events executor.
But if it's needed I think it's fine to add it.

> Maybe a new on_time_jump_callback could be created, along the on_reset_callback.

IMO, that sounds better.
It's weird to have a on_reset callback that's also triggered on jump.
Could the [clock jump callbacks](https://github.com/ros2/rclcpp/blob/33dae5d679751b603205008fcb31755986bcee1c/rclcpp/include/rclcpp/clock.hpp#L188) be used for that though?

mauropasse:
> Could the clock jump callbacks be used for that though?

Yes, no need to add an extra callback for it. 
I removed the callback call on `_rcl_timer_time_jump`. That function is actually used as a [clock jump callback](https://github.com/ros2/rclcpp/blob/33dae5d679751b603205008fcb31755986bcee1c/rclcpp/include/rclcpp/clock.hpp#L188) you mentioned.

ivanpauno (approved):
LGTM!

It would be nice to have a review of another maintainer before merging, maybe @audrow or @wjwwood ?

audrow (approved):
Looks good to me with green CI!

PR #999, opened 2022-08-17
Support yaml string tag '!!str'

Barry-Xu-2018:
Address ros2/rclcpp#1983

Support
```bash
ros2 run pkg_name exe_name --ros-args -p 'calib:=!!str yes'
```


rcl_yaml_param_parser/test/correct_config.yaml
```diff
@@ -49,3 +49,7 @@ intel:
   ros__parameters:
     num_cores: 8
     arch: x86_64
+light:
+  ros__parameters:
+    is_on: yes
+    on_status: !!str yes
```
fujitatomoya, line 55:
```suggestion
light:
  ros__parameters:
    is_on: yes
    on_status: !!str yes
    string_int: !!str 1234
    string_float: !!str 12.34
```

i think the bottom 2 should be recognized as string as well. it would be nice to add test?

fujitatomoya (approved):
lgtm with one confirmation for test.

rcl_yaml_param_parser/src/parse.c
```diff
@@ -89,7 +90,8 @@ void * get_value(
 
   /// Check if it is bool
   if (style != YAML_SINGLE_QUOTED_SCALAR_STYLE &&
-    style != YAML_DOUBLE_QUOTED_SCALAR_STYLE)
+    style != YAML_DOUBLE_QUOTED_SCALAR_STYLE &&
+    (tag == NULL || strcmp(YAML_STR_TAG, (char *)tag) != 0))
```
iuhilnehc-ynos, line 94: It seems the `strcmp` will be called multiple times if using `!!str` in the yaml.

fujitatomoya: Nice catch, that we can improve!

rcl_yaml_param_parser/test/correct_config.yaml
Barry-Xu-2018: Yes. I will add tests.

Barry-Xu-2018:
@clalancette Could you help to review this PR?

rcl_yaml_param_parser/src/parse.c
```diff
@@ -87,9 +88,12 @@ void * get_value(
...
     &allocator, "allocator is invalid", return NULL);
 
+  bool str_tag_enabled = tag != NULL && strcmp(YAML_STR_TAG, (char *)tag) == 0;
+
   /// Check if it is bool
   if (style != YAML_SINGLE_QUOTED_SCALAR_STYLE &&
-    style != YAML_DOUBLE_QUOTED_SCALAR_STYLE)
+    style != YAML_DOUBLE_QUOTED_SCALAR_STYLE &&
+    !str_tag_enabled)
   {
```
clalancette, line 97: Adding this check everywhere seems to be brittle.  If someone adds a new type, but forgets to add the `!str_tag_enabled`, it will fail in very specific circumstances.

Instead, can we just do the following at the top of the function:

```
if (tag != NULL && strcmp(YAML_STR_TAG, (char *)tag) == 0) {
  *val_type = DATA_TYPE_STRING;
  return rcutils_strdup(value, allocator);
}
```

rcl_yaml_param_parser/test/correct_config.yaml
```diff
@@ -49,3 +49,8 @@ intel:
   ros__parameters:
     num_cores: 8
     arch: x86_64
+string_tag:
+  ros__parameters:
+    string_bool: !!str yes
+    string_int: !!str 1234
+    string_double: !!str 12.34
```
clalancette, line 56: Nit: add a newline at the end of this file.

clalancette (changes requested):

rcl_yaml_param_parser/test/correct_config.yaml
Barry-Xu-2018: will update

Barry-Xu-2018:
Rebase code

Barry-Xu-2018:
@clalancette Thanks for your comments.   
I updated code.

rcl_yaml_param_parser/src/parse.c
iuhilnehc-ynos: It seems a duplicated code [there](https://github.com/ros2/rcl/blob/6b516f0455c4739b64ff0dfa11387712ce6648d6/rcl_yaml_param_parser/src/parse.c#L212-L213), can we use `goto string_value;` instead?

clalancette (approved):
This looks good to me with green CI.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17864/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12432/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18535/)

ivanpauno:
Do we really need this?

Using "yes" or "no" (i.e. between quotes) should be enough.
If the quoted version is being converted to a bool, then that should be fixed.
But the addition of !!str seems unnecessary IMHO.
It's still fine to have it though, but it may be confusing to only support the !!str tag.

clalancette:
> If the quoted version is being converted to a bool, then that should be fixed.

The problem is that there is fundamentally no way to fix it.  The YAML specification states that if it sees "yes" or "no", it automatically casts the value to a boolean type.  Further, the YAML specification specially allows for `!!str`, exactly to handle this situation.  See the spec in https://yaml.org/spec/1.2.2/ .

The YAML specification actually allows for forcing *any* type, via something like `!!bool`, which this isn't supporting.  There is an argument to be made that we should support that as well, but I think that `!!str` is by far the most likely one to be practically needed.  I'd leave the rest of it for future work.

ivanpauno:
> The problem is that there is fundamentally no way to fix it. The YAML specification states that if it sees "yes" or "no", it automatically casts the value to a boolean type. 

That only applies if the value is unquoted.
e.g.

```python3
import yaml

yaml.safe_load("{string_val: 'yes', boolean_val: yes}")
```

The regex of the specification also only matches unquoted values, see https://yaml.org/spec/1.2.2/#1022-tag-resolution (10.2.2 doesn't support `yes` as a boolean though, that's from older specs).
That's why supporting !!str seems unnecessary IMO.

The library we're using allows you to check if the value is quoted or not: https://github.com/ros2/rcl/blob/6b516f0455c4739b64ff0dfa11387712ce6648d6/rcl_yaml_param_parser/src/parse.c#L98-L99
We're actually using that for booleans, so I'm not sure if there's something wrong in the logic.

fujitatomoya:
> That only applies if the value is unquoted.

as mentioned https://github.com/ros2/rclcpp/issues/1983#issuecomment-1220399274, this is correct.

> That's why supporting !!str seems unnecessary IMO.

direct solution against https://github.com/ros2/rclcpp/issues/1983, it would be easier just to add quote for user? at least that what i would do instead of adding `!!str` tag before string. having `!!str` in rcl yaml parse is nothing wrong, and probably unnecessary for most users, but someone still might do that since it is in yaml specification tag?

@Barry-Xu-2018 @iuhilnehc-ynos what do you think?

ivanpauno:
> having !!str in rcl yaml parse is nothing wrong, and probably unnecessary for most users, but someone still might do that since it is in yaml specification tag?

Yes, it's also fine to have tags.
It's confusing to only support one tag though.
But it still seems okay.

I wasn't sure if the option of quoting the value was mentioned or not.

Barry-Xu-2018:
> direct solution against https://github.com/ros2/rclcpp/issues/1983, it would be easier just to add quote for user?

Yeah. It is more convenient for users.   

> but someone still might do that since it is in yaml specification tag?

'!!str' is defined in YAML specification. So user may use it.  If YAML tag isn't supported, we should tell users this limitation and workaround way in the user manual.   
If tag is supported, we should support all tags defined in YAML specification.

ivanpauno (approved):
LGTM!

fujitatomoya:
created follow-up issue on https://github.com/ros2/rcl/issues/1026

PR #508, opened 2019-09-23
Parse CLI parameters and YAML files

hidmic:
This pull request adds parameter parsing functionality to `rcl`. Parameters overrides provided directly through the command line and indirectly through parameter YAML files are combined in a single `rcl_params_t` structure -- last override always takes precedence.

Depends on #507.

PR #1004, opened 2022-08-26
Modifies timers API to select autostart state

Voldivh:
Signed-off-by: Voldivh <eloyabmfcv@gmail.com>

Part of ros2/rclcpp#1980

Basically it gives the timer initialization a variable that enables or disables the autostart on initialization. The variable `autostart` is defaulted to `True` in order to avoid any breaks in current use of the API.


rcl/include/rcl/timer.h
```diff
@@ -157,7 +158,8 @@ rcl_timer_init(
   rcl_context_t * context,
   int64_t period,
   const rcl_timer_callback_t callback,
-  rcl_allocator_t allocator);
+  rcl_allocator_t allocator,
+  bool autostart);
```
tfoote, line 162: I think that we want to add a second method with the old API and mark it as deprecated so that there's a cycle where we don't break users of the current API.

tfoote (changes requested):
Two thoughts: 

* We should add a backwards compatible function.
* We should make sure that there's new test coverage of the `autostart = false` path in the timer unit tests.

rcl/include/rcl/timer.h
tfoote: I forgot that in addition to no default arguments you can't overload either in pure C. Nevermind.

Voldivh: Yeah, I believe there are other packages using this method. Should we create PR's solving this issue as well before merging this one?

clalancette: Because this is such a core function, we should definitely tick-tock deprecate this one before we change it.

Towards that end, we should make a new API with a similar name, but different enough in the eyes of C.  Some ideas:

* `rcl_timer_init2`
* `rcl_timer_init_autostart`

(I don't love those, but other things along that line would be fine as well).  Then we should mark the old API as deprecated, and then finally we should update all of the in-core users of the old API to the new API.

rcl/test/rcl/test_timer.cpp
```diff
@@ -12,6 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Suppress deprecated function warning
+#ifndef _WIN32
+# pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#else  // defined(_WIN32)
+# pragma warning(disable: 4996)
+#endif
```
Voldivh, line 20: I added this declarations in order to avoid the warnings of deprecation during the CI (which were causing it to 'fail'). On the test, I added it for the whole file because there are several places where `rcl_timer_init` was used and would reduce readability to add it one by one. Let me know if you think this is ok or if we want to add the `# pragma GCC diagnostic ignored "-Wdeprecated-declarations"` only to the places where `rcl_timer_init` is being called.

clalancette: So the thing is, we don't want to do either.  If we are going to deprecate an API, then all of the in-core uses of it should use the new API, not the old one.

So the correct fix here is to go through and fix all of the callers of the old API to the new one.  Unfortunately, this also means that we'll probably have to change `rclcpp` and `rclpy` as well.

Voldivh: Oh ok, I misunderstood what we wanted to do I'll do the modifications. Regarding the other packages, yes it affects `rclcpp` package (I already made the changes there on this [PR](https://github.com/ros2/rclcpp/pull/2005) and regarding `rclpy` I'll do the changes and comment that PR here.

Voldivh: This are the repositories with their corresponding PR that were modified due to this change:
- rclcpp -> [#2005](https://github.com/ros2/rclcpp/pull/2005)
- rclpy -> [#999](https://github.com/ros2/rclpy/pull/999)
- system_test -> [#508](https://github.com/ros2/system_tests/pull/508)

fujitatomoya (commented):
although this is an easy way to support autostart capability for timer, i was expecting we were going to extend the capability like `start` and `stop` for the timer. and then autostart can rely on those capabilities?

i was thinking that,

- with current implementation including https://github.com/ros2/rclcpp/pull/2005/
  - creating already canceled timer and start the timer with `reset` is kinda weird?
- having `start` and `stop` capabilities for the timer.
  - cannot start / stop the canceled timer.
  - canceled timer can be reset only.
  - `next_call_time` should be calculated and managed internally. (for example, period is 30s timer. after 15s, it can call `stop`, then left time should be kept. if it starts the timer, 1st period will be 15s and after that 30s period. this is the difference from reset.)
  - `autostart` flag can call `start` API internally. in this case, we would add `rcl_timer_start/stop` APIs as extension, probably avoid to add `rcl_timer_init2` to keep the backward compatibility?

There could have been discussion something like this, but i would like to have feedback.

thanks,

fujitatomoya:
CC: @ivanpauno @wjwwood

ivanpauno:
> having start and stop capabilities for the timer.

How would they work exactly?
Those sound like reset and cancel to me, only with different names.
The period of the timer can also be modified.
With that and the possibility of starting a timer "stopped" as this PR is adding, all the start/stop capabilities that I can imagine of a timer are supported.

ivanpauno:
> for example, period is 30s timer. after 15s, it can call stop, then left time should be kept. if it starts the timer, 1st period will be 15s and after that 30s period. this is the difference from reset.

Ah sorry, you already explained my question.
I think that adding support for that is fine, but I don't think it's related to the functionality added here.
Naming of the new function(s) to achieve that is going to be confusing though ...

Voldivh:
Taking into account the proposal of extending the capabilities of timer and some concerns on this [PR](https://github.com/ros2/rclcpp/pull/2006#discussion_r973768620) regarding the use of the `reset` method. I wanted to show a simple flowchart on how this autostart feature is working at the moment. 
image
Basically I'm giving a choice and using current 'states' of the timer to achieve the feature. I do think that in order to start a timer you have to reset it, sounds weird (semantically). However, my concern is that implementing a `start()` method would imply to basically code the same `reset()` method or call that method inside a new one (I'm not sure on what would be preferred on this case). Keeping that in mind and assuming we are going to add that `start()` method, we could differentiate it from the reset function by extending it in order to do the following:
image
That gives the possibility to reset the timer without it being autostarted and the start method is just a way to "uncancel" the timer so it goes to an ongoing state. Let me know what you guys think about this @fujitatomoya @ivanpauno

fujitatomoya:
I believe that the latter one makes more sense. (related to https://github.com/ros2/rclcpp/pull/2006#discussion_r973768620, adding `start` / `stop` would be useful to support these features.)

@ros2/team friendly ping.

rcl/include/rcl/timer.h
fujitatomoya: i am okay with taking `rcl_timer_init2` but this needs to be `tick-tock` every time we have this kind of API changes. how about adding the arguments configuration option like a bitmap as interfaces and implementation can be done so internally?

fujitatomoya (commented):
overall lgtm, sorry to be late to get back to this.

@Voldivh are you still working on this? there are conflicts what needs to be addressed.

Voldivh:
> overall lgtm, sorry to be late to get back to this.
> 
> @Voldivh are you still working on this? there are conflicts what needs to be addressed.

Hey @fujitatomoya I stopped working on this some time ago. However, I believe it only requires a little final push to get this and the follow up PR's in. I believe I can get back on this later this week. Give me a couple of days to recall the implementation I was doing and solve the conflicts we currently have.

rcl/include/rcl/timer.h
Voldivh: Just to be clear, your suggestion is to instead of passing a boolean for each configuration feature the timer have (right now it would just be the autostart) we pass an array of booleans which are then later used on the implementation for whatever means they have? Creating the appropiate documentation of course. 
If that is the case I see the benefit of it from a developer point of view, however, wouldn't it be kinda bothersome for the user?

fujitatomoya: yeah, that is correct, probably we can include API negotiation for that. user can know which feature is enabled, and then using common structure with configuration mode, just call `rcl_timer_configure(COMMAND, common)` to the initialized timer. so that we can avoid this `tick-tock` every time we change the API.

but this is suggestion here, and not directly related to this specific feature. maybe we can consider with different issue 1st.

fujitatomoya:
@Voldivh thanks for taking care of this, there is build failure https://build.ros2.org/job/Rpr__rcl__ubuntu_jammy_amd64/350/console

Voldivh:
> @Voldivh thanks for taking care of this, there is build failure https://build.ros2.org/job/Rpr__rcl__ubuntu_jammy_amd64/350/console

Yeah I noticed, however, I believe the error is coming from this [commit](https://github.com/ros2/rcl/commit/c43c7abb0a254f4b60c42c4c2993917eb08cb0ec) made into rolling. I'm not sure if I should modify something here.

fujitatomoya:
@christophebedard Rpr job is supposed to fail because of https://github.com/ros2/rcl/commit/c43c7abb0a254f4b60c42c4c2993917eb08cb0ec?

christophebedard:
Yeah, it will fail until `tracetools` is released and available in the testing repo.

fujitatomoya:
@Voldivh never mind the rpr job failure, i wll start the CI to verify.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18881/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13419/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19637/)

with

- https://github.com/ros2/rcl/pull/1004
- https://github.com/ros2/rclcpp/pull/2005
- https://github.com/ros2/rclpy/pull/999
- https://github.com/ros2/system_tests/pull/508

Voldivh:
Hey @fujitatomoya, I believe the CI build failed because of the branch for https://github.com/ros2/system_tests/pull/508 was outdated. I already sync it with the latest changes from rolling, I think it should fix it.

fujitatomoya:
@Voldivh thanks for checking!

CI: 🤞 
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18944/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13475/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19694/)

Voldivh:
@fujitatomoya well that didn't do the trick :c. I'm not sure on what to do next, this is the error we are getting https://ci.ros2.org/job/ci_linux/18944/console#:~:text=12%3A18%3A46%20%3D%3D%3E,with%20return%20code%20%271%27

clalancette:
> @fujitatomoya well that didn't do the trick :c. I'm not sure on what to do next, this is the error we are getting https://ci.ros2.org/job/ci_linux/18944/console#:~:text=12%3A18%3A46%20%3D%3D%3E,with%20return%20code%20%271%27

There's no package called `system_tests`, it is called `test_rclcpp`.

Voldivh:
> There's no package called system_tests, it is called test_rclcpp.

That makes a lot of sense, is that something that is defined when running the CI or do I have to modify something?

fujitatomoya:
@Voldivh @clalancette Ah, sorry 😓 i will restart the CI.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18949/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13480/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19701/)

fujitatomoya:
@Voldivh i think we need to rebase rclcpp, please check https://ci.ros2.org/job/ci_linux/18949/console.

Voldivh:
> @Voldivh i think we need to rebase rclcpp, please check https://ci.ros2.org/job/ci_linux/18949/console.

I agree, it should be fine now after [this commit](https://github.com/ros2/rclcpp/pull/2005/commits/9597b9359dc7e66030a3e8314a2dbae9f78aed8e).

emersonknapp:
Pulls: ros2/rcl#1004, ros2/rclcpp#2005, ros2/system_tests#508, ros2/rclpy#999
Gist: https://gist.githubusercontent.com/emersonknapp/f9353e435d78a1ff796a38b74eaf0547/raw/2a9e6c7645332dcc2c4a6d0fb6f7b66e6b428bb7/ros2.repos
BUILD args:  --packages-above-and-dependencies rcl rcl_action rclcpp rclpy test_rclcpp
TEST args:  --packages-above rcl rcl_action rclcpp rclpy test_rclcpp
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12240
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18956/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13489/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19709/)

Voldivh:
> * Linux [Build Status](http://ci.ros2.org/job/ci_linux/18956/)
> * Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13489/)
> * Windows [Build Status](http://ci.ros2.org/job/ci_windows/19709/)

I think we had the same issue as before but now with rclpy. I already updated the branch. All branches from the different PR's should be up to date now with the latest updates of rolling.

emersonknapp:
Rebuilds build #[12240](https://ci.ros2.org/job/ci_launcher/12240)
Running as SYSTEM
Building on the built-in node in workspace /var/lib/jenkins/jobs/ci_launcher/workspace
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18968/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13496/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19715/)

fujitatomoya (approved):

fujitatomoya:
@clalancette can you review and merge these PRs? they need to be go in at the same time, but i do not have merge permission for `system_tests`.

- https://github.com/ros2/rcl/pull/1004
- https://github.com/ros2/rclcpp/pull/2005
- https://github.com/ros2/rclpy/pull/999
- https://github.com/ros2/system_tests/pull/508

CC: @emersonknapp @Voldivh

rcl/include/rcl/timer.h
```diff
@@ -160,6 +162,22 @@ rcl_get_zero_initialized_timer(void);
...
+  const rcl_timer_callback_t callback,
+  rcl_allocator_t allocator,
+  bool autostart);
+
+/**
+ * \deprecated `rcl_timer_init` implementation was removed.
+ *   Refer to `rcl_timer_init2`.
+ */
+RCL_PUBLIC
+RCUTILS_DEPRECATED_WITH_MSG("Calling rcl_timer_init2")
```
clalancette:
```suggestion
RCUTILS_DEPRECATED_WITH_MSG("Call rcl_timer_init2 instead")
```

rcl/include/rcl/timer.h
```diff
@@ -125,7 +126,7 @@ rcl_get_zero_initialized_timer(void);
  * // ... error handling
  *
  * rcl_timer_t timer = rcl_get_zero_initialized_timer();
- * ret = rcl_timer_init(
+ * ret = rcl_timer_init2(
  *   &timer, &clock, context, RCL_MS_TO_NS(100), my_timer_callback, allocator);
```
clalancette:
```suggestion
 *   &timer, &clock, context, RCL_MS_TO_NS(100), my_timer_callback, allocator, true);
```

rcl/src/rcl/timer.c
```diff
@@ -22,7 +22,6 @@ extern "C"
 #include <inttypes.h>
 
 #include "rcl/error_handling.h"
-#include "rcutils/logging_macros.h"
```
clalancette: Why the removal of this include?  Those macros are used in the file.

clalancette (changes requested):
I've left three small things to fix.

rcl/src/rcl/timer.c
Voldivh: You are right, I believe this removal was a mistake.

Voldivh: Done.

rcl/include/rcl/timer.h
Voldivh: Done.

rcl/include/rcl/timer.h
Voldivh: Done.

fujitatomoya:
Final? CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19006/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13531/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19760/)

clalancette (approved):

PR #1012, opened 2022-10-19
Reduce result_timeout to 10 seconds.

clalancette:

I spent a bit of time investigating https://github.com/ros2/ros2/issues/1074 .  One problem I found is that if an action client calls an action server repeatedly, the time it takes to complete that action gets slower and slower, eventually ending up taking a good portion of a second.

After doing some debugging, I found out that at least part of the problem is that the number of goal handles (as stored by the rcl_action_server_t structure) continually goes up.  Since rclcpp_action goes and copies all of the goal handles during [publish_status](https://github.com/ros2/rclcpp/blob/b9b1468d15c7ddc697c079e6934d54f183294280/rclcpp_action/src/server.cpp#L597-L626), this causes the amount of time necessary to call publish_status to go up and up.

But looking at it further, it seems like the main problem is that goal_handles are never being cleared out when they are done being used.  Instead, it looks like they are meant to "expire", but this only happens by default every 15 minutes.  Quite a lot of goal handles can be accumulated in that time.

This PR is a not-serious attempt to remedy this by reducing the default expiration time to 10 seconds.  With this in place, things still end up going slower for 10 seconds when first starting up, but then the expiration logic kicks in and keeps the number of goal handles to a reasonable number.  I don't really intend this to go in as-is, but I would like to have a discussion about whether we should remove goals automatically once `goal_handle->succeed` is called, whether `publish_status` should really publish status on *all* goals, and whether we should reduce the default time expiration period for action goal handles.

### DIFF
--- a/rcl_action/include/rcl_action/action_server.h
+++ b/rcl_action/include/rcl_action/action_server.h
@@ -225,7 +225,7 @@ rcl_action_server_fini(rcl_action_server_t * action_server, rcl_node_t * node);
  * - feedback_topic_qos = rmw_qos_profile_default;
  * - status_topic_qos = rcl_action_qos_profile_status_default;
  * - allocator = rcl_get_default_allocator();
- * - result_timeout = RCUTILS_S_TO_NS(15 * 60);  // 15 minutes
+ * - result_timeout = RCUTILS_S_TO_NS(10);  // 10 seconds
  */
 RCL_ACTION_PUBLIC
 RCL_WARN_UNUSED
--- a/rcl_action/src/rcl_action/action_server.c
+++ b/rcl_action/src/rcl_action/action_server.c
@@ -173,7 +173,7 @@ rcl_action_server_init(
   // Store reference to clock
   action_server->impl->clock = clock;
 
-// Initialize Timer
+  // Initialize Timer
   ret = rcl_timer_init(
     &action_server->impl->expire_timer, action_server->impl->clock, node->context,
     options->result_timeout.nanoseconds, NULL, allocator);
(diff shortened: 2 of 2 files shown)


clalancette:
After some discussion, this may be a reasonable thing to do.  So I'm going to go ahead and open this as a PR for real feedback.

clalancette:

fujitatomoya (approved):
good to go.

clalancette:

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17662/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12214/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18235/)

fujitatomoya:
https://ci.ros2.org/job/ci_windows/18331/testReport/junit/(root)/projectroot/test_publisher_subscriber_cpp__rmw_fastrtps_cpp__BoundedPlainSequences/ windows failure is unrelated.

fujitatomoya:
@clalancette do we want to backport this to humble?

clalancette:
> @clalancette do we want to backport this to humble?

I'm worried that this is a big semantic change, and it is really hard to debug.  I'd rather let it soak in Rolling for a while before we even consider that.

SteveMacenski:
By the way, this caused alot of our tests to become flaky that we're not just realizing: https://github.com/ros-planning/navigation2/issues/3765

I believe this should be set to max and/or disabled by default since its very common for actions to be very long running. I think the method for expiring of goals so they don't pile up needs to be done another way -- its throwing out currently running goals now and returning statuses of `UNKNOWN` back to the application when we have things running > 10 s.

PR #1013, opened 2022-10-20
Small cleanups to rcl

clalancette:
This PR does two things:

1.  It cleans up logging_rosout.c to get rid of some unnecessary macros.  This gives us better type checking, as well as slightly shrinks the resulting library.
2. Fix a memory leak in the tests.  This is useful mostly for trying to detect *other* memory leaks.

### DIFF
--- a/rcl/src/rcl/logging_rosout.c
+++ b/rcl/src/rcl/logging_rosout.c
@@ -28,51 +28,42 @@
 #include "rcutils/types/rcutils_ret.h"
 #include "rosidl_runtime_c/string_functions.h"
 
-#ifdef __cplusplus
-extern "C"
-{
-#endif
-
 #define ROSOUT_TOPIC_NAME "/rosout"
 
-/* Return RCL_RET_OK from this macro because we won't check throughout rcl if rosout is
- * initialized or not and in the case it's not we want things to continue working.
- */
-#define RCL_LOGGING_ROSOUT_VERIFY_INITIALIZED \
-  if (!__is_initialized) { \
-    return RCL_RET_OK; \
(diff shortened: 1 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17518/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12046/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18072/)

fujitatomoya (approved):
lgtm

Note, changing `inline`ed macro into `call`ed function might affect the performance especially on following rosout logging. but i think there would be almost no performance degression for this change.

https://github.com/ros2/rcl/blob/edd65bfc5fca267c8ec842e8e66c404e5bcb0138/rcl/src/rcl/logging_rosout.c#L260

clalancette:
> Note, changing `inline`ed macro into `call`ed function might affect the performance especially on following rosout logging. but i think there would be almost no performance degression for this change.

Yes, that's a good point.  Theoretically, there could be a slowdown from this since we now have a function call (with a preamble, `call` instruction, and postfix).

That said, all of that stuff should be fast.  And an optimizing pass in the compiler *might* decide to inline this anyway (since it is `static` within the compilation unit).  Finally, there should theoretically be less cache pressure on the machine because of this change (since the resulting code is smaller), which may or may not offset any slowdowns from making this a call.

I think in this case, the resulting slowdown will be immeasurable.  My opinion is that the readability and type checking is a stronger case for this code.

Thanks for the review and bringing up an interesting question!  I'm going to go ahead with this.

PR #1018, opened 2022-11-08
Implement validity checks for time points

methylDragon:
Precursor for https://github.com/ros2/rclcpp/pull/2040

Codifies the notion of zero-time being invalid as defined in the design docs, will be used in downstream rclcpp validity checks.

One question I had was whether a time point with non-zero time but associated with an uninitialized clock type should also be uninitialized? I interpreted the docs to mean that all that matters is the time point value.

### DIFF
--- a/rcl/include/rcl/time.h
+++ b/rcl/include/rcl/time.h
@@ -168,6 +168,49 @@ typedef struct rcl_time_point_s
 // } rcl_rate_t;
 // TODO(tfoote) integrate rate and timer implementations
 
+/// Check if the time point value is valid.
+/**
+ * This function returns true if the time point value is non-zero.
+ * Note that if data is uninitialized it may give a false positive.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | Yes
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] time_point_value the time point value which is being queried
+ * \return true if the source is believed to be valid, otherwise return false.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+bool
+rcl_time_point_value_valid(rcl_time_point_value_t time_point_value);
+
+/// Check if the time point is valid.
+/**
+ * This function returns true if the time point appears to be valid.
+ * It will check that the type is not uninitialized, and that the time point value is non-zero.
+ * Note that if data is uninitialized it may give a false positive.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | Yes
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] time_point the time point which is being queried
+ * \return true if the source is believed to be valid, otherwise return false.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+bool
+rcl_time_point_valid(rcl_time_point_t * time_point);
+
 /// Check if the clock has valid values.
 /**
  * This function returns true if the time source appears to be valid.
--- a/rcl/src/rcl/time.c
+++ b/rcl/src/rcl/time.c
@@ -73,6 +73,22 @@ rcl_get_ros_time(void * data, rcl_time_point_value_t * current_time)
 }
 
 bool
+rcl_time_point_value_valid(rcl_time_point_value_t time_point_value)
+{
+  return time_point_value > 0;
+}
+
+bool
+rcl_time_point_valid(rcl_time_point_t * time_point)
+{
+  if (time_point == NULL)
+  {
+    return false;
+  }
+  return rcl_time_point_value_valid(time_point->nanoseconds);
+}
+
+bool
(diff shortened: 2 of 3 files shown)


methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17606/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12147/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18176/)

sloretz (commented):
LGTM once linter is happy

rcl/src/rcl/time.c
```diff
@@ -72,6 +72,22 @@ rcl_get_ros_time(void * data, rcl_time_point_value_t * current_time)
   return RCL_RET_OK;
 }
 
+bool
+rcl_time_point_value_valid(rcl_time_point_value_t time_point_value)
```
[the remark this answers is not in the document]
methylDragon: Interestingly enough, [`rcl_clock_get_now`](https://github.com/ros2/rcl/blob/4eccc3cbe65f5b58981f22efaf612a65731cb2f0/rcl/src/rcl/time.c#L253) only allows you to get an `rcl_time_point_value_t`, so I was provisioning this one for that usage path.

methylDragon:

sloretz (approved):

rcl/test/rcl/test_time.cpp
```diff
@@ -205,6 +205,29 @@ TEST_F(CLASSNAME(TestTimeFixture, RMW_IMPLEMENTATION), test_ros_clock_initially_
   EXPECT_EQ(0, query_now);
 }
 
+TEST(CLASSNAME(rcl_time, RMW_IMPLEMENTATION), time_point_validation) {
+  ASSERT_FALSE(rcl_time_point_valid(NULL));
+
+  rcl_time_point_t uninitialized;
+  (void) uninitialized;
+  // Not reliably detectable due to random values.
+  // ASSERT_FALSE(rcl_clock_valid(&uninitialized));
```
wjwwood, line 214: As we talked about in person @methylDragon, we should just not have stuff in the tests like this (I know you're copying from the clock tests, but I consider them to be not ideal as well) and instead we should document that using uninitialized values with any of our functions is undefined behavior, e.g.:

https://github.com/ros2/rcl/blob/e47ed58d8d2018ee0fdb309b130bffa1e18c0c3d/rcl/include/rcl/node.h#L96-L99

Node is just an example, we should ideally document that all structures have to be zero initialized before being used (unless we have specific cases where we want to avoid that for performance reasons).

I don't see the `rcl_..._zero_initialized_...()` like functions for time/clock objects. We should probably clean them up to have these, again unless there are specific scenarios where we want to avoid that for performance reasons. Like it might be ok to use an uninitialized time point if passing it to get time (which is write only maybe). Avoiding an unnecessary initialization state. But functions that read the struct, like "is valid" should definitely be documented to not work with uninitialized values. In which case we don't need commented blocks like these in the tests.

rcl/src/rcl/time.c
```diff
@@ -72,6 +72,21 @@ rcl_get_ros_time(void * data, rcl_time_point_value_t * current_time)
...
+{
+  return time_point_value > 0;
```
fujitatomoya, line 78: I think this implementation should be in `rcutils`?

rcl/include/rcl/time.h
```diff
@@ -168,6 +168,49 @@ typedef struct rcl_time_point_s
...
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+bool
+rcl_time_point_value_valid(rcl_time_point_value_t time_point_value);
+
+/// Check if the time point is valid.
+/**
+ * This function returns true if the time point appears to be valid.
+ * It will check that the type is not uninitialized, and that the time point value is non-zero.
```
fujitatomoya, line 195: it does not seem to check the `clock_type` is initialized or not...?

rcl/src/rcl/time.c
```diff
@@ -72,6 +72,21 @@ rcl_get_ros_time(void * data, rcl_time_point_value_t * current_time)
...
+}
+
+bool
+rcl_time_point_valid(rcl_time_point_t * time_point)
```
fujitatomoya, line 82: even if `rcl_clock_type_t` is `RCL_CLOCK_UNINITIALIZED`, this will return true if `time_point->nanoseconds` is greater than zero.

rcl/test/rcl/test_time.cpp
```diff
@@ -205,6 +205,29 @@ TEST_F(CLASSNAME(TestTimeFixture, RMW_IMPLEMENTATION), test_ros_clock_initially_
...
+
+  rcl_time_point_t valid, invalid;
+
+  valid.nanoseconds = 1000;
+  valid.clock_type = RCL_ROS_TIME;
```
fujitatomoya, line 219: missing test case for `RCL_CLOCK_UNINITIALIZED`, so that we can detect that implementation is not completed?

fujitatomoya (commented):
according to the comment https://github.com/ros2/rcl/pull/1018#discussion_r1017107275, i think you already have consensus of this change, right. that is okay but it seems that this needs to be fixed.

PR #1021, opened 2022-12-05
Implement rcl_clock_time_started and revert #1018

methylDragon:
Signed-off-by: methylDragon <methylDragon@gmail.com>

PR #1022, opened 2022-12-05
Revert "Implement validity checks for time points (#1018)"

methylDragon:
Splitting out this change from https://github.com/ros2/rcl/pull/1021

This was from an internal discussion that concluded that checking for time validity on a time point doesn't make sense (std::chrono's time points have no concept of validity, a time point reflecting (0, 0) is valid.) Instead, what matters is if a clock has a time point or not.

### DIFF
--- a/rcl/include/rcl/time.h
+++ b/rcl/include/rcl/time.h
@@ -168,49 +168,6 @@ typedef struct rcl_time_point_s
 // } rcl_rate_t;
 // TODO(tfoote) integrate rate and timer implementations
 
-/// Check if the time point value is valid.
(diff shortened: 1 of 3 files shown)


methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17751/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12293/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18397/)

fujitatomoya (approved):

PR #1027, opened 2023-01-17
Calculate type hash from TypeDescription (rep2011)

emersonknapp:
Related to ros-infrastructure/rep#358
Part of ros2/ros2#1159 
Depends on ros2/rosidl#729

This PR introduces a utility function to provide a stable reproducible hash for `type_description_interfaces/msg/TypeDescription` for use in type comparison/advertising in REP-2011

### DIFF
--- a/rcl/include/rcl/type_hash.h
+++ b/rcl/include/rcl/type_hash.h
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef RCL__TYPE_VERSION_HASH_H_
-#define RCL__TYPE_VERSION_HASH_H_
+#ifndef RCL__TYPE_HASH_H_
+#define RCL__TYPE_HASH_H_
 
 #ifdef __cplusplus
 extern "C"
@@ -28,7 +28,7 @@ extern "C"
 
 /// Given a TypeDescription, output a string of the hashable JSON representation of that data.
 /**
- * The output here is generally hashed via rcl_calculate_type_version_hash below.
+ * The output here is generally hashed via rcl_calculate_type_hash below.
  * Compare this reference implementation with rosidl_parser.type_hash.idl_to_hashable_json.
  * Both must produce the same output for the same types, providing a stable reference for
  * external implementations of the ROS 2 Type Version Hash.
@@ -65,7 +65,7 @@ rcl_type_description_to_hashable_json(
  */
 RCL_PUBLIC
 rcl_ret_t
-rcl_calculate_type_version_hash(
+rcl_calculate_type_hash(
   const type_description_interfaces__msg__TypeDescription * type_description,
   rosidl_type_hash_t * out_type_hash);
 
@@ -73,4 +73,4 @@ rcl_calculate_type_version_hash(
 }
 #endif
 
-#endif  // RCL__TYPE_VERSION_HASH_H_
+#endif  // RCL__TYPE_HASH_H_
--- a/rcl/include/rcl/type_hash.h
+++ b/rcl/include/rcl/type_hash.h
@@ -28,8 +28,9 @@ extern "C"
 
 /// Given a TypeDescription, output a string of the hashable JSON representation of that data.
 /**
- * The output here is generally hashed via rcl_calculate_type_hash below.
- * Compare this reference implementation with rosidl_parser.type_hash.idl_to_hashable_json.
+ * The output here is generally hashed via rcl_calculate_type_hash() below.
+ * Compare this reference implementation with the .json output files from
+ * `rosidl_generator_type_description.generate_type_hash`.
  * Both must produce the same output for the same types, providing a stable reference for
  * external implementations of the ROS 2 Type Version Hash.
  *
--- a/rcl/include/rcl/type_hash.h
+++ b/rcl/include/rcl/type_hash.h
@@ -43,7 +43,8 @@ extern "C"
  *   the JSON representation of type_description. Note that output_repr will have a
  *   terminating null character, which should be omitted from hashing. To do so, use
  *   (output_repr.buffer_length - 1) or strlen(output_repr.buffer) for the size of data to hash.
- * \return RCL_RET_OK on success, RCL_RET_ERROR if any problems occur in translation
+ * \return #RCL_RET_OK on success, or
+ * \return #RCL_RET_ERROR if any problems occur in translation.
  */
 RCL_PUBLIC
 rcl_ret_t
@@ -61,7 +62,8 @@ rcl_type_description_to_hashable_json(
  *
  * \param[in] msg Prefilled TypeDescription message describing the type to be hashed
  * \param[out] message_digest Preallocated buffer, to be filled with calculated checksum
- * \return RCL_RET_OK on success, RCL_RET_ERROR if any problems occur while hashing
+ * \return #RCL_RET_OK on success, or
+ * \return #RCL_RET_ERROR if any problems occur while hashing.
  */
 RCL_PUBLIC
 rcl_ret_t
--- a/rcl/include/rcl/type_hash.h
+++ b/rcl/include/rcl/type_hash.h
@@ -57,7 +57,7 @@ rcl_type_description_to_hashable_json(
  * For design motivations leading to this implementation, see REP-2011.
  *
  * This convenience wrapper calls rcl_type_description_to_hashable_json,
- * then runs sha256 hash on the result
+ * then runs sha256 hash on the result.
  *
  * \param[in] msg Prefilled TypeDescription message describing the type to be hashed
  * \param[out] message_digest Preallocated buffer, to be filled with calculated checksum
--- a/rcl/include/rcl/type_version_hash.h
+++ b/rcl/include/rcl/type_version_hash.h
@@ -1,0 +1,76 @@
+// Copyright 2023 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCL__TYPE_VERSION_HASH_H_
+#define RCL__TYPE_VERSION_HASH_H_
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#include "rcl/types.h"
+#include "rcl/visibility_control.h"
+#include "rcutils/sha256.h"
+#include "rosidl_runtime_c/type_hash.h"
+#include "type_description_interfaces/msg/type_description.h"
+
+/// Given a TypeDescription, output a string of the hashable JSON representation of that data.
+/**
+ * The output here is generally hashed via rcl_calculate_type_version_hash below.
+ * Compare this reference implementation with rosidl_parser.type_hash.idl_to_hashable_json.
+ * Both must produce the same output for the same types, providing a stable reference for
+ * external implementations of the ROS 2 Type Version Hash.
+ *
+ * The JSON representation contains all types and fields of the original message, but excludes:
+ * - Default values
+ * - Comments
+ * - The input plaintext files that generated the TypeDescription
+ *
+ * \param[in] type_description Prefilled TypeDescription message to be translated
+ * \param[out] output_repr An initialized empty char array that will be filled with
+ *   the JSON representation of type_description. Note that output_repr will have a
+ *   terminating null character, which should be omitted from hashing. To do so, use
+ *   (output_repr.buffer_length - 1) or strlen(output_repr.buffer) for the size of data to hash.
+ * \return RCL_RET_OK on success, RCL_RET_ERROR if any problems occur in translation
+ */
+RCL_PUBLIC
+rcl_ret_t
+rcl_type_description_to_hashable_json(
+  const type_description_interfaces__msg__TypeDescription * type_description,
+  rcutils_char_array_t * output_repr);
+
+/// Calculate the Type Version Hash for a given TypeDescription.
+/**
+ * This function produces a stable hash value for a ROS communication interface type.
+ * For design motivations leading to this implementation, see REP-2011.
+ *
+ * This convenience wrapper calls rcl_type_description_to_hashable_json,
+ * then runs sha256 hash on the result
+ *
+ * \param[in] msg Prefilled TypeDescription message describing the type to be hashed
+ * \param[out] message_digest Preallocated buffer, to be filled with calculated checksum
+ * \return RCL_RET_OK on success, RCL_RET_ERROR if any problems occur while hashing
+ */
+RCL_PUBLIC
+rcl_ret_t
+rcl_calculate_type_version_hash(
+  const type_description_interfaces__msg__TypeDescription * type_description,
+  rosidl_type_hash_t * out_type_hash);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif  // RCL__TYPE_VERSION_HASH_H_
--- a/rcl/src/rcl/type_hash.c
+++ b/rcl/src/rcl/type_hash.c
@@ -67,7 +67,9 @@ static int emit_key(yaml_emitter_t * emitter, const char * key)
   yaml_event_t event;
   return
     yaml_scalar_event_initialize(
-    &event, NULL, NULL, (yaml_char_t *)key, strlen(key), 0, 1, YAML_DOUBLE_QUOTED_SCALAR_STYLE) &&
+    &event, NULL, NULL,
+    (yaml_char_t *)key, (int)strlen(key),
+    0, 1, YAML_DOUBLE_QUOTED_SCALAR_STYLE) &&
     yaml_emitter_emit(emitter, &event);
 }
 
@@ -88,7 +90,7 @@ static int emit_int(yaml_emitter_t * emitter, size_t val, const char * fmt)
   return
     yaml_scalar_event_initialize(
     &event, NULL, NULL,
-    (yaml_char_t *)decimal_buf, strlen(decimal_buf),
+    (yaml_char_t *)decimal_buf, (int)strlen(decimal_buf),
     1, 0, YAML_PLAIN_SCALAR_STYLE) &&
     yaml_emitter_emit(emitter, &event);
 }
@@ -99,7 +101,7 @@ static int emit_str(yaml_emitter_t * emitter, const rosidl_runtime_c__String * v
   return
     yaml_scalar_event_initialize(
     &event, NULL, NULL,
-    (yaml_char_t *)val->data, val->size,
+    (yaml_char_t *)val->data, (int)val->size,
     0, 1, YAML_DOUBLE_QUOTED_SCALAR_STYLE) &&
     yaml_emitter_emit(emitter, &event);
 }
--- a/rcl/src/rcl/type_hash.c
+++ b/rcl/src/rcl/type_hash.c
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <stdio.h>
+
 #include <yaml.h>
 
 #include "rcl/allocator.h"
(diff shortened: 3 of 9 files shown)


rcl/include/rcl/type_version_hash.h
```diff
@@ -0,0 +1,70 @@
...
+extern "C"
+{
+#endif
+
+#include "rcl/types.h"
+#include "rcl/visibility_control.h"
+#include "rcutils/sha256.h"
+#include "type_description_interfaces/msg/type_description.h"
+
+/// Convert
```
clalancette, line 28: Can we make this more descriptive?  Something like:

```suggestion
/// Given a TypeDescription message, output a string of the hashable JSON data.
```

rcl/include/rcl/type_version_hash.h
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Compare this reference implementation with rosidl_parser.type_hash.idl_to_hashable_json.
+ * Both must produce the same output for the same types, providing a stable reference for
+ * external implementations of the ROS 2 Type Version Hash.
```
clalancette, line 32:
```suggestion
 * The output here is generally hashed via rcl_calculate_type_version_hash below.
 * Compare this reference implementation with rosidl_parser.type_hash.idl_to_hashable_json.
 * Both must produce the same output for the same types, providing a stable reference for
 * external implementations of the ROS 2 Type Version Hash.
```

rcl/include/rcl/type_version_hash.h
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * Excludes:
+ * - Default values
+ * - The input plaintext IDL files that generated the TypeDescription
```
clalancette, line 36:
```suggestion
 * The JSON representation contains all types and fields of the original message, but excludes:
 * - Default values
 * - Comments
 * - The input plaintext IDL files that generated the TypeDescription
```

rcl/src/rcl/type_version_hash.c
```diff
@@ -0,0 +1,257 @@
...
+
+  // type_name
+  EMIT_KEY("type_name");
+  EMIT_STR_VAL(
+    individual_type_description->type_name.data,
+    individual_type_description->type_name.size);
+
+  // fields
+  EMIT_KEY("fields");
+  {
```
clalancette, line 134: Why the extra scope here?

rcl/src/rcl/type_version_hash.c
```diff
@@ -0,0 +1,257 @@
...
+  const type_description_interfaces__msg__TypeDescription * type_description,
+  rcutils_char_array_t * output_repr)
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(type_description, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(output_repr, RCL_RET_INVALID_ARGUMENT);
+
+  yaml_emitter_t emitter;
+  yaml_event_t event;
+
+  if (!yaml_emitter_initialize(&emitter)) {goto error;}
```
clalancette, line 190: Style:

```suggestion
  if (!yaml_emitter_initialize(&emitter)) {
    goto error;
  }
```

clalancette: (same a few times below)

rcl/src/rcl/type_version_hash.c
```diff
@@ -0,0 +1,257 @@
...
+  if (
+    !yaml_document_end_event_initialize(&event, 1) ||
+    !yaml_emitter_emit(&emitter, &event)) {goto error;}
+  if (
+    !yaml_stream_end_event_initialize(&event) ||
+    !yaml_emitter_emit(&emitter, &event)) {goto error;}
+
+  yaml_emitter_delete(&emitter);
+  // Remove null-terminator character which should not be part of hashed content
+  output_repr->buffer_length--;
```
clalancette, line 216: I'm not sure that I agree with doing this here.  For one thing, this likely messes up the `rcutils_char_array` internal bookkeeping.  For another, this is probably surprising to external users of this API (i.e. not `rcl_calculate_type_version_hash`).

My suggestion here is to leave the `rcutils_char_array` as-is in this function, and deal with removing the `\0` in `rcl_calculate_type_version_hash`.

rcl/src/rcl/type_version_hash.c
```diff
@@ -0,0 +1,257 @@
...
+  rcutils_sha256_update(&sha_ctx, (const uint8_t *)msg_repr.buffer, msg_repr.buffer_length);
+  rcutils_sha256_final(&sha_ctx, output_hash);
+
+  result = rcutils_char_array_fini(&msg_repr);
+  return result;
+
+error:
+  result = rcutils_char_array_fini(&msg_repr);
+  return RCL_RET_ERROR;
+}
```
clalancette, line 253: This can probably be simplified to:

```suggestion
  if (result == RCL_RET_OK) {
    rcutils_sha256_ctx_t sha_ctx;
    rcutils_sha256_init(&sha_ctx);
    // Last item in char_array is null terminator, which should not be hashed.
    rcutils_sha256_update(&sha_ctx, (const uint8_t *)msg_repr.buffer, msg_repr.buffer_length);
    rcutils_sha256_final(&sha_ctx, output_hash);
  }
  result = rcutils_char_array_fini(&msg_repr);
  return result;
}
```

rcl/src/rcl/type_version_hash.c
```diff
@@ -0,0 +1,257 @@
...
+    char decimal_buf[21]; \
+    FAILFAST( \
+      (size_t)snprintf(decimal_buf, sizeof(decimal_buf), fmt, value) < sizeof(decimal_buf)); \
+    FAILFAST( \
+      yaml_scalar_event_initialize( \
+        &event, NULL, NULL, \
+        (yaml_char_t *)decimal_buf, strlen(decimal_buf), \
+        1, 0, YAML_PLAIN_SCALAR_STYLE)); \
+    EMIT(); \
+}
```
clalancette, line 63: I have to say I'm not a huge fan of all of these macros. I know this is how we've done it in some other places, but macros like this just make reading and debugging this stuff hard later on. It might be good to figure out if we can avoid at least some of the macros here. But I don't have a great alternative, so this isn't a blocking comment from me.

rcl/src/rcl/type_version_hash.c
```diff
@@ -0,0 +1,257 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
```
clalancette, line 18: This is not necessary in the definition, just the declaration. So just remove it here.

rcl/include/rcl/type_version_hash.h
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * This function produces a stable hash value for a ROS communication interface type.
+ * For design motivations leading to this implementation, see REP-2011.
+ *
+ * This convenience wrapper calls rcl_type_description_to_hashable_json,
+ * then runs sha256 hash on the result
+ *
+ * \param[in] msg Prefilled TypeDescription message describing the type to be hashed.
+ * \param[out] message_digest Preallocated buffer, filled with calculated checksum calculated.
+ * \return
```
clalancette, line 58: The return value should be filled out more completely.

rcl/include/rcl/type_version_hash.h
```diff
@@ -0,0 +1,70 @@
...
+ * Both must produce the same output for the same types, providing a stable reference for
+ * external implementations of the ROS 2 Type Version Hash.
+ *
+ * Excludes:
+ * - Default values
+ * - The input plaintext IDL files that generated the TypeDescription
+ *
+ * \param[in] type_description
+ * \param[out] output_repr
+ * \return
```
clalancette, line 40: These should all be filled in more completely.

clalancette (changes requested):
I've left a few suggestions for improvement inline, along with some questions.

Besides that, this PR should be retargeted to `rolling`, which is our default branch.  Also, this can't be merged until we merge in the `type_description_interfaces` package.

emersonknapp:
Changed the base branch, Yes, I've marked the dependency in the description - though github doesn't understand that language, it has only the "closes" behavior semantic for cross-linking issues.

Addressing other comments shortly.

clalancette:
> Changed the base branch, Yes, I've marked the dependency in the description - though github doesn't understand that language, it has only the "closes" behavior semantic for cross-linking issues.

Thanks.  Yeah, it would be nice if we could express these dependencies in GitHub somehow.  The good news is that this will totally fail to compile in CI until `type_description_interfaces` is available, so we should be OK.

rcl/src/rcl/type_version_hash.c
emersonknapp: I will note that it does not mess up the internal bookkeeping - there are two values, `buffer_capacity`, and `buffer_length`. Length marks how long the content currently is and can be manipulated arbitrarily, so long as it is not made longer than the capacity. The capacity on the other hand, needs to be carefully modified via APIs. There are no such APIs for modification of the length.

However, it's fine with me to shift the responsibility - that's how I had it in my first draft. The end user just has to know to use either `strlen(msg_repr.buffer`), or `msg_repr.buffer_length - 1` for the size of data to hash. I will note it in the docstring

rcl/src/rcl/type_version_hash.c
emersonknapp: Part of me is perversely disappointed to note that my updated version has actually shortened the file length by one line. It seems this is a better pattern, let me know what you think of it now.

rcl/src/rcl/type_version_hash.c
emersonknapp: no longer relevant with rewrite

rcl/src/rcl/type_version_hash.c
```diff
@@ -0,0 +1,256 @@
...
+  yaml_event_t event;
+  return
+    yaml_scalar_event_initialize(
+    &event, NULL, NULL, (yaml_char_t *)key, strlen(key), 0, 1, YAML_DOUBLE_QUOTED_SCALAR_STYLE) &&
+    yaml_emitter_emit(emitter, &event);
+}
+
+static int emit_int(yaml_emitter_t * emitter, size_t val, const char * fmt)
+{
+  // longest uint64 is 21 decimal digits
```
clalancette, line 74: More accurately, the longest uint64 is 20 decimal digits, with one byte for the trailing `\0`.

rcl/src/rcl/type_version_hash.c
```diff
@@ -0,0 +1,256 @@
...
+  char decimal_buf[21];
+  yaml_event_t event;
+  if ((size_t)snprintf(decimal_buf, sizeof(decimal_buf), fmt, val) >= sizeof(decimal_buf)) {
+    emitter->problem = "Decimal buffer overflow";
+    return 0;
+  }
```
clalancette, line 80: `snprintf` can return a negative result on certain errors (other than truncation).  While this check works because that would get promoted to a very large `uint64`, it is somewhat unintuitive.  I'll suggest:

```suggestion
  int ret = snprintf(decimal_buf, sizeof(decimal_buf), fmt, val);
  if (ret < 0) {
    emitter->problem = "Failed expanding integer";
    return 0;
  }
  if ((size_t)ret >= sizeof(decimal_buf)) {
    emitter->problem = "Decimal buffer overflow";
    return 0;
  }
```

emersonknapp:
OK - I think I've addressed everything. I appreciate the feedback, round 2 welcome :)

clalancette (changes requested):
Overall, I *love* this new structure; so much better than the macros.  Thanks for taking the time to investigate this!

I've left a couple of minor comments inline; otherwise, I think this is ready to go modulo getting the `type_description_interface` package in.

emersonknapp:
Moving this back to draft while working out the type hash struct - which this calculator should also provide

rcl/test/rcl/test_type_version_hash.cpp
```diff
@@ -0,0 +1,212 @@
...
+  0xaf, 0x49, 0x7e, 0x7c, 0x94, 0x3f, 0xf7, 0xf8,
+  0xdb, 0x1d, 0x50, 0x1d, 0xb0, 0x9d, 0x0a, 0xad,
+  0x44, 0x12, 0x4e, 0xd7, 0xdf, 0x43, 0xc1, 0xe8,
+  0x6c, 0x64, 0x26, 0x02, 0xd0, 0x6c, 0xbb, 0x7c,
+}};
+
+static void init_individual_type_description(
+  type_description_interfaces__msg__IndividualTypeDescription * itd,
+  const char * name,
+  const std::vector<std::tuple<const char *, uint8_t, const char *>> & fields)
```
clalancette, line 35: Nit: `#include <tuple>` and `#include <vector>` at the top of the file.

rcl/test/rcl/test_type_version_hash.cpp
```diff
@@ -0,0 +1,212 @@
...
+    rcutils_sha256_final(&sha, hash_from_repr.value);
+
+    res = rcutils_char_array_fini(&msg_repr);
+    ASSERT_EQ(res, RCL_RET_OK);
+  }
+
+  auto validation_hash = sensor_msgs__msg__PointCloud2__TYPE_VERSION_HASH__copy;
+  ASSERT_EQ(direct_hash.version, hash_from_repr.version);
+  ASSERT_EQ(direct_hash.version, validation_hash.version);
+  ASSERT_EQ(0, memcmp(direct_hash.value, hash_from_repr.value, ROSIDL_TYPE_HASH_SIZE));
```
clalancette, line 210: Nit: `#include <cstring>` at the top of the file for this.

clalancette (changes requested):
I left a few more nits to fix here, otherwise looks pretty good to me.

emersonknapp:
@clalancette thanks for the review pass - I think the plan might be to move the functionality of this utility down to `rosidl_runtime_c` so that @methylDragon could use it below the `rmw` boundary (please weigh in Brandon if that's true or not) However that shouldn't affect the code, probably just a few types from like `rcl_ret_t` -> `rcutils_ret_t` . That would also have the nice side effect of containing the two hashing implementations both in the `rosidl` repo. Maybe then we'd want to expose it up here as a redefine or wrapper, not sure yet.

methylDragon:
> @clalancette thanks for the review pass - I think the plan might be to move the functionality of this utility down to `rosidl_runtime_c` so that @methylDragon could use it below the `rmw` boundary (please weigh in Brandon if that's true or not) However that shouldn't affect the code, probably just a few types from like `rcl_ret_t` -> `rcutils_ret_t` . That would also have the nice side effect of containing the two hashing implementations both in the `rosidl` repo. Maybe then we'd want to expose it up here as a redefine or wrapper, not sure yet.

Just noting here for posterity.
From our internal discussions yesterday, we determined that the actual hash structs don't need to go into the rmw layer (as long as the hash string itself can be passed down.) So I think it's fine for this struct to live in `rcl`!

rcl/include/rcl/type_hash.h
```diff
@@ -0,0 +1,76 @@
...
+ * - Default values
+ * - Comments
+ * - The input plaintext files that generated the TypeDescription
+ *
+ * \param[in] type_description Prefilled TypeDescription message to be translated
+ * \param[out] output_repr An initialized empty char array that will be filled with
+ *   the JSON representation of type_description. Note that output_repr will have a
+ *   terminating null character, which should be omitted from hashing. To do so, use
+ *   (output_repr.buffer_length - 1) or strlen(output_repr.buffer) for the size of data to hash.
+ * \return RCL_RET_OK on success, RCL_RET_ERROR if any problems occur in translation
```
wjwwood, line 46: nitpick: you might want to have a look at how the other rcl functions document return types, I think they use one per `\return` and wrap the return types in ``` `RCL_RET_...` ``` for a reason, but I can't remember them off-hand. Thought I do think it's best to just emulate them.

rcl/src/rcl/type_hash.c
```diff
@@ -0,0 +1,264 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <stdio.h>
+
+#include <yaml.h>
```
wjwwood, line 17: This concerns me, I think it's coming in implicitly through `rcl_yaml_param_parser`. I actually don't love depending on libyaml from rcl proper like this, though the effect is the same as the current situation, since we hard depend on the yaml stuff in `rcl_yaml_param_parser`.

wjwwood: I don't think we should do something different, I think we should use libyaml, but I do think we need to explicitly depend on libyaml, via package.xml and CMake.

emersonknapp: Done

rcl/include/rcl/type_hash.h
emersonknapp: Done (they use 
```
\return #RCL_RET_... for reason, or
\return #RCL_RET_... for a final reason, period.
``` 

so I did that

clalancette (changes requested):
Two small comment updates, then this looks good to me.

rcl/include/rcl/type_hash.h
```diff
@@ -0,0 +1,78 @@
...
+#include "rcl/types.h"
+#include "rcl/visibility_control.h"
+#include "rcutils/sha256.h"
+#include "rosidl_runtime_c/type_hash.h"
+#include "type_description_interfaces/msg/type_description.h"
+
+/// Given a TypeDescription, output a string of the hashable JSON representation of that data.
+/**
+ * The output here is generally hashed via rcl_calculate_type_hash below.
+ * Compare this reference implementation with rosidl_parser.type_hash.idl_to_hashable_json.
```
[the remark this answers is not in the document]
emersonknapp: Good catch - this was an old comment at this point

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/ff276a843492f27fd1afac2806fdb849/raw/e273147f65788680698e21401ce10184f120cdfd/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11684
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18325/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12847/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18991/)

clalancette (approved):

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/e163a4a54a5fa58279b131748d54a31b/raw/e273147f65788680698e21401ce10184f120cdfd/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11688
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18331/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12852/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18998/)

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18344/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12865/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19012/)

I noticed that the other project using this dependency, `rcl_yaml_param_parser`, switched over to using `target_link_libraries` away from `ament_target_dependencies`. That allowed it to hide `yaml` behind a `PRIVATE` dependency. I was hesitant to try that conversion and confuse the rest of this review, though

clalancette:
> I noticed that the other project using this dependency, `rcl_yaml_param_parser`, switched over to using `target_link_libraries` away from `ament_target_dependencies`. That allowed it to hide `yaml` behind a `PRIVATE` dependency. I was hesitant to try that conversion and confuse the rest of this review, though

I *almost* think it is worth it, but there is enough going on here that we can punt it.  Assuming what you did already fixes the problem for now, I'll do a follow-up where we switch to `target_link_libraries`, since it makes it a lot nicer.

emersonknapp:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19019/)

Issue #977, opened 2022-04-11
Improve performance for ROS_DISABLE_LOANED_MESSAGES


fujitatomoya:
#### Description

originally this came up with https://github.com/ros2/rosbag2/pull/981#issuecomment-1095330189

reading environmental variable every single time to publish the data is not cost effective for performance.

#### Implementation considerations

- extend `rcl_publisher_t` and `rcl_subscription_t` to cache the force flag to disable `LoanedMessages`.
- read environmental variable just once for corresponding publisher or subscription construction in `rcl`.
- return false immediately in `rcl_publisher_can_loan_messages` and `rcl_subscription_can_loan_messages`.

#### Related Issues / PRs

- https://github.com/ros2/rcl/issues/945
- https://github.com/ros2/rcl/pull/949
- https://github.com/ros2/rosbag2/pull/981

PR #1029, opened 2023-01-20
Cache disable flag to avoid reading environmental variable.

fujitatomoya:
address https://github.com/ros2/rcl/issues/977

### DIFF
--- a/rcl/include/rcl/publisher.h
+++ b/rcl/include/rcl/publisher.h
@@ -49,6 +49,8 @@ typedef struct rcl_publisher_options_s
   rcl_allocator_t allocator;
   /// rmw specific publisher options, e.g. the rmw implementation specific payload.
   rmw_publisher_options_t rmw_publisher_options;
+  /// Disable flag to LoanedMessage, initialized via environmental variable.
+  bool disable_loaned_message;
 } rcl_publisher_options_t;
 
 /// Return a rcl_publisher_t struct with members set to `NULL`.
--- a/rcl/include/rcl/publisher.h
+++ b/rcl/include/rcl/publisher.h
@@ -37,6 +37,8 @@ typedef struct rcl_publisher_s
 {
   /// Pointer to the publisher implementation
   rcl_publisher_impl_t * impl;
+  /// Disable flag to loan messages
+  bool disable_loaned_message;
 } rcl_publisher_t;
 
 /// Options available for a rcl publisher.
--- a/rcl/include/rcl/publisher.h
+++ b/rcl/include/rcl/publisher.h
@@ -37,8 +37,6 @@ typedef struct rcl_publisher_s
 {
   /// Pointer to the publisher implementation
   rcl_publisher_impl_t * impl;
-  /// Disable flag to loan messages
-  bool disable_loaned_message;
 } rcl_publisher_t;
 
 /// Options available for a rcl publisher.
--- a/rcl/include/rcl/publisher.h
+++ b/rcl/include/rcl/publisher.h
@@ -196,6 +196,7 @@ rcl_publisher_fini(rcl_publisher_t * publisher, rcl_node_t * node);
  * - qos = rmw_qos_profile_default
  * - allocator = rcl_get_default_allocator()
  * - rmw_publisher_options = rmw_get_default_publisher_options()
+ * - disable_loaned_message = false, true only if ROS_DISABLE_LOANED_MESSAGES=1
  *
  * \return A structure with the default publisher options.
  */
--- a/rcl/src/rcl/publisher.c
+++ b/rcl/src/rcl/publisher.c
@@ -203,6 +203,16 @@ rcl_publisher_get_default_options()
   default_options.qos = rmw_qos_profile_default;
   default_options.allocator = rcl_get_default_allocator();
   default_options.rmw_publisher_options = rmw_get_default_publisher_options();
+
+  // Load disable flag to LoanedMessage via environmental variable.
+  bool disable_loaned_message = false;
+  rcl_ret_t ret = rcl_get_disable_loaned_message(&disable_loaned_message);
+  if (ret == RCL_RET_OK) {
+    default_options.disable_loaned_message = disable_loaned_message;
+  } else {
+    default_options.disable_loaned_message = false;
+  }
+
   return default_options;
 }
 
@@ -441,9 +451,7 @@ rcl_publisher_can_loan_messages(const rcl_publisher_t * publisher)
     return false;  // error message already set
   }
 
-  bool disable_loaned_message = false;
-  rcl_ret_t ret = rcl_get_disable_loaned_message(&disable_loaned_message);
-  if (ret == RCL_RET_OK && disable_loaned_message) {
+  if (publisher->impl->options.disable_loaned_message) {
     return false;
   }
 
--- a/rcl/src/rcl/publisher.c
+++ b/rcl/src/rcl/publisher.c
@@ -210,6 +210,9 @@ rcl_publisher_get_default_options()
   if (ret == RCL_RET_OK) {
     default_options.disable_loaned_message = disable_loaned_message;
   } else {
+    RCUTILS_SAFE_FWRITE_TO_STDERR("Failed to get disable_loaned_message: ");
+    RCUTILS_SAFE_FWRITE_TO_STDERR(rcl_get_error_string().str);
+    rcl_reset_error();
     default_options.disable_loaned_message = false;
   }
 
--- a/rcl/src/rcl/publisher.c
+++ b/rcl/src/rcl/publisher.c
@@ -104,13 +104,6 @@ rcl_publisher_init(
   RCL_CHECK_FOR_NULL_WITH_MSG(
     publisher->impl, "allocating memory failed", ret = RCL_RET_BAD_ALLOC; goto cleanup);
 
-  // Cache disable flag to loan messasges on this publisher
-  bool disable_loaned_message = false;
-  ret = rcl_get_disable_loaned_message(&disable_loaned_message);
-  if (ret == RCL_RET_OK && disable_loaned_message) {
-    publisher->disable_loaned_message = true;
-  }
-
   // Fill out implementation struct.
   // rmw handle (create rmw publisher)
   // TODO(wjwwood): pass along the allocator to rmw when it supports it
@@ -448,7 +441,9 @@ rcl_publisher_can_loan_messages(const rcl_publisher_t * publisher)
     return false;  // error message already set
   }
 
-  if (publisher->disable_loaned_message) {
+  bool disable_loaned_message = false;
+  rcl_ret_t ret = rcl_get_disable_loaned_message(&disable_loaned_message);
+  if (ret == RCL_RET_OK && disable_loaned_message) {
     return false;
   }
 
--- a/rcl/src/rcl/publisher.c
+++ b/rcl/src/rcl/publisher.c
@@ -104,6 +104,13 @@ rcl_publisher_init(
   RCL_CHECK_FOR_NULL_WITH_MSG(
     publisher->impl, "allocating memory failed", ret = RCL_RET_BAD_ALLOC; goto cleanup);
 
+  // Cache disable flag to loan messasges on this publisher
+  bool disable_loaned_message = false;
+  ret = rcl_get_disable_loaned_message(&disable_loaned_message);
+  if (ret == RCL_RET_OK && disable_loaned_message) {
+    publisher->disable_loaned_message = true;
+  }
+
   // Fill out implementation struct.
   // rmw handle (create rmw publisher)
   // TODO(wjwwood): pass along the allocator to rmw when it supports it
@@ -441,9 +448,7 @@ rcl_publisher_can_loan_messages(const rcl_publisher_t * publisher)
     return false;  // error message already set
   }
 
-  bool disable_loaned_message = false;
-  rcl_ret_t ret = rcl_get_disable_loaned_message(&disable_loaned_message);
-  if (ret == RCL_RET_OK && disable_loaned_message) {
+  if (publisher->disable_loaned_message) {
     return false;
   }
 
--- a/rcl/src/rcl/subscription.c
+++ b/rcl/src/rcl/subscription.c
@@ -95,6 +95,14 @@ rcl_subscription_init(
     1, sizeof(rcl_subscription_impl_t), allocator->state);
   RCL_CHECK_FOR_NULL_WITH_MSG(
     subscription->impl, "allocating memory failed", ret = RCL_RET_BAD_ALLOC; goto cleanup);
+
+  // Cache disable flag to loan messasges on this subscription
+  bool disable_loaned_message = false;
+  ret = rcl_get_disable_loaned_message(&disable_loaned_message);
+  if (ret == RCL_RET_OK && disable_loaned_message) {
+    subscription->disable_loaned_message = true;
+  }
+
   // Fill out the implemenation struct.
   // rmw_handle
   // TODO(wjwwood): pass allocator once supported in rmw api.
@@ -736,9 +744,7 @@ rcl_subscription_can_loan_messages(const rcl_subscription_t * subscription)
     return false;  // error message already set
   }
 
-  bool disable_loaned_message = false;
-  rcl_ret_t ret = rcl_get_disable_loaned_message(&disable_loaned_message);
-  if (ret == RCL_RET_OK && disable_loaned_message) {
+  if (subscription->disable_loaned_message) {
     return false;
   }
 
--- a/rcl/src/rcl/subscription.c
+++ b/rcl/src/rcl/subscription.c
@@ -206,6 +206,16 @@ rcl_subscription_get_default_options()
   default_options.qos = rmw_qos_profile_default;
   default_options.allocator = rcl_get_default_allocator();
   default_options.rmw_subscription_options = rmw_get_default_subscription_options();
+
+  // Load disable flag to LoanedMessage via environmental variable.
+  bool disable_loaned_message = false;
+  rcl_ret_t ret = rcl_get_disable_loaned_message(&disable_loaned_message);
+  if (ret == RCL_RET_OK) {
+    default_options.disable_loaned_message = disable_loaned_message;
+  } else {
+    default_options.disable_loaned_message = false;
+  }
+
   return default_options;
 }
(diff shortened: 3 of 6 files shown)


fujitatomoya:
@clalancette minor performance improvement. this can avoid reading environmental variable every time it publishes data with using `LoanedMessage`.

CC: @MichaelOrlov would be interested for you, as performance improvement.

rcl/include/rcl/publisher.h
```diff
@@ -37,6 +37,8 @@ typedef struct rcl_publisher_s
 {
   /// Pointer to the publisher implementation
   rcl_publisher_impl_t * impl;
+  /// Disable flag to loan messages
+  bool disable_loaned_message;
```
fujitatomoya, line 41: Or maybe this can be in `rcl_publisher_option`? currently we do not have this option via rclcpp, so integrated in `rcl` layer with environmental variable would be good enough?

rcl/src/rcl/publisher.c
```diff
@@ -104,6 +104,13 @@ rcl_publisher_init(
   RCL_CHECK_FOR_NULL_WITH_MSG(
     publisher->impl, "allocating memory failed", ret = RCL_RET_BAD_ALLOC; goto cleanup);
 
+  // Cache disable flag to loan messasges on this publisher
+  bool disable_loaned_message = false;
+  ret = rcl_get_disable_loaned_message(&disable_loaned_message);
+  if (ret == RCL_RET_OK && disable_loaned_message) {
+    publisher->disable_loaned_message = true;
+  }
```
clalancette, line 112: I think we need to deal with errors better here:

```suggestion
  // Cache disable flag to loan messages on this publisher
  bool disable_loaned_message = false;
  ret = rcl_get_disable_loaned_message(&disable_loaned_message);
  if (ret != RCL_RET_OK) {
    goto cleanup;
  }
  publisher->disable_loaned_message = disable_loaned_message;
```

Same in the subscription below.

rcl/include/rcl/publisher.h
clalancette: It would be nice to make this a publisher option, so we *could* control it via rclcpp if we wanted in the future.  I'd vote for moving it there.  Same goes for the subscription.

clalancette (changes requested):

rcl/src/rcl/publisher.c
fujitatomoya: thanks, this is a bug. i will fix it.

rcl/include/rcl/publisher.h
fujitatomoya: okay, so we will do.

- extend option for publisher and subscription that can be set by user application via rclcpp.
- environmental variable check during initialization for publisher and subscription, and if environmental disable flag is on, that prevails option setting in the argument.

MichaelOrlov: @fujitatomoya @clalancette I agree and it was my original proposal to add `disable_loaned_message` to the `rcl_publisher_options_t` and `rcl_subscription_options_t`.
Perhaps we can read out environment variable and init `disable_loaned_message` in `rcl_publisher_get_default_options()`
https://github.com/ros2/rcl/blob/d651a5cc77d3b81b1d779e611cae651238960206/rcl/src/rcl/publisher.c#L197-L207
and in `rcl_subscription_get_default_options()` https://github.com/ros2/rcl/blob/d651a5cc77d3b81b1d779e611cae651238960206/rcl/src/rcl/subscription.c#L200-L210
respectively.

fujitatomoya: addressed by https://github.com/ros2/rcl/pull/1029/commits/3b25f4b8fa1043d8c83c490d2d6d6c41f1780132

fujitatomoya:
@clalancette requesting 2nd review, modification for test code is bigger than fix itself...

fujitatomoya:
can we backport this humble? probably not, since c structure is extended, this is API compatibility but ABI?

fujitatomoya:
https://build.ros2.org/job/Rpr__rcl__ubuntu_jammy_amd64/154/ is related to this PR, but cannot reproduce this failure on my local environment.

fujitatomoya:
either @Barry-Xu-2018 or @iuhilnehc-ynos can you run the test with your local environment? I cannot reproduce the failure reported https://build.ros2.org/job/Rpr__rcl__ubuntu_jammy_amd64/159/.

```bash
colcon test --event-handlers console_direct+ --packages-select rcl
```

iuhilnehc-ynos:
@fujitatomoya 

I can reproduce this issue in the ros:rolling image with this PR. The failure in Rpc is because not using the latest rmw_fastrtps.
Please use CI directly.

fujitatomoya:
> It means the ros-rolling-rmw-fastrtps-cpp should be updated in http://repo.ros2.org/ubuntu/testing.

Ah, okay so using latest repo for rolling and CI should be fine. i will start the CI to verify. @iuhilnehc-ynos thanks for the information.

@clalancette i thought Rpr job base runtime is updated nightly, but obviously it is not. How often does it get updated?

clalancette:
> @clalancette i thought Rpr job base runtime is updated nightly, but obviously it is not. How often does it get updated?

Only when we do releases into Rolling.  We are way overdue for that; we will spend some time next week doing releases into Rolling.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18021/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12547/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18656/)

fujitatomoya:
Window is unstable, i will restart it later.

fujitatomoya:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18670/)

rcl/test/rcl/test_publisher.cpp
```diff
@@ -372,38 +372,69 @@ TEST_F(CLASSNAME(TestPublisherFixture, RMW_IMPLEMENTATION), test_publisher_loan)
...
+    OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+    {
+      rcl_ret_t ret = rcl_publisher_fini(&publisher, this->node_ptr);
+      EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+    });
+    if (is_fastdds) {
+      EXPECT_TRUE(rcl_publisher_can_loan_messages(&publisher));
+    } else {
+      EXPECT_FALSE(rcl_publisher_can_loan_messages(&publisher));
+    }
```
MichaelOrlov, line 423: @fujitatomoya Could you please explain why there are different behavior for `rmw_fastrtps`?  
I think we shouldn't rely on default value of the environment variable in tests since it tend to be flaky or false negative/positive.
Instead we need explicitly set `publisher_options.disable_loaned_message` in code before publisher init to verify that our explicit options settings will override environment variable settings.

MichaelOrlov (changes requested):
@fujitatomoya Thanks for moving to rcl_options and checking environment variable inside get_default_options methods. Implementation looks good to me.

However I have a question and suggestion for unit tests.
- Need to check that explicitly settled up value for the `publisher_options.disable_loaned_message` and `subscriber_options.disable_loaned_message` will not be overridden by the environment variable after calling pub/sub init.
- And we shouldn't rely on default value of the environment variable. We should set environment variable explicitly for each test.

rcl/test/rcl/test_publisher.cpp
fujitatomoya: @MichaelOrlov rmw_fastrtps just uses shared memory if possible as long as data type is POD, that is the default behavior for rmw_fastrtps, and this default behavior depends on rmw implementation. this default behavior verification was there before this PR, and i think we can keep those as well.

> explicitly set publisher_options.disable_loaned_message in code before publisher init to verify that our explicit options settings will override environment variable settings.

makes sense, i will add more test accordingly.

MichaelOrlov: @fujitatomoya I had in mind that if you will explicitly set `publisher_options.disable_loaned_message = true` before calling publisher init the further call for the `rcl_publisher_can_loan_messages(&publisher)` should return always false - regardless of the internal `rmw_fastrtps` implementation.
This is actually what we need to check.

fujitatomoya: addressed by https://github.com/ros2/rcl/pull/1029/commits/c6f9fc3e78c8afdf5fd2f1ad274fb7825e05f36f

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18028/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12553/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18672/)

fujitatomoya:
@iuhilnehc-ynos can you review this?

MichaelOrlov (changes requested):
@fujitatomoya Thanks for addressing comments from review.
I still have a strong opinion that we should not rely on default settings for `ROS_DISABLE_LOANED_MESSAGES` environment variable in `test_publisher_loan_disable` and `test_subscription_loan_disable` when doing check for `EXPECT_FALSE(publisher_options.disable_loaned_message);` right after reading default options.  
We need either explicitly set environment variable in test or even better remove check for `EXPECT_FALSE(publisher_options.disable_loaned_message);` after getting default options.  And always forcefully set `publisher_options.disable_loaned_message` value as we needed for test case.
The rationale for that is:
1. It's not safe to rely (not reliable to assume) that `ROS_DISABLE_LOANED_MESSAGES` environment variable will not be set by default on testing machine and environment. It could be leftovers from another tests. It might work currently but could be broken in future if order of tests will be changed or some new tests would be added.
1. You've already have dedicated tests for checking `publisher_options.disable_loaned_message` after getting default options with different values of the `ROS_DISABLE_LOANED_MESSAGES` environment variable.

rcl/include/rcl/publisher.h
```diff
@@ -49,6 +49,8 @@ typedef struct rcl_publisher_options_s
   rcl_allocator_t allocator;
   /// rmw specific publisher options, e.g. the rmw implementation specific payload.
   rmw_publisher_options_t rmw_publisher_options;
+  /// Disable flag to LoanedMessage, initialized via environmental variable.
```
iuhilnehc-ynos, line 52: To make the comment clear, do you think it's necessary to add something like "True only if `ROS_DISABLE_LOANED_MESSAGES=1`"? (some cases might be used, e.g:  ROS_DISABLE_LOANED_MESSAGES=TRUE)

https://github.com/ros2/rcl/blob/c6f9fc3e78c8afdf5fd2f1ad274fb7825e05f36f/rcl/include/rcl/publisher.h#L198 should be updated

```
 * - rmw_publisher_options = rmw_get_default_publisher_options()
 * - disable_loaned_message = ?
```
Sorry, I have no good description for this.

rcl/src/rcl/publisher.c
```diff
@@ -203,6 +203,16 @@ rcl_publisher_get_default_options()
...
   default_options.allocator = rcl_get_default_allocator();
   default_options.rmw_publisher_options = rmw_get_default_publisher_options();
+
+  // Load disable flag to LoanedMessage via environmental variable.
+  bool disable_loaned_message = false;
+  rcl_ret_t ret = rcl_get_disable_loaned_message(&disable_loaned_message);
+  if (ret == RCL_RET_OK) {
+    default_options.disable_loaned_message = disable_loaned_message;
+  } else {
+    default_options.disable_loaned_message = false;
```
iuhilnehc-ynos, line 213: To call `rcl_reset_error` in `else block` might be better because `rcl_get_disable_loaned_message` sets the error message when returns ERROR, right?

What about
```
  bool disable_loaned_message = false;
  rcl_ret_t ret = rcl_get_disable_loaned_message(&disable_loaned_message);
  if (ret != RCL_RET_OK) {
    // debug log
    rcl_reset_error();
  }
  default_options.disable_loaned_message = disable_loaned_message;
```

iuhilnehc-ynos (commented):
LGTM with some minor comments.

fujitatomoya:
> @fujitatomoya Thanks for addressing comments from review. I still have a strong opinion that we should not rely on default settings for `ROS_DISABLE_LOANED_MESSAGES` environment variable in `test_publisher_loan_disable` and `test_subscription_loan_disable` when doing check for `EXPECT_FALSE(publisher_options.disable_loaned_message);` right after reading default options. We need either explicitly set environment variable in test or even better remove check for `EXPECT_FALSE(publisher_options.disable_loaned_message);` after getting default options. And always forcefully set `publisher_options.disable_loaned_message` value as we needed for test case. The rationale for that is:
> 
> 1. It's not safe to rely (not reliable to assume) that `ROS_DISABLE_LOANED_MESSAGES` environment variable will not be set by default on testing machine and environment. It could be leftovers from another tests. It might work currently but could be broken in future if order of tests will be changed or some new tests would be added.
> 2. You've already have dedicated tests for checking `publisher_options.disable_loaned_message` after getting default options with different values of the `ROS_DISABLE_LOANED_MESSAGES` environment variable.

@MichaelOrlov I see, probably it would be better, not to have flaky test 👍

rcl/src/rcl/publisher.c
fujitatomoya: yeah, right. i will fix this as you suggested.

fujitatomoya: actually i just wanted to use local value only when API succeeds, but i will take some of your suggestion here.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18029/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12554/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18673/)

fujitatomoya:
@MichaelOrlov @iuhilnehc-ynos thanks for the review, all comments are addressed. requesting final check.

rcl/src/rcl/publisher.c
```diff
@@ -203,6 +203,19 @@ rcl_publisher_get_default_options()
...
   default_options.rmw_publisher_options = rmw_get_default_publisher_options();
+
+  // Load disable flag to LoanedMessage via environmental variable.
+  bool disable_loaned_message = false;
+  rcl_ret_t ret = rcl_get_disable_loaned_message(&disable_loaned_message);
+  if (ret == RCL_RET_OK) {
+    default_options.disable_loaned_message = disable_loaned_message;
+  } else {
+    RCUTILS_SAFE_FWRITE_TO_STDERR("Failed to get disable_loaned_message: ");
+    RCUTILS_SAFE_FWRITE_TO_STDERR(rcl_get_error_string().str);
```
iuhilnehc-ynos, line 214: I will use `RCUTILS_SAFE_FWRITE_TO_STDERR_WITH_FORMAT_STRING` instead, but it's OK.

iuhilnehc-ynos (approved):

fujitatomoya:
@iuhilnehc-ynos thanks for the review.

CI(windows unstable):
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18676/)

rcl/src/rcl/subscription.c
```diff
@@ -213,6 +213,9 @@ rcl_subscription_get_default_options()
   if (ret == RCL_RET_OK) {
     default_options.disable_loaned_message = disable_loaned_message;
   } else {
+    RCUTILS_SAFE_FWRITE_TO_STDERR("Failed to get disable_loaned_message: ");
+    RCUTILS_SAFE_FWRITE_TO_STDERR(rcl_get_error_string().str);
+    rcl_reset_error();
```
MichaelOrlov, line 218: The same comment in regards removing logging to the error stream

MichaelOrlov (changes requested):
@fujitatomoya Please remove logging to the stderr stream when rcl_get_disable_loaned_message(&disable_loaned_message) returning failure.
This is a normal situation when `ROS_DISABLE_LOANED_MESSAGES` is not defined and `ret != RCL_RET_OK` in this particular situation.

rcl/src/rcl/subscription.c
fujitatomoya: >  And by default it will be absent in most configurations.

I think that this error is important cz using `ROS_DISABLE_LOANED_MESSAGES` is meant to disable the feature. so this is strong user intention. in that case if unexpected error happens, it should be able to notify the user that `ROS_DISABLE_LOANED_MESSAGES` is not effectively working.

> This error string will appear each time we creating new pub/sub - it would be pretty annoying and could also affect performance.

I believe there should be. I think it is unlikely that this logging affects the performance, because it only happens in initialization. I think originally this PR is targeting send/receive performance improvement but initialization or finalization?

what do you think?

MichaelOrlov: @fujitatomoya Ahh never mind. 
rcl_get_disable_loaned_message(bool * disable_loaned_message) can return `RCL_RET_ERROR` only if input parameters are wrong which is not a our case any way.  
You can keep your output to the error string but error will never happened.
I was wrong on assumption that `rcl_get_disable_loaned_message(bool * disable_loaned_message)` that will return some error code if `ROS_DISABLE_LOANED_MESSAGES` env variable not defined. Sorry for confusion.

MichaelOrlov (approved):

rcl/src/rcl/subscription.c
fujitatomoya: > You can keep your output to the error stream but error will never happened.

yeah you are right, this error might print here if `something` really unexpected happens. probably never... 😅 

not at all, really appreciate your review and comments, thank you so much 👍

fujitatomoya:
@clalancette i am putting a hold on this until https://github.com/ros2/rcl/pull/1029#issuecomment-1414028600 is resolved. after rolling release, i will kick the Rpr job to make sure it passes, and then i will merge this into rolling.

fujitatomoya:

fujitatomoya:
rpr job is happy now, CI is green.

@clalancette i will be waiting for the final review on this.

clalancette (approved):

fujitatomoya:
@clalancette @MichaelOrlov @iuhilnehc-ynos thank you for the review, i will go ahead to merge this. since this changes ABI, this will be available on Iron or later.

PR #1030, opened 2023-01-30
Make sure to check the return value of rcl_clock_init in tests.

clalancette:
This is more correct, and also makes clang static analysis happier.


audrow (approved):
LGTM!

PR #1031, opened 2023-01-31
Update rcl to C++17.

clalancette:
The two reasons to do this are:

1. So that we can compile rcl with the clang static analyzer. As of clang++-14 (what is in Ubuntu 22.04), the default still seems to be C++14, so we need to specify C++17 so that new things in the rclcpp headers work properly.
2. So we can build with a newer version of gtest which requires this.

Further, due to reasons I don't fully understand, I needed to set CMAKE_CXX_STANDARD_REQUIRED in order for clang to really use that version. So set this as well.


fujitatomoya (approved):

mjcarroll (approved):

audrow (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18009/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12537/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18646/)

PR #1033, opened 2023-02-15
Implement matched event

Barry-Xu-2018:
Address ros2/rmw#330

### DIFF
--- a/rcl/test/rcl/test_events.cpp
+++ b/rcl/test/rcl/test_events.cpp
@@ -872,7 +872,7 @@ void event_callback(const void * user_data, size_t number_of_events)
 }
 
 /*
- * Basic test of publisher matched/unmatched event
+ * Basic test of publisher matched event
  */
 TEST_F(TestEventFixture, test_pub_matched_unmatched_event)
 {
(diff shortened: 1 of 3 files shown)


rcl/test/rcl/test_events.cpp
```diff
@@ -853,3 +854,303 @@ INSTANTIATE_TEST_SUITE_P(
...
+
+    rmw_unmatched_status_t unmatched_status;
+    EXPECT_EQ(RMW_RET_OK, rcl_take_event(&pub_unmatched_event, &unmatched_status));
+    EXPECT_EQ(0, unmatched_status.current_matched_count);
+    EXPECT_EQ(1, unmatched_status.current_count_change);
+  }
+}
+
+/*
+ * Basic test of publisher matched/unmatched event
```
fujitatomoya, line 1016: just a typo

```suggestion
 * Basic test of subscription matched/unmatched event
```

rcl/test/rcl/test_events.cpp
```diff
@@ -853,3 +854,303 @@ INSTANTIATE_TEST_SUITE_P(
...
+  matched_data.event_count = std::make_shared<std::atomic_size_t>(0);
+  struct EventUserData unmatched_data;
+  unmatched_data.event_count = std::make_shared<std::atomic_size_t>(0);
+
+  // rmw_connextdds doesn't support rmw_event_set_callback() interface
+  if (std::string(rmw_get_implementation_identifier()).find("rmw_connextdds") != 0) {
+    ret = rcl_event_set_callback(&pub_matched_event, event_callback, &matched_data);
+    ASSERT_EQ(RMW_RET_OK, ret);
+
+    ret = rcl_event_set_callback(&pub_unmatched_event, event_callback, &unmatched_data);
```
fujitatomoya, line 918: I think there should be test case that this is called after creating the subscription. it is expected that callback should not be fired because the matched event has happened before.

fujitatomoya (changes requested):
overall lgtm, but i believe one test case should be added as following.

rcl/test/rcl/test_events.cpp
Barry-Xu-2018: will update

rcl/test/rcl/test_events.cpp
Barry-Xu-2018: Yes. I willl add test cases.

fujitatomoya (approved):

Barry-Xu-2018:
Rebase was done.

PR #1035, opened 2023-03-01
avoid unnecessary copy for rcutils_char_array_vsprintf.

fujitatomoya:
address https://github.com/ros2/rcutils/issues/409

### DIFF
--- a/rcl/src/rcl/logging.c
+++ b/rcl/src/rcl/logging.c
@@ -190,12 +190,7 @@ rcl_logging_ext_lib_output_handler(
     .allocator = g_logging_allocator
   };
 
-  va_list args_clone;
-  // The args are initialized, but clang-tidy cannot tell.
-  // It may be related to this bug: https://bugs.llvm.org/show_bug.cgi?id=41311
-  va_copy(args_clone, *args);  // NOLINT(clang-analyzer-valist.Uninitialized)
-  status = rcutils_char_array_vsprintf(&msg_array, format, args_clone);
-  va_end(args_clone);
+  status = rcutils_char_array_vsprintf(&msg_array, format, *args);
 
   if (RCL_RET_OK == status) {
     status = rcutils_logging_format_message(
--- a/rcl/src/rcl/logging_rosout.c
+++ b/rcl/src/rcl/logging_rosout.c
@@ -348,12 +348,7 @@ void rcl_logging_rosout_output_handler(
       .allocator = __rosout_allocator
     };
 
-    va_list args_clone;
-    // The args are initialized, but clang-tidy cannot tell.
-    // It may be related to this bug: https://bugs.llvm.org/show_bug.cgi?id=41311
-    va_copy(args_clone, *args);  // NOLINT(clang-analyzer-valist.Uninitialized)
-    status = rcl_ret_from_rcutils_ret(rcutils_char_array_vsprintf(&msg_array, format, args_clone));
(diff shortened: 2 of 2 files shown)


fujitatomoya:
@clalancette minor improvement, can you review on this?

fujitatomoya:
CC: @iuhilnehc-ynos

iuhilnehc-ynos (approved):

rcl/src/rcl/logging.c
```diff
@@ -190,12 +190,7 @@ rcl_logging_ext_lib_output_handler(
     .allocator = g_logging_allocator
   };
 
-  va_list args_clone;
-  // The args are initialized, but clang-tidy cannot tell.
-  // It may be related to this bug: https://bugs.llvm.org/show_bug.cgi?id=41311
-  va_copy(args_clone, *args);  // NOLINT(clang-analyzer-valist.Uninitialized)
-  status = rcutils_char_array_vsprintf(&msg_array, format, args_clone);
-  va_end(args_clone);
+  status = rcutils_char_array_vsprintf(&msg_array, format, *args);
```
clalancette, line 193: It is not clear to me that it is safe to do this.  The man page for `vsnprintf` (which is what `rcutils_char_array_vsprintf` calls) says this:

```
       The functions vprintf(), vfprintf(), vdprintf(), vsprintf(), vsnprintf() are equivalent to the functions printf(), fprintf(), dprintf(), sprintf(),
       snprintf(), respectively, except that they are called with a va_list instead of a variable number of arguments.  These functions do  not  call  the
       va_end macro.  Because they invoke the va_arg macro, the value of ap is undefined after the call.  See stdarg(3).
```

And further in the man page for stdarg:
```
If ap is passed to a function that uses va_arg(ap,type), then the value of ap is undefined after the return of that function.
```

clalancette (changes requested):
I don't think that this is safe to do.  But I'm willing to hear what others have to say about it.

rcl/src/rcl/logging.c
iuhilnehc-ynos: Please notice there is a [va_copy](https://github.com/ros2/rcutils/blob/23dae4430d5707ca5732adfedf45c16e0d1279a0/src/char_array.c#L155) in the `_rcutils_char_array_vsprintf` which is necessary because `rcutils_char_array_vsprintf` might call `_rcutils_char_array_vsprintf`([1](https://github.com/ros2/rcutils/blob/23dae4430d5707ca5732adfedf45c16e0d1279a0/src/char_array.c#L169), [2](https://github.com/ros2/rcutils/blob/23dae4430d5707ca5732adfedf45c16e0d1279a0/src/char_array.c#L185)) twice, so there is unnecessary to make an extra `va_list` by `va_copy` before calling `rcutils_char_array_vsprintf`.

clalancette: Oh, you are totally right.  In that case, yes, I agree that this is an unnecessary copy, thank you!

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18155/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12692/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18813/)

fujitatomoya:
@mergifyio backport humble

PR #1042, opened 2023-03-06
Fix some warnings from clang.

clalancette:
1. The "expected" variable should be default initialized, otherwise it might have a garbage value when being checked.
2. Mark some variables in tests as unused.
3. Call rcl_reset_error in a few more places.


fujitatomoya (approved):
lgtm

mjcarroll (approved):

clalancette:
The yellow tests on Windows are unrelated to this, so going ahead and merging.

PR #1052, opened 2023-03-29
Add `~/get_type_description` service (rep2011)

achim-k:
Part of https://github.com/ros2/ros2/issues/1159

~Depends on https://github.com/ros2/rosidl/pull/727 (https://github.com/ros2/rosidl/pull/735)~

Features:
- Implements the [~/get_type_description](https://github.com/ros-infrastructure/rep/blob/ba2f96c5163b77385956d51065607201ae3f71e7/rep-2011.rst#sending-the-type-description) which will be offered by each node (can be disabled via node options)
- Adds a type cache for each node which holds all types that are registered by subscribers, publishers, services and action servers
- Adds conversion functions to convert between relevant `rosidl_runtime_c` and `type_description_interfaces` structs

### DIFF
--- a/rcl/include/rcl/node.h
+++ b/rcl/include/rcl/node.h
@@ -549,6 +549,57 @@ RCL_PUBLIC
 rcl_ret_t
 rcl_get_disable_loaned_message(bool * disable_loaned_message);
 
+/// Initialize the node's ~/get_type_description service.
+/**
+ * This function initializes the node's private ~/get_type_description service
+ * which can be used to retrieve information about types used by the node's
+ * publishers, subscribers, services or actions.
+ *
+ * Note that this function will be called in `rcl_init_node` if the node option
+ * `enable_type_description_service` is set to true.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] node handle to the node for which to initialize the service
+ * \return #RCL_RET_OK if the service was successfully initialized, or
+ * \return #RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
+ * \return #RCL_RET_ALREADY_INIT if the service is already initialized, or
+ * \return #RCL_RET_BAD_ALLOC if memory allocation for the service failed, or
+ * \return #RCL_RET_ERROR if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t rcl_node_type_description_service_init(rcl_node_t * node);
+
+/// Finalizes the node's ~/get_type_description service.
+/**
+ * This function finalizes the node's private ~/get_type_description service.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] node the handle to the node whose type cache should be initialized
+ * \return #RCL_RET_OK if service was deinitialized successfully, or
+ * \return #RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
+ * \return #RCL_RET_SERVICE_INVALID if the service is invalid, or
+ * \return #RCL_RET_NODE_INVALID if the node is invalid, or
+ * \return #RCL_RET_ERROR if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t rcl_node_type_description_service_fini(rcl_node_t * node);
+
 #ifdef __cplusplus
 }
 #endif
--- a/rcl/include/rcl/node.h
+++ b/rcl/include/rcl/node.h
@@ -33,9 +33,12 @@ extern "C"
 #include "rcl/types.h"
 #include "rcl/visibility_control.h"
 
+#include "type_description_interfaces/srv/get_type_description.h"
+
 extern const char * const RCL_DISABLE_LOANED_MESSAGES_ENV_VAR;
 
 typedef struct rcl_node_impl_s rcl_node_impl_t;
+typedef struct rcl_service_s rcl_service_t;
 
 /// Structure which encapsulates a ROS Node.
 typedef struct rcl_node_s
@@ -551,12 +554,15 @@ rcl_get_disable_loaned_message(bool * disable_loaned_message);
 
 /// Initialize the node's ~/get_type_description service.
 /**
- * This function initializes the node's private ~/get_type_description service
+ * This function initializes the node's ~/get_type_description service
  * which can be used to retrieve information about types used by the node's
  * publishers, subscribers, services or actions.
  *
- * Note that this function will be called in `rcl_init_node` if the node option
- * `enable_type_description_service` is set to true.
+ * Note that this will not register any callback for the service, client-level code
+ * must register rcl_node_type_description_service_handle_request or a custom callback
+ * to handle incoming requests, via that client's executor/waitset capabilities.
+ *
+ * This will initialize the node's type cache, if it has not been initialized already.
  *
  * <hr>
  * Attribute          | Adherence
@@ -599,6 +605,61 @@ rcl_ret_t rcl_node_type_description_service_init(rcl_node_t * node);
 RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_ret_t rcl_node_type_description_service_fini(rcl_node_t * node);
+
+
+/// Returns a pointer to the node's ~/get_type_description service.
+/**
+ * On success, sets service_out to the initialized service.
+ * rcl_node_type_description_service_init must be called before this.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] node the handle to the node
+ * \param[out] service_out Handle to pointer that will be set
+ * \return #RCL_RET_OK if valid service was returned successfully, or
+ * \return #RCL_RET_NODE_INVALID if node is invalid, or
+ * \return #RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
+ * \return #RCL_RET_NOT_INIT if the service hasn't yet been initialized, or
+ * \return #RCL_RET_ERROR if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t rcl_node_get_type_description_service(
+  const rcl_node_t * node,
+  rcl_service_t ** service_out);
+
+
+/// Process a single pending request to the GetTypeDescription service.
+/**
+ * This function may be called to handle incoming requests by any client starting the service.
+ * It is not intended to be called directly by users.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] node the handle to the node
+ * \param[in] request_header ID of the incoming request
+ * \param[in] request Request that came in to the service
+ * \param[out] response Allocated, uninitialized response to the request
+ * \return void
+ */
+RCL_PUBLIC
+void rcl_node_type_description_service_handle_request(
+  rcl_node_t * node,
+  const rmw_request_id_t * request_header,
+  const type_description_interfaces__srv__GetTypeDescription_Request * request,
+  type_description_interfaces__srv__GetTypeDescription_Response * response);
 
 #ifdef __cplusplus
 }
--- a/rcl/include/rcl/node_type_cache.h
+++ b/rcl/include/rcl/node_type_cache.h
@@ -1,0 +1,172 @@
+// Copyright 2023 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCL__NODE_TYPE_CACHE_H_
+#define RCL__NODE_TYPE_CACHE_H_
+
+#include "rcl/node.h"
+#include "rcl/types.h"
+#include "rcl/visibility_control.h"
+#include "rosidl_runtime_c/type_description/type_description__struct.h"
+#include "rosidl_runtime_c/type_description/type_source__struct.h"
+#include "type_description_interfaces/msg/type_description.h"
+#include "type_description_interfaces/msg/type_source.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct rcl_type_info_t
+{
+  type_description_interfaces__msg__TypeDescription * type_description;
+  type_description_interfaces__msg__TypeSource__Sequence * type_sources;
+} rcl_type_info_t;
+
+/// Initialize the node's type cache.
+/**
+ * This function initializes hash map of the node's type cache such that types
+ * can be registered and retrieved.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] node the handle to the node whose type cache should be initialized
+ * \return #RCL_RET_OK if the node's type cache was successfully initialized, or
+ * \return #RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
+ * \return #RCL_RET_NODE_INVALID if the given `node` is invalid, or
+ * \return #RCL_RET_ERROR if an unspecified error occurs.
+ */
+RCL_WARN_UNUSED
+rcl_ret_t rcl_node_type_cache_init(rcl_node_t * node);
+
+/// Finalize the node's type cache.
+/**
+ * This function clears the hash map of the node's type cache and deallocates
+ * used memory.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] node the handle to the node whose type cache should be finalized
+ * \return #RCL_RET_OK if the node's type cache was successfully finalized, or
+ * \return #RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
+ * \return #RCL_RET_NODE_INVALID if the given `node` is invalid, or
+ * \return #RCL_RET_ERROR if an unspecified error occurs.
+ */
+RCL_WARN_UNUSED
+rcl_ret_t rcl_node_type_cache_fini(rcl_node_t * node);
+
+/// Register a type with the node's type cache.
+/**
+ * This function registers the given type, uniquely identified by the type_hash,
+ * with the node with the node's type cache. Multiple registrations of the same
+ * type will increment its registration count.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] node the handle to the node whose type cache should be finalized
+ * \param[in] type_hash hash of the type
+ * \param[in] type_description type description struct
+ * \param[in] type_description_sources type description sources struct
+ * \return #RCL_RET_OK if the type was successfully registered, or
+ * \return #RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
+ * \return #RCL_RET_NODE_INVALID if the given `node` is invalid, or
+ * \return #RCL_RET_ERROR if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t rcl_node_type_cache_register_type(
+  const rcl_node_t * node, const rosidl_type_hash_t * type_hash,
+  const rosidl_runtime_c__type_description__TypeDescription * type_description,
+  const rosidl_runtime_c__type_description__TypeSource__Sequence * type_description_sources
+);
+
+/// Unregister a message type from the node's type cache.
+/**
+ * This function uses the given `type_hash` to unregister the associated type in
+ * the node's type cache. If the type has been registered multiple times, the
+ * type will only be removed if its registration count reaches 0.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] node the handle to the node whose type cache should be finalized
+ * \param[in] type_hash type hash
+ * \return #RCL_RET_OK if the type was successfully registered, or
+ * \return #RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
+ * \return #RCL_RET_NODE_INVALID if the given `node` is invalid, or
+ * \return #RCL_RET_ERROR if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t rcl_node_type_cache_unregister_type(
+  const rcl_node_t * node, const rosidl_type_hash_t * type_hash);
+
+/// Retrieve type information from the node's type cache.
+/**
+ * This function returns the desired type information from the node's type cache
+ *
+ * The `type_info` field must point to an allocated `rcl_type_info_t` object to
+ * which the type information will be written.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] node the handle to the node whose type cache should be queried
+ * \param[in] type_hash type hash
+ * \param[out] type_info pointer to the type info struct that will be populated
+ * \return #RCL_RET_OK if type information was retrieved successfully
+ * \return #RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
+ * \return #RCL_RET_NODE_INVALID if the given `node` is invalid, or
+ * \return #RCL_RET_NOT_INIT if node's type cache has not been initialized, or
+ * \return #RCL_RET_ERROR if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t rcl_node_type_cache_get_type_info(
+  const rcl_node_t * node,
+  const rosidl_type_hash_t * type_hash,
+  rcl_type_info_t * type_info);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif  // RCL__NODE_TYPE_CACHE_H_
--- a/rcl/include/rcl/node_type_cache.h
+++ b/rcl/include/rcl/node_type_cache.h
@@ -55,6 +55,27 @@ typedef struct rcl_type_info_t
 RCL_WARN_UNUSED
 rcl_ret_t rcl_node_type_cache_init(rcl_node_t * node);
 
+/// Check whether node's type cache has been initialized.
+/**
+ * Based on configuration, the node may optionally disable the type cache,
+ * this convenience function allows a standardized check for it.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] node the handle to the node whose type cache to check
+ * \return true if the node's type cache was successfully initialized, or
+ * \return false for any other case.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+bool rcl_node_type_cache_is_valid(const rcl_node_t * node);
+
 /// Finalize the node's type cache.
 /**
  * This function clears the hash map of the node's type cache and deallocates
--- a/rcl/include/rcl/node_type_cache.h
+++ b/rcl/include/rcl/node_type_cache.h
@@ -54,30 +54,8 @@ typedef struct rcl_type_info_t
  * \return #RCL_RET_NODE_INVALID if the given `node` is invalid, or
  * \return #RCL_RET_ERROR if an unspecified error occurs.
  */
-RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_ret_t rcl_node_type_cache_init(rcl_node_t * node);
-
-/// Check whether node's type cache has been initialized.
-/**
- * Based on configuration, the node may optionally disable the type cache,
- * this convenience function allows a standardized check for it.
- *
- * <hr>
- * Attribute          | Adherence
- * ------------------ | -------------
- * Allocates Memory   | No
- * Thread-Safe        | No
- * Uses Atomics       | No
- * Lock-Free          | Yes
- *
- * \param[in] node the handle to the node whose type cache to check
- * \return true if the node's type cache was successfully initialized, or
- * \return false for any other case.
- */
-RCL_PUBLIC
-RCL_WARN_UNUSED
-bool rcl_node_type_cache_is_valid(const rcl_node_t * node);
 
 /// Finalize the node's type cache.
 /**
@@ -98,7 +76,6 @@ bool rcl_node_type_cache_is_valid(const rcl_node_t * node);
  * \return #RCL_RET_NODE_INVALID if the given `node` is invalid, or
  * \return #RCL_RET_ERROR if an unspecified error occurs.
  */
-RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_ret_t rcl_node_type_cache_fini(rcl_node_t * node);
 
--- a/rcl/include/rcl/node_type_cache.h
+++ b/rcl/include/rcl/node_type_cache.h
@@ -37,6 +37,8 @@ typedef struct rcl_type_info_t
 /**
  * This function initializes hash map of the node's type cache such that types
  * can be registered and retrieved.
+ * Note that to correctly capture all types used by a node, this needs to be called
+ * before any "builtin" publishers or services are created.
  *
  * <hr>
  * Attribute          | Adherence
@@ -52,6 +54,7 @@ typedef struct rcl_type_info_t
  * \return #RCL_RET_NODE_INVALID if the given `node` is invalid, or
  * \return #RCL_RET_ERROR if an unspecified error occurs.
  */
+RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_ret_t rcl_node_type_cache_init(rcl_node_t * node);
 
@@ -95,6 +98,7 @@ bool rcl_node_type_cache_is_valid(const rcl_node_t * node);
  * \return #RCL_RET_NODE_INVALID if the given `node` is invalid, or
  * \return #RCL_RET_ERROR if an unspecified error occurs.
  */
+RCL_PUBLIC
 RCL_WARN_UNUSED
 rcl_ret_t rcl_node_type_cache_fini(rcl_node_t * node);
 
--- a/rcl/include/rcl/type_description_conversions.h
+++ b/rcl/include/rcl/type_description_conversions.h
@@ -1,0 +1,133 @@
+// Copyright 2023 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCL__TYPE_DESCRIPTION_CONVERSIONS_H_
+#define RCL__TYPE_DESCRIPTION_CONVERSIONS_H_
+
+#include "rcl/macros.h"
+#include "rcl/visibility_control.h"
+#include "rosidl_runtime_c/type_description/type_description__struct.h"
+#include "rosidl_runtime_c/type_description/type_source__struct.h"
+#include "type_description_interfaces/msg/type_description.h"
+#include "type_description_interfaces/msg/type_source.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/// Convert type description runtime struct to msg struct.
+/**
+ * This function converts a rosidl_runtime_c__type_description__TypeDescription
+ * to the corresponding type_description_interfaces/msg/TypeDescription struct.
+ * The retrieved pointer shall be destroyed with
+ * `type_description_interfaces__msg__TypeDescription__destroy()` after use.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] runtime_description the pointer to the runtime type description struct
+ * \return a valid type_description_interfaces/msg/TypeDescription pointer, or
+ * \return NULL if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+type_description_interfaces__msg__TypeDescription *
+rcl_convert_type_description_runtime_to_msg(
+  const rosidl_runtime_c__type_description__TypeDescription * runtime_description);
+
+/// Convert type description msg struct to a rosidl runtime struct.
+/**
+ * This function converts a type_description_interfaces/msg/TypeDescription
+ * to the corresponding rosidl_runtime_c__type_description__TypeDescription
+ * struct. The retrieved pointer shall be destroyed with
+ * `rosidl_runtime_c__type_description__TypeDescription__destroy()` after use.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] description_msg the pointer to the msg type description struct
+ * \return a valid pointer to the runtime struct, or
+ * \return NULL if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rosidl_runtime_c__type_description__TypeDescription *
+rcl_convert_type_description_msg_to_runtime(
+  const type_description_interfaces__msg__TypeDescription * description_msg);
+
+/// Convert type sources sequence runtime struct to msg struct.
+/**
+ * This function converts a rosidl_runtime_c__type_description__TypeSource__Sequence
+ * struct to the corresponding type_description_interfaces__msg__TypeSource__Sequence
+ * struct. The retrieved pointer shall be destroyed with
+ * `type_description_interfaces__msg__TypeSource__Sequence__destroy()` after use.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] runtime_type_sources the pointer to the type sources sequence struct
+ * \return a valid pointer to the msg struct, or
+ * \return NULL if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+type_description_interfaces__msg__TypeSource__Sequence *
+rcl_convert_type_source_sequence_runtime_to_msg(
+  const rosidl_runtime_c__type_description__TypeSource__Sequence * runtime_type_sources);
+
+/// Convert type sources sequece msg struct to a rosidl runtime struct.
+/**
+ * This function converts a rosidl_runtime_c__type_description__TypeSource__Sequence
+ * struct to the corresponding type_description_interfaces__msg__TypeSource__Sequence
+ * struct. The retrieved pointer shall be destroyed with
+ * `type_description_interfaces__msg__TypeSource__Sequence__destroy()` after use.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] type_sources_msg the pointer to the type sources sequence struct
+ * \return a valid pointer to the msg struct, or
+ * \return NULL if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rosidl_runtime_c__type_description__TypeSource__Sequence *
+rcl_convert_type_source_sequence_msg_to_runtime(
+  const type_description_interfaces__msg__TypeSource__Sequence * type_sources_msg);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif  // RCL__TYPE_DESCRIPTION_CONVERSIONS_H_
--- a/rcl/src/rcl/node.c
+++ b/rcl/src/rcl/node.c
@@ -52,18 +52,9 @@ extern "C"
 #include "tracetools/tracetools.h"
 
 #include "./context_impl.h"
+#include "./node_impl.h"
 
 const char * const RCL_DISABLE_LOANED_MESSAGES_ENV_VAR = "ROS_DISABLE_LOANED_MESSAGES";
-
-struct rcl_node_impl_s
-{
-  rcl_node_options_t options;
-  rmw_node_t * rmw_node_handle;
-  rcl_guard_condition_t * graph_guard_condition;
-  const char * logger_name;
-  const char * fq_name;
-};
-
 
 /// Return the logger name associated with a node given the validated node name and namespace.
 /**
--- a/rcl/src/rcl/node.c
+++ b/rcl/src/rcl/node.c
@@ -306,12 +306,6 @@ rcl_node_init(
       goto fail;
     }
   }
-  if (node->impl->options.enable_type_description_service) {
-    RCUTILS_LOG_WARN_NAMED(
-      ROS_PACKAGE_NAME,
-      "Requested ~/get_type_description service enabled, but feature is not yet implemented. "
-      "Service is not created.");
-  }
   RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME, "Node initialized");
   ret = RCL_RET_OK;
   TRACETOOLS_TRACEPOINT(
--- a/rcl/src/rcl/node.c
+++ b/rcl/src/rcl/node.c
@@ -30,6 +30,7 @@ extern "C"
 #include "rcl/localhost.h"
 #include "rcl/logging.h"
 #include "rcl/logging_rosout.h"
+#include "rcl/node_type_cache.h"
 #include "rcl/rcl.h"
 #include "rcl/remap.h"
 #include "rcl/security.h"
@@ -43,17 +44,21 @@ extern "C"
 #include "rcutils/repl_str.h"
 #include "rcutils/snprintf.h"
 #include "rcutils/strdup.h"
+#include "rcutils/types/hash_map.h"
 
 #include "rmw/error_handling.h"
 #include "rmw/security_options.h"
 #include "rmw/rmw.h"
 #include "rmw/validate_namespace.h"
 #include "rmw/validate_node_name.h"
+#include "rosidl_runtime_c/string_functions.h"
+#include "rosidl_runtime_c/type_description/type_description__functions.h"
+#include "rosidl_runtime_c/type_description/type_source__functions.h"
 #include "tracetools/tracetools.h"
+#include "type_description_interfaces/srv/get_type_description.h"
 
 #include "./context_impl.h"
 #include "./node_impl.h"
-#include "./node_type_cache_init.h"
 
 const char * const RCL_DISABLE_LOANED_MESSAGES_ENV_VAR = "ROS_DISABLE_LOANED_MESSAGES";
 
@@ -200,6 +205,7 @@ rcl_node_init(
   node->impl->fq_name = NULL;
   node->impl->options = rcl_node_get_default_options();
   node->impl->registered_types_by_type_hash = rcutils_get_zero_initialized_hash_map();
+  node->impl->get_type_description_service = rcl_get_zero_initialized_service();
   node->context = context;
   // Initialize node impl.
   ret = rcl_node_options_copy(options, &(node->impl->options));
@@ -283,6 +289,14 @@ rcl_node_init(
     // error message already set
     goto fail;
   }
+  // Initialize ~/get_type_description service
+  if (node->impl->options.enable_type_description_service) {
+    ret = rcl_node_type_description_service_init(node);
+    if (ret != RCL_RET_OK) {
+      // error message already set
+      goto fail;
+    }
+  }
   // The initialization for the rosout publisher requires the node to be in initialized to a point
   // that it can create new topic publishers
   if (rcl_logging_rosout_enabled() && node->impl->options.enable_rosout) {
@@ -309,6 +323,11 @@ rcl_node_init(
   goto cleanup;
 fail:
   if (node->impl) {
+    ret = rcl_node_type_description_service_fini(node);
+    RCUTILS_LOG_ERROR_EXPRESSION_NAMED(
+      (ret != RCL_RET_OK && ret != RCL_RET_NOT_INIT), ROS_PACKAGE_NAME,
+      "Failed to fini get_type_description service for node: %i", ret);
+
     ret = rcl_node_type_cache_fini(node);
     RCUTILS_LOG_ERROR_EXPRESSION_NAMED(
       (ret != RCL_RET_OK && ret != RCL_RET_NOT_INIT), ROS_PACKAGE_NAME,
@@ -390,6 +409,11 @@ rcl_node_fini(rcl_node_t * node)
       RCL_SET_ERROR_MSG("Unable to fini publisher for node.");
       result = RCL_RET_ERROR;
     }
+  }
+  rcl_ret = rcl_node_type_description_service_fini(node);
+  if (rcl_ret != RCL_RET_OK && rcl_ret != RCL_RET_NOT_INIT) {
+    RCL_SET_ERROR_MSG("Unable to fini ~/get_type_description service for node.");
+    result = RCL_RET_ERROR;
   }
   rcl_ret = rcl_node_type_cache_fini(node);
   if (rcl_ret != RCL_RET_OK && rcl_ret != RCL_RET_NOT_INIT) {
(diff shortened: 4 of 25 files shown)


rcl/src/rcl/node_type_cache.c
```diff
@@ -0,0 +1,251 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifdef __cplusplus
+extern "C" {
+#endif
```
emersonknapp, line 17:
```suggestion
```

You don't need these in `.c` files, or anywhere in `src/`. I think there are one or two things that do it in here, but they don't have to.

rcl/src/rcl/type_description_conversions.c
```diff
@@ -0,0 +1,332 @@
...
+#include "rosidl_runtime_c/type_description/field__functions.h"
+#include "rosidl_runtime_c/type_description/individual_type_description__functions.h"
+#include "rosidl_runtime_c/type_description/type_description__functions.h"
+#include "rosidl_runtime_c/type_description/type_source__functions.h"
+#include "type_description_interfaces/msg/detail/field__functions.h"
+#include "type_description_interfaces/msg/individual_type_description.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
```
emersonknapp, line 27:
```suggestion
```

emersonknapp (changes requested):

rcl/src/rcl/node.c
```diff
@@ -536,6 +565,194 @@ rcl_get_disable_loaned_message(bool * disable_loaned_message)
...
+    RCL_SET_ERROR_MSG("Service already initialized");
+    return RCL_RET_ALREADY_INIT;
+  }
+
+  char * serviceName = NULL;
+  const rosidl_service_type_support_t * type_support =
+    ROSIDL_GET_SRV_TYPE_SUPPORT(
+    type_description_interfaces, srv,
+    GetTypeDescription);
+  // TODO(achim-k): Pass in as argument?
```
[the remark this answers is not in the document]
achim-k: The service options in the line below. Right now I'm using the default options. Do you think that's ok?

rcl/src/rcl/node_type_cache.c
```diff
@@ -0,0 +1,251 @@
...
+
+  /// The actual type info.
+  rcl_type_info_t type_info;
+} rcl_type_info_with_registration_count_t;
+
+size_t get_type_hash_hashmap_key(const void * key)
+{
+  // Reinterpret-cast the first sizeof(size_t) bytes of the hash value
+  const rosidl_type_hash_t * type_hash = key;
+  return (size_t)type_hash->value[0];
```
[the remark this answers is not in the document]
achim-k: That's an important one, good catch

rcl/src/rcl/node.c
```diff
@@ -536,6 +565,194 @@ rcl_get_disable_loaned_message(bool * disable_loaned_message)
...
+  rcl_ret_t ret =
+    rcl_node_type_cache_get_type_info(node, &type_hash, &type_info);
+
+  if (RCUTILS_RET_OK == ret) {
+    response.successful = type_description_interfaces__msg__TypeDescription__copy(
+      type_info.type_description, &response.type_description);
+
+    if (request.include_type_sources) {
+      response.successful = type_description_interfaces__msg__TypeSource__Sequence__copy(
+        type_info.type_sources, &response.type_sources);
```
achim-k, line 620:
```suggestion
      response.successful &= type_description_interfaces__msg__TypeSource__Sequence__copy(
        type_info.type_sources, &response.type_sources);
```

rcl/src/rcl/node.c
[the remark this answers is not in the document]
emersonknapp: I see. I think it is better to leave it the way it is right now - the Node decides what options the service should have, the user can only choose to turn it on or off. Unless, do the parameter services allow any customization? If they do then maybe this should too

rcl/include/rcl/type_description_conversions.h
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | Yes
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] in the handle to the runtime type description struct
```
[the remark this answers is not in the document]
emersonknapp: Looks like this was marked resolved but I don't see a diff in the file fixing it (or a comment explaining a decision not to do it)

rcl/src/rcl/node.c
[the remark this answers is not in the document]
achim-k:
```suggestion
```

rcl/src/rcl/node.c
```diff
@@ -536,6 +564,173 @@ rcl_get_disable_loaned_message(bool * disable_loaned_message)
...
+      &node->impl->get_type_description_service,
+      &request_header, &request))
+  {
+    RCUTILS_LOG_ERROR_NAMED(ROS_PACKAGE_NAME, "Failed to take request");
+    goto cleanup;
+  }
+
+  rosidl_type_hash_t type_hash;
+  if (RCUTILS_RET_OK !=
+    rosidl_parse_type_hash_string(request.type_hash.data, &type_hash))
```
achim-k, line 603: `request.type_name` is currently ununsed

james-rms:
Gist: https://gist.githubusercontent.com/james-rms/22e9369a4fd54e4d18d0e0323d88c489/raw/47868239244ef18091c0f06c166de7a54e361caa/ros2.repos
BUILD args:  --packages-above-and-dependencies rcl rcl_action
TEST args:  --packages-above rcl rcl_action
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11831
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18480/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13028/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19200/)

james-rms:
@clalancette @wjwwood are you able to take a look at this PR?

emersonknapp (approved):
LGTM

emersonknapp:
We probably won't get any reviews from the others until at least tomorrow - today is rmw freeze day, but the deadline for this PR is next Monday so we're OK. I'm actually going to hold off on re-running CI too just to leave the build servers open for more pressing issues for today's EOD freeze.

achim-k:
Rebased to include #1060. I have also squashed commits to make it easier reviewable.

rcl/include/rcl/node.h
```diff
@@ -549,6 +549,57 @@ RCL_PUBLIC
 rcl_ret_t
 rcl_get_disable_loaned_message(bool * disable_loaned_message);
 
+/// Initialize the node's ~/get_type_description service.
+/**
+ * This function initializes the node's private ~/get_type_description service
+ * which can be used to retrieve information about types used by the node's
+ * publishers, subscribers, services or actions.
+ *
+ * Node that this function will be called in `rcl_init_node` if the node option
```
clalancette:
```suggestion
 * Note that this function will be called in `rcl_init_node` if the node option
```

rcl/include/rcl/type_description_conversions.h
```diff
@@ -0,0 +1,133 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/// Convert type description runtime struct to msg struct.
+/**
+ * This function converts a rosidl_runtime_c__type_description__TypeDescription
+ * to the corresponding type_description_interfaces/msg/TypeDescription struct.
+ * The retrieved pointer shall be destroyed with
+ * `type_description_interfaces__msg__TypeDescription__destroy() after` use.
```
clalancette:
```suggestion
 * `type_description_interfaces__msg__TypeDescription__destroy()` after use.
```

rcl/include/rcl/type_description_conversions.h
```diff
@@ -0,0 +1,133 @@
...
+type_description_interfaces__msg__TypeDescription *
+rcl_convert_type_description_runtime_to_msg(
+  const rosidl_runtime_c__type_description__TypeDescription * runtime_description);
+
+/// Convert type description msg struct to a rosidl runtime struct.
+/**
+ * This function converts a type_description_interfaces/msg/TypeDescription
+ * to the corresponding rosidl_runtime_c__type_description__TypeDescription
+ * struct. The retrieved pointer shall be destroyed with
+ * `rosidl_runtime_c__type_description__TypeDescription__destroy() after` use.
```
clalancette:
```suggestion
 * `rosidl_runtime_c__type_description__TypeDescription__destroy()` after use.
```

rcl/include/rcl/type_description_conversions.h
```diff
@@ -0,0 +1,133 @@
...
+rosidl_runtime_c__type_description__TypeDescription *
+rcl_convert_type_description_msg_to_runtime(
+  const type_description_interfaces__msg__TypeDescription * description_msg);
+
+/// Convert type sources sequence runtime struct to msg struct.
+/**
+ * This function converts a rosidl_runtime_c__type_description__TypeSource__Sequence
+ * struct to the corresponding type_description_interfaces__msg__TypeSource__Sequence
+ * struct. The retrieved pointer shall be destroyed with
+ * `type_description_interfaces__msg__TypeSource__Sequence__destroy() after` use.
```
clalancette:
```suggestion
 * `type_description_interfaces__msg__TypeSource__Sequence__destroy()` after use.
```

rcl/include/rcl/type_description_conversions.h
```diff
@@ -0,0 +1,133 @@
...
+type_description_interfaces__msg__TypeSource__Sequence *
+rcl_convert_type_source_sequence_runtime_to_msg(
+  const rosidl_runtime_c__type_description__TypeSource__Sequence * runtime_type_sources);
+
+/// Convert type sources sequece msg struct to a rosidl runtime struct.
+/**
+ * This function converts a rosidl_runtime_c__type_description__TypeSource__Sequence
+ * struct to the corresponding type_description_interfaces__msg__TypeSource__Sequence
+ * struct. The retrieved pointer shall be destroyed with
+ * `type_description_interfaces__msg__TypeSource__Sequence__destroy() after` use.
```
clalancette:
```suggestion
 * `type_description_interfaces__msg__TypeSource__Sequence__destroy()` after use.
```

rcl/src/rcl/node.c
```diff
@@ -284,6 +283,20 @@ rcl_node_init(
     // error message already set
     goto fail;
   }
+  // Initialize the node type cache hash map
+  ret = rcl_node_type_cache_init(node);
```
clalancette, line 289: I may just be missing something, but `rcl_node_type_cache_fini` doesn't seem to be getting called in a failure case.

rcl/src/rcl/node_impl.h
```diff
@@ -0,0 +1,81 @@
...
+ * Lock-Free          | Yes
+ *
+ * \param[in] node the handle to the node whose type cache should be initialized
+ * \return #RCL_RET_OK if the node's type cache was successfully initialized, or
+ * \return #RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
+ * \return #RCL_RET_NODE_INVALID if the given `node` is invalid, or
+ * \return #RCL_RET_ERROR if an unspecified error occurs.
+ */
+RCL_WARN_UNUSED
+rcl_ret_t rcl_node_type_cache_init(rcl_node_t * node);
```
clalancette: It doesn't really fit our pattern to have `rcl_node_type_cache_init/rcl_node_type_cache_fini` declared in `node_impl.h`, but then defined in `node_type_cache.c`.  I'll suggest moving these to `node_type_cache.h`.

rcl/src/rcl/node_type_cache.c
```diff
@@ -0,0 +1,240 @@
...
+#include "rcutils/logging_macros.h"
+#include "rcutils/types/hash_map.h"
+
+#include "./context_impl.h"
+#include "./node_impl.h"
+
+typedef struct rcl_type_info_with_registration_count_t
+{
+  /// Counter to keep track of registrations
+  size_t numRegistrations;
```
clalancette:
```suggestion
  size_t num_registrations;
```

rcl/src/rcl/node_type_cache.c
```diff
@@ -0,0 +1,240 @@
...
+
+  /// The actual type info.
+  rcl_type_info_t type_info;
+} rcl_type_info_with_registration_count_t;
+
+size_t get_type_hash_hashmap_key(const void * key)
```
clalancette:
```suggestion
static size_t get_type_hash_hashmap_key(const void * key)
```

rcl/src/rcl/node_type_cache.c
```diff
@@ -0,0 +1,240 @@
...
+{
+  // Reinterpret-cast the first sizeof(size_t) bytes of the hash value
+  const rosidl_type_hash_t * type_hash = key;
+  return *(size_t *)type_hash->value;
+}
+
+int cmp_type_hash(const void * val1, const void * val2)
```
clalancette:
```suggestion
static int cmp_type_hash(const void * val1, const void * val2)
```

rcl/src/rcl/node_type_cache.c
```diff
@@ -0,0 +1,240 @@
...
+    return RCL_RET_NOT_INIT;
+  }
+
+  return RCL_RET_ERROR;
+}
+
+rcl_ret_t rcl_node_type_cache_register_type(
+  const rcl_node_t * node, const rosidl_type_hash_t * type_hash,
+  const rosidl_runtime_c__type_description__TypeDescription
+  * type_description,
```
clalancette: This is a nit, but can we have this on the same line?

rcl/src/rcl/node_type_cache.c
```diff
@@ -0,0 +1,240 @@
...
+      type_hash))
+  {
+    if (RCUTILS_RET_OK !=
+      rcutils_hash_map_get(
+        &node->impl->registered_types_by_type_hash,
+        type_hash, &type_info_with_registrations))
+    {
+      RCL_SET_ERROR_MSG("Failed to retrieve type info");
+      return RCL_RET_ERROR;
+    }
```
clalancette: This is somewhat minor, but I think you can get rid of the call to `rcutils_hash_map_key_exists` and instead just check whether `rcutils_hash_map_get` returns `RCUTILS_RET_OK` or `RCUTILS_RET_NOT_FOUND`.

rcl/src/rcl/node_type_cache.c
```diff
@@ -0,0 +1,240 @@
...
+    RCL_CHECK_FOR_NULL_WITH_MSG(
+      type_info_with_registrations.type_info.type_description,
+      "converting type description struct failed", return RCL_RET_ERROR);
+
+    // Convert type sources struct to type sources message struct.
+    type_info_with_registrations.type_info.type_sources =
+      rcl_convert_type_source_sequence_runtime_to_msg(type_description_sources);
+    RCL_CHECK_FOR_NULL_WITH_MSG(
+      type_info_with_registrations.type_info.type_description,
+      "converting type sources struct failed", return RCL_RET_ERROR);
```
clalancette: If we fail here, I think we are going to leak `type_info_with_registrations.type_info.type_description`.

rcl/src/rcl/node_type_cache.c
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  // Update the hash map entry.
+  if (RCUTILS_RET_OK !=
+    rcutils_hash_map_set(
+      &node->impl->registered_types_by_type_hash,
+      type_hash, &type_info_with_registrations))
+  {
+    RCL_SET_ERROR_MSG("Failed to update type info");
+    return RCL_RET_ERROR;
```
clalancette, line 194: Similar here; if we are adding in a new type, and this fails, I think we are going to leak both the `type_info_with_registrations.type_info.type_description` as well as `type_info_with_registrations.type_info.type_sources`.

rcl/src/rcl/node_type_cache.c
```diff
@@ -0,0 +1,240 @@
...
+
+  /// The actual type info.
+  rcl_type_info_t type_info;
+} rcl_type_info_with_registration_count_t;
+
+size_t get_type_hash_hashmap_key(const void * key)
+{
+  // Reinterpret-cast the first sizeof(size_t) bytes of the hash value
+  const rosidl_type_hash_t * type_hash = key;
+  return *(size_t *)type_hash->value;
```
clalancette, line 38: This seems somewhat suspicious to me.  First of all, the size of `size_t` can vary platform-to-platform (particularly on 32-bit vs 64-bit).  Second, can you explain more about why are we only looking at the first 4 or 8 bytes of the hash value?

rcl/src/rcl/service.c
```diff
@@ -118,7 +120,6 @@ rcl_service_init(
   RCL_CHECK_FOR_NULL_WITH_MSG(
     service->impl, "allocating memory failed",
     return RCL_RET_BAD_ALLOC;);
-
```
clalancette: Nit: let's leave this whitespace intact.

rcl/src/rcl/service.c
```diff
@@ -138,6 +139,18 @@ rcl_service_init(
...
+  if (RCL_RET_OK !=
+    rcl_node_type_cache_register_type(
+      node, type_support->get_type_hash_func(type_support),
+      type_support->get_type_description_func(type_support),
+      type_support->get_type_description_sources_func(type_support)))
+  {
+    rcutils_reset_error();
+    RCL_SET_ERROR_MSG("Failed to register type for service");
+    ret = RCL_RET_ERROR;
+    goto free_remapped_service_name;
```
clalancette: Because of this change, you actually need to change the cleanup below so that you successfully call `unregister_type` on failure, and then change the error handling for `rmw_create_service` to jump there instead.

rcl/src/rcl/subscription.c
```diff
@@ -110,6 +111,17 @@ rcl_subscription_init(
...
+  if (RCL_RET_OK !=
+    rcl_node_type_cache_register_type(
+      node, type_support->get_type_hash_func(type_support),
+      type_support->get_type_description_func(type_support),
+      type_support->get_type_description_sources_func(type_support)))
+  {
+    rcutils_reset_error();
+    RCL_SET_ERROR_MSG("Failed to register type for subscription");
+    goto fail;
+  }
```
clalancette: During the failure case, you need to `unregister_type`.

rcl/src/rcl/type_description_conversions.c
```diff
@@ -0,0 +1,324 @@
...
+#include "type_description_interfaces/msg/detail/field__functions.h"
+#include "type_description_interfaces/msg/individual_type_description.h"
+
+static bool individual_type_description_runtime_to_msg(
+  const rosidl_runtime_c__type_description__IndividualTypeDescription * in,
+  type_description_interfaces__msg__IndividualTypeDescription * out)
+{
+  if (NULL == in) {
+    return false;
+  }
```
clalancette: We should have a similar check for `out`.  Additionally, we may want to consider using the `RCL_CHECK_ARGUMENT_FOR_NULL` macros here.

rcl/src/rcl/type_description_conversions.c
```diff
@@ -0,0 +1,324 @@
...
+  return false;
+}
+
+static bool individual_type_description_msg_to_runtime(
+  const type_description_interfaces__msg__IndividualTypeDescription * in,
+  rosidl_runtime_c__type_description__IndividualTypeDescription * out)
+{
+  if (NULL == in) {
+    return false;
+  }
```
clalancette: We should have a similar check for out. Additionally, we may want to consider using the RCL_CHECK_ARGUMENT_FOR_NULL macros here.

rcl/src/rcl/type_description_conversions.c
```diff
@@ -0,0 +1,324 @@
...
+  return false;
+}
+
+type_description_interfaces__msg__TypeDescription *
+rcl_convert_type_description_runtime_to_msg(
+  const rosidl_runtime_c__type_description__TypeDescription * runtime_description)
+{
+  if (NULL == runtime_description) {
+    return NULL;
+  }
```
clalancette: Consider using the RCL_CHECK_ARGUMENT_FOR_NULL macros here.

rcl/src/rcl/type_description_conversions.c
```diff
@@ -0,0 +1,324 @@
...
+  return NULL;
+}
+
+rosidl_runtime_c__type_description__TypeDescription *
+rcl_convert_type_description_msg_to_runtime(
+  const type_description_interfaces__msg__TypeDescription * description_msg)
+{
+  if (NULL == description_msg) {
+    return NULL;
+  }
```
clalancette: Consider using the RCL_CHECK_ARGUMENT_FOR_NULL macros here.

rcl/src/rcl/type_description_conversions.c
```diff
@@ -0,0 +1,324 @@
...
+  return NULL;
+}
+
+type_description_interfaces__msg__TypeSource__Sequence *
+rcl_convert_type_source_sequence_runtime_to_msg(
+  const rosidl_runtime_c__type_description__TypeSource__Sequence * runtime_type_sources)
+{
+  if (NULL == runtime_type_sources) {
+    return NULL;
+  }
```
clalancette: Consider using the RCL_CHECK_ARGUMENT_FOR_NULL macros here.

rcl/src/rcl/type_description_conversions.c
```diff
@@ -0,0 +1,324 @@
...
+  return NULL;
+}
+
+rosidl_runtime_c__type_description__TypeSource__Sequence *
+rcl_convert_type_source_sequence_msg_to_runtime(
+  const type_description_interfaces__msg__TypeSource__Sequence * type_sources_msg)
+{
+  if (NULL == type_sources_msg) {
+    return NULL;
+  }
```
clalancette: Consider using the RCL_CHECK_ARGUMENT_FOR_NULL macros here.

rcl/src/rcl/node.c
```diff
@@ -542,6 +564,173 @@ rcl_get_disable_loaned_message(bool * disable_loaned_message)
...
+  }
+
+  rosidl_type_hash_t type_hash;
+  if (RCUTILS_RET_OK !=
+    rosidl_parse_type_hash_string(request.type_hash.data, &type_hash))
+  {
+    RCUTILS_LOG_ERROR_NAMED(
+      ROS_PACKAGE_NAME, "Failed to parse type hash '%s'",
+      request.type_hash.data);
+  }
```
clalancette, line 590: I think we need some additional error handling here.  In particular, we don't want to pass potentially uninitialized data (in `type_hash`) to `rcl_node_type_cache_get_type_info` below if this fails.

rcl/src/rcl/node.c
```diff
@@ -542,6 +564,173 @@ rcl_get_disable_loaned_message(bool * disable_loaned_message)
...
+      rosidl_runtime_c__String__assign(
+        &response.failure_reason,
+        "Failed to populate response");
+    }
+  } else {
+    response.successful = false;
+    RCUTILS_LOG_INFO_NAMED(
+      ROS_PACKAGE_NAME,
+      "Type '%s' not found in type cache",
+      request.type_hash.data);
```
clalancette: I don't think we should put this print here.  This allows a remote node to spam the console of this node, which is probably not what we want.

rcl/src/rcl/node.c
```diff
@@ -542,6 +564,173 @@ rcl_get_disable_loaned_message(bool * disable_loaned_message)
...
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(node, RCL_RET_INVALID_ARGUMENT);
+  RCL_CHECK_ARGUMENT_FOR_NULL(node->impl, RCL_RET_NODE_INVALID);
+
+  if (rcl_service_is_valid(&node->impl->get_type_description_service)) {
+    return RCL_RET_ALREADY_INIT;
+  }
+  rcl_reset_error();  // Reset the error message set by rcl_service_is_valid()
+
+  char * serviceName = NULL;
```
clalancette:
```suggestion
  char * service_name = NULL;
```

rcl/src/rcl/node.c
```diff
@@ -542,6 +564,173 @@ rcl_get_disable_loaned_message(bool * disable_loaned_message)
...
+  rcl_allocator_t allocator = node->context->impl->allocator;
+
+  // Construct service name
+  rcl_ret_t ret = rcl_node_resolve_name(
+    node, "~/get_type_description",
+    allocator, true, true, &serviceName);
+  if (RCL_RET_OK != ret) {
+    RCL_SET_ERROR_MSG(
+      "Failed to construct ~/get_type_description service name");
+    goto cleanup;
```
clalancette: I think we can just return here, as if this failed we haven't allocated any memory up to this point.

rcl/src/rcl/node.c
```diff
@@ -542,6 +564,173 @@ rcl_get_disable_loaned_message(bool * disable_loaned_message)
...
+    RCL_SET_ERROR_MSG(
+      "Failed to set callback for ~/get_type_description service");
+  }
+
+cleanup:
+  if (serviceName) {
+    allocator.deallocate(serviceName, allocator.state);
+  }
+
+  return ret;
```
clalancette: I think you can revamp the error handling here so that this is simpler.  In particular:
```suggestion
  ret = rcl_service_init(
    &node->impl->get_type_description_service, node,
    type_support, serviceName, &service_ops);
  allocator.deallocate(serviceName, allocator.state);
  if (RCL_RET_OK != ret) {
    return ret;
  }

  // Set service callback
  ret = rcl_service_set_on_new_request_callback(
    &node->impl->get_type_description_service,
    rcl_node_get_type_description_service_callback, node);
  if (RCL_RET_OK != ret) {
    goto fail_fini_service;
  }

  return RCL_RET_OK;

fail_fini_service:
  if (RCL_RET_OK !=
    rcl_service_fini(&node->impl->get_type_description_service, node))
  {
    RCUTILS_LOG_ERROR_NAMED(
      ROS_PACKAGE_NAME,
      "Failed to deinitialize ~/get_type_description service.");
  }

  return ret;
```

rcl/src/rcl/node_options.c
```diff
@@ -36,7 +36,7 @@ rcl_node_get_default_options()
     .arguments = rcl_get_zero_initialized_arguments(),
     .enable_rosout = true,
     .rosout_qos = rcl_qos_profile_rosout_default,
-    .enable_type_description_service = false,
+    .enable_type_description_service = true,
```
clalancette: As I've said in our weekly meetings, this is my biggest concern with this PR.  This means that for every new node a user constructs, we will now, by default, create the following services:

* `describe_parameters`
* `get_parameter_types`
* `get_parameters`
* `list_parameters`
* `set_parameters`
* `set_parameters_atomically`
* `get_type_description_service`

With each of those nominally taking 2 topics, that means that every node, by default, will now use 14 topics for services.  Add to that the default `/parameter_events`, and `/rosout` topics that are published/subscribed to, and we now have every node creating 16 topics before any user topics are included.  And given the fully-connected graph nature of DDS, that means a *lot* of discovery traffic at startup.

So the issue here is a system/integration issue, rather than anything particularly wrong with this PR.  There are a couple of ideas around this that we have been kicking around for a while, but that would definitely help mitigate this:

1.  All of those `*_parameter` services are excessive.  We honestly only need two: `get_parameters` and `set_parameters`.  `get_parameters`, with some appropriate configuration, could be used for `describe_parameters`, `get_parameter_types`, `get_parameters`, and `list_parameters`.  And `set_parameters` could be used for both `set_parameters` and `set_parameters_atomically`.  If we did this, then even with the addition of this PR we would be down to 8 topics per node.  This is my preferred solution to this issue, though it will probably be a bunch of work to fix this in the core.
2.  We have also had this idea of "lazy" services.  That is, the services would somehow advertise that they are available, but not actually go through the entire discovery process on startup.  They would only do full discovery when they want to be called.  I'm actually unsure of the details of how this one would work; @wjwwood might have more thoughts.

clalancette (changes requested):

rcl/src/rcl/node.c
achim-k:
```suggestion
```

rcl/src/rcl/node.c
achim-k:
```suggestion
    return ret;
```

rcl/src/rcl/node_type_cache.c
achim-k: @emersonknapp can you comment on Chris' question? That's your magic code ;)

rcl/src/rcl/node.c
achim-k: I have changed this to send an error response and then return when the type hash could not be parsed

achim-k:
@clalancette thanks for your review. I have addressed most of your comments and added two commits that contain the changes. What's the policy here, is it OK to do a rebase? If so, I could squash the changes into previous commits to make it easier to review (commit by commit). Also, the base is rather outdated (959ef2836e8cb88bcc1805752ad857127684ac4d), which might be a reason why CI is failing

clalancette:
> @clalancette thanks for your review. I have addressed most of your comments and added two commits that contain the changes. What's the policy here, is it OK to do a rebase? If so, I could squash the changes into previous commits to make it easier to review (commit by commit). Also, the base is rather outdated ([959ef28](https://github.com/ros2/rcl/commit/959ef2836e8cb88bcc1805752ad857127684ac4d)), which might be a reason why CI is failing

Please feel free to rebase as you need.

rcl/src/rcl/node_type_cache.c
emersonknapp: Not magic, I promise!

The signature for the `rcutils_hash_map` key hashing function is `size_t (const void * key)`  - this is sensible because the hashmap is just an array of buckets, and an array is indexed/addressed by `size_t`, which yes is system-dependent in size. So that explains why it's a `size_t`, that's the required (eventual) hash type in order to use the `rcutils_hash_map`.

That being the constraint - our keys for this hash map are a RIHS01 value, which is a SHA256 value. This is already a hash of real content, it would be a waste of computation to perform further hashing on it, it's as "random" as it could be for the information represented - this includes within the hash itself. None of the bits of a SHA256 hash should be considered privileged to any others in terms of information content, so for a `size_t` amount of uniqueness, we can just lop off a `size_t`-worth of the full 256 bits to use as our hash. The beginning, middle, or end should make no difference.

I'll note that I'm not so positive on the above paragraph that I'd use it without further study for security purposes, but in this case it's just for a simple lookup table so I expect on average no more collisions between "first N bytes of SHA256 value" vs "SHA256 put through some hashing function to output an N byte hash", and the former option requires no computation.

rcl/src/rcl/node_options.c
emersonknapp: While it's definitely not my preferred outcome - I would be open to an initial merge of this PR with `enable_type_description_service = false` by default.

Option 1 would be a good improvement on the current state, I agree it's gotten bloated. Do you mean we'd get down to 6 topics? `get_parameters`, `set_parameters`, `get_type_description` = 3 * 2.

Option 2 - "lazy" services would be great - since for _all_ of these services, I don't think anybody cares about them independently - you learn about the existence of a Node first, then you care about calling these node-specific services.

clalancette: After a lot of discussion, what we ended up deciding to do here is to merge this with the type description service disabled to start with.  Then we have a further plan to enable this by default, but adding in a parameter so it is easy to disable.  And then follow all of that up with lazy services.  So I'm going to resolve this particular conversation.

rcl/src/rcl/node_type_cache.c
clalancette: I see, that makes sense.  Thanks for the explanation!

rcl/src/rcl/node.c
```diff
@@ -284,6 +283,20 @@ rcl_node_init(
     // error message already set
     goto fail;
   }
+  // Initialize the node type cache hash map
+  ret = rcl_node_type_cache_init(node);
+  if (ret != RCL_RET_OK) {
+    // error message already set
+    goto fail;
+  }
```
clalancette: Sorry, one thing I just noticed here.  If the `get_type_description_service` service is disabled, we don't need the type cache at all, right?  That is, when it is disabled we can save memory by not allocating the cache and then not adding things to it.  So I'll suggest we move this inside of the `if (node->impl->options.enable_type_description_service) {`, and then have a similar `if` statement inside of `publisher_init`/`subscription_init`/`service_init`.

rcl/src/rcl/node.c
```diff
@@ -310,6 +317,16 @@ rcl_node_init(
...
   if (node->impl) {
+    ret = rcl_node_type_description_service_fini(node);
+    RCUTILS_LOG_ERROR_EXPRESSION_NAMED(
+      (ret != RCL_RET_OK && ret != RCL_RET_NOT_INIT), ROS_PACKAGE_NAME,
+      "Failed to fini get_type_description service for node: %i", ret);
+
+    ret = rcl_node_type_cache_fini(node);
+    RCUTILS_LOG_ERROR_EXPRESSION_NAMED(
+      (ret != RCL_RET_OK && ret != RCL_RET_NOT_INIT), ROS_PACKAGE_NAME,
+      "Failed to fini node_type_cache for node: %i", ret);
```
clalancette: Similarly, I think we only need to do this teardown if we initialized it to begin with.  This is not a huge deal, but might avoid some unrelated spam on the console when tearing down from a failure.

rcl/src/rcl/node.c
```diff
@@ -387,6 +404,16 @@ rcl_node_fini(rcl_node_t * node)
...
+  rcl_ret = rcl_node_type_description_service_fini(node);
+  if (rcl_ret != RCL_RET_OK && rcl_ret != RCL_RET_NOT_INIT) {
+    RCL_SET_ERROR_MSG("Unable to fini ~/get_type_description service for node.");
+    result = RCL_RET_ERROR;
+  }
+  rcl_ret = rcl_node_type_cache_fini(node);
+  if (rcl_ret != RCL_RET_OK && rcl_ret != RCL_RET_NOT_INIT) {
+    RCL_SET_ERROR_MSG("Unable to fini type cache for node.");
+    result = RCL_RET_ERROR;
+  }
```
clalancette: And the same here; we should only teardown if we initialized to begin with.

clalancette (changes requested):
I've left inline one more minor conceptual thing I think we should consider fixing.  Once that is done, I think this is good for CI.

rcl/src/rcl/node.c
emersonknapp: These two return `RCL_NOT_INIT` if the object is not initialized, and the error printing checks for `ret != RCL_RET_OK && ret != RCL_RET_NOT_INIT` - so if the objects weren't initialized they will move on silently.

rcl/src/rcl/node.c
emersonknapp: Same here - `fini` checks for and returns `RCL_NOT_INIT`, and the error printing checks for that case

rcl/src/rcl/node.c
emersonknapp: One concern with this - if we allow the service to be turned on and off at runtime (something we discussed), then if the service was turned on after some sub/pub/srv creation, those types won't be available in its type cache.

clalancette: We did discuss that, but I remember us saying that this should be a read-only parameter.  That is, you can only set it at startup time and then never again.  Mostly I want to make it so that people who don't want to use the type_description service don't pay for anything (including in memory).  Further, I think that if we wanted to make this a runtime-configurable option in the future, it would be a relatively simple change to make that parameter read-write, and then move this stuff back to always being tracked.

rcl/src/rcl/node.c
emersonknapp: Sorry - you're right, it will still leave an error state that will add spam in _other_ failure teardown. Will address

rcl/src/rcl/node.c
emersonknapp: Sorry - you're right, it will still leave an error state that will add spam in _other_ failure teardown. Will address

rcl/src/rcl/node.c
emersonknapp: OK yeah, I'm back on board

emersonknapp:
Pulls: ros2/rcl#1052
Gist: https://gist.githubusercontent.com/emersonknapp/518d9f5873376cab656059c63d00995b/raw/47868239244ef18091c0f06c166de7a54e361caa/ros2.repos
BUILD args:  --packages-above-and-dependencies rcl rcl_action
TEST args:  --packages-above rcl rcl_action
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12251
* <s>Linux</s> [Build Status](http://ci.ros2.org/job/ci_linux/18974/) (infra failure, auto-rebuild)
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/18977/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13501/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19720/)

emersonknapp:
Looks like the test is just failing on connext... investigating now. Really strange, it's giving "bad take" for the client response. In debugging, Connext is hanging on `rcl_wait` for the service, even with a timeout, so something weird is going on.

rcl/test/rcl/test_get_type_description_service.cpp
```diff
@@ -0,0 +1,318 @@
...
+  type_description_interfaces__srv__GetTypeDescription_Request__fini(&client_request);
+  ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+
+  // Initialize the response owned by the client and take the response.
+  type_description_interfaces__srv__GetTypeDescription_Response client_response;
+  type_description_interfaces__srv__GetTypeDescription_Response__init(&client_response);
+
+  // Retrieve the response.
+  rmw_service_info_t header;
+  ret = rcl_take_response_with_info(&client, &header, &client_response);
```
emersonknapp, line 269: Here - I think we need to wait for the response to be ready before taking it. In Connext, the response isn't ready yet at the time this is called. However, if I add a `wait_for_client_to_be_ready` there is some deadlock inside the Connext binaries (non-public code from the deb) so it hangs indefinitely, even with the timeout set. Here's the stack trace where it is hanging:

```
#0  futex_wait (private=0, expected=2, futex_word=0x55555596e938) at ../sysdeps/nptl/futex-internal.h:146
#1  __GI___lll_lock_wait (futex=futex@entry=0x55555596e938, private=0) at ./nptl/lowlevellock.c:49
#2  0x00007ffff7246082 in lll_mutex_lock_optimized (mutex=0x55555596e938) at ./nptl/pthread_mutex_lock.c:48
#3  ___pthread_mutex_lock (mutex=0x55555596e938) at ./nptl/pthread_mutex_lock.c:93
#4  0x00007ffff4ed1139 in RTIOsapiSemaphore_take () from /opt/rti.com/rti_connext_dds-6.0.1/lib/x64Linux4gcc7.3.0/libnddscore.so
#5  0x00007ffff4eb1ace in REDAWorker_enterExclusiveArea () from /opt/rti.com/rti_connext_dds-6.0.1/lib/x64Linux4gcc7.3.0/libnddscore.so
#6  0x00007ffff4eb4465 in REDACursor_modifyReadWriteArea () from /opt/rti.com/rti_connext_dds-6.0.1/lib/x64Linux4gcc7.3.0/libnddscore.so
#7  0x00007ffff4bf8d21 in PRESPsReader_readOrTakeInstanceI () from /opt/rti.com/rti_connext_dds-6.0.1/lib/x64Linux4gcc7.3.0/libnddscore.so
#8  0x00007ffff4bf982c in PRESPsReader_takeInstance () from /opt/rti.com/rti_connext_dds-6.0.1/lib/x64Linux4gcc7.3.0/libnddscore.so
#9  0x00007ffff564c4d3 in DDS_DataReader_read_or_take_instance_untypedI () from /opt/rti.com/rti_connext_dds-6.0.1/lib/x64Linux4gcc7.3.0/libnddsc.so
#10 0x00007ffff5de43dd in rmw_connextdds_take_samples(RMW_Connext_Subscriber*) () from /ws/install/r/rmw_connextdds_common/lib/librmw_connextdds_common_pro.so
#11 0x00007ffff5dabcf8 in RMW_Connext_Subscriber::loan_messages(bool) () from /ws/install/r/rmw_connextdds_common/lib/librmw_connextdds_common_pro.so
#12 0x00007ffff5dbbf97 in RMW_Connext_Subscriber::has_data() () from /ws/install/r/rmw_connextdds_common/lib/librmw_connextdds_common_pro.so
#13 0x00007ffff5db7f42 in RMW_Connext_WaitSet::detach(rmw_subscriptions_s*, rmw_guard_conditions_s*, rmw_services_s*, rmw_clients_s*, rmw_events_s*, unsigned long&) ()
   from /ws/install/r/rmw_connextdds_common/lib/librmw_connextdds_common_pro.so
#14 0x00007ffff5db8f55 in RMW_Connext_WaitSet::wait(rmw_subscriptions_s*, rmw_guard_conditions_s*, rmw_services_s*, rmw_clients_s*, rmw_events_s*, rmw_time_s const*) ()
   from /ws/install/r/rmw_connextdds_common/lib/librmw_connextdds_common_pro.so
#15 0x00007ffff7f5753e in rcl_wait () from /ws/install/r/rcl/lib/librcl.so
#16 0x00005555555a3fd0 in wait_for_client_to_be_ready(rcl_client_s*, rcl_context_s*, unsigned long, long) ()
#17 0x0000555555568921 in TestGetTypeDescSrvEnabledFixture__rmw_connextdds_test_service_nominal_Test::TestBody() ()
```

Note, this doesn't happen with the other two implementations, this test works on those with and without the `wait`

emersonknapp: More context: the `rcl_node_get_type_description_service_callback` does get called (from a Connext-created thread), but hangs on attempt to actually get the request message. Trying to find what's creating the deadlock

```
Thread 9 (Thread 0x7ffff1a4f640 (LWP 404837) "rR02000a59fd640"):                                                                                                                                                                                              
#0  futex_wait (private=0, expected=2, futex_word=0x555555f53738) at ../sysdeps/nptl/futex-internal.h:146                                                                                                                                                     
#1  __GI___lll_lock_wait (futex=futex@entry=0x555555f53738, private=0) at ./nptl/lowlevellock.c:49                                                                                                                                                            
#2  0x00007ffff7246082 in lll_mutex_lock_optimized (mutex=0x555555f53738) at ./nptl/pthread_mutex_lock.c:48                                                                                                                                                   
#3  ___pthread_mutex_lock (mutex=0x555555f53738) at ./nptl/pthread_mutex_lock.c:93                                                                                                                                                                            
#4  0x00007ffff71a39c3 in __gthread_mutex_lock(pthread_mutex_t*) () from /ws/install/r/rcl_logging_spdlog/lib/librcl_logging_spdlog.so                                                                                                                        
#5  0x00007ffff71a3a7e in std::mutex::lock() () from /ws/install/r/rcl_logging_spdlog/lib/librcl_logging_spdlog.so                                                                                                                                            
#6  0x00007ffff71a48cc in std::lock_guard<std::mutex>::lock_guard(std::mutex&) () from /ws/install/r/rcl_logging_spdlog/lib/librcl_logging_spdlog.so                                                                                                          
#7  0x00007ffff5d67150 in RMW_Connext_Condition::update_state<RMW_Connext_SubscriberStatusCondition::set_data_available(bool)::{lambda()#1}>(RMW_Connext_SubscriberStatusCondition::set_data_available(bool)::{lambda()#1}&&, bool) (this=0x555555f53730, upda
te_condition=..., notify=true) at /ws/src/ros2/rmw_connextdds/rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp:110                                                                                                                            
#8  0x00007ffff5d65d63 in RMW_Connext_SubscriberStatusCondition::set_data_available (this=0x555555f53730, available=true) at /ws/src/ros2/rmw_connextdds/rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp:709                                 
#9  0x00007ffff5d5bf8d in RMW_Connext_Subscriber::loan_messages (this=0x555555f536c0, update_condition=true) at /ws/src/ros2/rmw_connextdds/rmw_connextdds_common/src/common/rmw_impl.cpp:1621                                                                
#10 0x00007ffff5d66699 in RMW_Connext_Subscriber::loan_messages_if_needed (this=0x555555f536c0, update_condition=true) at /ws/src/ros2/rmw_connextdds/rmw_connextdds_common/include/rmw_connextdds/rmw_impl.hpp:397                                           
#11 0x00007ffff5d5c161 in RMW_Connext_Subscriber::take_next (this=0x555555f536c0, ros_messages=0x7ffff1a468d0, message_infos=0x7ffff1a46950, max_samples=1, taken=0x7ffff1a468c8, serialized=false, request_writer_handle=0x0) at /ws/src/ros2/rmw_connextdds/
rmw_connextdds_common/src/common/rmw_impl.cpp:1692                                                                                                                                                                                                            
#12 0x00007ffff5d5bd0d in RMW_Connext_Subscriber::take_message (this=0x555555f536c0, ros_message=0x7ffff1a46920, message_info=0x7ffff1a46950, taken=0x7ffff1a4691b, request_writer_handle=0x0) at /ws/src/ros2/rmw_connextdds/rmw_connextdds_common/src/common
/rmw_impl.cpp:1522                                                                                                                                                                                                                                            
#13 0x00007ffff5d62c0c in RMW_Connext_Service::take_request (this=0x5555556424e0, request_header=0x7ffff1a48330, ros_request=0x7ffff1a48550, taken=0x7ffff1a46a5f) at /ws/src/ros2/rmw_connextdds/rmw_connextdds_common/src/common/rmw_impl.cpp:2943          
#14 0x00007ffff5d7c6fa in rmw_api_connextdds_take_request (service=0x555555f419f0, request_header=0x7ffff1a48330, ros_request=0x7ffff1a48550, taken=0x7ffff1a46a5f) at /ws/src/ros2/rmw_connextdds/rmw_connextdds_common/src/common/rmw_service.cpp:69        
#15 0x00007ffff7eec577 in rmw_take_request (service=0x555555f419f0, request_header=0x7ffff1a48330, ros_request=0x7ffff1a48550, taken=0x7ffff1a46a5f) at /ws/src/ros2/rmw_connextdds/rmw_connextdds/src/rmw_api_impl_ndds.cpp:544                              
#16 0x00007ffff7f47d52 in rcl_take_request_with_info (service=0x5555555e7168, request_header=0x7ffff1a48330, ros_request=0x7ffff1a48550) at /ws/src/ros2/rcl/rcl/src/rcl/service.c:344                                                                        
#17 0x00007ffff7f4803d in rcl_take_request (service=0x5555555e7168, request_header=0x7ffff1a48630, ros_request=0x7ffff1a48550) at /ws/src/ros2/rcl/rcl/src/rcl/service.c:381                                                                                  
#18 0x00007ffff7f3f94d in rcl_node_get_type_description_service_callback (user_data=0x55555566abd0, number_of_events=1) at /ws/src/ros2/rcl/rcl/src/rcl/node.c:601                                                                                            
#19 0x00007ffff5d6be46 in operator() (__closure=0x7ffff1a4cf70) at /ws/src/ros2/rmw_connextdds/rmw_connextdds_common/src/common/rmw_impl_waitset_std.cpp:880                                                                                                  
--Type <RET> for more, q to quit, c to continue without paging--                                                                                                                                                                                              
#20 0x00007ffff5d6de0f in RMW_Connext_Condition::perform_action_and_update_state<RMW_Connext_SubscriberStatusCondition::notify_new_data()::<lambda()>, RMW_Connext_SubscriberStatusCondition::notify_new_data()::<lambda()> >(struct {...} &&, struct {...} &&
) (this=0x555555f53730, action=..., update_condition=...) at /ws/src/ros2/rmw_connextdds/rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp:136                                                                                                 
#21 0x00007ffff5d6bef1 in RMW_Connext_SubscriberStatusCondition::notify_new_data (this=0x555555f53730) at /ws/src/ros2/rmw_connextdds/rmw_connextdds_common/src/common/rmw_impl_waitset_std.cpp:868                                                           
#22 0x00007ffff5d679f8 in RMW_Connext_DataReaderListener_on_data_available (listener_data=0x555555f53730, reader=0x55555591d6e0) at /ws/src/ros2/rmw_connextdds/rmw_connextdds_common/src/common/rmw_impl_waitset_std.cpp:138                                 
#23 0x00007ffff54d36c3 in DDS_DataReaderListener_forward_onDataAvailable (listener=0x7ffff1a4d040, presReader=0x55555591d4b8, worker=0x5555559e1dc0) at DataReaderListener.c:363                                                                              
#24 0x00007ffff54e845e in DDS_DataReader_impl_forward_onDataAvailable (presListener=0x7ffff5b6f900 <DDS_DataReader_PRESENTATION_LISTENER>, reader=0x55555591d4b8, worker=0x5555559e1dc0) at DataReader.c:6870                                                 
#25 0x00007ffff48eeb84 in PRESPsService_readerNotifyOfReaderQueueChanges (service=0x5555557a4950, committalInfoList=0x555555f691d0, workerStat=0x0, timestamp=0x7ffff1a4d6c8, receptionTimestamp=0x7ffff1a4d6c8, notifySampleRejected=1, topicQuery=0, worker=
0x5555559e1dc0) at PsServiceImpl.c:2167                                                                                                                                                                                                                       
#26 0x00007ffff48f19ab in PRESPsService_readerSampleListenerOnNewData (listener=0x5555557a5048, firstRelevantSn=0x555555a4aee0, nextRelevantRangeStartSn=0x7ffff1a4d6a8, data=0x7ffff1a4da30, reservedCount=1, timestamp=0x7ffff1a4d6c8, storage=0x555555a4ae6
0, worker=0x5555559e1dc0) at PsServiceImpl.c:2893                                                                                                                                                                                                             
#27 0x00007ffff4accc50 in COMMENDSrReaderService_onSubmessage (listener=0x555555a43990, context=0x7fffd4000c30, timestamp=0x7ffff1a4d6c8, storage=0x55555582cc30, worker=0x5555559e1dc0) at SrReaderService.c:2601                                            
#28 0x00007ffff4b44337 in MIGInterpreter_parse (me=0x55555582bf30, context=0x7fffd4000c30, msg=0x7ffff1a4dd70, worker=0x5555559e1dc0) at Interpreter.c:687                                                                                                    
#29 0x00007ffff4a9607f in COMMENDActiveFacadeReceiver_loop (param=0x5555559d1d60) at ActiveFacade.c:647                                                                                                                                                       
#30 0x00007ffff4d419f3 in RTIOsapiThreadChild_onSpawned (param=0x5555559e3e80) at Thread.c:1726                                                                                                                                                               
#31 0x00007ffff7242b43 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:442                                                                                                                                                                   
#32 0x00007ffff72d4a00 in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:81                
```

emersonknapp: OK - that above stack trace actually does have the requisite information. This is a reentrant condition on the same mutex.

`notify_new_data` at frame 21 calls `RMW_ConnextCondition::perform_action_and_update_state` which locks its `mutex_internal`. That calls the registered callback function, which tries to `take_request`, which makes it down to the same Condition which then tries to lock, causing a reentrant deadlock.

What I _don't_ understand, then, is how services work in Connext at all

emersonknapp: If I change https://github.com/ros2/rmw_connextdds/blob/rolling/rmw_connextdds_common/include/rmw_connextdds/rmw_waitset_std.hpp#L141 to `std::recursive_mutex mutex_internal` then the test starts passing. What I don't know is if this is the correct thing to do.

I think I've identified that there are _no_ tests calling any RMW API within a `rcl_service_set_on_new_request_callback` callback. The only test that calls this API is doing a simple variable increment to check that it was called. The only real service tests are at the `rclcpp` and `rclpy` layers where they are using a totally separate callback framework with their client-level waitsets. `rclcpp` exposes this API, but nothing is testing it, the callback passed to the Service constructor goes through this alternate path.

I _believe_ that it should be totally acceptable to call `rmw_take` during the service callback, as it is not a blocking operation, it just returns an error code if data is not available. So, this looks like a bug in `rmw_connextdds` - unfortunately a variable type issue in a public header :facepalm: so it would cause an ABI break to fix it.

clalancette: > I _believe_ that it should be totally acceptable to call `rmw_take` during the service callback, as it is not a blocking operation, it just returns an error code if data is not available. So, this looks like a bug in `rmw_connextdds` - unfortunately a variable type issue in a public header facepalm so it would cause an ABI break to fix it.

If we think that is the right fix, then let's go ahead and open a pull request for `rmw_connextdds` to make that fix.  We can worry about backporting it to Iron afterwards.  It would also be nice to have a test for this situation, though I think that test would probably want to end up here in `rcl` (or maybe in `rmw`) so we can test it for all RMW implementations.

rcl/include/rcl/node.h
```diff
@@ -549,6 +549,57 @@ RCL_PUBLIC
...
+ * \param[in] node the handle to the node whose type cache should be initialized
+ * \return #RCL_RET_OK if service was deinitialized successfully, or
+ * \return #RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
+ * \return #RCL_RET_SERVICE_INVALID if the service is invalid, or
+ * \return #RCL_RET_NODE_INVALID if the node is invalid, or
+ * \return #RCL_RET_ERROR if an unspecified error occurs.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t rcl_node_type_description_service_fini(rcl_node_t * node);
```
emersonknapp, line 607: Will need two more APIs here:
* `get_get_type_description_service` to retrieve the service pointer
* `get_get_type_descripiton_service_calback` to retrieve the callback that should be attached to the service

rcl/src/rcl/node.c
```diff
@@ -542,6 +570,167 @@ rcl_get_disable_loaned_message(bool * disable_loaned_message)
   *disable_loaned_message = (strcmp(env_val, "1") == 0);
   return RCL_RET_OK;
 }
+
+static void rcl_node_get_type_description_service_callback(
```
emersonknapp: not static - we'll expose this

rcl/src/rcl/node.c
```diff
@@ -310,6 +316,18 @@ rcl_node_init(
   goto cleanup;
 fail:
   if (node->impl) {
+    if (node->impl->options.enable_type_description_service) {
```
emersonknapp: rm

rcl/src/rcl/node.c
```diff
@@ -387,6 +405,16 @@ rcl_node_fini(rcl_node_t * node)
       result = RCL_RET_ERROR;
     }
   }
+  if (node->impl->options.enable_type_description_service) {
```
emersonknapp: rm

rcl/src/rcl/node.c
```diff
@@ -284,6 +283,19 @@ rcl_node_init(
     // error message already set
     goto fail;
   }
+  // Initialize ~/get_type_description service and type cache hash map
+  if (node->impl->options.enable_type_description_service) {
```
emersonknapp: rm

rcl/src/rcl/node.c
```diff
@@ -542,6 +570,167 @@ rcl_get_disable_loaned_message(bool * disable_loaned_message)
   *disable_loaned_message = (strcmp(env_val, "1") == 0);
   return RCL_RET_OK;
 }
+
+static void rcl_node_get_type_description_service_callback(
+  const void * user_data,
+  size_t number_of_events)
+{
+  (void)number_of_events;
```
emersonknapp: if there is more than one event, we need to handle them, this should loop

rcl/src/rcl/node.c
```diff
@@ -542,6 +570,167 @@ rcl_get_disable_loaned_message(bool * disable_loaned_message)
...
+  ret = rcl_service_init(
+    &node->impl->get_type_description_service, node,
+    type_support, service_name, &service_ops);
+  allocator.deallocate(service_name, allocator.state);
+  if (RCL_RET_OK != ret) {
+    return ret;
+  }
+
+  // Set service callback
+  ret = rcl_service_set_on_new_request_callback(
```
emersonknapp: turns out we shouldn't do this - this API was added for use by the events executor and shouldn't do any work at all, just put things into queue. rm this line

rcl/src/rcl/node.c
```diff
@@ -542,6 +570,167 @@ rcl_get_disable_loaned_message(bool * disable_loaned_message)
...
+  }
+
+cleanup:
+  type_description_interfaces__srv__GetTypeDescription_Request__fini(&request);
+  type_description_interfaces__srv__GetTypeDescription_Response__fini(
+    &response);
+}
+
+rcl_ret_t rcl_node_type_description_service_init(rcl_node_t * node)
+{
```
emersonknapp, line 625: maybe just init the type cache in here?

rcl/src/rcl/node.c
```diff
@@ -542,6 +570,167 @@ rcl_get_disable_loaned_message(bool * disable_loaned_message)
...
+      ROS_PACKAGE_NAME,
+      "Failed to deinitialize ~/get_type_description service.");
+  }
+
+  return ret;
+}
+
+rcl_ret_t rcl_node_type_description_service_fini(rcl_node_t * node)
+{
+  RCL_CHECK_ARGUMENT_FOR_NULL(node, RCL_RET_INVALID_ARGUMENT);
```
emersonknapp, line 665: if init type cache in init, fini it here

emersonknapp (commented):
After discussing this with @wjwwood today, we determined that the callback approach we are taking here is not a good idea. `rcl_service_set_on_new_request_callback` is not intended to do any serious work, it exists for the events executor and should only put things onto a work queue - not fetch the type description and send a response.

That's the reason there isn't any other code using this API - it's not really intended for this use. In this case the thread that is calling this is an RMW-created thread rather than a user-created one, so there is no possibility for user control over the run environment, and the RMW implementation may very well block on getting new data while the callback is being handled.

The solution we landed on as the "right way" is to provide `get_type_description_service_init` and `_fini` functions, and do not call them from here in `rcl`, instead relying on `rclcpp` and `rclpy` nodes to call that initialization, and give them access to the service and callback function, to register them with those client libraries' wait sets and executors. This'll also allow us to use parameters to turn the service on and off.

I've added comments here as notes to self of what needs to change for that.

@clalancette let me know if you disagree with that assessment, but I'm pretty convinced that's the way to go.

I will open an issue against `rmw_connextdds` and provide a failing test, as I think this use case _should_ have worked, even if it's not the right thing to do in this case.

rcl/src/rcl/node.c
emersonknapp: I've unfortunately got to turn this ship _back_ around.

The new design allows for the service to be enabled via Parameters, and initialization is therefore called from the instantiating client. However, in order for the type cache to have _all_ types the node provides - the type cache has to have been initialized before any interface is created (including the parameter topics). This has to be done in `rcl_node_init` because the rosout publisher is created in this function.

This comment originally banked on using NodeOptions for enabling, but if there is parameter-based configuration then there is no way to avoid always initializing the hashmap here.

I'll argue the overhead is negligible - no hashing calculation is done since the hashes are precomputed, and the data stored is `4 * sizeof(size_t)` per interface, which just doesn't add up to much

emersonknapp:
@clalancette this is ready for next pass review now, its new state is successfully utilized by https://github.com/ros2/rclcpp/pull/2224. See https://github.com/achim-k/rcl/pull/5 for the exact changes made, or just the squashed commit https://github.com/ros2/rcl/pull/1052/commits/48a611a2fa1ca32d387082133e99e687866acfee

emersonknapp:
Pulls: ros2/rcl#1052, ros2/rclcpp#2224
Gist: https://gist.githubusercontent.com/emersonknapp/c4a4fc1d37b707ec959808ee6dd2d189/raw/f6081a66740b791e891c43d39ace0c33004c8312/ros2.repos
BUILD args:  --packages-above-and-dependencies rcl rcl_action rclcpp rclcpp_lifecycle
TEST args:  --packages-above rcl rcl_action rclcpp rclcpp_lifecycle
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12306
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19047/)
* Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/13568/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19790/)

rcl/include/rcl/node_type_cache.h
```diff
@@ -0,0 +1,197 @@
...
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] node the handle to the node whose type cache to check
+ * \return true if the node's type cache was successfully initialized, or
+ * \return false for any other case.
+ */
+RCL_PUBLIC
+RCL_WARN_UNUSED
+bool rcl_node_type_cache_is_valid(const rcl_node_t * node);
```
clalancette: Given that we are unconditionally calling `rcl_node_type_cache_init` during `rcl_node_init` now, do we need this function?

rcl/src/rcl/node.c
clalancette: I think we may have lost a bug fix here, because this seems to be an issue again.  For instance, if `rcl_logging_rosout_init_publisher_for_node` fails below, then we'll leak any memory associated with the node type cache.

rcl/src/rcl/node.c
```diff
@@ -387,6 +388,20 @@ rcl_node_fini(rcl_node_t * node)
       result = RCL_RET_ERROR;
     }
   }
+  rcl_ret = rcl_node_type_description_service_fini(node);
+  if (rcl_ret == RCL_RET_NOT_INIT) {
+    rcl_reset_error();
+  } else if (rcl_ret != RCL_RET_OK) {
+    RCL_SET_ERROR_MSG("Unable to fini ~/get_type_description service for node.");
+    result = RCL_RET_ERROR;
+  }
```
clalancette: It seems somewhat asymmetric to `fini` the service here when we didn't `init` the service in `rcl_node_init`.  Do we actually need to call this here, or should we leave it to the client libraries to do this?

rcl/src/rcl/publisher.c
```diff
@@ -137,6 +140,19 @@ rcl_publisher_init(
     (const void *)publisher->impl->rmw_handle,
     remapped_topic_name,
     options->qos.depth);
+  // Register type.
+  if (rcl_node_type_cache_is_valid(node)) {
```
clalancette: If I recall our conversation from earlier this week, I think we determined that we are just going to unconditionally register types.  If that recollection is correct, then we can just remove this conditional.

rcl/src/rcl/publisher.c
```diff
@@ -187,6 +203,13 @@ rcl_publisher_fini(rcl_publisher_t * publisher, rcl_node_t * node)
       RCL_SET_ERROR_MSG(rmw_get_error_string().str);
       result = RCL_RET_ERROR;
     }
+    // Unregister type
+    if (rcl_node_type_cache_is_valid(node)) {
```
clalancette: Ditto.

rcl/src/rcl/service.c
```diff
@@ -186,13 +188,28 @@ rcl_service_init(
...
+  service->impl->type_hash = *type_support->get_type_hash_func(type_support);
   RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME, "Service initialized");
   TRACETOOLS_TRACEPOINT(
     rcl_service_init,
     (const void *)service,
     (const void *)node,
     (const void *)service->impl->rmw_handle,
     service->impl->remapped_service_name);
+  // Register type.
+  if (rcl_node_type_cache_is_valid(node)) {
```
clalancette: Ditto.

rcl/src/rcl/service.c
```diff
@@ -248,6 +265,14 @@ rcl_service_fini(rcl_service_t * service, rcl_node_t * node)
       result = RCL_RET_ERROR;
     }
 
+    // Unregister type
+    if (rcl_node_type_cache_is_valid(node)) {
```
clalancette: Ditto.

rcl/src/rcl/subscription.c
```diff
@@ -131,6 +134,19 @@ rcl_subscription_init(
     (const void *)subscription->impl->rmw_handle,
     remapped_topic_name,
     options->qos.depth);
+  // Register type.
+  if (rcl_node_type_cache_is_valid(node)) {
```
clalancette: Ditto.

rcl/src/rcl/subscription.c
```diff
@@ -192,6 +208,17 @@ rcl_subscription_fini(rcl_subscription_t * subscription, rcl_node_t * node)
       result = RCL_RET_ERROR;
     }
 
+    // Unregister type
+    if (rcl_node_type_cache_is_valid(node)) {
```
clalancette: Ditto.

rcl/test/rcl/test_get_type_description_service.cpp
```diff
@@ -0,0 +1,357 @@
...
+# define CLASSNAME_(NAME, SUFFIX) NAME ## __ ## SUFFIX
+# define CLASSNAME(NAME, SUFFIX) CLASSNAME_(NAME, SUFFIX)
+#else
+# define CLASSNAME(NAME, SUFFIX) NAME
+#endif
+
+constexpr char GET_TYPE_DESCRIPTION_SRV_TYPE_NAME[] =
+  "type_description_interfaces/srv/GetTypeDescription";
+
+static bool string_in_array(rcutils_string_array_t * array, const char * pattern)
```
clalancette, line 42: `#include <rcutils/types/string_array.h>` at the top of the file for this.

rcl/test/rcl/test_get_type_description_service.cpp
```diff
@@ -0,0 +1,357 @@
...
+{
+  for (size_t i = 0; i < array->size; ++i) {
+    if (strcmp(array->data[i], pattern) == 0) {
```
clalancette, line 45: `#include <cstring>` at the top of the file for this.

rcl/test/rcl/test_get_type_description_service.cpp
clalancette: Given that we've totally revamped how this works, I'm going to resolve this conversation.  Feel free to reopen if you think that was the wrong thing to do.

rcl_action/src/rcl_action/action_server.c
```diff
@@ -192,6 +195,20 @@ rcl_action_server_init(
     ret = RCL_RET_BAD_ALLOC;
     goto fail;
   }
+
+  // Register type.
+  if (rcl_node_type_cache_is_valid(node)) {
```
clalancette: Same comment here that we might not need to have this conditional at all.

rcl_action/src/rcl_action/action_server.c
```diff
@@ -249,6 +266,14 @@ rcl_action_server_fini(rcl_action_server_t * action_server, rcl_node_t * node)
     }
     allocator.deallocate(action_server->impl->goal_handles, allocator.state);
     action_server->impl->goal_handles = NULL;
+    // Unregister type
+    if (rcl_node_type_cache_is_valid(node)) {
```
clalancette: Ditto.

clalancette (changes requested):
I've gone through another pass here.  Things generally look good to me, though I've left a few things to take a look at.

rcl/include/rcl/node_type_cache.h
emersonknapp: Nope - removing

rcl/src/rcl/node.c
emersonknapp: I suppose that `rmw_destroy_node` will destroy the service anyways - and invalidating `node->impl` will get rid of the pointer to this service, so there's no harm in not calling `fini` here

EDIT: `rmw_destroy_node` requires that everything is destroyed first. Makes sense to require client libraries to `fini` the service if they `init`ed it, just like with everything else on a node.

emersonknapp:
Pulls: ros2/rcl#1052, ros2/rclcpp#2224
Gist: https://gist.githubusercontent.com/emersonknapp/e64c966d59d7e2081d4d296fa756a1e3/raw/f6081a66740b791e891c43d39ace0c33004c8312/ros2.repos
BUILD args:  --packages-above-and-dependencies rcl rcl_action rclcpp rclcpp_lifecycle
TEST args:  --packages-above rcl rcl_action rclcpp rclcpp_lifecycle
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12316
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19059/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13584/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19811/)

emersonknapp:
Pulls: ros2/rcl#1052, ros2/rclcpp#2224, ros2/system_tests#520
Gist: https://gist.githubusercontent.com/emersonknapp/b7f98e834558aa5d256554506c5ab18b/raw/e787e5a7cbeadb0ba5865e439578867d581b682d/ros2.repos
BUILD args:  --packages-above-and-dependencies rcl rcl_action rclcpp rclcpp_lifecycle test_rclcpp
TEST args:  --packages-above rcl rcl_action rclcpp rclcpp_lifecycle test_rclcpp
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12317
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19060/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13586/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19812/)

emersonknapp:
Pulls: ros2/rcl#1052, ros2/rclcpp#2224, ros2/system_tests#520
Gist: https://gist.githubusercontent.com/emersonknapp/b250eee074072eeb11c1a6d544866179/raw/e787e5a7cbeadb0ba5865e439578867d581b682d/ros2.repos
BUILD args:  --packages-above-and-dependencies rcl rcl_action rclcpp rclcpp_lifecycle test_rclcpp
TEST args:  --packages-above rcl rcl_action rclcpp rclcpp_lifecycle test_rclcpp
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12318
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19061/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13587/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19813/)

emersonknapp:
Pulls: ros2/rcl#1052, ros2/rclcpp#2224, ros2/system_tests#520
Gist: https://gist.githubusercontent.com/emersonknapp/b250eee074072eeb11c1a6d544866179/raw/e787e5a7cbeadb0ba5865e439578867d581b682d/ros2.repos
BUILD args:  --packages-above-and-dependencies rcl rcl_action rclcpp rclcpp_lifecycle test_rclcpp
TEST args:  --packages-above rcl rcl_action rclcpp rclcpp_lifecycle test_rclcpp
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12318
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19061/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13587/)
* [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/13596/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19813/)

rcl/src/rcl/node.c
```diff
@@ -542,6 +554,147 @@ rcl_get_disable_loaned_message(bool * disable_loaned_message)
...
   return RCL_RET_OK;
 }
+
+void rcl_node_type_description_service_handle_request(
+  rcl_node_t * node,
+  const rmw_request_id_t * request_header,
+  const type_description_interfaces__srv__GetTypeDescription_Request * request,
+  type_description_interfaces__srv__GetTypeDescription_Response * response)
+{
+  (void)request_header;
```
clalancette: Nit; I don't think we need this, since we are using the type in the `RCL_CHECK_FOR_NULL_WITH_MSG` call below.

clalancette (approved):
I've left one extremely trivial nit, so I'll approve anyway.

emersonknapp:
Pulls: ros2/rcl#1052, ros2/rclcpp#2224, ros2/system_tests#520
Gist: https://gist.githubusercontent.com/emersonknapp/61666ed4bfef7aa7b2f349cd87122a42/raw/e787e5a7cbeadb0ba5865e439578867d581b682d/ros2.repos
BUILD args:  --packages-above-and-dependencies rcl rcl_action rclcpp rclcpp_lifecycle test_rclcpp
TEST args:  --packages-above rcl rcl_action rclcpp rclcpp_lifecycle test_rclcpp
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12324
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19071/)
  * [Build Status](https://ci.ros2.org/job/ci_linux/19074/)
  * [Build Status](https://ci.ros2.org/job/ci_linux/19076/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13600/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19819/)

emersonknapp:
@mergifyio backport iron

emersonknapp:
Rolling post-merge, no repos file
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/19086/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13612/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19831/)

PR #1055, opened 2023-04-06
Use get_type_hash_func for typesupports

emersonknapp:
contributes https://github.com/ros2/rcl/pull/1053 to the correct branch `rolling`, instead of `master`

Part of ros2/ros2#1159
Depends on ros2/rosidl#727 (and is bundled with)


christophebedard (approved):
For what it's worth, this fixes `rcl` source builds for me.

mjcarroll (approved):

Issue #1046, opened 2023-03-16
rcl_node_init uses wrong return code


Deric-W:
## Bug report

Operation System: Windows 10
Client library: N/A
Version: Humble

#### Steps to reproduce issue

1. Call `rcl_node_init` with `NULL` as context
2. Check return code

#### Expected behavior

The return code should be `RCL_RET_NOT_INIT` since the context is invalid.

#### Actual behavior

The return code is `RCL_RET_INVALID_ARGUMENT`.

#### Additional information

The [documentation](http://docs.ros.org/en/humble/p/rcl/generated/function_node_8h_1a404531a35e68e71f4900ff48544682b5.html#function-rcl-node-init) suggests that `RCL_RET_NOT_INIT` should be returned since [`rcl_context_is_valid(NULL)`](http://docs.ros.org/en/humble/p/rcl/generated/function_context_8h_1ab6249347bd221d620d7e9432ff2cadd0.html#function-rcl-context-is-valid) returns `false`.
If this is not desirable then the documentation should be updated to mention this.

fujitatomoya:
I think current behavior is correct.

> [RCL_RET_NOT_INIT](http://docs.ros.org/en/humble/p/rcl/generated/define_types_8h_1a61039a0ea73e2c8685572446b237a827.html#types_8h_1a61039a0ea73e2c8685572446b237a827) if the given context is invalid, or

I believe that there is clear difference between `context is invalid` and `argument is invalid`.

- If the context is null via argument, that means argument is invalid, not context. that should return `RCL_RET_INVALID_ARGUMENT`.
- If the context is invalid, arguments are okay, but content of context is not right. that should return `RCL_RET_NOT_INIT`.

Deric-W:
Your reasoning makes sense, I thought `RCL_RET_NOT_INIT` would be returned since the `rcl_{publisher,subscription,client,service}_init` functions behave this way when returning `RCL_RET_NODE_INVALID`.
Would it make sense to mention this in the documentation?

Deric-W:
Implemented in #1066

PR #1066, opened 2023-04-28
clarify `rcl_node_init` return code

Deric-W:
This is to prevent confusion that I encountered in #1046.

### DIFF
--- a/rcl/include/rcl/node.h
+++ b/rcl/include/rcl/node.h
@@ -137,7 +137,7 @@ rcl_get_zero_initialized_node(void);
  * \return #RCL_RET_OK if the node was initialized successfully, or
  * \return #RCL_RET_ALREADY_INIT if the node has already be initialized, or
  * \return #RCL_RET_NOT_INIT if the given context is invalid, or
- * \return #RCL_RET_INVALID_ARGUMENT if any arguments are invalid, or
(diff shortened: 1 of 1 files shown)


rcl/include/rcl/node.h
```diff
@@ -137,7 +137,7 @@ rcl_get_zero_initialized_node(void);
  * \return #RCL_RET_OK if the node was initialized successfully, or
  * \return #RCL_RET_ALREADY_INIT if the node has already be initialized, or
  * \return #RCL_RET_NOT_INIT if the given context is invalid, or
```
fujitatomoya: if i am not mistaken, you mean that this line is the confusion, right? context is one of the arguments, so that it should return `RCL_RET_INVALID_ARGUMENT`?

if that is the case, how about trying to address with following doc?

```suggestion
 * \return #RCL_RET_NOT_INIT if the given context is not initialized, or
```

could it be less confusion?

fujitatomoya (approved):

fujitatomoya:
@Deric-W can you check DCO error?

clalancette (approved):

clalancette:
Since this is an update to a comment, I'm going to merge it without CI; the Rpr job checked the linters for us.

PR #1075, opened 2023-06-06
feat: Thread configuration prototype

smorita-esol:
This is a prototype implementation of RCL for discussion about the thread configuration feature to receive and apply a set of scheduling parameters for the threads controlled by the ROS 2 executor.

Our basic idea is as below.
 1. Implement a new class rclcpp::thread and modify rclcpp to use it. This class has the same function set as the std::thread but also additional features to control its thread attributions.
 2. Modify the rcl layer to receive a set of scheduling parameters. The parameters are described in YAML format and passed via command line parameters, environment variables, or files.
 3. the rclcpp reads the parameters from rcl and applies them to each thread in the thread pool.

There have been some discussions about this pull request, as below. 

[ROS Discourse]
https://discourse.ros.org/t/adding-thread-attributes-configuration-in-ros-2-framework/30701 

[ROS 2 Real-Time Working Group]
https://github.com/ros-realtime/ros-realtime.github.io/issues/18


ros-discourse:

https://discourse.ros.org/t/adding-thread-attributes-configuration-in-ros-2-framework/30701/2

wjwwood:
See my comment here: https://github.com/ros2/rclcpp/pull/2205#issuecomment-1593832758

This seems useful, but we should probably adjust where the code lives and how we extend rcl/rclcpp to accommodate them.

smorita-esol:
I've shared the draft version of the REP related to this pull request, as provided below. 
I would greatly appreciate it if you could kindly share your feedback or point out any areas that need improvement.
https://discourse.ros.org/t/adding-thread-attributes-configuration-in-ros-2-framework/30701/5?u=smorita-esol

PR #1081, opened 2023-06-22
[WIP] Add a service test that fetches the service request in the set callback

emersonknapp:
Currently this is just a demonstration, it's not really a recommended workflow but it highlights a deadlock in `rmw_connext`.

### DIFF
--- a/rcl/test/rcl/test_service.cpp
+++ b/rcl/test/rcl/test_service.cpp
@@ -678,3 +678,80 @@ TEST_F(CLASSNAME(TestServiceFixture, RMW_IMPLEMENTATION), test_fail_send_respons
     rcl_reset_error();
   }
 }
+
+static void service_callback(const void * user_data, size_t number_of_events)
+{
+  rcl_ret_t ret;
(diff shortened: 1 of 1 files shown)


rcl/test/rcl/test_service.cpp
```diff
@@ -678,3 +678,80 @@ TEST_F(CLASSNAME(TestServiceFixture, RMW_IMPLEMENTATION), test_fail_send_respons
...
+  printf("Service callback %zu!\n", number_of_events);
+  rcl_service_t * service = (rcl_service_t *)user_data;
+
+  test_msgs__srv__BasicTypes_Request request;
+  ret = test_msgs__srv__BasicTypes_Request__init(&request);
+  test_msgs__srv__BasicTypes_Response response;
+  ret = test_msgs__srv__BasicTypes_Response__init(&response);
+
+  rmw_request_id_t header;
+  ret = rcl_take_request(service, &header, &request);
```
emersonknapp, line 694: Right here is where the deadlock hang occurs

PR #1084, opened 2023-08-01
Cleanup error reporting in the type hash code.

clalancette:
In particular, using the RCL_CHECK_* macros in both rcl_node_type_cache_register_type() and
rcl_convert_type_description_runtime_to_msg() meant that we would overwrite the error message of the latter in the former.  This would end up with an rcutils warning about losing error messages

Avoid that by just doing an open-coded check for NULL if rcl_node_type_cache_register_type(), passing along the error we got from the lower layers.

While we are in here, also very slightly revamp the client_init code to not have two separate calls to type_support->get_type_hash_func().

### DIFF
--- a/rcl/src/rcl/client.c
+++ b/rcl/src/rcl/client.c
@@ -178,9 +178,15 @@ rcl_client_init(
   client->impl->options = *options;
   atomic_init(&client->impl->sequence_number, 0);
 
+  const rosidl_type_hash_t * hash = type_support->get_type_hash_func(type_support);
+  if (hash == NULL) {
+    RCL_SET_ERROR_MSG("Failed to get the type hash");
+    ret = RCL_RET_INVALID_ARGUMENT;
+    goto destroy_client;
+  }
+
   if (RCL_RET_OK != rcl_node_type_cache_register_type(
-      node, type_support->get_type_hash_func(type_support),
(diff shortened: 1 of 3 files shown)


clalancette:
I should also say that this is the first in a large-ish series of patches that will reduce `rcl` test time substantially.  Locally, with this series in place, my test time went from ~8 minutes to ~5 minutes.

fujitatomoya (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19275/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13816/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20004/)

emersonknapp (approved):
:+1: I'm surprised that the error message handling incurs such overhead. Thanks for fixing it!

clalancette:
> +1 I'm surprised that the error message handling incurs such overhead. Thanks for fixing it!

Ah, I should have been more clear.  The error message handling code does incur a *bit* of overhead, but that wasn't the main thing that allowed me to remove time.  This (and a whole host of other patches to come) were mostly just things I noticed while shrinking the time down.

PR #1085, opened 2023-08-03
Reset errors in tests to reduce warnings

clalancette:
When we have a test that fails, we always want to do an `rcl_reset_error` after it to reset the internal error that is set.  That prepares it for the next test.  We were missing this in a few places in our tests, so add them as needed.


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19296/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13829/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20016/)

audrow (approved):
Looks good to me with green CI.

fujitatomoya (approved):

PR #1086, opened 2023-08-06
Simplify the various 'is_valid' calls

clalancette:
Currently if you call one of the rcl "is_valid" functions on an uninitialized pointer of the correct type, it will return false but it will also set an error.  But this doesn't seem correct; just asking the question of whether it is valid should never return an error.  This PR fixes that.

Note that if you pass a NULL pointer into the "is_valid" function, you'll still get an error set.  That seems reasonable since you are basically asking a nonsense question.

### DIFF
--- a/rcl/src/rcl/client.c
+++ b/rcl/src/rcl/client.c
@@ -421,11 +421,8 @@ bool
 rcl_client_is_valid(const rcl_client_t * client)
 {
   RCL_CHECK_FOR_NULL_WITH_MSG(client, "client pointer is invalid", return false);
-  RCL_CHECK_FOR_NULL_WITH_MSG(
-    client->impl, "client's rmw implementation is invalid", return false);
-  RCL_CHECK_FOR_NULL_WITH_MSG(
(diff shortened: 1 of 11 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19301/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13833/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20022/)

fujitatomoya (approved):
lgtm

> just asking the question of whether it is valid should never return an error.

makes sense to me.

clalancette:
The more I look at this, the less I like it.

If you are looking at these `is_valid` calls as part of the external `rcl` API, then I think this change makes sense.  However, the vast majority of calls to `is_valid` comes from within `rcl` itself.  And in that case, it becomes difficult to determine the split of responsibilities between the `is_valid` function itself, and the caller, because sometimes the caller needs to set the error, and sometimes the caller needs to reset the error and set something else.

I'm going to turn this into a draft for now until I come up with something better.

Issue #984, opened 2022-05-15
Add identifying information to the log message when duplicate nodes are created


jlack1987:
When creating duplicate nodes(in my experience from rclcpp) a message is printed to the console,
```
[rcl.logging_rosout]: Publisher already registered for provided node name. If this is due to multiple nodes with the same name then all logs for that logger name will go out over the existing publisher. As soon as any node with that name is destructed it will unregister the publisher, preventing any further logs for that name from being published on the rosout topic.
```
It can be very difficult from this message to figure out where the offending line(s) of code may be in our very large stack of code. Is there any way to get some sort of identifying information into this? I took a brief glance at the code and even putting the `logger_name` in the message would be really helpful. If the node name is available that would be optimal, but anything to help the developer figure out where it's coming from would be a step in the right direction. I'm willing to put in a fix for this if a dev can recommend something that would be acceptable.

It looks to me like the chunk of code where this originates from is,

```
if (rcutils_hash_map_key_exists(&__logger_map, &logger_name)) {
    // @TODO(nburek) Update behavior to either enforce unique names or work with non-unique
    // names based on the outcome here: https://github.com/ros2/design/issues/187
    RCUTILS_LOG_WARN_NAMED(
      "rcl.logging_rosout",
      "Publisher already registered for provided node name. If this is due to multiple nodes "
      "with the same name then all logs for that logger name will go out over the existing "
      "publisher. As soon as any node with that name is destructed it will unregister the "
      "publisher, preventing any further logs for that name from being published on the rosout "
      "topic.");
    return RCL_RET_OK;
}
```

fujitatomoya:
> putting the logger_name in the message would be really helpful.

:+1: that sounds reasonable to me.

CursedRock17:
This issue can be closed if the PR was enough.

fujitatomoya:
@CursedRock17 good eye 👁️ closing.

PR #1088, opened 2023-08-11
Adding duplicate node information

CursedRock17:
This resolves issue #984 just bringing more information to an error message when there are multiple nodes that are being published. Originally the issue wanted to just add the `logger name` to make it easier to read, but I also got the `node name` because I figured to user should want to know both, making an easier to resolve the duplicate node.

That should solve the issue, but there is still a TODO, that's been there since [2019](https://github.com/ros2/rcl/commit/af49d6b3bf13925ac61f3c950391cc63e15211e8), which talks about the design docs and the instability of duplicate nodes. Do we want to add anything now, like mentioned in [the doc](https://github.com/ros2/design/issues/187):
> For a default (non-production) behavior I think having the newer node raise an exception makes sense

Or is it better to leave it be?

### DIFF
--- a/rcl/src/rcl/logging_rosout.c
+++ b/rcl/src/rcl/logging_rosout.c
@@ -12,9 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "rcl/logging_rosout.h"
+
 #include "rcl/allocator.h"
 #include "rcl/error_handling.h"
-#include "rcl/logging_rosout.h"
 #include "rcl/node.h"
 #include "rcl/publisher.h"
 #include "rcl/time.h"
@@ -250,13 +251,20 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(rcl_node_t * node)
   if (rcutils_hash_map_key_exists(&__logger_map, &logger_name)) {
     // @TODO(nburek) Update behavior to either enforce unique names or work with non-unique
     // names based on the outcome here: https://github.com/ros2/design/issues/187
+    const char * node_name = rcl_node_get_name(node);
+    if (NULL == node_name) {
+      node_name = "unknown node";
+    }
+
     RCUTILS_LOG_WARN_NAMED(
       "rcl.logging_rosout",
-      "Publisher already registered for provided node name. If this is due to multiple nodes "
-      "with the same name then all logs for that logger name will go out over the existing "
-      "publisher. As soon as any node with that name is destructed it will unregister the "
-      "publisher, preventing any further logs for that name from being published on the rosout "
-      "topic.");
+      "Publisher already registered for node name: '%s'. If this is due to multiple nodes "
+      "with the same name then all logs for the logger named '%s' will go out over "
+      "the existing publisher. As soon as any node with that name is destructed "
+      "it will unregister the publisher, preventing any further logs for that name from "
+      "being published on the rosout topic.",
+      node_name,
+      logger_name);
     return RCL_RET_OK;
   }
 


rcl/src/rcl/logging_rosout.c
```diff
@@ -243,20 +246,31 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(rcl_node_t * node)
   // Verify input and make sure it's not already initialized
   RCL_CHECK_ARGUMENT_FOR_NULL(node, RCL_RET_NODE_INVALID);
   logger_name = rcl_node_get_logger_name(node);
+  const char * node_name = rcl_node_get_name(node);
```
clalancette: I think this should be moved into the `if (rcutils_hash_map_key_exists)` section, since we don't need it elsewhere.  We should also double-check whether the `node_name` is NULL before using it, as that is an (unlikely) situation that can happen.

rcl/src/rcl/logging_rosout.c
```diff
@@ -243,20 +246,31 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(rcl_node_t * node)
...
+      "then all logs for: ";
+    const char * third_part = " will go out over the existing publisher. As soon as any "
+      "Node with that name is destructed it will unregister the publisher, preventing a "
+      "any further logs for that name from being published on the rosout topic.";
+
+    const size_t MsgSize = strlen(logger_name) + strlen(node_name) +
+      strlen(first_part) + strlen(second_part) + strlen(third_part) + 1;
+    char message[MsgSize];
+
+    snprintf(
```
clalancette: We generally prefer to use `rcutils_snprintf`, as that handles differences between the platforms.

rcl/src/rcl/logging_rosout.c
```diff
@@ -243,20 +246,31 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(rcl_node_t * node)
...
+    const char * first_part = "Publisher already registered for: ";
+    const char * second_part = ". If this is due to multiple nodes with the same name "
+      "then all logs for: ";
+    const char * third_part = " will go out over the existing publisher. As soon as any "
+      "Node with that name is destructed it will unregister the publisher, preventing a "
+      "any further logs for that name from being published on the rosout topic.";
+
+    const size_t MsgSize = strlen(logger_name) + strlen(node_name) +
+      strlen(first_part) + strlen(second_part) + strlen(third_part) + 1;
+    char message[MsgSize];
```
clalancette: This isn't very consistent with our style.  This should be something like:

```suggestion
    const size_t msg_size = strlen(logger_name) + strlen(node_name) +
      strlen(first_part) + strlen(second_part) + strlen(third_part) + 1;
    char message[msg_size];
```

(also, today I learned that `strlen` on a constant string is actually evaluated at compile time, not runtime.  Neat!)

clalancette (changes requested):
I've left some comments on things to fix inline.

To reply to:
> That should solve the issue, but there is still a TODO, that's been there since [2019](https://github.com/ros2/rcl/commit/af49d6b3bf13925ac61f3c950391cc63e15211e8), which talks about the design docs and the instability of duplicate nodes. Do we want to add anything now, like mentioned in https://github.com/ros2/design/issues/187:
>
>  >  For a default (non-production) behavior I think having the newer node raise an exception makes sense
>
> Or is it better to leave it be?

Certainly we should do it in a separate PR if we want to attempt to address it.

That said, it is likely a large project.  I don't think it will be a lot of code change, but there will have to be a lot of research done both in the code (composition, SROS2, etc), and into how the community is using node names.  And since a change like that will almost certainly break some downstream users, we'll have to do community outreach to make sure people are aware that the change is coming.

rcl/src/rcl/logging_rosout.c
CursedRock17: I figure it should work here since everything is known at compile time and the code doesn't throw any warning/error, but `cpplint` complains about using a variable array with that style for some reason. Test looks like this:
> logging_rosout.c:270:  Do not use variable-length arrays. Use an appropriately named ('k' followed by CamelCase) compile-time constant for the size

clalancette: Oh, but the `node_name` and `logger_name` is not known at compile-time, which is why it is complaining.

CursedRock17: In that case, should I just make `msg_size` const, by removing `strlen(node_name)` and `strlen(logger_name)`, then replacing that will a constant value like 100. It should be enough for both the logger and node name for the most part, while not being too large. This takes away a variable sized array, because it would be all constant.

clalancette: No, I think we should just allocate on the heap here.

rcl/src/rcl/logging_rosout.c
```diff
@@ -250,13 +252,40 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(rcl_node_t * node)
...
+    rcutils_char_array_strncat(&message_array, logger_name, strlen(logger_name) + 1);
+    rcutils_char_array_strncat(&message_array, third_part, strlen(third_part) + 1);
+
     RCUTILS_LOG_WARN_NAMED(
       "rcl.logging_rosout",
-      "Publisher already registered for provided node name. If this is due to multiple nodes "
-      "with the same name then all logs for that logger name will go out over the existing "
-      "publisher. As soon as any node with that name is destructed it will unregister the "
-      "publisher, preventing any further logs for that name from being published on the rosout "
-      "topic.");
```
fujitatomoya, line 259: Can we just do the following format?

```c
    RCUTILS_LOG_WARN_NAMED(
      "rcl.logging_rosout",
      "Publisher already registered for: '%s'. If this is due to multiple nodes with the same name "
      "then all logs for: '%s' will go out over the existing publisher. As soon as any "
      "node with that name is destructed it will unregister the publisher, preventing a "
      "any further logs for that name from being published on the rosout topic.",
      node_name, logger_name);
```

CursedRock17: Oh, that would make this a lot cleaner, as long as the buffer length wouldn't get messed up within the `RCUTILS_LOG_WARN_NAMED` function. I was about to ask if I should implement a `strncat_s` type function, with rcutils that would clean up the process, but this seems much better.

fujitatomoya (approved):

fujitatomoya:
@CursedRock17 thanks for the contribution, i am good to go. 

@clalancette can you do the review?

rcl/src/rcl/logging_rosout.c
```diff
@@ -250,13 +251,19 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(rcl_node_t * node)
...
+      "Publisher already registered for: '%s'. If this is due to multiple nodes with the same name "
+      "then all logs for: '%s' will go out over the existing publisher. As soon as any "
+      "node with that name is destructed it will unregister the publisher, preventing a "
```
iuhilnehc-ynos:
```suggestion
      "node with that name is destructed it will unregister the publisher, preventing "
```

rcl/src/rcl/logging_rosout.c
```diff
@@ -250,13 +251,19 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(rcl_node_t * node)
   if (rcutils_hash_map_key_exists(&__logger_map, &logger_name)) {
     // @TODO(nburek) Update behavior to either enforce unique names or work with non-unique
     // names based on the outcome here: https://github.com/ros2/design/issues/187
+    const char * node_name = rcl_node_get_name(node);
+    if (NULL == node_name) {
+      RCL_SET_ERROR_MSG("Node name was null.");
+      return RCL_RET_ERROR;
+    }
```
clalancette: I think we should do this error handling a bit differently.  If we fail to get it, we can still go to the `RCUTILS_LOG_WARN_NAMED` line below, we will just have an "unknown" name.  So maybe something like:

```suggestion
    if (NULL == node_name) {
      node_name = "unknown node";
    }
```

(to be perfectly honest if this ever fails, there will probably be other failures shortly after, but we may as well handle this as best we can)

rcl/src/rcl/logging_rosout.c
```diff
@@ -250,13 +251,19 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(rcl_node_t * node)
...
+
     RCUTILS_LOG_WARN_NAMED(
       "rcl.logging_rosout",
-      "Publisher already registered for provided node name. If this is due to multiple nodes "
-      "with the same name then all logs for that logger name will go out over the existing "
-      "publisher. As soon as any node with that name is destructed it will unregister the "
-      "publisher, preventing any further logs for that name from being published on the rosout "
-      "topic.");
+      "Publisher already registered for: '%s'. If this is due to multiple nodes with the same name "
+      "then all logs for: '%s' will go out over the existing publisher. As soon as any "
```
clalancette:
```suggestion
      "then all logs for the logger named '%s' will go out over the existing publisher. As soon as any "
```

rcl/src/rcl/logging_rosout.c
```diff
@@ -250,13 +251,19 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(rcl_node_t * node)
...
+    }
+
     RCUTILS_LOG_WARN_NAMED(
       "rcl.logging_rosout",
-      "Publisher already registered for provided node name. If this is due to multiple nodes "
-      "with the same name then all logs for that logger name will go out over the existing "
-      "publisher. As soon as any node with that name is destructed it will unregister the "
-      "publisher, preventing any further logs for that name from being published on the rosout "
-      "topic.");
+      "Publisher already registered for: '%s'. If this is due to multiple nodes with the same name "
```
clalancette:
```suggestion
      "Publisher already registered for node name '%s'. If this is due to multiple nodes with the same name "
```

clalancette (changes requested):

rcl/src/rcl/logging_rosout.c
```diff
@@ -250,13 +251,18 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(rcl_node_t * node)
...
+
     RCUTILS_LOG_WARN_NAMED(
       "rcl.logging_rosout",
-      "Publisher already registered for provided node name. If this is due to multiple nodes "
-      "with the same name then all logs for that logger name will go out over the existing "
-      "publisher. As soon as any node with that name is destructed it will unregister the "
-      "publisher, preventing any further logs for that name from being published on the rosout "
-      "topic.");
+      "Publisher already registered for node name: '%s'. If this is due to multiple nodes with the same name "
+      "then all logs for the logger named: '%s' will go out over the existing publisher. As soon as any "
```
clalancette: Linters are complaining that these two lines are too long.  Once that is fixed, I'm happy to approve and run full CI on this.

clalancette (changes requested):

rcl/src/rcl/logging_rosout.c
CursedRock17: Yup going through the extra tests right now, should be done soon.

rcl/src/rcl/logging_rosout.c
```diff
@@ -250,13 +251,20 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(rcl_node_t * node)
...
+
     RCUTILS_LOG_WARN_NAMED(
       "rcl.logging_rosout",
-      "Publisher already registered for provided node name. If this is due to multiple nodes "
-      "with the same name then all logs for that logger name will go out over the existing "
-      "publisher. As soon as any node with that name is destructed it will unregister the "
-      "publisher, preventing any further logs for that name from being published on the rosout "
-      "topic.");
+      "Publisher already registered for node name: '%s'. If this is due to multiple nodes "
+      "with the same name then all logs for the logger named: '%s' will go out over"
```
clalancette:
```suggestion
      "with the same name then all logs for the logger named '%s' will go out over "
```

rcl/src/rcl/logging_rosout.c
```diff
@@ -250,13 +251,20 @@ rcl_ret_t rcl_logging_rosout_init_publisher_for_node(rcl_node_t * node)
...
+      "the existing publisher. As soon as any node with that name is destructed "
+      "it will unregister the publisher, preventing any further logs for that name from"
```
clalancette:
```suggestion
      "it will unregister the publisher, preventing any further logs for that name from "
```

clalancette (changes requested):
Two more small fixes, then we are ready for CI.

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19381/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13906/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20101/)

PR #1089, opened 2023-08-15
Change next_call_time computation of rcl_timer to always result in a future timestamp

kohrt:
In the current implementation `next_call_time` can be equal to `now` after the timer has been called. This can lead to consecutive callback calls with the same timestamp when using sim time.

With this change `next_call_time` will always be greater than `now` after the timer has been called.

### DIFF
--- a/rcl/src/rcl/timer.c
+++ b/rcl/src/rcl/timer.c
@@ -294,7 +294,7 @@ rcl_timer_call(rcl_timer_t * timer)
   // between the timer being ready and the callback being triggered
   next_call_time += period;
   // in case the timer has missed at least once cycle
-  if (next_call_time < now) {
+  if (next_call_time <= now) {
     if (0 == period) {
       // a timer with a period of zero is considered always ready
       next_call_time = now;
@@ -302,7 +302,7 @@ rcl_timer_call(rcl_timer_t * timer)
       // move the next call time forward by as many periods as necessary
       int64_t now_ahead = now - next_call_time;
       // rounding up without overflow
-      int64_t periods_ahead = 1 + (now_ahead - 1) / period;
+      int64_t periods_ahead = 1 + now_ahead / period;
       next_call_time += periods_ahead * period;
     }
   }


rcl/src/rcl/timer.c
```diff
@@ -294,15 +294,15 @@ rcl_timer_call(rcl_timer_t * timer)
   // between the timer being ready and the callback being triggered
   next_call_time += period;
   // in case the timer has missed at least once cycle
-  if (next_call_time < now) {
+  if (next_call_time <= now) {
     if (0 == period) {
       // a timer with a period of zero is considered always ready
       next_call_time = now;
```
fujitatomoya, line 300: period of zero case is handled here.

rcl/src/rcl/timer.c
```diff
@@ -294,15 +294,15 @@ rcl_timer_call(rcl_timer_t * timer)
   // between the timer being ready and the callback being triggered
   next_call_time += period;
   // in case the timer has missed at least once cycle
-  if (next_call_time < now) {
+  if (next_call_time <= now) {
```
fujitatomoya, line 297: what if the period of one case? technically we expect executor calls timer on each tick as much as possible, this will lead the much less cyclic to call the timer?

kohrt: With a period of one nanosecond, `next_call_time` will be `now + 1ns`. On the next tick it will get called again, if the clock advanced at least one nanosecond.

rcl/src/rcl/timer.c
kohrt: I only changed the behaviour for periods greater zero. With zero it will get called every tick, even if the sim clock did not advance. But that may be expected. Not expected is, that a timer with a period gets called twice without the clock advancing.

kohrt:
I think, I didn't make it clear in the beginning. We have the issue that timer callbacks of our nodes get called twice with the same timestamp regularly when used in simulation with use_sim_time set to true. I tracked down the problem and tried to create a minimal change that fixes it.

rcl/src/rcl/timer.c
fujitatomoya: > next_call_time will be now + 1ns

i think this is wrong, it should be `next_call_time = next_call_time + period`.

and adding if statement `(next_call_time == now)`, means either we should call the timer callback or we already miss the timer at least one cycle.
i think the former is correct, timer should be called with `(next_call_time == now)` case. So `periods_ahead` should not be added.

fujitatomoya:
> I think, I didn't make it clear in the beginning. We have the issue that timer callbacks of our nodes get called twice with the same timestamp regularly when used in simulation with use_sim_time set to true. I tracked down the problem and tried to create a minimal change that fixes it.

@kohrt thanks for sharing the issue. is that possible to share the reproducible test project? that would be helpful to understand details and do further debug.

rcl/src/rcl/timer.c
kohrt: > i think this is wrong, it should be next_call_time = next_call_time + period.

I was just referring to your example of `period = 1`, that would result in a `next_call_time` equal to `now + 1`, because `next_call_time` will get increased by a multiple of `period` until its greater than `now`.

next_call_time will be set to `next_call_time += period` in line 295. But if the result is still in the past (or equal to now with my change) it will get increased by a multiple of `period` so that the result will not be in the past.

> and adding if statement (next_call_time == now), means either we should call the timer callback or we already miss the timer at least one cycle.

We are currently in the timer call, the callback will get called in line 313. If `next_call_time == now` we missed the timer already, but since we are about to call it the next call should definitely be in the future. If not, the timer will get called with the same timestemp again, in case of using sim_time. (Because sim_time is updated rarely compared to system clock)

kohrt:
Here is the example project: https://github.com/cellumation/timer_test.git

Compile and run `ros2 launch timer_test timer_test.yaml`. It will output if calls with the same timestamps occure:
```
[timer_test-1] [WARN] [1692347432.710947829] [timer_test]: same timestamp: 2600000000
[timer_test-1] [WARN] [1692347432.810938363] [timer_test]: same timestamp: 2700000000
[timer_test-1] [WARN] [1692347432.910990966] [timer_test]: same timestamp: 2800000000
[timer_test-1] [WARN] [1692347433.011015172] [timer_test]: same timestamp: 2900000000
[timer_test-1] [WARN] [1692347433.111107161] [timer_test]: same timestamp: 3000000000
[timer_test-1] [WARN] [1692347433.211355448] [timer_test]: same timestamp: 3100000000
[timer_test-1] [WARN] [1692347439.224457634] [timer_test]: same timestamp: 9110000000
```

rcl/src/rcl/timer.c
```diff
@@ -294,15 +294,15 @@ rcl_timer_call(rcl_timer_t * timer)
...
     if (0 == period) {
       // a timer with a period of zero is considered always ready
       next_call_time = now;
     } else {
       // move the next call time forward by as many periods as necessary
       int64_t now_ahead = now - next_call_time;
       // rounding up without overflow
-      int64_t periods_ahead = 1 + (now_ahead - 1) / period;
+      int64_t periods_ahead = 1 + now_ahead / period;
```
fujitatomoya, line 305: I think this should be fixed as suggested.

`next_call_time < now` is true, so it cannot be set to `next_call_time == now` since it is already in the past 100%.
the reported problem happens if this situation happens, because next execution loop takes this timer is ready.

rcl/src/rcl/timer.c
fujitatomoya: > We are currently in the timer call, the callback will get called in line 313.

i don't think so. according to your sample, you are using `rclcpp`, right? so that this should be NULL.

user callback will be called `rclcpp` here,

https://github.com/ros2/rclcpp/blob/fbe8f28cd13710c5c643a4e7149e509f3a952677/rclcpp/include/rclcpp/timer.hpp#L279

> If next_call_time == now we missed the timer already,

for me, this means that we should not set the bias. instead, we go to the next execution right away to call the timer to avoid missing the timer cyclic as less as possible. and then period increases `next_call_time`.

fujitatomoya (commented):
@kohrt thanks for providing the test project, that really helped me to understand.

i did some experimental check and left a couple of comments, could you share your thought?

CC: @ros2/team

fujitatomoya:
My local test project, see https://github.com/fujitatomoya/ros2_test_prover/commit/544e61545ebf38686a9a69a124bfcd8ffdf088eb

with the following commands, we can reproduce the problem.

```bash
> colcon build --symlink-install --packages-select prover_rclcpp

### Terminal-1
> ros2 run prover_rclcpp rcl_1089 --ros-args -p use_sim_time:=true --ros-args --log-level debug

### Terminal-2
> ros2 run prover_rclcpp sim_clock_publisher 
```

rcl/src/rcl/timer.c
kohrt: > i don't think so. according to your sample, you are using rclcpp, right? so that this should be NULL.

True, rclcpp does the callback differently, I just wanted to stay in scope of `rcl`. Anyway the decision for calling is based on `next_call_time` in both implementations and `rcl_timer_call` is called in both cases to update the timer internals.

> for me, this means that we should not set the bias. instead, we go to the next execution right away to call the timer to avoid missing the timer cyclic as less as possible. and then period increases next_call_time.

To prevent the issue from occurring `next_call_time` must be different from the current call time, which is `now`.

I think the current implementation tries to stick to the period between cycles, even if previous cycles were missed. That's why `next_call_time` is always incremented by a multiple of `period` instead of just setting it to `now + 1` to minimise missed cycles.

But I am fine with any solution, as long as multiple callback calls with the same timestamp are prevented.

fujitatomoya: right, let's keep it open for other reviewers to get more opinion.

kohrt: How can we procced? Should we @mention someone to ask for a second opinion?

fujitatomoya: @kohrt yes, that i would like to do. actually i had a quick chat with @ros2/team we will get the feedback, please give us some time.

clalancette: I spent a bit of time looking at this, and I have a question and some comments.

First, I want to ask about this statement:

> If next_call_time == now we missed the timer already,

Is that really true?  If we look at the implementation of `rcl_timer_is_ready` (the next function down), we determine that the timer is ready if the (next_call_time - now **<=** 0).  So it seems to me that the timer is considered ready (and thus expired) in both the case where the next call time is prior to now, and when they are exactly equal.  Does that match your understanding?

If what I say above is true, then I think it makes sense for us to change this check to **<=** as well, since `next_call_time == now` is considered "ready".  We could consider changing `rcl_timer_is_ready` to change that **<=** to a **<**, but I think that would cause us to miss more timer periods.

Getting slightly more formal here, we should consider what happens in this bit of code for all three cases of `next_call_time < now`, `next_call_time == now`, and `next_call_time > now`.  The third case is illogical; we should never get here if that is the case.  Nevertheless, it might be a good idea to check for that in a separate clause just to cover all of the bases.  The first case is already covered by the current code.  That leaves the equal case, and like I said it seems to make sense to me to make this change because we consider **<=** to be ready below.

kohrt: > First, I want to ask about this statement:
> 
> > If next_call_time == now we missed the timer already,
> 
> Is that really true? If we look at the implementation of `rcl_timer_is_ready` (the next function down), we determine that the timer is ready if the (next_call_time - now **<=** 0). So it seems to me that the timer is considered ready (and thus expired) in both the case where the next call time is prior to now, and when they are exactly equal. Does that match your understanding?

Yes, that is true.

But maybe we are talking about different things. Let me try to clarify the issue:

We are at timestamp `now` and the current timer had been checked for expiration, `rcl_timer_is_ready` was called and set `is_ready = true`, therefor `rcl_timer_call` is called.

That means that we are in the state that the timer is expired and the timer in currently in the process of being called.

One part of calling the timer is to update the timestamps for the next timer call.

`next_call_time` will be increased by the timer period: `next_call_time += period;`

Afterwards it is checked, if the `next_call_time` is still in the past: `if (next_call_time < now)`.
*Note: If it's the case, it means that the timer should have been called at least period time earlier, but the period was missed.*

If `next_call_time` is in the past, it is increased by a multiple of period so that the result is `>= now`.

This can lead to the situation that the current call time is `now` and `next_call_time` is also `now`. In case the system time does not update as fast as the timer is checked for expiration, the timer will to be called twice with the same timestamp. This can be the case when sim_time is used.

Therefore I suggested to change the update of  `next_call_time` to always give results that are in the future. That's why I changed the check to `<=` and increased update by +1 period.

Currently `next_call_time` is always increased by a multiple of period, even if a previous period was missed. I think this is done on purpose, therefore my suggestion keeps it that way.

Alternatively, one could set `next_call_time = now + 1;` in case of a missed period, to minimise the overall missed periods.

*Note: The special case of a timer with a period of 0 is not considered yet. But if we change it, to always set the 'next_call_time' to a future timestamps, we should consider changing the special case update in line 300 to `next_call_time = now + 1;` as well.*

clalancette: > Afterwards it is checked, if the `next_call_time` is still in the past: `if (next_call_time < now)`.

Yes, thank you.  This was the part that I missed in my previous review.

> Therefore I suggested to change the update of `next_call_time` to always give results that are in the future. That's why I changed the check to `<=` and increased update by +1 period.

Right.  This makes sense.

Going back to @fujitatomoya 's statement:

> for me, this means that we should not set the bias. instead, we go to the next execution right away to call the timer to avoid missing the timer cyclic as less as possible. and then period increases next_call_time.

I don't think I agree with this.  Regardless of the math to determine the next call, we are going to call the registered callback right now with the amount of time since the last callback.  That is going to be >= 1 period.  But the important point is that the callback has the information it needs to know that it missed a callback.  I don't think we need to additionally attempt to "catch up" and call the callback multiple times.  That could lead to starvation of the rest of the system in the case where we missed multiple periods.

All of this is to say that I agree with the solution that @kohrt has laid out here.  I'm going to approve it and run CI, but @fujitatomoya I'd appreciate another look from you to see whether you agree.

clalancette (approved):

clalancette:
> and run CI

Actually, that won't work because we've added some APIs in the meantime.  @kohrt Can you please go ahead and rebase this?  Once that is done I'll run CI.

rcl/src/rcl/timer.c
fujitatomoya: @clalancette thanks for taking a look at this.

I am good to go with you and @kohrt , i just wanted to have the 2nd opinion since me and @kohrt were only having this discussion 😄 appreciate it.

@kohrt thanks for being patient on this 👍

fujitatomoya (approved):

kohrt:
Should be rebased now. Thanks for the review @fujitatomoya and @clalancette

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19649/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14157/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20373/)

PR #1092, opened 2023-08-21
Shorten wait time for a subscription not being ready.

clalancette:
We shouldn't wait around for 10 seconds for a subscription to not be ready; that can take literally forever.  Instead, just wait one second.  While we are in here, also make the case where we *are* waiting for a subscription to be ready more reactive by reducing the wait time to 100ms, with more tries.

### DIFF
--- a/rcl/test/rcl/test_subscription.cpp
+++ b/rcl/test/rcl/test_subscription.cpp
@@ -984,9 +984,9 @@ TEST_F(
   }
 
   if (is_cft_support) {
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19366/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13894/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20088/)

fujitatomoya (approved):

PR #1095, opened 2023-08-22
Revamp the test_get_type_description_service.

clalancette:
1. Make it more robust to the network taking time to do discovery.
2. Shorten the delay time when we expect things to fail.
3. Fix a possible leak during service_exists(), pointed out by valgrind.
4. Remove unnecessary heap allocations in service_exists() (we can just use the stack here).
5. Fix a definite leak during test teardown, where we were forgetting to fini the node_type_description_service.

@emersonknapp @achim-k FYI

### DIFF
--- a/rcl/test/rcl/test_get_type_description_service.cpp
+++ b/rcl/test/rcl/test_get_type_description_service.cpp
@@ -14,7 +14,9 @@
 
 #include <gtest/gtest.h>
 
+#include <chrono>
 #include <cstring>
+#include <thread>
 
 #include "rcl/error_handling.h"
 #include "rcl/graph.h"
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19372/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13899/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20093/)

rcl/test/rcl/test_get_type_description_service.cpp
```diff
@@ -51,87 +53,61 @@ static bool string_in_array(rcutils_string_array_t * array, const char * pattern
...
-
-  const bool srv_name_found = string_in_array(
-    &srv_names_and_types->names,
-    service_name);
+    if (
+      RCL_RET_OK != rcl_get_service_names_and_types(
+        node_ptr,
+        &allocator, &srv_names_and_types))
+    {
+      return false;
```
fujitatomoya, line 70: `OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT` should be called before this if that returns false?

clalancette: It isn't actually necessary in this case.

When you call `rcl_get_zero_initialized_names_and_types`, it returns a copy of the statically-zeroed structure, which we are storing on the stack.  When you call `rcl_get_service_names_and_types`, it expects to receive a zeroed `rcl_names_and_types_t` structure, and actually fails if it does not.  So if you call `rcl_names_and_types_init` first, `rcl_get_service_names_and_types` will fail.  So you only need to call `rcl_names_and_types_fini` after a successful `rcl_get_service_names_and_types`.

To be honest, I don't really like this API.  It operates differently from most of the rest of our APIs, which typically assume that you have run `init` first before doing anything else with the structure.  But it would be difficult to change (since it goes through the RMW layer), and is orthogonal to this change anyway.

Does this make sense to you?

fujitatomoya: understood, thanks for the explanation.

fujitatomoya (approved):

PR #1096, opened 2023-08-24
Reduce the number of tests we run

clalancette:
The rcl APIs for arguments, time, and lexing never touch the RMW layer at all.  Therefore, there is no need to run the tests for them against all of the individual RMWs; we can get away with just running them once.  This should (slightly) speed up our compilations and tests.

### DIFF
--- a/rcl/test/CMakeLists.txt
+++ b/rcl/test/CMakeLists.txt
@@ -86,14 +86,6 @@ function(test_target_function)
     AMENT_DEPENDENCIES ${rmw_implementation} "osrf_testing_tools_cpp"
   )
 
-  rcl_add_custom_gtest(test_lexer_lookahead${target_suffix}
-    SRCS rcl/test_lexer_lookahead.cpp
-    ENV ${rmw_implementation_env_var}
-    APPEND_LIBRARY_DIRS ${extra_lib_dirs}
-    LIBRARIES ${PROJECT_NAME}
-    AMENT_DEPENDENCIES ${rmw_implementation} "osrf_testing_tools_cpp"
-  )
-
   rcl_add_custom_gtest(test_graph${target_suffix}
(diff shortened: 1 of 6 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19384/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13910/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20104/)

rcl/test/CMakeLists.txt
```diff
@@ -62,14 +62,6 @@ function(test_target_function)
...
 
-  rcl_add_custom_gtest(test_time${target_suffix}
-    SRCS rcl/test_time.cpp
-    ENV ${rmw_implementation_env_var} ${memory_tools_ld_preload_env_var}
-    APPEND_LIBRARY_DIRS ${extra_lib_dirs}
-    LIBRARIES ${PROJECT_NAME} osrf_testing_tools_cpp::memory_tools
-    AMENT_DEPENDENCIES ${rmw_implementation}
-  )
-
   rcl_add_custom_gtest(test_timer${target_suffix}
```
fujitatomoya, line 65: what about the timer? i think `rcl` also deals with this too, but RMW implementation.

clalancette: The main reason I didn't change timer is because a timer instantiates a guard condition, which makes a call down into the RMW layer.  So theoretically there could be a bug in one of the RMW implementations, but not another, which this test would find.

fujitatomoya: sounds reasonable. thanks!

fujitatomoya (approved):

audrow (approved):
Looks good to me!

PR #1097, opened 2023-08-25
Simplify local_namespace handling in rcl_node_init.

clalancette:
There is a comment in there about avoiding a memory allocation, but that doesn't seem to make much sense to me; we are going to be doing an allocation directly below, so avoiding it here seems unnecessary and makes the code harder to follow.  Simplify everything by always making an allocation

### DIFF
--- a/rcl/src/rcl/node.c
+++ b/rcl/src/rcl/node.c
@@ -12,11 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifdef __cplusplus
-extern "C"
-{
-#endif
-
 #include "rcl/node.h"
 
 #include <limits.h>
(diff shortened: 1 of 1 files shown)


fujitatomoya (approved):
makes sense to me.

i think it allocated the memory for `"/"` anyway (if we really want to avoid memory allocation, i would store it in the code area)... instead of this, we can go straight-forward to keep it simple. besides, this is one of the configuration APIs.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19392/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13921/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20113/)

PR #1099, opened 2023-08-28
Cleanup the error handling in rcl_node_init.

clalancette:
Only undo things that we've done before.  Not only does this result in less code, it also avoids spurious warnings from trying to cleanup things that we haven't initialized.

### DIFF
--- a/rcl/src/rcl/node.c
+++ b/rcl/src/rcl/node.c
@@ -118,7 +118,6 @@ rcl_node_init(
     rcl_guard_condition_get_default_options();
   rcl_ret_t ret;
   rcl_ret_t fail_ret = RCL_RET_ERROR;
-  char * remapped_node_name = NULL;
 
   // Check options and allocator first, so allocator can be used for errors.
   RCL_CHECK_ARGUMENT_FOR_NULL(options, RCL_RET_INVALID_ARGUMENT);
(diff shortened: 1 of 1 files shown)


rcl/src/rcl/node.c
```diff
@@ -270,96 +277,83 @@ rcl_node_init(
...
     rcl_node_get_name(node),
     rcl_node_get_namespace(node));
-  goto cleanup;
-fail:
-  if (node->impl) {
-    if (rcl_logging_rosout_enabled() &&
-      node->impl->options.enable_rosout &&
-      node->impl->logger_name)
-    {
-      ret = rcl_logging_rosout_fini_publisher_for_node(node);
```
fujitatomoya, line 307: is this missing to be called with new code?

fujitatomoya (commented):
having the inverted water fall order to finalize with `goto` statement works and do only necessary finalization. on the other hand, it would be complicated to add the process afterwards not to miss the correct jump point for code maintenance. (this could easily go to memory leak issue.)

i would prefer the single failing jump point and check if what is necessary to call with current implementation.

rcl/src/rcl/node.c
clalancette: Oh, yikes.  I missed that, good call.

clalancette:
> i would prefer the single failing jump point and check if what is necessary to call with current implementation.

That's fair.  While this definitely cleans some things up, maybe it isn't the best strategy (even if I find it to be the easiest to understand).  Let me try something else here.

rcl/src/rcl/node.c
clalancette: Oh wait, no.  Now that I look at this again, I removed this on purpose.  `rcl_logging_rosout_init_publisher_for_node` is the last substantial thing we do.  So there is no way to get into the error path after having called that.  Thus, we never need to clean it up.

fujitatomoya: i think this assumption is based on implementation detail here? what if someone modifies `rcl_logging_rosout_init_publisher_for_node` later, which leads that to fail here possibly? according to API, i think we can keep `rcl_logging_rosout_fini_publisher_for_node` to be called?

btw,

https://github.com/ros2/rcl/blob/41b83c8ebfe8344a80c0093083b7abdfded79337/rcl/include/rcl/logging_rosout.h#L104

i cannot find `rcl_logging_destroy_rosout_publisher_for_node`, we probably need to change it into `rcl_logging_rosout_fini_publisher_for_node`.

clalancette: > i think this assumption is based on implementation detail here?

Yes and no.  All of the error handling in here is based on an implementation detail, namely that if a call fails it cleans up after itself.  If any of the init-style functions don't follow that, we'll probably leak or be in an inconsistent state anyway.  So even if `rcl_logging_rosout_init_publisher_for_node`, fails, it should clean up after itself.  There is just nothing (currently) that happens after it that would cause us to need to finalize it.

fujitatomoya: understood, thanks!

clalancette:
All right, I've totally revamped this to go another direction.  In particular, I keep the `fail` handling as it was, but just make it a lot more careful about only undoing things that have already been done.  I think this turns out pretty nice, should still remove some of the bogus warning messages we get, and still removes a bunch of code.  Please take another look when you get a chance!

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19525/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14043/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20245/)

fujitatomoya (approved):

fujitatomoya:
CI(retry):
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14049/)

PR #1101, opened 2023-09-11
Fix a clang warning for suspicious string concatentation.

clalancette:
clang warns when strings are concatenated within array declarations; this can help to catch bugs.

However, in the cases here we actually mean to do string concatenation.  Signal this to the compiler by making a separate variable where we do the concatenation, which quiets the clang warning and should have no effect on operation.

### DIFF
--- a/rcl/test/rcl/test_logging.cpp
+++ b/rcl/test/rcl/test_logging.cpp
@@ -91,8 +91,8 @@ TEST(TestLogging, test_configure_with_bad_arguments) {
 
 TEST(TestLogging, test_logging_rosout_enabled) {
   {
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19491/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14011/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20210/)

mjcarroll (approved):

fujitatomoya (approved):

PR #1106, opened 2023-10-04
Remove an arbitrary sleep of 500 milliseconds.

clalancette:
The code below already attempts to take the messages for up to 10 seconds, so there is no reason for this wait.

### DIFF
--- a/rcl/test/rcl/test_subscription.cpp
+++ b/rcl/test/rcl/test_subscription.cpp
@@ -533,9 +533,14 @@ TEST_F(
     ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
   }
 
+  // Give a brief moment for publications to go through.
+  // TODO(clalancette): We should have a way to "peek" at how much data is available, without
+  // taking it.  Then we could block until we see that amount, letting us remove this sleep and
+  // the do..while loop below.
+  std::this_thread::sleep_for(std::chrono::milliseconds(500));
   // Take fewer messages than are available in the subscription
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19604/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14114/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20328/)
* Linux repeated [Build Status](http://ci.ros2.org/job/ci_linux/19605/)

fujitatomoya (approved):
lgtm. (it is weird that we have this sleep only in here... it can deal with `do...while` wait loop anyway.)

clalancette:
So it turns out that we actually do indeed need this 500 millisecond timeout.  The point of the test is to test that we are only taking 3 of the messages when 5 are available.  And for that, we need to make sure the 5 are actually available, which this 500 millisecond sleep (mostly) accomplishes.

Honestly, though, what we really need is another API to tell us when we have 5 subscription messages available to take.  That way we can just wait until that happens, and then get rid of the sleep.  Since that would involve work at the RMW layer, I decided to just add a TODO to this effect.  I don't love it, but at least we won't wonder why this is in here again.

sloretz:
> [...] what we really need is another API to tell us when we have 5 subscription messages available to take [...]

Opened issue describing that feature: https://github.com/ros2/rcl/issues/1109

Closing for now.

PR #1112, opened 2023-10-26
Return service from node_type_description_service_init

mjcarroll:
This changes the mechanism of `rcl_node_type_description_service_init` to populate a user-provided service.

This makes it the calling user's responsibility to call `rcl_service_fini` on the returned service, making it more in line with the rest of the `rcl` API.

This was discovered when working with the new `rclcpp::WaitSet` implementation.  

The main issue is that the executor needs to share ownership with objects that it is `wait`-ing on for the entire duration of the `wait`, even if a node is removed from the executor at that time.  That is, the executor doesn't have knowledge of the nodes so much as the waitable objects in the collection.  

In this current implementation, the type_description was being `fini`-ed during the `NodeTypeDescriptions` destructor here: https://github.com/ros2/rclcpp/blob/fcbe64cff4bea3109531254ceb2955dc4b1bb320/rclcpp/src/rclcpp/node_interfaces/node_type_descriptions.cpp#L128-L139.  This could cause a situation where a service is cleaned up while we are still waiting on it, and yields a segfault afterwards.

After discussion with @clalancette, it seems the better approach is to have the user init the service and then be responsible for managing the lifetime of that service through their client library (eg `rclcpp::Service`)

### DIFF
--- a/rcl/include/rcl/node.h
+++ b/rcl/include/rcl/node.h
@@ -562,10 +562,8 @@ rcl_get_disable_loaned_message(bool * disable_loaned_message);
  * must register rcl_node_type_description_service_handle_request or a custom callback
  * to handle incoming requests, via that client's executor/waitset capabilities.
  *
- * Note that the underlying service must be cleaned up by the caller by calling
- * rcl_node_get_type_description_service and rcl_service_fini.
- *
- * This will initialize the node's type cache, if it has not been initialized already.
+ * Note that the returned service must be cleaned up by the caller by calling
+ * rcl_service_fini.
  *
  * <hr>
  * Attribute          | Adherence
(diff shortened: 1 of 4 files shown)


rcl/src/rcl/node.c
```diff
@@ -628,13 +628,7 @@ rcl_ret_t rcl_node_type_description_service_fini(rcl_node_t * node)
...
   }
 
-  const rcl_ret_t ret =
-    rcl_service_fini(&node->impl->get_type_description_service, node);
-  if (RCL_RET_OK == ret) {
-    node->impl->get_type_description_service = rcl_get_zero_initialized_service();
-  }
-
-  return ret;
+  return RCL_RET_OK;
```
fujitatomoya: > This was causing a segmentation fault in the new rclcpp::WaitSet implementation as we were waiting on a service that was being cleaned up by the fini call.

i am not sure about this problem, can you elaborate it? btw, after the change, this function does pretty much nothing? in that case, we can just remove this function?

mjcarroll: Sorry, I opened this to get @clalancette's opinion quickly.  I have updated the PR as well as the description to more clearly explain the problem and proposed solution.

rcl/include/rcl/node.h
```diff
@@ -562,7 +562,8 @@ rcl_get_disable_loaned_message(bool * disable_loaned_message);
  * must register rcl_node_type_description_service_handle_request or a custom callback
  * to handle incoming requests, via that client's executor/waitset capabilities.
  *
- * This will initialize the node's type cache, if it has not been initialized already.
```
mjcarroll, line 565: Note that this is not true.  The type cache is actually initialized when the node is initialized.

fujitatomoya: i might be mistaken... I think this is correct in `rcl` aspect? client library `rclcpp` needs to call this function to initialize type description service? in client library `rclcpp`, this is not true.

rcl/include/rcl/node.h
```diff
@@ -581,59 +582,9 @@ rcl_get_disable_loaned_message(bool * disable_loaned_message);
...
- * \return #RCL_RET_ERROR if an unspecified error occurs.
- */
-RCL_PUBLIC
-RCL_WARN_UNUSED
-rcl_ret_t rcl_node_get_type_description_service(
-  const rcl_node_t * node,
-  rcl_service_t ** service_out);
-
+rcl_ret_t rcl_node_type_description_service_init(
+  rcl_service_t * service,
```
fujitatomoya, line 587: doc section needs to be update accordingly?

```c
 * \param[in] service the handle to the type description service to be initialized
 * \param[in] node the handle to the node whose type cache should be initialized
```

rcl/include/rcl/node.h
mjcarroll: This call will initialize the service, but the cache is already created in `rcl_node_init` here: https://github.com/ros2/rcl/blob/45fc9520216013007c2d4f3bd69d778af9051aa3/rcl/src/rcl/node.c#L278-L283

rcl/include/rcl/node.h
mjcarroll: [6f7da3a](https://github.com/ros2/rcl/pull/1112/commits/6f7da3a16387b59e799e33463cb6daecb4de7a19)

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19694/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14200/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20424/)

fujitatomoya (approved):
lgtm

mjcarroll:
One more run after some rebase/force-push:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19696/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14202/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20426/)

Issue #1103, opened 2023-09-23
rcl_action: result_timeout should be started on goal completion


fujitatomoya:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04 (Any)
- Installation type:
  - Any but Iron and Rolling
- Version or commit hash:
  - N.A
- DDS implementation:
  - Any
- Client library (if applicable):
  - Any (rclcpp, rclpy)

### Description

If the user application uses default `rcl_action_server_options_t`, `goal_handle` will be considered as expired after 10 seconds.

as described in https://github.com/ros-planning/navigation2/issues/3765, that is so likely to take more than 10 seconds to set the goal result before expired(10 seconds) once accepted. https://github.com/ros-planning/navigation2/issues/3765#issuecomment-1689951628 analyzes and work-around this issue with setting 30 seconds via `rcl_action_server_options_t.result_timeout`.

### Consideration / Proposal

- `rcl_action` : increase the timeout from 10 seconds to 1 minute in default. (15 minutes are too long though.) https://github.com/ros2/rcl/pull/1012 reduced the timeout into 10 seconds, but thinking about the use case such as Nav2 relies on ROS 2 action, 10 seconds is short in default. (**backport required to Iron**)
- `rclpy_action` : result_timeout default should be set to rcl's default accordingly. (currently this is set to `900` seconds.) 

### Related Issue

- https://github.com/ros2/rclcpp/issues/2101 (`timeout == 0` does not work)

fujitatomoya:
@SteveMacenski @clalancette @AlexeyMerzlyakov what do you think?

SteveMacenski:
One minute is better yet than 10 seconds, but doesn't really address the underlying problem. We need a different mechanic to expire goals that isn't based on request time (e.g. last-updated time? last-result-requested time?) 

But, I'll take incremental improvements where I can get it. Nav2's workaround of increasing the time solves **my** immediate problems like this, but still leaves every other user that isn't extremely well plugged into the on-goings of Nav2 / rclcpp development in the lurch. So for their sake to help mask more of the problem in the meantime, I would be very supportive of a move up to 1 minute. 

+1 this is a good suggestion

fujitatomoya:
> last-updated time? last-result-requested time?

i see, maybe we can rely on server process or event that resets the timeout. (e.g `publish_feedback`, `publish_status` or `result_request_received`) thanks for the idea.

the followings are PRs just change the default value.

[...]

PR #1121, opened 2023-11-29
Change the starting time of the goal expiration timeout

Barry-Xu-2018:
Address one issue mentioned #1103 

The expire timeout [is being calculated from the time the goal was accepted](https://github.com/ros2/rcl/blob/9a05fa65926673b2986175ff055215861e29bf5b/rcl_action/src/rcl_action/action_server.c#L644-L645). According to the design https://design.ros2.org/articles/actions.html#result-caching, it should be calculated from the time the goal is completed.

### DIFF
--- a/rcl/include/rcl/types.h
+++ b/rcl/include/rcl/types.h
@@ -128,7 +128,7 @@ typedef rmw_ret_t rcl_ret_t;
 
 // rcl action specific ret codes in 40XX
 /// No terminal timestamp for the goal as it has not reached a terminal state.
-#define RCL_RET_NOT_TERMINATED_YET 4001
+#define RCL_ACTION_RET_NOT_TERMINATED_YET 4001
 
 /// typedef for rmw_serialized_message_t;
 typedef rmw_serialized_message_t rcl_serialized_message_t;
--- a/rcl/include/rcl/types.h
+++ b/rcl/include/rcl/types.h
@@ -126,6 +126,10 @@ typedef rmw_ret_t rcl_ret_t;
 /// rcl_lifecycle state not registered
 #define RCL_RET_LIFECYCLE_STATE_NOT_REGISTERED 3001
 
+// rcl action specific ret codes in 40XX
+/// No terminal timestamp for the goal as it has not reached a terminal state.
+#define RCL_RET_NOT_TERMINATED_YET 4001
+
 /// typedef for rmw_serialized_message_t;
 typedef rmw_serialized_message_t rcl_serialized_message_t;
 
--- a/rcl_action/include/rcl_action/goal_handle.h
+++ b/rcl_action/include/rcl_action/goal_handle.h
@@ -24,6 +24,7 @@ extern "C"
 #include "rcl_action/types.h"
 #include "rcl_action/visibility_control.h"
 #include "rcl/allocator.h"
+#include "rcl/time.h"
 
 
 /// Internal rcl action goal implementation struct.
@@ -190,6 +191,56 @@ rcl_action_goal_handle_get_status(
   const rcl_action_goal_handle_t * goal_handle,
   rcl_action_goal_state_t * status);
 
+/// Get the goal terminal timestamp.
+/**
+ * This is a non-blocking call.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | No
+ *
+ * \param[in] goal_handle struct containing the goal and metadata
+ * \param[out] timestamp a preallocated struct where goal terminal timestamp is copied
+ * \return `RCL_RET_OK` if the goal ID was accessed successfully, or
+ * \return `RCL_RET_ACTION_GOAL_HANDLE_INVALID` if the goal handle is invalid, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if the timestamp argument is invalid
+ */
+RCL_ACTION_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_action_goal_handle_get_goal_terminal_timestamp(
+  const rcl_action_goal_handle_t * goal_handle,
+  rcl_time_point_value_t * timestamp);
+
+/// Set the goal terminal timestamp.
+/**
+ * This is a non-blocking call.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | No
+ *
+ * \param[in] goal_handle struct containing the goal and metadata
+ * \param[in] timestamp The timestamp of goal termination
+ * \return `RCL_RET_OK` if the goal ID was accessed successfully, or
+ * \return `RCL_RET_ACTION_GOAL_HANDLE_INVALID` if the goal handle is invalid, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if the timestamp argument is invalid
+ */
+RCL_ACTION_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_action_goal_handle_set_goal_terminal_timestamp(
+  const rcl_action_goal_handle_t * goal_handle,
+  rcl_time_point_value_t timestamp);
+
 /// Check if a goal is active using a rcl_action_goal_handle_t.
 /**
  * This is a non-blocking call.
@@ -200,7 +251,7 @@ rcl_action_goal_handle_get_status(
  * Allocates Memory   | No
  * Thread-Safe        | No
  * Uses Atomics       | No
- * Lock-Free          | Yes
+ * Lock-Free          | No
  *
  * \param[in] goal_handle struct containing the goal and metadata
  * \return `true` if the goal is in one of the following states: ACCEPTED, EXECUTING, or CANCELING, or
@@ -222,7 +273,7 @@ rcl_action_goal_handle_is_active(const rcl_action_goal_handle_t * goal_handle);
  * Allocates Memory   | No
  * Thread-Safe        | No
  * Uses Atomics       | No
- * Lock-Free          | Yes
+ * Lock-Free          | No
  *
  * \param[in] goal_handle struct containing the goal and metadata
  * \return `true` if the goal can be transitioned to CANCELING from its current state, or
--- a/rcl_action/include/rcl_action/goal_handle.h
+++ b/rcl_action/include/rcl_action/goal_handle.h
@@ -201,7 +201,7 @@ rcl_action_goal_handle_get_status(
  * Allocates Memory   | No
  * Thread-Safe        | No
  * Uses Atomics       | No
- * Lock-Free          | No
+ * Lock-Free          | Yes
  *
  * \param[in] goal_handle struct containing the goal and metadata
  * \param[out] timestamp a preallocated struct where goal terminal timestamp is copied
@@ -226,7 +226,7 @@ rcl_action_goal_handle_get_goal_terminal_timestamp(
  * Allocates Memory   | No
  * Thread-Safe        | No
  * Uses Atomics       | No
- * Lock-Free          | No
+ * Lock-Free          | Yes
  *
  * \param[in] goal_handle struct containing the goal and metadata
  * \param[in] timestamp The timestamp of goal termination
@@ -251,7 +251,7 @@ rcl_action_goal_handle_set_goal_terminal_timestamp(
  * Allocates Memory   | No
  * Thread-Safe        | No
  * Uses Atomics       | No
- * Lock-Free          | No
+ * Lock-Free          | Yes
  *
  * \param[in] goal_handle struct containing the goal and metadata
  * \return `true` if the goal is in one of the following states: ACCEPTED, EXECUTING, or CANCELING, or
@@ -273,7 +273,7 @@ rcl_action_goal_handle_is_active(const rcl_action_goal_handle_t * goal_handle);
  * Allocates Memory   | No
  * Thread-Safe        | No
  * Uses Atomics       | No
- * Lock-Free          | No
+ * Lock-Free          | Yes
  *
  * \param[in] goal_handle struct containing the goal and metadata
  * \return `true` if the goal can be transitioned to CANCELING from its current state, or
--- a/rcl_action/include/rcl_action/goal_handle.h
+++ b/rcl_action/include/rcl_action/goal_handle.h
@@ -38,7 +38,7 @@ typedef struct rcl_action_goal_handle_s
 } rcl_action_goal_handle_t;
 
 /// Define invaild value for goal terminal timestamp
-#define INVAILD_GOAL_TERMINAL_TIMESTAMP 0
+#define INVAILD_GOAL_TERMINAL_TIMESTAMP -1
 
 /// Return a rcl_action_goal_handle_t struct with members set to `NULL`.
 /**
@@ -208,7 +208,7 @@ rcl_action_goal_handle_get_status(
  *
  * \param[in] goal_handle struct containing the goal and metadata
  * \param[out] timestamp a preallocated struct where goal terminal timestamp is copied.
-     Return `INVAILD_GOAL_TERMINAL_TIMESTAMP` if goal terminal timestamp isn't set.
+     Return `INVAILD_GOAL_TERMINAL_TIMESTAMP` if the goal has not reached terminal state.
  * \return `RCL_RET_OK` if the goal ID was accessed successfully, or
  * \return `RCL_RET_ACTION_GOAL_HANDLE_INVALID` if the goal handle is invalid, or
  * \return `RCL_RET_INVALID_ARGUMENT` if the timestamp argument is invalid
@@ -219,31 +219,6 @@ rcl_ret_t
 rcl_action_goal_handle_get_goal_terminal_timestamp(
   const rcl_action_goal_handle_t * goal_handle,
   rcl_time_point_value_t * timestamp);
-
-/// Set the goal terminal timestamp.
-/**
- * This is a non-blocking call.
- *
- * <hr>
- * Attribute          | Adherence
- * ------------------ | -------------
- * Allocates Memory   | No
- * Thread-Safe        | No
- * Uses Atomics       | No
- * Lock-Free          | Yes
- *
- * \param[in] goal_handle struct containing the goal and metadata
- * \param[in] timestamp The timestamp of goal termination
- * \return `RCL_RET_OK` if the goal ID was accessed successfully, or
- * \return `RCL_RET_ACTION_GOAL_HANDLE_INVALID` if the goal handle is invalid, or
- * \return `RCL_RET_INVALID_ARGUMENT` if the timestamp argument is invalid
- */
-RCL_ACTION_LOCAL
-RCL_WARN_UNUSED
-rcl_ret_t
-rcl_action_goal_handle_set_goal_terminal_timestamp(
-  const rcl_action_goal_handle_t * goal_handle,
-  rcl_time_point_value_t timestamp);
 
 /// Check if a goal is active using a rcl_action_goal_handle_t.
 /**
--- a/rcl_action/include/rcl_action/goal_handle.h
+++ b/rcl_action/include/rcl_action/goal_handle.h
@@ -208,10 +208,10 @@ rcl_action_goal_handle_get_status(
  *
  * \param[in] goal_handle struct containing the goal and metadata
  * \param[out] timestamp a preallocated struct where goal terminal timestamp is copied.
-     Return `INVAILD_GOAL_TERMINAL_TIMESTAMP` if the goal has not reached terminal state.
  * \return `RCL_RET_OK` if the goal ID was accessed successfully, or
  * \return `RCL_RET_ACTION_GOAL_HANDLE_INVALID` if the goal handle is invalid, or
- * \return `RCL_RET_INVALID_ARGUMENT` if the timestamp argument is invalid
+ * \return `RCL_RET_INVALID_ARGUMENT` if the timestamp argument is invalid or
+ * \return `RCL_RET_NOT_TERMINATED_YET` if the goal has not reached terminal state
  */
 RCL_ACTION_PUBLIC
 RCL_WARN_UNUSED
--- a/rcl_action/include/rcl_action/goal_handle.h
+++ b/rcl_action/include/rcl_action/goal_handle.h
@@ -37,6 +37,9 @@ typedef struct rcl_action_goal_handle_s
   rcl_action_goal_handle_impl_t * impl;
 } rcl_action_goal_handle_t;
 
+/// Define invaild value for goal terminal timestamp
+#define INVAILD_GOAL_TERMINAL_TIMESTAMP 0
+
 /// Return a rcl_action_goal_handle_t struct with members set to `NULL`.
 /**
  * Should be called to get a null rcl_action_goal_handle_t before passing to
@@ -204,7 +207,8 @@ rcl_action_goal_handle_get_status(
  * Lock-Free          | Yes
  *
  * \param[in] goal_handle struct containing the goal and metadata
- * \param[out] timestamp a preallocated struct where goal terminal timestamp is copied
+ * \param[out] timestamp a preallocated struct where goal terminal timestamp is copied.
+     Return `INVAILD_GOAL_TERMINAL_TIMESTAMP` if goal terminal timestamp isn't set.
  * \return `RCL_RET_OK` if the goal ID was accessed successfully, or
  * \return `RCL_RET_ACTION_GOAL_HANDLE_INVALID` if the goal handle is invalid, or
  * \return `RCL_RET_INVALID_ARGUMENT` if the timestamp argument is invalid
@@ -234,7 +238,7 @@ rcl_action_goal_handle_get_goal_terminal_timestamp(
  * \return `RCL_RET_ACTION_GOAL_HANDLE_INVALID` if the goal handle is invalid, or
  * \return `RCL_RET_INVALID_ARGUMENT` if the timestamp argument is invalid
  */
-RCL_ACTION_PUBLIC
+RCL_ACTION_LOCAL
 RCL_WARN_UNUSED
 rcl_ret_t
 rcl_action_goal_handle_set_goal_terminal_timestamp(
--- a/rcl_action/include/rcl_action/goal_handle.h
+++ b/rcl_action/include/rcl_action/goal_handle.h
@@ -211,7 +211,7 @@ rcl_action_goal_handle_get_status(
  * \return `RCL_RET_OK` if the goal ID was accessed successfully, or
  * \return `RCL_RET_ACTION_GOAL_HANDLE_INVALID` if the goal handle is invalid, or
  * \return `RCL_RET_INVALID_ARGUMENT` if the timestamp argument is invalid or
- * \return `RCL_RET_NOT_TERMINATED_YET` if the goal has not reached terminal state
+ * \return `RCL_ACTION_RET_NOT_TERMINATED_YET` if the goal has not reached terminal state
  */
 RCL_ACTION_PUBLIC
 RCL_WARN_UNUSED
--- a/rcl_action/src/rcl_action/action_server.c
+++ b/rcl_action/src/rcl_action/action_server.c
@@ -631,6 +631,7 @@ rcl_action_expire_goals(
   rcl_ret_t ret_final = RCL_RET_OK;
   const int64_t timeout = (int64_t)action_server->impl->options.result_timeout.nanoseconds;
   rcl_action_goal_handle_t * goal_handle;
+  rcl_action_goal_info_t goal_info;
   rcl_time_point_value_t goal_terminal_timestamp;
   size_t num_goal_handles = action_server->impl->num_goal_handles;
   for (size_t i = 0u; i < num_goal_handles; ++i) {
@@ -639,6 +640,19 @@ rcl_action_expire_goals(
       break;
     }
     goal_handle = action_server->impl->goal_handles[i];
+    // Expiration only applys to terminated goals
+    if (rcl_action_goal_handle_is_active(goal_handle)) {
+      continue;
+    }
+    rcl_action_goal_info_t * info_ptr = &goal_info;
+    if (output_expired) {
+      info_ptr = &(expired_goals[num_goals_expired]);
+    }
+    ret = rcl_action_goal_handle_get_info(goal_handle, info_ptr);
+    if (RCL_RET_OK != ret) {
+      ret_final = RCL_RET_ERROR;
+      continue;
+    }
 
     ret = rcl_action_goal_handle_get_goal_terminal_timestamp(goal_handle, &goal_terminal_timestamp);
     if (RCL_RET_NOT_TERMINATED_YET == ret) {
--- a/rcl_action/src/rcl_action/action_server.c
+++ b/rcl_action/src/rcl_action/action_server.c
@@ -631,7 +631,6 @@ rcl_action_expire_goals(
   rcl_ret_t ret_final = RCL_RET_OK;
   const int64_t timeout = (int64_t)action_server->impl->options.result_timeout.nanoseconds;
   rcl_action_goal_handle_t * goal_handle;
-  rcl_action_goal_info_t goal_info;
   rcl_time_point_value_t goal_terminal_timestamp;
   size_t num_goal_handles = action_server->impl->num_goal_handles;
   for (size_t i = 0u; i < num_goal_handles; ++i) {
@@ -640,19 +639,6 @@ rcl_action_expire_goals(
       break;
     }
     goal_handle = action_server->impl->goal_handles[i];
-    // Expiration only applys to terminated goals
-    if (rcl_action_goal_handle_is_active(goal_handle)) {
-      continue;
-    }
-    rcl_action_goal_info_t * info_ptr = &goal_info;
-    if (output_expired) {
-      info_ptr = &(expired_goals[num_goals_expired]);
-    }
-    ret = rcl_action_goal_handle_get_info(goal_handle, info_ptr);
-    if (RCL_RET_OK != ret) {
-      ret_final = RCL_RET_ERROR;
-      continue;
-    }
 
     ret = rcl_action_goal_handle_get_goal_terminal_timestamp(goal_handle, &goal_terminal_timestamp);
     if (RCL_RET_NOT_TERMINATED_YET == ret) {
--- a/rcl_action/src/rcl_action/action_server.c
+++ b/rcl_action/src/rcl_action/action_server.c
@@ -458,7 +458,7 @@ _recalculate_expire_timer(
       rcl_time_point_value_t goal_terminal_timestamp;
       ret = rcl_action_goal_handle_get_goal_terminal_timestamp(
         goal_handle, &goal_terminal_timestamp);
-      if (RCL_RET_NOT_TERMINATED_YET == ret) {
+      if (RCL_ACTION_RET_NOT_TERMINATED_YET == ret) {
         continue;
       }
       if (RCL_RET_OK != ret) {
@@ -654,7 +654,7 @@ rcl_action_expire_goals(
     }
 
     ret = rcl_action_goal_handle_get_goal_terminal_timestamp(goal_handle, &goal_terminal_timestamp);
-    if (RCL_RET_NOT_TERMINATED_YET == ret) {
+    if (RCL_ACTION_RET_NOT_TERMINATED_YET == ret) {
       continue;
     }
     if (RCL_RET_OK != ret) {
@@ -737,7 +737,7 @@ rcl_action_notify_goal_done(
       rcl_time_point_value_t goal_terminal_timestamp;
       rcl_ret_t ret = rcl_action_goal_handle_get_goal_terminal_timestamp(
         goal_handle, &goal_terminal_timestamp);
-      if (RCL_RET_NOT_TERMINATED_YET == ret) {
+      if (RCL_ACTION_RET_NOT_TERMINATED_YET == ret) {
         ret = rcl_action_goal_handle_set_goal_terminal_timestamp(goal_handle, current_time);
         if (RCL_RET_OK != ret) {
           return RCL_RET_ERROR;
--- a/rcl_action/src/rcl_action/action_server.c
+++ b/rcl_action/src/rcl_action/action_server.c
@@ -631,7 +631,6 @@ rcl_action_expire_goals(
   rcl_ret_t ret_final = RCL_RET_OK;
   const int64_t timeout = (int64_t)action_server->impl->options.result_timeout.nanoseconds;
   rcl_action_goal_handle_t * goal_handle;
-  rcl_action_goal_info_t goal_info;
   rcl_time_point_value_t goal_terminal_timestamp;
   size_t num_goal_handles = action_server->impl->num_goal_handles;
   for (size_t i = 0u; i < num_goal_handles; ++i) {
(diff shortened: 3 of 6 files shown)


Barry-Xu-2018:
This patch doesn't introduce new interfaces. As a result, the current modification is not very clear for setting goal terminal timestamp.  
To achieve a clearer implementation, I am considering adding a new interface to set the goal terminal timestamp for a specific goal handle. However, this would require modifications in both rclcpp and rclpy.     
Would you please share your opinions?

rcl_action/include/rcl_action/goal_handle.h
```diff
@@ -190,6 +191,56 @@ rcl_action_goal_handle_get_status(
...
+/**
+ * This is a non-blocking call.
+ *
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | No
```
fujitatomoya: ? I do not see any lock mechanism in there...?

```suggestion
 * Lock-Free          | Yes
```

rcl_action/include/rcl_action/goal_handle.h
```diff
@@ -200,7 +251,7 @@ rcl_action_goal_handle_get_status(
  * Allocates Memory   | No
  * Thread-Safe        | No
  * Uses Atomics       | No
- * Lock-Free          | Yes
+ * Lock-Free          | No
```
fujitatomoya: ? why this change needed? implementation has not been modified?

rcl_action/src/rcl_action/goal_handle.c
```diff
@@ -68,6 +69,8 @@ rcl_action_goal_handle_init(
   goal_handle->impl->state = GOAL_STATE_ACCEPTED;
   // Copy the allocator
   goal_handle->impl->allocator = allocator;
+  // Set invaild time
```
SteveMacenski:
```suggestion
  // Set invalid time
```

SteveMacenski (commented):
How is `goal_handle->impl->goal_terminal_timestamp` set initially?

If my understanding of your code is correct (I don't live in `rcl`) when calling `_recalculate_expire_timer`, which I assume is periodically called during an action server's normal operations, we reset to `now` the clock ticking down when its expired. 

I don't understand the `goal_terminal_timestamp == INT64_MAX` condition then --> shouldn't we _always_ be resetting the clock if the action is still active? That would make the most sense to me; while the goal is still actively being processed, we stave off the timeout to remove it by another full timeout length. Then when the last update / sending of the result comes so we don't call it anymore, the full countdown is considered before removal.

rcl_action/src/rcl_action/goal_handle.c
```diff
@@ -68,6 +69,8 @@ rcl_action_goal_handle_init(
...
+  goal_handle->impl->goal_terminal_timestamp = INT64_MAX;
```
fujitatomoya: I think `INT64_MAX` can be still valid time? how about using `0` for initialized value instead?

```suggestion
  // Set invaild time
  goal_handle->impl->goal_terminal_timestamp = 0;
```

fujitatomoya (commented):
i have a few comments to confirm.

Barry-Xu-2018:
@SteveMacenski 

> How is `goal_handle->impl->goal_terminal_timestamp` set initially?
> 
> If my understanding of your code is correct (I don't live in `rcl`) when calling `_recalculate_expire_timer`, which I assume is periodically called during an action server's normal operations, we reset to `now` the clock ticking down when its expired.
> 
> I don't understand the `goal_terminal_timestamp == INT64_MAX` condition then --> shouldn't we _always_ be resetting the clock if the action is still active? That would make the most sense to me; while the goal is still actively being processed, we stave off the timeout to remove it by another full timeout length. Then when the last update / sending of the result comes so we don't call it anymore, the full countdown is considered before removal.

First, let me explain when _recalculate_expire_timer() is called.   
Currently, there are two scenarios in which this function is called.

1. While goal state is changed to terminal state (`success` or `cancel` or `abort`), rcl_action_notify_goal_done() will be called. And _recalculate_expire_timer() will be called in rcl_action_notify_goal_done() for setting a timer for the next expiring goal.  

2. While timer for goal expire is triggered, rcl_action_expire_goals() will be called. And _recalculate_expire_timer() will be called for setting a timer for the next expiring goal.

I use scenario 1 to set goal_terminal_timestamp.    
In order to avoid scenario 2, the condition to set goal_terminal_timestamp include 
- The state of goal should be one of `success` or `cancel` and `abort`.
 
    So the code to set goal_terminal_timestamp should be in this condition. This can avoid wrong setting goal_terminal_timestamp in scenario 2. 
    
https://github.com/ros2/rcl/blob/a17bc95988910ef372c84b311fb30cac6accd91e/rcl_action/src/rcl_action/action_server.c#L447
  
- goal_terminal_timestamp isn't set before. 

    I want to use INT64_MAX as invalid value (unset state). 
    So, only if goal_terminal_timestamp is INT64_MAX, goal_terminal_timestamp can be updated to current time.
    But invalid value is unsuitable (From Fujita-san' review comment). I will change it to 0.

rcl_action/include/rcl_action/goal_handle.h
Barry-Xu-2018: It is a mistake to change.

rcl_action/src/rcl_action/goal_handle.c
Barry-Xu-2018: Yes.

rcl_action/src/rcl_action/goal_handle.c
Barry-Xu-2018: Updated

rcl_action/include/rcl_action/goal_handle.h
Barry-Xu-2018: Updated

fujitatomoya (approved):
lgtm, i will try CI.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19950/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14422/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20656/)

SteveMacenski:
> setting a timer for the next expiring goal.

In this case, why is there any timer still set for expiring the goals? That seems like the root cause of the problem; our goals are being expired before they're done due to a terminal condition if they're long-running. 

In this design, that seems to me should be completely removed and this check in terminal states (success or cancel or abort) is all that should be required. This is where the timer to count down the expiration of the goal handle should be initiated.

Alternatively, the timer for expiring the next goal should be adjusted during action server operations from `checking` to `resetting` the clock to the full timeout. That way the timeout would only trigger when its become stale. I think the first way is better though. That removes alot of unnecessary complication that would be removed with starting the countdown where you've made changes alone. God knows simplifying actions would reduce the likelihood of future issues

fujitatomoya:
@Barry-Xu-2018 can you rebase your development branch? it is failing build https://ci.ros2.org/job/ci_linux/19950/console since it does not include some changes in the mainline. https://github.com/Barry-Xu-2018/rcl/compare/review/topic-use-goal-terminal-timestamp-to-expire-check...ros2%3Arcl%3Arolling

Barry-Xu-2018:
> can you rebase your development branch? it is failing build https://ci.ros2.org/job/ci_linux/19950/console since it does not include some changes in the mainline. https://github.com/Barry-Xu-2018/rcl/compare/review/topic-use-goal-terminal-timestamp-to-expire-check...ros2:rcl:rolling

Rebase was done.  @fujitatomoya

Barry-Xu-2018:
@SteveMacenski 

> > setting a timer for the next expiring goal.
> 
> In this case, why is there any timer still set for expiring the goals? That seems like the root cause of the problem; our goals are being expired before they're done due to a terminal condition if they're long-running.

Sorry, I'm not quite clear about the scenarios you described. Could you please describe them again?
The setting of the expire timer begins when a goal enters the terminal state. It is not related to how long the goal has been running.

https://github.com/ros2/rcl/pull/1121#issuecomment-1844405572 
In the two scenarios described in my comment, it may be necessary to reset the expire timer:

- When the state of a goal enters the terminal state, if the expire timer has not started yet (e.g., for the first terminal goal, or if the previous expire timer was triggered earlier but there was no subsequent expire time, so the timer was canceled), then the expire timer should be started. If the expire timer already started, nothing need to be done.  

- When the expire timer is triggered, at this point, we should calculate the nearest upcoming expire time and set the expire timer accordingly.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19961/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14431/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20667/)

Barry-Xu-2018:
There is a failure in windows test.
projectroot.src.core.ddsc.tests.CUnit_ddsc_whc_check_end_state  
After checking the log, it seems that the failure is unrelated to this PR.

fujitatomoya:
@sloretz @iuhilnehc-ynos either of you, can you take a look at this? i would like to have another review on this.

SteveMacenski:
Your two situations this method is called: 

> While goal state is changed to terminal state (success or cancel or abort), rcl_action_notify_goal_done() will be called. And _recalculate_expire_timer() will be called in rcl_action_notify_goal_done() for setting a timer for the next expiring goal.

I'll shorthand this as achieving a terminal state

> While timer for goal expire is triggered, rcl_action_expire_goals() will be called. And _recalculate_expire_timer() will be called for setting a timer for the next expiring goal.

I'll shorthand the timer for goal expiring as "the ticking clock"

---

My assertion is that you found a clever place to make modifications to the goal handle's state: at achieving the terminal state. We know that once we achieve this state, the goal is done processing and anytime before this point, the goal is still processing (or was dropped on the floor from a users' crappy application, which no matter what is going to cause problems). 

If we have a place (e.g. _here_) we can modify a ticking clock, that calls into question the need for the ticking clock prior to this point at all. The issue we want to solve is that ticking clock expiring still-processing goals -- thus why even _have_ the ticking clock have the _ability_ to expire still-processing goals? I'd argue that it shouldn't even be possible. We should be waiting until the terminal state to trigger the clock to start its countdown until the handle / result are deleted. Of which, the default of 10s or what-have-you is totally appropriate and fully solves the problem.

It also comes with the added benefit of simplifying the action logic since there (1) doesn't need to be a ticking clock checked while the goal is processing and (2) the modification you require don't need to reset it, it merely needs to start it

Barry-Xu-2018:
@SteveMacenski 

Thank you for the further explanation.

> If we have a place (e.g. _here_) we can modify a ticking clock, that calls into question the need for the ticking clock prior to this point at all. The issue we want to solve is that ticking clock expiring still-processing goals -- thus why even _have_ the ticking clock have the _ability_ to expire still-processing goals? I'd argue that it shouldn't even be possible. We should be waiting until the terminal state to trigger the clock to start its countdown until the handle / result are deleted. Of which, the default of 10s or what-have-you is totally appropriate and fully solves the problem.
> 
> It also comes with the added benefit of simplifying the action logic since there (1) doesn't need to be a ticking clock checked while the goal is processing and (2) the modification you require don't need to reset it, it merely needs to start it

Your current question revolves around "the ticking clock", but I haven't fully caught your idea yet. 
To clarify, I use an example for the following discussion.  

image

For "the ticking clock", while the timer is triggered (At 2 in above diagram), the next expire time need to be calculated.   

> It also comes with the added benefit of simplifying the action logic since there (1) doesn't need to be a ticking clock checked while the goal is processing and (2) the modification you require don't need to reset it, it merely needs to start it

There are currently two time points where "the ticking clock" is checked. One is when a goal enters the termination state, and the other is when the ticking clock timer is triggered. When you mentioned "the goal is processing," in what situation does that occur?

If expire timer has been run, a goal enters terminal state to call _recalculate_expire_timer(). At this case, calculating and setting the timer is an unnecessary operation.   
While the expire timer is triggered (At 2 in above diagram), I think calculating and setting the timer is a necessary operation.

SteveMacenski:
> When you mentioned "the goal is processing," in what situation does that occur?

That's a question you're most suited to answer than me. With the current behavior before this PR, where was the goal being expired resulting in the bug that this is meant to address (which was expiring still-processing goals causing issues)? That was happening which triggered this discussion. My suggestion is removing entirely where-ever that is happening if we're able to start the count down to removing a goal at a terminal state. You've shown the part of the code where that terminal state is found to be able to reset a timer. Instead, we could just start it here and remove the previous problematic countdown that previously caused us issues.

I ack that perhaps I'm missing something and that my suggestion isn't fully informed with the ins-and-outs of rcl. What I'm trying to suggest at its core is this, regardless of the specific details:
- We previously had a problem where goals were expiring mid-execution and we weren't getting the results back because they were no longer in existence. That required us in Nav2 to take somewhat drastic measures to [expose this not-well-documented timeout](https://github.com/ros-planning/navigation2/blob/main/nav2_controller/src/controller_server.cpp#L220-L224) to users as a temporary work around.
- You've identified a part of the code where we know the goal has just reached its terminal state
- Thus: my suggestion is to completely rip out the existing logic that caused the bug in the first place with a timeout existing at all during normal goal processing -- and replace it with logic in the goal-terminal-state-reached section of the code to start the timeout clock before expiring the goal's information. 

To my eye, this PR resets some timeouts, but still leaves the original timeout in place that caused the problem in the past - which could still cause a problem moving forward as well if you have a timeout of 10s but an action length of 10 minutes (which, we have).

Barry-Xu-2018:
@SteveMacenski 

Thank you for detailed explanation.  
I understand the concerns you mentioned.  
Before this patch, when a goal handle is created, this time point is already recorded as the starting time for expire calculation. Of course, the expire timer is not set at this time. When the goal enters the terminal state, this time point is used to calculate and set the time on the expire timer.  In this scenario, if the time it takes to process the goal is longer than the set timeout, when the goal enters the terminal state, the result will not be retained for the timeout duration; instead, it will be immediately removed.

This PR ensures that the current time is recorded as the start time for expiration only when the goal enters the terminal state. The expire timer is set based on this time. Therefore, with the current modifications, it is impossible for the goal to expire while being processed. In rcl_action_expire_goals(), it will get time of goal termination by calling rcl_action_goal_handle_get_goal_terminal_timestamp() to check if expiration occurs.

SteveMacenski:
OK, got it. That makes sense, you can ignore my comments then :-) This is a very important fix for us in Nav2 so I wanted to make sure beyond the shadow of my ignorance that this was going to fully resolve the problem :wink: I very much appreciate your time here!

fujitatomoya:
@Barry-Xu-2018 since you added the fix from last CI, i started the full CI.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19967/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14437/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20677/)

@SteveMacenski thank you for taking your time for the review!

fujitatomoya:
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14442/)

SteveMacenski:
What's the good word?

fujitatomoya:
we would like to have 2nd review on this.

CC: @wjwwood @clalancette @alsora @mjcarroll

Bjoernolav:
Any hope for getting this merged soon? A fix for this is highly appreciated.

fujitatomoya:
@Barry-Xu-2018 can you rebase this?

Barry-Xu-2018:
@fujitatomoya Rebase was done.

fujitatomoya:
@alsora @wjwwood @mjcarroll we would like to request the review from either of you before CI.

Barry-Xu-2018:
Rebase was done.

fujitatomoya:
@alsora @wjwwood @clalancette @mjcarroll  i think this is one of the major issue for action, sadly we missed this for jazzy release. could you take a look at this?

alsora (approved):

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21288/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15606/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/787/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22024/)

alsora:
New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21308/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15624/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/807/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22040/)

Barry-Xu-2018:
Re-run Linux-rhel

- Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/817/)

alsora:
It looks like Linux-rhel is broken across the board, see also https://github.com/ros2/rclcpp/pull/2564#issuecomment-2184136893

fujitatomoya:
https://ci.ros2.org/job/ci_windows/22040/testReport/ is unrelated failure.

fujitatomoya:
Re-run Linux-rhel, maybe we just got unlucky.

- Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/821/)

rcl_action/src/rcl_action/action_server.c
```diff
@@ -451,13 +451,28 @@ _recalculate_expire_timer(
...
+      // If goal_terminal_timestamp of goal is invaild, this goal is just in terminal state.
+      // So current time is set to goal_terminal_timestamp of this goal.
+      if (goal_terminal_timestamp == 0) {
+        goal_terminal_timestamp = current_time;
+        ret = rcl_action_goal_handle_set_goal_terminal_timestamp(
+          goal_handle, goal_terminal_timestamp);
+        if (RCL_RET_OK != ret) {
+          return RCL_RET_ERROR;
+        }
+      }
```
jmachowinski: This block should not be in the rcl_action_expire_goals function. I would move it into rcl_action_notify_goal_done function.

rcl_action/include/rcl_action/goal_handle.h
```diff
@@ -190,6 +191,56 @@ rcl_action_goal_handle_get_status(
...
+ * \param[in] goal_handle struct containing the goal and metadata
+ * \param[in] timestamp The timestamp of goal termination
+ * \return `RCL_RET_OK` if the goal ID was accessed successfully, or
+ * \return `RCL_RET_ACTION_GOAL_HANDLE_INVALID` if the goal handle is invalid, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if the timestamp argument is invalid
+ */
+RCL_ACTION_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_action_goal_handle_set_goal_terminal_timestamp(
```
jmachowinski: This is an internal thing, do we really want to expose an API, to modify this time ?
From my point of view, this value should only be modified by the internal logic.

rcl_action/src/rcl_action/goal_handle.c
```diff
@@ -68,6 +69,8 @@ rcl_action_goal_handle_init(
   goal_handle->impl->state = GOAL_STATE_ACCEPTED;
   // Copy the allocator
   goal_handle->impl->allocator = allocator;
+  // Set invalid time
+  goal_handle->impl->goal_terminal_timestamp = 0;
```
jmachowinski: 0 is technically a valid time, I guess -1 or some define INVALID_TIME_POINT would be nicer.

rcl_action/include/rcl_action/goal_handle.h
```diff
@@ -190,6 +191,56 @@ rcl_action_goal_handle_get_status(
...
+ * <hr>
+ * Attribute          | Adherence
+ * ------------------ | -------------
+ * Allocates Memory   | No
+ * Thread-Safe        | No
+ * Uses Atomics       | No
+ * Lock-Free          | Yes
+ *
+ * \param[in] goal_handle struct containing the goal and metadata
+ * \param[out] timestamp a preallocated struct where goal terminal timestamp is copied
```
jmachowinski: It should be mentioned, that this value can be not set yet. And that in this case the returned value is set to some magic value.

Hm, perhaps make this an explicit return code ?

rcl_action/src/rcl_action/goal_handle.c
```diff
@@ -26,6 +26,7 @@ typedef struct rcl_action_goal_handle_impl_s
 {
   rcl_action_goal_info_t info;
   rcl_action_goal_state_t state;
+  rcl_time_point_value_t goal_terminal_timestamp;
```
jmachowinski, line 31: There are some magic values that may be set, to indicate, that the value has not been set yet, this should be documented here,

jmachowinski (changes requested):

rcl_action/src/rcl_action/action_server.c
Barry-Xu-2018: Sorry, I didn't quite understand what you meant.   
Do you mean ?
```c++
             if (RCL_RET_OK != ret) {
                 return ret;
             }
```

rcl_action/include/rcl_action/goal_handle.h
Barry-Xu-2018: Yes, it's currently for internal use only.

rcl_action/src/rcl_action/goal_handle.c
Barry-Xu-2018: This value comes from the current clock time, so it can't be 0.  
Using macro definitions for unavailable values makes the code easier to understand.  
I want to use INVAILD_GOAL_TERMINAL_TIMESTAMP

rcl_action/include/rcl_action/goal_handle.h
Barry-Xu-2018: Yes. I will update the description.

rcl_action/src/rcl_action/goal_handle.c
Barry-Xu-2018: Okay.

Barry-Xu-2018:
@jmachowinski  

Thanks for your review comments. I have addressed them. Please review again.

rcl_action/include/rcl_action/goal_handle.h
jmachowinski: I guess then we should remove this from the public headers...

rcl_action/src/rcl_action/action_server.c
jmachowinski: I mean this block : 
```c++
 if (goal_terminal_timestamp == 0) {
        goal_terminal_timestamp = current_time;
        ret = rcl_action_goal_handle_set_goal_terminal_timestamp(
          goal_handle, goal_terminal_timestamp);
        if (RCL_RET_OK != ret) {
          return RCL_RET_ERROR;
        }
      }
``` 
This one should only be executed, if a goal finishes. Having it in the expire function makes the code flow harder to read, and may have unintended side effects.

rcl_action/src/rcl_action/goal_handle.c
```diff
@@ -26,6 +26,8 @@ typedef struct rcl_action_goal_handle_impl_s
 {
   rcl_action_goal_info_t info;
   rcl_action_goal_state_t state;
+  // If goal_terminal_timestamp isn't set, INVAILD_GOAL_TERMINAL_TIMESTAMP is the initial value.
```
jmachowinski:
```suggestion
  // As long as the goal has not reached terminal state, this variable is set to  
  // INVAILD_GOAL_TERMINAL_TIMESTAMP. 
```

rcl_action/include/rcl_action/goal_handle.h
```diff
@@ -190,6 +194,57 @@ rcl_action_goal_handle_get_status(
...
+ * \return `RCL_RET_OK` if the goal ID was accessed successfully, or
+ * \return `RCL_RET_ACTION_GOAL_HANDLE_INVALID` if the goal handle is invalid, or
+ * \return `RCL_RET_INVALID_ARGUMENT` if the timestamp argument is invalid
+ */
+RCL_ACTION_PUBLIC
+RCL_WARN_UNUSED
+rcl_ret_t
+rcl_action_goal_handle_get_goal_terminal_timestamp(
+  const rcl_action_goal_handle_t * goal_handle,
+  rcl_time_point_value_t * timestamp);
```
jmachowinski, line 221: I still undecided on the API here. 
You get RCL_RET_OK returned for a value, that may be invalid. 
What about adding something like RCL_RET_NOT_TERMINATED_YET ?

If the rest is fine withe the API, then we can leave it as it. @fujitatomoya  @alsora  your thoughts ?

rcl_action/src/rcl_action/goal_handle.c
jmachowinski: > This value comes from the current clock time, so it can't be 0. 

I am not sure what kind of clock is used here, but if it is a RCL_ROS_TIME clock, 0 is a valid value for the current time.

Barry-Xu-2018:
Rebase was done.

Barry-Xu-2018:
Friendly ping @jmachowinski 
Please review again.

jmachowinski (changes requested):

jmachowinski:
damn, I reviewed this last week, but did not press submit...

rcl_action/include/rcl_action/goal_handle.h
Barry-Xu-2018: Since there isn't a header file specifically for internal use, I just changed the declaration. This file is meant for external use, and this function doesn't really belong here. I'll create a new header file to place this internal function instead.

rcl_action/src/rcl_action/action_server.c
Barry-Xu-2018: > This one should only be executed, if a goal finishes. Having it in the expire function makes the code flow harder to read, and may have unintended side effects.

Due to performance considerations, I didn't put the code into rcl_action_notify_goal_done(). If it were placed in rcl_action_notify_goal_done(), it would require looping through all the goal_handles in both rcl_action_notify_goal_done() and _recalculate_expire_timer(), and rcl_action_goal_handle_get_goal_terminal_timestamp() would be called twice.  

Of course, this does impact the readability of the code.

jmachowinski: Goals do no finish in high frequencies, therefore I would go for readability over performance.

Also, from my point of view, this should NOT be executed in _recalculate_expire_timer().
This code sets the timestamp of the goal termination, why would you want to do this in the expire timer ?

rcl_action/include/rcl_action/goal_handle.h
jmachowinski: I would just move the function declaration into the head of the .cpp file...

Barry-Xu-2018: > I would just move the function declaration into the head of the .cpp file...

Thanks.  This is a simpler way.

rcl_action/src/rcl_action/action_server.c
Barry-Xu-2018: > Goals do no finish in high frequencies, therefore I would go for readability over performance.

Yes.

> Also, from my point of view, this should NOT be executed in _recalculate_expire_timer().
> This code sets the timestamp of the goal termination, why would you want to do this in the expire timer ?

I put it in _recalculate_expire_timer because I considered performance issues. I agree with what you said above, so I'll move this part of the code to rcl_action_notify_goal_done().

rcl_action/src/rcl_action/goal_handle.c
Barry-Xu-2018: After checking rcl_get_ros_time(), rcl_get_steady_time() and rcl_get_system_time(), they should not return negative value.  So it is better to set INVAILD_GOAL_TERMINAL_TIMESTAMP as -1.

rcl_action/include/rcl_action/goal_handle.h
alsora: +1 on adding a dedicated error code.

Barry-Xu-2018:
@jmachowinski I have updated code. Please review again.

rcl_action/src/rcl_action/action_server.c
```diff
@@ -451,13 +455,14 @@ _recalculate_expire_timer(
     if (!rcl_action_goal_handle_is_active(goal_handle)) {
       ++num_inactive_goals;
 
-      rcl_action_goal_info_t goal_info;
-      ret = rcl_action_goal_handle_get_info(goal_handle, &goal_info);
+      rcl_time_point_value_t goal_terminal_timestamp;
+      ret = rcl_action_goal_handle_get_goal_terminal_timestamp(
+        goal_handle, &goal_terminal_timestamp);
       if (RCL_RET_OK != ret) {
```
jmachowinski, line 464:
```suggestion
      if (RCL_RET_NOT_TERMINATED_YET == ret) {
      continue;
      }
      if (RCL_RET_OK != ret) {
```

rcl_action/src/rcl_action/action_server.c
```diff
@@ -451,13 +455,14 @@ _recalculate_expire_timer(
     if (!rcl_action_goal_handle_is_active(goal_handle)) {
```
jmachowinski, line 455: I guess we can replace this now by rcl_action_goal_handle_get_goal_terminal_timestamp ?

rcl_action/src/rcl_action/action_server.c
```diff
@@ -645,8 +650,14 @@ rcl_action_expire_goals(
       ret_final = RCL_RET_ERROR;
       continue;
     }
-    goal_time = _goal_info_stamp_to_nanosec(info_ptr);
-    if ((current_time - goal_time) > timeout) {
+
+    ret = rcl_action_goal_handle_get_goal_terminal_timestamp(goal_handle, &goal_terminal_timestamp);
+    if (RCL_RET_OK != ret) {
```
jmachowinski, line 660:
```suggestion
    if (RCL_RET_NOT_TERMINATED_YET == ret)
    {
    continues;
    }
    if (RCL_RET_OK != ret) {
```

rcl_action/src/rcl_action/action_server.c
```diff
@@ -645,8 +650,14 @@ rcl_action_expire_goals(
...
+      ret_final = RCL_RET_ERROR;
+      continue;
+    }
+
+    if ((goal_terminal_timestamp != 0) && (current_time - goal_terminal_timestamp) > timeout) {
```
jmachowinski:
```suggestion
    if ( (current_time - goal_terminal_timestamp) > timeout) {
```

rcl_action/src/rcl_action/action_server.c
```diff
@@ -645,8 +650,14 @@ rcl_action_expire_goals(
       ret_final = RCL_RET_ERROR;
       continue;
```
jmachowinski: 640-651 can be deleted ?

jmachowinski (changes requested):

rcl_action/src/rcl_action/action_server.c
```diff
@@ -706,6 +717,34 @@ rcl_action_notify_goal_done(
...
+  }
+
+  // Set current time to goal_terminal_timestamp of goal which has reached terminal state
+  for (size_t i = 0; i < action_server->impl->num_goal_handles; ++i) {
+    rcl_action_goal_handle_t * goal_handle = action_server->impl->goal_handles[i];
+    if (!rcl_action_goal_handle_is_active(goal_handle)) {
+      rcl_time_point_value_t goal_terminal_timestamp;
+      rcl_ret_t ret = rcl_action_goal_handle_get_goal_terminal_timestamp(
+        goal_handle, &goal_terminal_timestamp);
+      if (RCL_RET_OK != ret) {
```
jmachowinski:
```suggestion
      if (RCL_RET_NOT_TERMINATED_YET != ret) {
        ret = rcl_action_goal_handle_set_goal_terminal_timestamp(goal_handle, current_time);
        if (RCL_RET_OK != ret) {
          return RCL_RET_ERROR;
        }
        continues;
      }
      if (RCL_RET_OK != ret) {
        return RCL_RET_ERROR;
      }
```

jmachowinski:
```suggestion
      if (RCL_RET_NOT_TERMINATED_YET != ret) {  
        ret = rcl_action_goal_handle_set_goal_terminal_timestamp(goal_handle, current_time);  
        if (RCL_RET_OK != ret) {  
          return RCL_RET_ERROR;  
        }  
        continues;  
      }  
      if (RCL_RET_OK != ret) {  
        return RCL_RET_ERROR;  
      }  
```

rcl_action/src/rcl_action/action_server.c
Barry-Xu-2018: Yes. It is better to use rcl_action_goal_handle_get_goal_terminal_timestamp().

rcl_action/src/rcl_action/action_server.c
Barry-Xu-2018: Do you means lines related to ` goal_time = _goal_info_stamp_to_nanosec(info_ptr);` ?

jmachowinski: the check if the goal is active, and the code around the info_ptr should now be not needed any more.

Barry-Xu-2018: Yes. These code can be deleted. 
I will update code. Thanks

Barry-Xu-2018:
@jmachowinski Thanks a lot for your review. I've updated the code based on your comments.

jmachowinski (approved):

fujitatomoya:
@jmachowinski @alsora thanks for your reviews, i will start the CI.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21368/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15675/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/878/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22104/)

Barry-Xu-2018:
I will check failed test cases.

Barry-Xu-2018:
Commit [a07e119](https://github.com/ros2/rcl/pull/1121/commits/a07e119073f1d94413c8840e40a2ab069e73477f) is wrong.   
Removed code is useful. It filled goal info for `expired_goals`.    
I will do more testing before running the CI.

Barry-Xu-2018:
CI:  
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21375/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15681/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/884/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22110/)

alsora:
@Barry-Xu-2018 please consider adding some comments around that code that we thought was not needed but broke the tests when removed, so that in the future we will immediately know why we need it.
Besides that, the PR looks good to me.

Barry-Xu-2018:
@alsora I added comments and optimized the code.  
Please take a look [03555f5](https://github.com/ros2/rcl/pull/1121/commits/03555f55f6c101b3c14b899434a9c494a02f407e).

rcl/include/rcl/types.h
```diff
@@ -126,6 +126,10 @@ typedef rmw_ret_t rcl_ret_t;
 /// rcl_lifecycle state not registered
 #define RCL_RET_LIFECYCLE_STATE_NOT_REGISTERED 3001
 
+// rcl action specific ret codes in 40XX
+/// No terminal timestamp for the goal as it has not reached a terminal state.
+#define RCL_RET_NOT_TERMINATED_YET 4001
```
fujitatomoya: i would change the name into the following if that is the dedicated ret code for `rcl_action`.

```suggestion
#define RCL_ACTION_RET_NOT_TERMINATED_YET 4001
```

rcl_action/src/rcl_action/goal_handle.c
```diff
@@ -172,6 +177,48 @@ rcl_action_goal_handle_is_valid(const rcl_action_goal_handle_t * goal_handle)
...
+  rcl_time_point_value_t timestamp)
+{
+  RCUTILS_CAN_RETURN_WITH_ERROR_OF(RCL_RET_ACTION_GOAL_HANDLE_INVALID);
+  RCUTILS_CAN_RETURN_WITH_ERROR_OF(RCL_RET_INVALID_ARGUMENT);
+
+  if (!rcl_action_goal_handle_is_valid(goal_handle)) {
+    return RCL_RET_ACTION_GOAL_HANDLE_INVALID;  // error message is set
+  }
+
+  if (timestamp == INVAILD_GOAL_TERMINAL_TIMESTAMP) {
```
fujitatomoya: Is this supposed to take any negative value except `-1`? i believe all negative values should be invalid argument to set.

```suggestion
  if (timestamp <= INVAILD_GOAL_TERMINAL_TIMESTAMP) {
```

rcl_action/src/rcl_action/goal_handle.c
```diff
@@ -172,6 +177,48 @@ rcl_action_goal_handle_is_valid(const rcl_action_goal_handle_t * goal_handle)
...
+  RCUTILS_CAN_RETURN_WITH_ERROR_OF(RCL_RET_INVALID_ARGUMENT);
+
+  if (!rcl_action_goal_handle_is_valid(goal_handle)) {
+    return RCL_RET_ACTION_GOAL_HANDLE_INVALID;  // error message is set
+  }
+  RCL_CHECK_ARGUMENT_FOR_NULL(timestamp, RCL_RET_INVALID_ARGUMENT);
+
+  if (goal_handle->impl->goal_terminal_timestamp == INVAILD_GOAL_TERMINAL_TIMESTAMP) {
+    return RCL_RET_NOT_TERMINATED_YET;
+  }
```
fujitatomoya: I think the precise way is to use `rcl_action_goal_handle_get_status` instead to check if the state is appropriate condition to call this function?

rcl_action/src/rcl_action/goal_handle.c
Barry-Xu-2018: Yes. Negative value is invalid.  
Although I don't think this value will be set to a negative number, I am fine to change this.

rcl/include/rcl/types.h
Barry-Xu-2018: Updated

rcl_action/src/rcl_action/goal_handle.c
Barry-Xu-2018: > I think the precise way is to use rcl_action_goal_handle_get_status instead to check if the state is appropriate condition to call this function?

Yes. There's a very short gap here.
First, set goal to terminal status, and then `goal_terminal_timestamp` will be set in rcl_action_notify_goal_done().
Even if rcl_action_goal_handle_get_status() is used, we have to check `goal_handle->impl->goal_terminal_timestamp == INVAILD_GOAL_TERMINAL_TIMESTAMP`.

fujitatomoya:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21507/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15789/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1003/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22223/)

Barry-Xu-2018:
The all unstable cases are related to this commit https://github.com/ros2/message_filters/commit/55515183fd38411ed57021217ee071673f5551e8  
It introduces warning macro in header files.  But test cases all pass.

For Windows, it is related to this PR.
This change lead to the below error. It is failed to build point_cloud_transport on Windows platform.
```
07:34:22 C:/ci/ws/install/include/message_filters\message_filters/simple_filter.h(18,1): fatal error C1021: invalid preprocessor command 'warning' (compiling source file C:\ci\ws\src\ros-perception\point_cloud_transport\point_cloud_transport\src\subscriber_filter.cpp) [C:\ci\ws\build\point_cloud_transport\point_cloud_transport.vcxproj]
```

Barry-Xu-2018:
Re-CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21538/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15822/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1033/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22255/)

Barry-Xu-2018:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22276/)

Barry-Xu-2018:
All CI tests passed.

SteveMacenski:
Will this be backported to Jazzy/Humble?

fujitatomoya:
I believe adding new function in C does not break ABI compatibility, so i say we can backport this to jazzy, iron and humble.

@clalancette @alsora what do you think?

SteveMacenski:
This is really great news, I know there's alot of interest in Nav2 users for the backport to remove that timeout! From my review as well, it looks backportable

SteveMacenski:
@Barry-Xu-2018 @fujitatomoya  can this be backported to Jazzy?

fujitatomoya:
This fix is API/ABI compatible, i think we can backport this to jazzy. (technically for humble too, but maybe code base could be different.)

fujitatomoya:
@Mergifyio backport jazzy

PR #1115, opened 2023-11-03
a rosout publisher of a node might not exist

iuhilnehc-ynos:
- [ ] https://github.com/ros2/rcl/pull/1115
- [ ] https://github.com/ros2/rclpy/pull/1196
- [ ] https://github.com/ros2/rclcpp/pull/2357

to fix https://github.com/ros2/rclpy/issues/1195

PR #1122, opened 2023-11-29
remove static function rcl_ret_from_rcutils_ret().

fujitatomoya:
refactoring to get rid of redundancy for code maintenance.

### DIFF
--- a/rcl/src/rcl/logging_rosout.c
+++ b/rcl/src/rcl/logging_rosout.c
@@ -33,41 +33,6 @@
 
 #define ROSOUT_TOPIC_NAME "/rosout"
 
-static rcl_ret_t rcl_ret_from_rcutils_ret(rcutils_ret_t rcutils_ret)
-{
-  rcl_ret_t rcl_ret_var;
-
-  if (RCUTILS_RET_OK != rcutils_ret) {
-    if (rcutils_error_is_set()) {
-      RCL_SET_ERROR_MSG(rcutils_get_error_string().str);
-    } else {
-      RCL_SET_ERROR_MSG_WITH_FORMAT_STRING("rcutils_ret_t code: %i", rcutils_ret);
-    }
-  }
-
-  switch (rcutils_ret) {
-    case RCUTILS_RET_OK:
-      rcl_ret_var = RCL_RET_OK;
-      break;
-    case RCUTILS_RET_ERROR:
-      rcl_ret_var = RCL_RET_ERROR;
-      break;
-    case RCUTILS_RET_BAD_ALLOC:
-      rcl_ret_var = RCL_RET_BAD_ALLOC;
-      break;
-    case RCUTILS_RET_INVALID_ARGUMENT:
-      rcl_ret_var = RCL_RET_INVALID_ARGUMENT;
-      break;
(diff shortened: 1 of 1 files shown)


rcl/src/rcl/logging_rosout.c
```diff
@@ -33,41 +33,6 @@
...
-    case RCUTILS_RET_BAD_ALLOC:
-      rcl_ret_var = RCL_RET_BAD_ALLOC;
-      break;
-    case RCUTILS_RET_INVALID_ARGUMENT:
-      rcl_ret_var = RCL_RET_INVALID_ARGUMENT;
-      break;
-    case RCUTILS_RET_NOT_INITIALIZED:
-      rcl_ret_var = RCL_RET_NOT_INIT;
-      break;
-    default:
```
fujitatomoya, line 64: `RCUTILS_RET_NOT_FOUND` maps to `RCL_RET_NOT_FOUND` is missing.

rcl/src/rcl/logging_rosout.c
```diff
@@ -33,41 +33,6 @@
...
-{
-  rcl_ret_t rcl_ret_var;
-
-  if (RCUTILS_RET_OK != rcutils_ret) {
-    if (rcutils_error_is_set()) {
-      RCL_SET_ERROR_MSG(rcutils_get_error_string().str);
-    } else {
-      RCL_SET_ERROR_MSG_WITH_FORMAT_STRING("rcutils_ret_t code: %i", rcutils_ret);
-    }
-  }
```
fujitatomoya, line 46: i am not sure why we need this. if needed, we can move the same logic into `rcl_convert_rcutils_ret_to_rcl_ret`?

fujitatomoya (commented):
@iuhilnehc-ynos can you take a look at this?

iuhilnehc-ynos:
To reduce the duplication seems good to me. I think that using the common API might be the correct way, which was the reason I used `rcl_convert_rcutils_ret_to_rcl_ret` instead of `rcl_ret_from_rcutils_ret` in https://github.com/ros2/rcl/pull/1115.

To set the error message explictly if needed is a better way because we would like to keep the same consistent behavior in rcl.

iuhilnehc-ynos (approved):
LGTM with green CI.

rcl/src/rcl/logging_rosout.c
fujitatomoya: > To set the error message explictly if needed is a better way because we would like to keep the same consistent behavior in rcl.

i was wondering if this message is really useful for user. adding message here is mainly to add the `rcutils`'s return code in the error message. but that would not be useful for the user, since that does not tell why, and there should have been error message in the buffer in the previous `rcutils`'s function call. i am inclined to make it simple and fast to return the converted code to the user instead of that.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19901/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14384/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20608/)

PR #1126, opened 2023-12-09
Fix for incorrect integer value conversion on Windows

### LINKED ISSUE #1125: rcl_yaml_param_parser incorrectly parsing integer values bigger than LONG_MAX = 2147483647
## Bug report
`rcl_yaml_param_parser` incorrectly parsing integer values bigger than LONG_MAX = 2147483647
- Relates https://github.com/ros2/rosbag2/pull/1419

**Required Info:**

- Operating System:  Windows 10 
- Installation type:  from source 
- Version or commit hash: rolling
- DDS implementation: N/A
- Client library (if applicable): rcl

#### Steps to reproduce issue

Change `id: 11` to the `id: 992147483647` in existing test
https://github.com/ros2/rcl/blob/246fd1de5c073e60573abb014a787f012a241419/rcl_yaml_param_parser/test/correct_config.yaml#L21

#### Expected behavior
The Value shall be correctly parsed to the integer_value

#### Actual behavior
Value parsing as a floating point value. i.e. the check ASSERT_TRUE(NULL != param_value->integer_value); is failing

#### Additional information
- First time found in newly added tests for composable recorder with an error message
```
C++ exception with description "parameter 'storage.max_bagfile_size' has invalid type: Wrong parameter type, parameter {storage.max_bagfile_size} is of type {integer}, setting it to {double} is not allowed." thrown in the test body.
```

#### RCA
It turns out that we are using `ival = strtol(value, &endptr, 0);` function when trying to parse integer value. 
On Linux and 64 bit, it works correctly for values bigger than LONG_MAX = 2147483647 because `sizeof(long)` returns 
**8 bytes**. However on Windows it is not true and `sizeof(long)` is 4 bytes and `strtol(value, &endptr, 0)` can't parse values bigger than `2147483647` and we fallback to the parsing value as a floating point value with `strtod(..)` and it is succeeds.

MichaelOrlov:
- Closes https://github.com/ros2/rcl/issues/1125

#### RCA
It turns out that we are using `ival = strtol(value, &endptr, 0);` function when trying to parse integer value.
On Linux and 64-bit, it works correctly for values bigger than `LONG_MAX = 2147483647` because `sizeof(long)` returns
**8 bytes**. However on Windows, it is not true and `sizeof(long)` is **4** bytes and `strtol(value, &endptr, 0)` can't parse values bigger than 2147483647 and we fallback to the parsing value as a floating point value with `strtod(..)` and it succeeds.

#### FIX
1. Replaced `strtol(..)` with `strtoll(..)`
2. Adjust values in the unit test to cover this use-case.

### DIFF
--- a/rcl_yaml_param_parser/src/parse.c
+++ b/rcl_yaml_param_parser/src/parse.c
@@ -153,7 +153,7 @@ void * get_value(
     errno = 0;
     ival = strtoll(value, &endptr, 0);
     if ((0 == errno) && (NULL != endptr)) {
-      if ((NULL != endptr) && (endptr != value)) {
+      if (endptr != value) {
         if (('\0' != *value) && ('\0' == *endptr)) {
           *val_type = DATA_TYPE_INT64;
           ret_val = allocator.zero_allocate(1U, sizeof(int64_t), allocator.state);
--- a/rcl_yaml_param_parser/src/parse.c
+++ b/rcl_yaml_param_parser/src/parse.c
@@ -151,7 +151,7 @@ void * get_value(
     style != YAML_DOUBLE_QUOTED_SCALAR_STYLE)
   {
     errno = 0;
-    ival = strtol(value, &endptr, 0);
+    ival = strtoll(value, &endptr, 0);
     if ((0 == errno) && (NULL != endptr)) {
       if ((NULL != endptr) && (endptr != value)) {
         if (('\0' != *value) && ('\0' == *endptr)) {
--- a/rcl_yaml_param_parser/test/correct_config.yaml
+++ b/rcl_yaml_param_parser/test/correct_config.yaml
@@ -18,7 +18,7 @@ lidar_ns:
         dy: 0.45
   lidar_2:
     ros__parameters:
-      id: 11
+      id: 992147483647
       name: back_lidar
       dy1: 0.003
       is_back: false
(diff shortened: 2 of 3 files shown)


rcl_yaml_param_parser/src/parse.c
```diff
@@ -151,7 +151,7 @@ void * get_value(
     style != YAML_DOUBLE_QUOTED_SCALAR_STYLE)
   {
     errno = 0;
-    ival = strtol(value, &endptr, 0);
+    ival = strtoll(value, &endptr, 0);
     if ((0 == errno) && (NULL != endptr)) {
       if ((NULL != endptr) && (endptr != value)) {
```
clalancette: I know this isn't directly related to what you are fixing here, but this `NULL != endptr` check is redundant; we can't get into this block if endptr == NULL.

```suggestion
      if (endptr != value) {
```

clalancette (changes requested):
One small change, this otherwise looks good to me.

rcl_yaml_param_parser/src/parse.c
MichaelOrlov: @clalancette  Done.

clalancette (approved):
Looks good to me with green CI.

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/2b6b58cb449fe2da0a220b18bca964cd/raw/df8f73acff15d43f9ecad2ac4c5d1920b013a47a/ros2.repos
BUILD args:  --packages-above-and-dependencies rcl_yaml_param_parser
TEST args:  --packages-above rcl_yaml_param_parser
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/13038
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19964/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14434/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20674/)

fujitatomoya (approved):
lgtm, @MichaelOrlov thanks!

fujitatomoya:
> [!NOTE]
> we would want to backport this to iron and humble.

clalancette:
@Mergifyio backport humble iron

PR #1127, opened 2023-12-13
support `--log-file-name` to ros args.

fujitatomoya:
address https://github.com/ros2/ros2cli/issues/856

depends on https://github.com/ros2/rcl_logging/pull/109

### DIFF
--- a/rcl/src/rcl/logging.c
+++ b/rcl/src/rcl/logging.c
@@ -66,6 +66,7 @@ rcl_logging_configure_with_output_handler(
   g_logging_allocator = *allocator;
   int default_level = -1;
   rcl_log_levels_t * log_levels = &global_args->impl->log_levels;
+  const char * file_name_prefix = global_args->impl->external_log_file_name_prefix;
   const char * config_file = global_args->impl->external_log_config_file;
   g_rcl_logging_stdout_enabled = !global_args->impl->log_stdout_disabled;
   g_rcl_logging_rosout_enabled = !global_args->impl->log_rosout_disabled;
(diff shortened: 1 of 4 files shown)


rcl/src/rcl/logging.c
```diff
@@ -100,7 +101,8 @@ rcl_logging_configure_with_output_handler(
     }
   }
   if (g_rcl_logging_ext_lib_enabled) {
-    status = rcl_logging_external_initialize(config_file, g_logging_allocator);
+    status = rcl_logging_external_initialize(
+      file_name_prefix, config_file, g_logging_allocator);
```
fujitatomoya, line 105: since `file_name_prefix` is in the global arguments, i do not think upstream packages need to care about this change as API/ABI. but probably there are some tests using `--log-level` in `launch`, `ros2cli` and so on, those should be updated with new test cases with `--log-file-name`. (besides `rclc` needs to be checked.)

fujitatomoya:
@iuhilnehc-ynos requesting review.

iuhilnehc-ynos:
I have a question, do most people think `python3_pid_ts.log` is a normal log file name? If yes, I think we can add an extra argument to generate a custom prefix name for log file, otherwise the `python3` should be replaced with another name. (e.g. the filename of `__main__`).

I am not sure if users would like to add this argument for most python executable in a launch file.

iuhilnehc-ynos (approved):

fujitatomoya:
> otherwise the python3 should be replaced with another name. (e.g. the filename of __main__).

yeah we can do that. maybe we can discuss on what could be default name when we get an executable name with `python3`. but here we just give the capability for user to change the log name as they like. i thought that can be more flexible for user. thanks for the review.

fujitatomoya:
Note: Rpr job is supposed to fail since it does not include https://github.com/ros2/rcl_logging/pull/109

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19988/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14454/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20697/)

clalancette (approved):

PR #1133, opened 2024-02-29
Re-order rcl_logging_interface include

christophebedard:
We're not very strict about includes, but I noticed a slight divergence as a result of https://github.com/ros2/rcl/pull/676


clalancette (approved):
OK by me, though as you say we don't truly have a defined style here.

clalancette:
Given that this is just updating the style, I'm going to go ahead and merge with the Rpr job passing green.

Issue #264, opened 2018-06-29
Inconsistent struct/pointer arguments for allocators


mosteo:
I don't see a reason why some functions expect a struct and others a pointer:

Pointer:
https://github.com/ros2/rcl/blob/d13a902d9ad7d5cc9f2b5bce084f823a2e66589d/rcl/include/rcl/graph.h#L86
https://github.com/ros2/rcl/blob/d13a902d9ad7d5cc9f2b5bce084f823a2e66589d/rcl/include/rcl/graph.h#L129

Struct:
https://github.com/ros2/rcl/blob/d13a902d9ad7d5cc9f2b5bce084f823a2e66589d/rcl/include/rcl/graph.h#L212
https://github.com/ros2/rcl/blob/d13a902d9ad7d5cc9f2b5bce084f823a2e66589d/rcl/include/rcl/timer.h#L138

dirk-thomas:
Thanks for reporting this inconsistency. There is no reason why some API takes a pointer and some takes a value. This should indeed be unified.

Just as a side note when choosing which one to use: if an `rcl_allocator_t` would ever store any data directly (rather then behind another pointer) the altering only the members of a copied value might be undesired...

wjwwood:
Indeed it needs to be cleaned up, though I don't think we necessarily need to use the same thing everywhere. The pointer should be preferred, as @dirk-thomas pointed out, so that if the allocator struct becomes non-trivial it is still efficient. However, in cases where the only possible error is that the allocator pointer is null (e.g. a "getter" where the allocator is used only for error processing), some functions return the value directly rather than an `rcl_ret_t`. So by making the allocator struct pass by value, it avoids the check for a null pointer in to an allocator struct.

In the end, making everything a pointer (and reducing the number of functions that need an allocator by doing things like making the macro to set the error message no longer allocate memory) might be the way to go, but I'd advise against a blanket policy despite good reasons to do the opposite in certain cases.

mosteo:
From my point of view as a binding implementer the main issue when passed as pointer is that, if copies are of the pointer are kept, I have to manage the lifetime on my side; whereas if it is always passed by copy the issue does not exist.

For example, just this week-end I realized (via segfaults) that `rcl_clock_t` indeed keeps a copy of the allocator pointer received during initialization. In the case of RCL types, which are themselves containers of pointers, I find it a bit more confusing the passing of pointers for no explicit reason.

[...]

PR #1137, opened 2024-03-15
document out parameters for rcl_get_node_names and rcl_get_node_names_with_enclaves

FelixPe:


rcl/include/rcl/graph.h
```diff
@@ -401,15 +401,15 @@ RCL_WARN_UNUSED
 rcl_ret_t
 rcl_names_and_types_fini(rcl_names_and_types_t * names_and_types);
 
-/// Return a list of available nodes in the ROS graph.
+/// Return a list of available node names in the ROS graph and a list of their namespaces.
```
clalancette: I think this might be a little clearer as:

```suggestion
/// Return a list of node names and their associated namespaces in the ROS graph.
```

rcl/include/rcl/graph.h
```diff
@@ -401,15 +401,15 @@ RCL_WARN_UNUSED
...
 /**
  * The `node` parameter must point to a valid node.
  *
- * The `node_names` parameter must be allocated and zero initialized.
- * `node_names` is the output for this function, and contains allocated memory.
+ * The `node_names` and `node_namespaces` parameter must be allocated and zero initialized.
```
clalancette:
```suggestion
 * The `node_names` and `node_namespaces` parameters must be allocated and zero initialized.
```

rcl/include/rcl/graph.h
```diff
@@ -401,15 +401,15 @@ RCL_WARN_UNUSED
...
+ * `node_names` and `node_namespaces` are the output for this function, and contain allocated memory.
  * Use rcutils_get_zero_initialized_string_array() for initializing an empty
  * rcutils_string_array_t struct.
- * This `node_names` struct should therefore be passed to rcutils_string_array_fini()
+ * This `node_names` and `node_namespaces` struct should therefore be passed to rcutils_string_array_fini()
  * when it is no longer needed.
```
clalancette:
```suggestion
 * These `node_names` and `node_namespaces` structs should therefore be passed to rcutils_string_array_fini()
 * when they are no longer needed.
```

rcl/include/rcl/graph.h
```diff
@@ -458,7 +464,7 @@ rcl_get_node_names(
   rcutils_string_array_t * node_names,
   rcutils_string_array_t * node_namespaces);
 
-/// Return a list of available nodes in the ROS graph, including their enclave names.
+/// Return a list of available node names in the ROS graph, of their namespaces and their enclaves.
```
clalancette:
```suggestion
/// Return a list of node names and their associated namespaces and enclaves in the ROS graph.
```

clalancette (changes requested):
Thanks for these improvements.

I've left a few wording changes.

rcl/include/rcl/graph.h
FelixPe: This is shows much clearer the relation between the two lists, thank you.

rcl/include/rcl/graph.h
```diff
@@ -401,24 +401,26 @@ RCL_WARN_UNUSED
...
 
-/// Return a list of available nodes in the ROS graph.
+/// Return a list of node names and their associated namespaces in the ROS graph.
 /**
  * The `node` parameter must point to a valid node.
  *
- * The `node_names` parameter must be allocated and zero initialized.
- * `node_names` is the output for this function, and contains allocated memory.
+ * The `node_names` and `node_namespaces` parameters must be allocated and zero initialized.
+ * `node_names` and `node_namespaces` are the output for this function, and contain allocated memory.
```
clalancette:
```suggestion
 * `node_names` and `node_namespaces` are the output for this function,
 * and contain allocated memory.
```

(the line is > 100 lines, and so will try up our linters)

rcl/include/rcl/graph.h
```diff
@@ -401,24 +401,26 @@ RCL_WARN_UNUSED
...
  * Use rcutils_get_zero_initialized_string_array() for initializing an empty
  * rcutils_string_array_t struct.
- * This `node_names` struct should therefore be passed to rcutils_string_array_fini()
- * when it is no longer needed.
+ * These `node_names` and `node_namespaces` structs should therefore be passed to rcutils_string_array_fini()
+ * when they are no longer needed.
```
clalancette:
```suggestion
 * These `node_names` and `node_namespaces` structs should therefore be passed
 * to rcutils_string_array_fini() when they are no longer needed.
```

clalancette: (for the linters)

clalancette (changes requested):
Sorry, two more changes for the linters, then I'll run CI.

FelixPe:
@clalancette Regarding the `rcl_allocator_t` parameter for `rcl_get_node_names` in graph.h it looks inconsistent to me: Why is it given for `rcl_get_topic_names_and_types` as pointer type `rcl_allocator_t*` ?

(The allocator is not yet in use for rcl_get_node_names, but it might improve consistency to switch to the pointer type here as well.)

clalancette:
> @clalancette Regarding the `rcl_allocator_t` parameter for `rcl_get_node_names` in graph.h it looks inconsistent to me: Why is it given for `rcl_get_topic_names_and_types` as pointer type `rcl_allocator_t*` ?

This is an ongoing discussion; see https://github.com/ros2/rcl/issues/264 .  There are tradeoffs between having the parameter be a pointer (where we only have to copy the pointer, but then have to check for NULL pointer in the callee), and having the parameter be a structure (where we have to copy the whole structure, but then don't have to check for NULL in the callee).  We still haven't resolved it one way or the other, so I don't think we should get bogged down in it here.

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20485/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14887/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21233/)

fujitatomoya (approved):
lgtm with green CI

PR #1140, opened 2024-03-26
Populate Transitions in Transition Events

### LINKED ISSUE #1019: transition events don't populate transition id, label, or timestamp
## Bug report

**Required Info:**

- Operating System:
  - N/A
- Installation type:
  - binaries
- Version or commit hash:
  - humble
- DDS implementation:
  -  CycloneDDS
  - N/A

#### Steps to reproduce issue
Not all the fields of `lifecycle_msgs/msg/TransitionEvent` are being utilized
```
ros2 topic echo /<node>/transition_event lifecycle_msgs/msg/TransitionEvent
```

#### Expected behavior
```
timestamp: 1668097069451777251
transition:
  id: 1
  label: configure
start_state:
  id: 1
  label: unconfigured
goal_state:
  id: 10
  label: configuring
```

#### Actual behavior
```
timestamp: 0
transition:
  id: 0
  label: ''
start_state:
  id: 1
  label: unconfigured
goal_state:
  id: 10
  label: configuring
```

CursedRock17:
Meant to resolve #1019 this PR adds the timestamp and transition portion of a `TransitionEvent` to the [`rcl_lifecycle_com_interface_publish_notification`](https://github.com/ros2/rcl/blob/rolling/rcl_lifecycle/src/com_interface.h#L80) function

Unfortunately, it looks like in order to do this we will have to break the ABI with either changes or additions to that function's parameters.

### DIFF
--- a/rcl_lifecycle/src/com_interface.c
+++ b/rcl_lifecycle/src/com_interface.c
@@ -328,7 +328,18 @@ rcl_lifecycle_com_interface_publish_notification(
   const char * transition_label, uint8_t transition_id,
   const rcl_lifecycle_state_t * start, const rcl_lifecycle_state_t * goal)
 {
-  com_interface->msg.timestamp = 0u;
+  // Get the current system time based on the rcl_clock
+  rcutils_time_point_value_t current_time;
+  rcutils_ret_t time_ret = rcutils_system_time_now(&current_time);
+  if (time_ret != RCUTILS_RET_OK) {
+    rcutils_error_string_t error = rcutils_get_error_string();
+    rcutils_reset_error();
+    RCL_SET_ERROR_MSG(error.str);
+    time_ret = RCL_RET_ERROR;
+    return time_ret;
+  }
+
+  com_interface->msg.timestamp = current_time;
   com_interface->msg.transition.id = transition_id;
   rosidl_runtime_c__String__assign(&com_interface->msg.transition.label, transition_label);
   com_interface->msg.start_state.id = start->id;
--- a/rcl_lifecycle/src/com_interface.c
+++ b/rcl_lifecycle/src/com_interface.c
@@ -324,9 +324,13 @@ rcl_lifecycle_com_interface_fini(
 
 rcl_ret_t
 rcl_lifecycle_com_interface_publish_notification(
-  rcl_lifecycle_com_interface_t * com_interface,
+  rcl_lifecycle_com_interface_t * com_interface, unsigned int timestamp,
+  const char * transition_label, uint8_t transition_id,
   const rcl_lifecycle_state_t * start, const rcl_lifecycle_state_t * goal)
 {
+  com_interface->msg.timestamp = timestamp;
+  com_interface->msg.transition.id = transition_id;
+  rosidl_runtime_c__String__assign(&com_interface->msg.transition.label, transition_label);
   com_interface->msg.start_state.id = start->id;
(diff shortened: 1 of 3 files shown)


rcl_lifecycle/src/com_interface.c
```diff
@@ -325,8 +325,23 @@ rcl_lifecycle_com_interface_fini(
 rcl_ret_t
 rcl_lifecycle_com_interface_publish_notification(
   rcl_lifecycle_com_interface_t * com_interface,
+  const char * transition_label, uint8_t transition_id,
   const rcl_lifecycle_state_t * start, const rcl_lifecycle_state_t * goal)
 {
+  // Get the current system time based on the rcl_clock
+  rcutils_time_point_value_t current_time;
+  rcutils_ret_t time_ret = rcutils_system_time_now(&current_time);
```
CursedRock17, line 333: The clock works, but will the result of `rcutils_system_time_now` serve as the correct time? The transition timestamp is meant to be 

> The time point at which this event occurred.

Which needs to be set during the period of notification. I can't image trying to pass down a `clock` through multiple layers of API so that the user could set the time that way though.

rcl_lifecycle/src/com_interface.h
```diff
@@ -79,6 +79,7 @@ rcl_ret_t
 RCL_WARN_UNUSED
 rcl_lifecycle_com_interface_publish_notification(
   rcl_lifecycle_com_interface_t * com_interface,
+  const char * transition_label, uint8_t transition_id,
   const rcl_lifecycle_state_t * start, const rcl_lifecycle_state_t * goal);
```
fujitatomoya, line 83: just a curiosity, does that make more sense if that takes `const rcl_lifecycle_transition_t * transition` which includes everything but `com_interface`?

rcl_lifecycle/src/com_interface.c
fujitatomoya: this will publish the system time only, but i think `LifecycleNode.clock_interface` should be used. more like, `rck_lifecycle_com_interface` should be initialized with the same clock with lifecycle node. so that if the node is configured with simulation time, we can also publish the simulation time in the message.

CursedRock17: I definitely agree that this would be the best route to go, but will it be problematic making API changes? As the [`rcl_lifecycle_com_interface`](https://github.com/ros2/rcl/blob/rolling/rcl_lifecycle/include/rcl_lifecycle/data_types.h#L73-#L91) is changed that means all functions initializing `com_interface` must be changed to include a `rcl_clock_t *` argument. This would include adding an argument to [`rcl_lifecycle_state_machine_init`](https://github.com/ros2/rcl/blob/rolling/rcl_lifecycle/src/rcl_lifecycle.c#L194) and all the functions it calls down the line.

rcl_lifecycle/src/com_interface.h
CursedRock17: It would, but originally I just wanted more transparency with what I was adding, that'll be fixed.

SuperJappie08:
@CursedRock17, are you still interested in fixing this?

CursedRock17:
@SuperJappie08 I don't have the time for this issue right now, so if you'd like to pick this up, that would be awesome!

SuperJappie08:
@CursedRock17, I might be able to pick it up.
However, I also have limited time right now (for at least the next few weeks).
But I will put it on my list. (If somebody wants to do it before I get to it, feel free to make the PR)

PR #1156, opened 2024-05-08
Fix up rmw_cyclonedds timestamp testing.

clalancette:
We are about to fix it so that rmw_cyclonedds has receive_timestamp support, so we also need to enable that support here in rcl.  We actually rewrite the logic a bit because now the only combination that doesn't work is rmw_connextdds on Windows.

This is the counterpart to https://github.com/ros2/rmw_cyclonedds/pull/491 , and must be merged at the same time.


marcoag (approved):

fujitatomoya (approved):
@clalancette lgtm with green CI.

wjwwood (approved):

clalancette:
CI for this is in https://github.com/ros2/rmw_cyclonedds/pull/491#issuecomment-2101288484 (since that is the companion PR).

clalancette:
@mergifyio backport jazzy

PR #1164, opened 2024-06-28
Cleanup test_count_matched test to handle non-DDS RMWs

clalancette:
The main goal of this PR is to update the test_count_matched test so that it handles non-DDS RMWs.  Along the way, it also does some cleanup in the test.  The patches are:

1.  Slightly refactor things so that `check_state` is a protected class method, rather than a free function.  That way it can use more of the internal details of the class, which it depends on anyway.
2.  Renames "ops" to "opts", as that better reflects that these are options.
3.  Uses scope_exit to do some cleanup in the second test.  We purposefully do *not* do this in the first test, as it specifically tests what happens as various things get destroyed.
4.  We add compatibility with non-DDS RMWs.  Other RMWs may have different matching semantics between different QoS settings.  In particular, `rmw_zenoh` will happily match BEST_EFFORT publishers to RELIABLE subscriptions, while the DDS RMWs will not.  Luckily, we have the function called `rmw_qos_profile_check_compatible` that we can call to tell whether the underlying RMW would match them, so we can use that here.

### DIFF
--- a/rcl/test/rcl/test_count_matched.cpp
+++ b/rcl/test/rcl/test_count_matched.cpp
@@ -28,73 +28,11 @@
 
 #include "rcl/error_handling.h"
 
-void check_state(
-  rcl_wait_set_t * wait_set_ptr,
-  rcl_publisher_t * publisher,
-  rcl_subscription_t * subscriber,
-  const rcl_guard_condition_t * graph_guard_condition,
-  size_t expected_subscriber_count,
-  size_t expected_publisher_count,
-  size_t number_of_tries)
-{
-  size_t subscriber_count = -1;
-  size_t publisher_count = -1;
-
-  rcl_ret_t ret;
-
-  for (size_t i = 0; i < number_of_tries; ++i) {
-    if (publisher) {
-      ret = rcl_publisher_get_subscription_count(publisher, &subscriber_count);
-      EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
-      rcl_reset_error();
-    }
-
-    if (subscriber) {
-      ret = rcl_subscription_get_publisher_count(subscriber, &publisher_count);
-      EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
-      rcl_reset_error();
-    }
-
-    if (
-      expected_publisher_count == publisher_count &&
-      expected_subscriber_count == subscriber_count)
-    {
-      RCUTILS_LOG_INFO_NAMED(ROS_PACKAGE_NAME, "  state correct!");
-      break;
-    }
-
-    if ((i + 1) == number_of_tries) {
-      // Don't wait for the graph to change on the last loop because we won't check again.
-      continue;
-    }
-
-    ret = rcl_wait_set_clear(wait_set_ptr);
-    ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21338/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15650/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/852/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22076/)

rcl/test/rcl/test_count_matched.cpp
```diff
@@ -141,47 +83,109 @@ class TestCountFixture : public ::testing::Test
...
+
+protected:
+  const rcl_guard_condition_t * graph_guard_condition;
+  rcl_context_t * context_ptr;
+  rcl_wait_set_t * wait_set_ptr;
+
+  void check_state(
+    rcl_publisher_t * publisher,
+    rcl_subscription_t * subscriber,
+    size_t expected_subscriber_count,
```
ahcorde, line 97: include `<cstddef>`

rcl/test/rcl/test_count_matched.cpp
```diff
@@ -141,47 +83,109 @@ class TestCountFixture : public ::testing::Test
...
+      }
+
+      ret = rcl_wait_set_clear(wait_set_ptr);
+      ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+      ret = rcl_wait_set_add_guard_condition(wait_set_ptr, graph_guard_condition, NULL);
+      ASSERT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
+      std::chrono::nanoseconds time_to_sleep = std::chrono::milliseconds(200);
+      RCUTILS_LOG_INFO_NAMED(
+        ROS_PACKAGE_NAME,
+        "  state wrong, waiting up to '%s' nanoseconds for graph changes... ",
```
ahcorde:
```suggestion
        " state wrong, waiting up to '%s' nanoseconds for graph changes... ",
```

ahcorde (approved):
Two minor comments, LGTM with green CI

rcl/test/rcl/test_count_matched.cpp
clalancette: Good call, fixed in 8d468ca.

rcl/test/rcl/test_count_matched.cpp
clalancette: Actually, we don't need the space at all.  `RCUTILS_LOG_INFO_NAMED` will automatically add a space.  So I removed both spaces, and also removed it in another call above in 8d468ca.

clalancette:
New CI after changes from review:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21339/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15652/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/853/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22077/)

fujitatomoya (approved):
lgtm with green CI

Yadunund (approved):

PR #1167, opened 2024-07-08
Removed deprecated rcl_init_timer()

ahcorde:
Removed deprecated rcl_init_timer() deprecated since `jazzy`

### DIFF
--- a/rcl/include/rcl/timer.h
+++ b/rcl/include/rcl/timer.h
@@ -178,21 +178,6 @@ rcl_timer_init2(
   rcl_allocator_t allocator,
   bool autostart);
 
-/**
- * \deprecated `rcl_timer_init` implementation was removed.
- *   Refer to `rcl_timer_init2`.
- */
-RCL_PUBLIC
-RCUTILS_DEPRECATED_WITH_MSG("Call rcl_timer_init2 instead")
-rcl_ret_t
-rcl_timer_init(
-  rcl_timer_t * timer,
-  rcl_clock_t * clock,
-  rcl_context_t * context,
-  int64_t period,
-  const rcl_timer_callback_t callback,
-  rcl_allocator_t allocator);
-
 /// Finalize a timer.
 /**
  * This function will deallocate any memory and make the timer invalid.
--- a/rcl/src/rcl/timer.c
+++ b/rcl/src/rcl/timer.c
@@ -127,20 +127,6 @@ void _rcl_timer_time_jump(
 }
 
 rcl_ret_t
-rcl_timer_init(
-  rcl_timer_t * timer,
-  rcl_clock_t * clock,
-  rcl_context_t * context,
-  int64_t period,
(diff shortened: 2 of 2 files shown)


ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21386/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15692/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/895/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22122/)

clalancette (changes requested):
I'm OK with removing this one, but I think we also need to update https://github.com/ros2/ros2_tracing/blob/74ff10b30809907b3799b85d6a0414c21198ab3c/tracetools_trace/tracetools_trace/tools/tracepoints.py#L35 .  FYI @christophebedard

christophebedard:
What needs to be updated? The `ros2:rcl_timer_init` tracepoint is already in the new `rcl_timer_init2()` function.

clalancette:
> What needs to be updated? The `ros2:rcl_timer_init` tracepoint is already in the new `rcl_timer_init2()` function.

Oh, I may have misinterpreted.  I thought maybe https://github.com/ros2/ros2_tracing/blob/74ff10b30809907b3799b85d6a0414c21198ab3c/tracetools_trace/tracetools_trace/tools/tracepoints.py#L35 needed to be updated to `ros2:rcl_timer_init2`?  But I definitely might be wrong about that.

christophebedard:
Yeah, that's a fair assumption. The tracepoint event name is usually very similar to the corresponding function name (because the function name itself is nicely descriptive), but it really doesn't have to.

In this case, the `rcl` function name changed because we added a new argument. I suppose we could record the `bool autostart` value, but I don't see a real need for it, unless someone wants that information.

clalancette:
> Yeah, that's a fair assumption. The tracepoint event name is usually very similar to the corresponding function name (because the function name itself is nicely descriptive), but it really doesn't have to.

Oh, I see.  I thought that the tracetools name needed to match the function name, but now I understand that it just needs to match the tracepoint name (which as you say, is similar to but not exactly the same in this case).

OK, in that case I'm fine with this removal.  Thanks for the info @christophebedard

clalancette (approved):
This looks good to me with green CI (it looks like Linux may have to be re-run).

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21422/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15707/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/922/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22138/)

Issue #1016, opened 2022-10-27
:farmer: Flaky test `test_graph__rmw_fastrtps_cpp` on ros2 buildfarm


Crola1702:
## Bug report

- Operating System:
  - Ubuntu 20.04 and 22.04
- Installation type:
  - Source
- Version or commit hash:
  - Rolling
- DDS implementation:
  - Fast-RTPS

#### Steps to reproduce issue
1. Run a build in one of the following jobs
   *  Nightlies repeated jobs (windows, arch64, linux, rhel)
   * Humble coverage or debug
   * Linux aarch64 Debug
2. See it fail (if lucky)

#### Expected behavior

Not failing

#### Actual behavior

This test is failing on repeated jobs since the last year (as they retest everything until fail, it is expected that this test fails there), 
However, I find it weird that is happening on other jobs different to repeated ones, because when the test fails there, it's rerun (opposite to repeated jobs)

This test has a 21% flaky ratio on Humble Coverage (7/33 builds checked, however just 1 of them is marked as unstable with this test failure).

#### Additional information

Reference build: https://ci.ros2.org/job/nightly_linux_humble_coverage/187/

Test regression: [rcl_action.test_graph__rmw_fastrtps_cpp.gtest.missing_result](https://ci.ros2.org/job/nightly_linux_humble_coverage/187/testReport/junit/(root)/rcl_action/test_graph__rmw_fastrtps_cpp_gtest_missing_result/)

Log output:
<details>

Log output:
```
      Start 10: test_graph__rmw_fastrtps_cpp

[...]

PR #1170, opened 2024-07-16
Added remapping resolution for action names

### LINKED ISSUE #1312: Remapping action names does not work
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- Installation type:
  - Docker container based on `ros:humble`

#### Steps to reproduce issue
Run the following launch file to launch the `fibonacci` action server from `action_tutorials_py` with remapping of the action name:
```python
from launch import LaunchDescription
from launch_ros.actions import Node


def generate_launch_description():
    node = Node(
        package='action_tutorials_py',
        executable='fibonacci_action_server',
        remappings=[
            ('fibonacci', '/remapped_fibonacci'),
            ('~/list_parameters', '/fibonacci_list_parameters'),
        ]
    )
    return LaunchDescription([
        node
    ])

```

#### Expected behavior
Launching the `fibonnaci_action_server` node running with the `/remapped_fibonacci` action server, and also exposing the `/fibonacci_list_parameters` service instead of the `/fibonacci_action_server/fibonacci_list_parameters` service.

#### Actual behavior
The service is remapped correctly, but the action is not.

Output from `ros2 action list`
```
➜  zb_ws ros2 action list
/fibonacci
```

Output from `ros2 service list`
```
➜  zb_ws ros2 service list
/fibonacci_action_server/describe_parameters
/fibonacci_action_server/get_parameter_types
/fibonacci_action_server/get_parameters
/fibonacci_action_server/set_parameters
/fibonacci_action_server/set_parameters_atomically
/fibonacci_list_parameters
```

JustusBraun:
Currently remapping of action names is not possible see  https://github.com/ros2/ros2/issues/1312 .
I have now stumbled over this multiple times and I think it is a useful feature, so i took a shot at implementing the remapping of action names. Please let me know if you have any comments or if I overlooked something.
Fixes https://github.com/ros2/ros2/issues/1312.

### DIFF
--- a/rcl_action/src/rcl_action/action_client.c
+++ b/rcl_action/src/rcl_action/action_client.c
@@ -110,7 +110,7 @@ _rcl_action_client_fini_impl(
 // \internal Initializes an action client specific service client.
 #define CLIENT_INIT(Type) \
   char * Type ## _service_name = NULL; \
-  ret = rcl_action_get_ ## Type ## _service_name(action_name, allocator, &Type ## _service_name); \
+  ret = rcl_action_get_ ## Type ## _service_name(resolved_action_name, allocator, &Type ## _service_name); \
   if (RCL_RET_OK != ret) { \
     rcl_reset_error(); \
     RCL_SET_ERROR_MSG("failed to get " #Type " service name"); \
@@ -137,7 +137,7 @@ _rcl_action_client_fini_impl(
 // \internal Initializes an action client specific topic subscription.
 #define SUBSCRIPTION_INIT(Type) \
   char * Type ## _topic_name = NULL; \
-  ret = rcl_action_get_ ## Type ## _topic_name(action_name, allocator, &Type ## _topic_name); \
+  ret = rcl_action_get_ ## Type ## _topic_name(resolved_action_name, allocator, &Type ## _topic_name); \
   if (RCL_RET_OK != ret) { \
     rcl_reset_error(); \
     RCL_SET_ERROR_MSG("failed to get " #Type " topic name"); \
@@ -195,8 +195,19 @@ rcl_action_client_init(
 
   // Avoid uninitialized pointers should initialization fail
   *action_client->impl = _rcl_action_get_zero_initialized_client_impl();
+
+  // Remap/Expand the action name
+  char* resolved_action_name = NULL;
+  ret = rcl_node_resolve_name(node, action_name, allocator, false, false, &resolved_action_name);
+  if (RCL_RET_OK != ret) {
+    if (RCL_RET_TOPIC_NAME_INVALID == ret || RCL_RET_SERVICE_NAME_INVALID == ret) {
+      ret = RCL_RET_ACTION_NAME_INVALID;
+    }
+    goto fail;
+  }
+
   // Copy action client name and options.
-  action_client->impl->action_name = rcutils_strdup(action_name, allocator);
+  action_client->impl->action_name = rcutils_strdup(resolved_action_name, allocator);
   if (NULL == action_client->impl->action_name) {
     RCL_SET_ERROR_MSG("failed to duplicate action name");
     ret = RCL_RET_BAD_ALLOC;
@@ -213,6 +224,10 @@ rcl_action_client_init(
   SUBSCRIPTION_INIT(feedback);
   SUBSCRIPTION_INIT(status);
 
+  // The resolved action name is no longer needed
+  allocator.deallocate(resolved_action_name, allocator.state);
+  resolved_action_name = NULL;
+
   ret = rcl_node_type_cache_register_type(
       node, type_support->get_type_hash_func(type_support),
       type_support->get_type_description_func(type_support),
@@ -227,6 +242,12 @@ rcl_action_client_init(
   RCUTILS_LOG_DEBUG_NAMED(ROS_PACKAGE_NAME, "Action client initialized");
   return ret;
 fail:
+  // Deallocate the resolved action name
+  if (NULL != resolved_action_name)
+  {
+    allocator.deallocate(resolved_action_name, allocator.state);
+  }
+
   fini_ret = _rcl_action_client_fini_impl(action_client, node, allocator);
   if (RCL_RET_OK != fini_ret) {
     RCL_SET_ERROR_MSG("failed to cleanup action client");
--- a/rcl_action/src/rcl_action/action_client.c
+++ b/rcl_action/src/rcl_action/action_client.c
@@ -110,7 +110,8 @@ _rcl_action_client_fini_impl(
 // \internal Initializes an action client specific service client.
 #define CLIENT_INIT(Type) \
   char * Type ## _service_name = NULL; \
-  ret = rcl_action_get_ ## Type ## _service_name(resolved_action_name, allocator, &Type ## _service_name); \
+  ret = rcl_action_get_ ## Type ## _service_name( \
+    resolved_action_name, allocator, &Type ## _service_name); \
   if (RCL_RET_OK != ret) { \
     rcl_reset_error(); \
     RCL_SET_ERROR_MSG("failed to get " #Type " service name"); \
@@ -137,7 +138,8 @@ _rcl_action_client_fini_impl(
 // \internal Initializes an action client specific topic subscription.
 #define SUBSCRIPTION_INIT(Type) \
   char * Type ## _topic_name = NULL; \
-  ret = rcl_action_get_ ## Type ## _topic_name(resolved_action_name, allocator, &Type ## _topic_name); \
+  ret = rcl_action_get_ ## Type ## _topic_name( \
+    resolved_action_name, allocator, &Type ## _topic_name); \
   if (RCL_RET_OK != ret) { \
     rcl_reset_error(); \
     RCL_SET_ERROR_MSG("failed to get " #Type " topic name"); \
@@ -197,7 +199,7 @@ rcl_action_client_init(
   *action_client->impl = _rcl_action_get_zero_initialized_client_impl();
 
   // Remap/Expand the action name
-  char* resolved_action_name = NULL;
+  char * resolved_action_name = NULL;
   ret = rcl_node_resolve_name(node, action_name, allocator, false, false, &resolved_action_name);
   if (RCL_RET_OK != ret) {
     if (RCL_RET_TOPIC_NAME_INVALID == ret || RCL_RET_SERVICE_NAME_INVALID == ret) {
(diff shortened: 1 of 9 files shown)


rcl_action/src/rcl_action/action_client.c
```diff
@@ -212,8 +214,19 @@ rcl_action_client_init(
...
   *action_client->impl = _rcl_action_get_zero_initialized_client_impl();
+
+  // Remap/Expand the action name
+  char * resolved_action_name = NULL;
+  ret = rcl_node_resolve_name(node, action_name, allocator, false, false, &resolved_action_name);
+  if (RCL_RET_OK != ret) {
+    if (RCL_RET_TOPIC_NAME_INVALID == ret || RCL_RET_SERVICE_NAME_INVALID == ret) {
+      ret = RCL_RET_ACTION_NAME_INVALID;
+    }
+    goto fail;
```
Barry-Xu-2018, line 216: rcl_node_resolve_name() can return many types of errors, so the return values should be simplified.
Refer to
https://github.com/ros2/rcl/blob/0723613c89d86f23a4aef70601994b812a6d15a2/rcl/src/rcl/service.c#L124-L139

rcl_action/src/rcl_action/action_server.c
```diff
@@ -163,6 +165,16 @@ rcl_action_server_init(
   action_server->impl->type_hash = rosidl_get_zero_initialized_type_hash();
 
   rcl_ret_t ret = RCL_RET_OK;
+  // Resolve action name
+  char * resolved_action_name = NULL;
+  ret = rcl_node_resolve_name(node, action_name, allocator, false, false, &resolved_action_name);
+  if (RCL_RET_OK != ret) {
+    if (RCL_RET_TOPIC_NAME_INVALID == ret || RCL_RET_SERVICE_NAME_INVALID == ret) {
+      ret = RCL_RET_ACTION_NAME_INVALID;
+    }
```
Barry-Xu-2018, line 169: rcl_node_resolve_name() can return many types of errors, so the return values should be simplified.

rcl_action/src/rcl_action/action_client.c
JustusBraun: Thank you for the feedback, i have implemented the recommended changes. The only error codes returned in case the name resolution fails are RCL_RET_ACTION_NAME_INVALID, RCL_RET_BAD_ALLOC and RCL_RET_ERROR.

rcl_action/src/rcl_action/action_server.c
JustusBraun: I have implemented the recommended changes.

rcl_action/src/rcl_action/action_server.c
```diff
@@ -189,12 +203,16 @@ rcl_action_server_init(
   }
 
   // Copy action name
-  action_server->impl->action_name = rcutils_strdup(action_name, allocator);
+  action_server->impl->action_name = rcutils_strdup(resolved_action_name, allocator);
```
Barry-Xu-2018: Change this will action_server_get_action_name().  This is why you also update test codes.   
I refer to the implementation of other functions rcl_client_get_service_name/rcl_service_get_service_name. They all return original service name (Not remapped).   
So I prefer not to modify action_name here. Is it necessary to add a new item to store the resolved action name?

JustusBraun: I looked into the implementation of `rcl_client_get_service_name()`/`rcl_service_get_service_name()`, they both return `impl->rmw_handle->service_name` which is the remapped and expanded name. I have tested this behavior by modifying the service and client rclc examples to accept remaps from the cmd line. Using `rcl_{client,service}_get_service_name()` returned the fully expanded and remapped service name in both cases.

https://github.com/ros2/rcl/blob/0723613c89d86f23a4aef70601994b812a6d15a2/rcl/src/rcl/service.c#L295-L304

`rcl_action_server_get_action_name()` returns `action_server->impl->action_name`, I think it would be more consistent with the services if this is also the remapped and expanded name. Maybe we could rename the field to `remapped_action_name` to make it clearer?

Barry-SH: You are right. It‘s my fault.  

> Maybe we could rename the field to remapped_action_name to make it clearer?

Totally agree.

Barry-Xu-2018: You are right. It's my fault.

> Maybe we could rename the field to remapped_action_name to make it clearer?

Totally agree.

JustusBraun:
I wasn't able to reproduce the test error locally. I think this might be related to/the same as https://github.com/ros2/rcl/issues/1016.

rcl_action/src/rcl_action/action_client.c
```diff
@@ -212,9 +214,22 @@ rcl_action_client_init(
 
   // Avoid uninitialized pointers should initialization fail
   *action_client->impl = _rcl_action_get_zero_initialized_client_impl();
+
+  // Remap/Expand the action name
+  char * resolved_action_name = NULL;
+  ret = rcl_node_resolve_name(node, action_name, allocator, false, false, &resolved_action_name);
```
fujitatomoya: Do we need the pointer variable on stack memory?

```suggestion
  // Remap/Expand the action name
  ret = rcl_node_resolve_name(node, action_name, allocator, false, false, &action_client->impl->remapped_action_name);
```

rcl_action/src/rcl_action/action_client.c
```diff
@@ -212,9 +214,22 @@ rcl_action_client_init(
...
+  if (RCL_RET_OK != ret) {
+    if (RCL_RET_TOPIC_NAME_INVALID == ret || RCL_RET_UNKNOWN_SUBSTITUTION == ret) {
+      ret = RCL_RET_ACTION_NAME_INVALID;
+    } else if (RCL_RET_BAD_ALLOC != ret) {
+      ret = RCL_RET_ERROR;
+    }
+    goto fail;
+  }
+
```
fujitatomoya, line 223: i would add debug log here aligned with service implementation.

```suggestion
  RCUTILS_LOG_DEBUG_NAMED(
    ROS_PACKAGE_NAME, "Expanded and remapped action name '%s'",
    action_client->impl->remapped_action_name);

```

rcl_action/src/rcl_action/action_client.c
```diff
@@ -212,9 +214,22 @@ rcl_action_client_init(
...
   // Copy action client name and options.
-  action_client->impl->action_name = rcutils_strdup(action_name, allocator);
-  if (NULL == action_client->impl->action_name) {
+  action_client->impl->remapped_action_name = rcutils_strdup(resolved_action_name, allocator);
```
fujitatomoya: this copy can be removed, can you take a look at the comment above?

rcl_action/src/rcl_action/action_client.c
```diff
@@ -212,9 +214,22 @@ rcl_action_client_init(
...
+  if (NULL == action_client->impl->remapped_action_name) {
```
fujitatomoya: this if statement does not make sense to me anymore, maybe we can remove this since this is for `rcutils_strdup` call which no longer required to copy.

rcl_action/src/rcl_action/action_client.c
```diff
@@ -230,6 +245,10 @@ rcl_action_client_init(
   SUBSCRIPTION_INIT(feedback);
   SUBSCRIPTION_INIT(status);
 
+  // The resolved action name is no longer needed
+  allocator.deallocate(resolved_action_name, allocator.state);
+  resolved_action_name = NULL;
+
```
fujitatomoya: same comment, this can be removed if we do not need `resolved_action_name` in stack memory.

fujitatomoya (changes requested):
test would be ideal, similar with https://github.com/ros2/rcl/blob/rolling/rcl/test/rcl/test_remap_integration.cpp.

most of the comments for the action client can be applied to server implementation as well.

rcl_action/src/rcl_action/action_client.c
JustusBraun: You are right, it is not actually needed, I removed the stack variable

JustusBraun:
I have added tests for action name remapping as requested. @fujitatomoya do you think it is necessary to add tests for local arguments passed to `rcl_node_init` also?

JustusBraun:
Hi, I was was busy for some time and forgot about this PR.

As far as I understand the errors, they are not related to my changes. There seem to be a lot of unexpected 'test_broadcaster_xy' nodes in the test environment of the rcl and projectroot test suites.

Sorry for leaving this pr stale for so long.

fujitatomoya:
@Mergifyio rebase

fujitatomoya:
@JustusBraun can you rebase this? i tried that via Mergifyio but failed.

JustusBraun:
@fujitatomoya done! Locally all test passed with `colcon test`

rcl_action/test/rcl_action/arg_macros.hpp
```diff
@@ -0,0 +1,81 @@
+// Copyright 2018 Open Source Robotics Foundation, Inc.
```
fujitatomoya, line 1: this is just a copy of `rcl/test/rcl/arg_macros.hpp`, basically we would want to avoid this redundancy for the maintenance cost with shared test package. i think this does not block this PR, but this is `TODO` with another PR in the future.

fujitatomoya (commented):
basically looks good to me.

> do you think it is necessary to add tests for local arguments passed to rcl_node_init also?

yes can you please add it? other endpoints such as publisher, subscription, service client and service server also have local argument tests.

fujitatomoya:
@Barry-Xu-2018 requesting review on this.

JustusBraun:
@fujitatomoya I have added the tests with local arguments.

fujitatomoya (approved):
@JustusBraun thanks for the update, lgtm with green CI.

fujitatomoya:
Pulls: ros2/rcl#1170
Gist: https://gist.githubusercontent.com/fujitatomoya/95f8ba959cca7319c9b6361710516c9c/raw/a56ce747ca67db7c529e958e565a96bd29f8a77f/ros2.repos
BUILD args:  --packages-above-and-dependencies rcl_action
TEST args:  --packages-above rcl_action
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15086
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22410/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16667/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1943/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23201/)

christophebedard:
@Mergifyio backport jazzy

Issue #1166, opened 2024-07-08
Lifecyle node failed to make transition


kjjpc:
## Bug report

- Operating System:
  - Ubuntu 22.04
- Installation type:
  - binaries
- Version or commit hash:
  - ros-jazzy-ros-base=0.11.0-1 of apt package
- DDS implementation:
  - Fast DDS (Default DDS of jazzy)
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue
When I launch multiple lifecycle nodes using LaunchService, some nodes failed to be activated.
This is a reproduction script.
```
import os
import signal

from launch import LaunchDescription
from launch.actions import EmitEvent
from launch.actions import RegisterEventHandler
from launch.event_handlers import OnProcessStart
import launch.events
from launch.launch_description_entity import LaunchDescriptionEntity
from launch.launch_service import LaunchService
import launch_ros
from launch_ros.actions import LifecycleNode
from launch_ros.events.lifecycle import ChangeState
from lifecycle_msgs.msg import Transition
from multiprocessing import Process
from typing import List


def up_launch(launch_description_list: List[LaunchDescriptionEntity]):
    ls = LaunchService()
    ls.include_launch_description(LaunchDescription(launch_description_list))
    ls.run()


def generate_launch_description(ns):
    lc_node = LifecycleNode(
        package='lifecycle_py',
        executable='lifecycle_talker',
        name='lc_talker',
        namespace=ns,
        output='screen',
    )
    emit_configure = RegisterEventHandler(
        OnProcessStart(
            target_action=lc_node,
            on_start=[
                EmitEvent(event=ChangeState(
                    lifecycle_node_matcher=launch.events.matches_action(lc_node),
                    transition_id=Transition.TRANSITION_CONFIGURE,
                )),
            ]))

    #  send activate when on_configure finished
    emit_active = RegisterEventHandler(
        launch_ros.event_handlers.OnStateTransition(
            target_lifecycle_node=lc_node,
            goal_state='inactive',
            # start_state='configuring',
            entities=[
                EmitEvent(event=ChangeState(
                    lifecycle_node_matcher=launch.events.matches_action(lc_node),
                    transition_id=Transition.TRANSITION_ACTIVATE,
                )),
            ],
        ))

    return [lc_node, emit_configure, emit_active]


if __name__ == '__main__':
    os.setpgrp()
    try:
        for i in range(12):
            ls = generate_launch_description('room'+str(i))
[...]

PR #1171, opened 2024-07-19
Make transition event topic reliable to avoid topic lost

kjjpc:
This PR is related to #1166.
Lifecycle action of launch_ros occasionally fails to transit because of topic event lost.
This PR make transition event topic reliable.
After the marge of this PR, I will make a PR for launch_ros.

### DIFF
--- a/rcl_lifecycle/src/com_interface.c
+++ b/rcl_lifecycle/src/com_interface.c
@@ -108,6 +108,9 @@ rcl_lifecycle_com_interface_publisher_init(
 
   // initialize publisher
   rcl_publisher_options_t publisher_options = rcl_publisher_get_default_options();
+  publisher_options.qos.reliability = RMW_QOS_POLICY_RELIABILITY_RELIABLE;
+  publisher_options.qos.durability = RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL;
+  publisher_options.qos.depth = 1;
   rcl_ret_t ret = rcl_publisher_init(
     &com_interface->pub_transition_event, node_handle,
     ts_pub_notify, pub_transition_event_topic, &publisher_options);
--- a/rcl_lifecycle/src/com_interface.c
+++ b/rcl_lifecycle/src/com_interface.c
@@ -108,7 +108,6 @@ rcl_lifecycle_com_interface_publisher_init(
 
   // initialize publisher
   rcl_publisher_options_t publisher_options = rcl_publisher_get_default_options();
-  publisher_options.qos.reliability = RMW_QOS_POLICY_RELIABILITY_RELIABLE;
   // transition event topic needs to be latched for the subscription joins later.
   publisher_options.qos.durability = RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL;
   publisher_options.qos.depth = 1;
--- a/rcl_lifecycle/src/com_interface.c
+++ b/rcl_lifecycle/src/com_interface.c
@@ -109,6 +109,7 @@ rcl_lifecycle_com_interface_publisher_init(
   // initialize publisher
   rcl_publisher_options_t publisher_options = rcl_publisher_get_default_options();
   publisher_options.qos.reliability = RMW_QOS_POLICY_RELIABILITY_RELIABLE;
+  // transition event topic needs to be latched for the subscription joins later.
   publisher_options.qos.durability = RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL;
   publisher_options.qos.depth = 1;
   rcl_ret_t ret = rcl_publisher_init(
--- a/rcl_lifecycle/src/com_interface.c
+++ b/rcl_lifecycle/src/com_interface.c
@@ -113,7 +113,6 @@ rcl_lifecycle_com_interface_publisher_init(
 
   // initialize publisher
   rcl_publisher_options_t publisher_options = rcl_publisher_get_default_options();
-  publisher_options.qos.reliability = RMW_QOS_POLICY_RELIABILITY_RELIABLE;
   // transition event topic needs to be latched for the subscription joins later.
   publisher_options.qos.durability = RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL;
   publisher_options.qos.depth = 1;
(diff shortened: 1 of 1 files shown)


rcl_lifecycle/src/com_interface.c
```diff
@@ -108,6 +108,9 @@ rcl_lifecycle_com_interface_publisher_init(
 
   // initialize publisher
   rcl_publisher_options_t publisher_options = rcl_publisher_get_default_options();
+  publisher_options.qos.reliability = RMW_QOS_POLICY_RELIABILITY_RELIABLE;
```
fujitatomoya: rmw_qos_profile_default already sets this `RMW_QOS_POLICY_RELIABILITY_RELIABLE`, so no need to overwrite this.

rcl_lifecycle/src/com_interface.c
```diff
@@ -108,6 +108,9 @@ rcl_lifecycle_com_interface_publisher_init(
...
+  publisher_options.qos.durability = RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL;
+  publisher_options.qos.depth = 1;
```
fujitatomoya, line 113:
```suggestion
  // transition event topic needs to be latched for the subscription joins later.
  publisher_options.qos.durability = RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL;
  publisher_options.qos.depth = 1;
```

changing publisher's QoS to transient local should not break any downstream subscriptions, https://docs.ros.org/en/rolling/Concepts/Intermediate/About-Quality-of-Service-Settings.html#id6.

IMO this transition event topic needs to be latched to deliver the latest state for the subscribers that join later. probably adding comments why we use transient local with depth 1 would be useful.

besides, [Managed Node Design](https://design.ros2.org/articles/node_lifecycle.html) also says the following.

> A topic should be provided to broadcast the new life cycle state when it changes. This topic must be latched.

rcl_lifecycle/src/com_interface.c
kjjpc: Thanks, I've removed the line.

mjcarroll:
@kjjpc Thanks for reporting.  We discussed this at the recent PMC maintainers' meeting and came to the following conclusion.

In this case, changing the durability of the transition event could have far reaching impacts on the system that are undesirable. The main concern is that if you had a late-joining lifecycle node, that it could be transitioned inadvertently by the latched topic.  This would be pretty un-intuitive and could cause issues in larger systems.

The real issue here stems from the fact that we are triggering the transition only when the process becomes available, rather than waiting for the node to be in the `unconfigured` state.

I think that the real solution here would be detect when the node is started in the `unconfigured` state in your launch file, like:

```
    emit_active = RegisterEventHandler(
        launch_ros.event_handlers.OnStateTransition(
            target_lifecycle_node=lc_node,
            goal_state='unconfigured',
            ...
        ))
```

The other alternative would be to add on `OnNodeReady` event to `launch_ros` so that we are more intelligently waiting for the node to start.

fujitatomoya:
@mjcarroll 

i might be mistaken, but i was thinking opposite...

> The main concern is that if you had a late-joining lifecycle node, that it could be transitioned inadvertently by the latched topic. This would be pretty un-intuitive and could cause issues in larger systems.

either that is lifecycle node or just a subscription, delivering the latest state change event via latched topic to the subscription that does specify the QoS with durability by the user application makes sense?
this is something user application wants to do that, so is specified by user.
if they do not want to receive the latest state change event via latched topic, they just do not need to specify the QoS durability on the subscription side? without this change, system cannot provide the choice for the user application.
besides, this does not break the communication between them, https://docs.ros.org/en/rolling/Concepts/Intermediate/About-Quality-of-Service-Settings.html#qos-compatibilities

mjcarroll:
> either that is lifecycle node or just a subscription, delivering the latest state change event via latched topic to the subscription that does specify the QoS with durability by the user application makes sense?

Hmm, maybe my understanding of the issue here was wrong.  I thought we were making the subscription `transient_local`, which would cause the issues I described.

kjjpc:
@mjcarroll 
Thanks for the discussion. The issue is that the transition event topic is volatile in the current implementation.
launch_ros occasionally fails to receive the change event topic because the topic is not latched.

Wiktor-99:
Hey @mjcarroll @mjcarroll, is this change going to be merged or discussion is ongoing? I'd like this feature to be reliable. If any additional work needs to be done I can help

wjwwood:
Pulls: ros2/rcl#1171
Gist: https://gist.githubusercontent.com/wjwwood/ed8b269a34311280adcfd2e8586e7643/raw/caf75df9e5aa18e9d0289b229a5e4b76c4377935/ros2.repos
BUILD args:  --packages-above-and-dependencies rcl_lifecycle
TEST args:  --packages-above rcl_lifecycle
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/20087
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30078/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22845/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9963/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28998/)

wjwwood:
I think @fujitatomoya is right that the behavior change would not occur unless the subscriber also requests transient_local, otherwise the subscriber will only get new messages when subscribing with volatile. So I think this should be safe. If we also change some common code that makes the subscription transient_local, then I think we could be in some danger of undesired behavior change.

Now whether or not this is the right fix for the original problem, that I'm not sure about. If it's a race between the state changing and the subscription being established, then this would fix the problem if you also change the subscription, but it has the side-effect that late joiners (really late joiners) will get a message on this topic indicating the state changed, even though the state change happened long in the past.

Instead, I think it would make more sense to establish a volatile pub-sub pair for event notification, but also use a service call to query the state explicitly when first starting to listen to the state of a node. That way you never get stale messages on the event topic, but also don't rely on racing that topic for the first state observed when starting a monitoring process.

I started a CI just to see what the status of the code is, but I think this needs more discussion. @fujitatomoya or @mjcarroll what do you think?

fujitatomoya:
@wjwwood i agree, maybe we can discuss what the original problem is before considering the fixes.

fujitatomoya:
@Mergifyio rebase

fujitatomoya:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30134/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22865/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9981/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29018/)

PR #1173, opened 2024-07-25
Fix a memory leak in test_rcl_lifecycle.

clalancette:
This one came about probably as a result of a bad merge. But essentially we were forcing the srv_change_state com_interface to be nullptr, but forgetting to save off the old pointer early enough.  Thus, we could never restore the old one before we went to "fini", and the memory would be leaked.  Fix this by remembering the impl pointer earlier.


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21555/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15840/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1052/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22273/)

fujitatomoya (approved):
good eye 👍 lgtm with green CI.

ahcorde (approved):

PR #1176, opened 2024-08-12
Update wait.h documentation for rcl_wait

FelixPe:
In documentation of rcl_wait add context as parameter for rcl_wait_set_init.


mjcarroll (approved):
Doc changes only, CI not needed.

PR #1202, opened 2024-12-24
Clean up error handling in many rcl{_action,_lifecycle} codepaths

clalancette:
The original impetus from this PR came from running the `rcl` tests, and seeing output like the following on the console:

```

  'failed to reallocate memory for new transitions, at /home/ubuntu/rolling_ws/src/ros2/rcl/rcl_lifecycle/src/transition_map.c:152, at /home/ubuntu/rolling_ws/src/ros2/rcl/rcl_lifecycle/src/default_state_machine.c:725'

with this new error message:

  'failed to reallocate memory for new transitions on state, at /home/ubuntu/rolling_ws/src/ros2/rcl/rcl_lifecycle/src/transition_map.c:169'

<<<
```

There are 2 main reasons why this can happen:
1.  The test itself is forcing an error condition, but then forgetting to call `rcl_reset_error()` before trying another condition.
2.  An error happens in a "lower-layer" method (like rcutils), which sets an error message, and then the rcl* layer also tries to set the error message.

This PR corrects almost all instances of this occurence in the rcl libraries, with the exception of `test_arguments`, which has quite a few of these and will require a separate PR.  Not only will this quite down many of the tests, it also fixes some bugs in error handling in this code.

### DIFF
--- a/rcl/src/rcl/publisher.c
+++ b/rcl/src/rcl/publisher.c
@@ -421,7 +421,11 @@ rcl_publisher_is_valid(const rcl_publisher_t * publisher)
     return false;  // error already set
   }
   if (!rcl_context_is_valid(publisher->impl->context)) {
-    RCL_SET_ERROR_MSG("publisher's context is invalid");
+    if (!rcl_error_is_set()) {
+      // rcl_context_is_valid can return false both in the error case, and when the context
+      // hasn't been initialized.  It will only set the error message in the first case.
+      RCL_SET_ERROR_MSG("publisher's context is invalid");
+    }
     return false;
   }
   RCL_CHECK_FOR_NULL_WITH_MSG(
--- a/rcl/src/rcl/subscription.c
+++ b/rcl/src/rcl/subscription.c
@@ -100,8 +100,11 @@ rcl_subscription_init(
     1, sizeof(rcl_subscription_impl_t), allocator->state);
   RCL_CHECK_FOR_NULL_WITH_MSG(
     subscription->impl, "allocating memory failed", ret = RCL_RET_BAD_ALLOC; goto cleanup);
+
+  // options
+  subscription->impl->options = *options;
+
   // Fill out the implemenation struct.
-  // rmw_handle
   // TODO(wjwwood): pass allocator once supported in rmw api.
   subscription->impl->rmw_handle = rmw_create_subscription(
     rcl_node_get_rmw_handle(node),
@@ -123,8 +126,6 @@ rcl_subscription_init(
   }
   subscription->impl->actual_qos.avoid_ros_namespace_conventions =
     options->qos.avoid_ros_namespace_conventions;
-  // options
-  subscription->impl->options = *options;
 
   if (RCL_RET_OK != rcl_node_type_cache_register_type(
       node, type_support->get_type_hash_func(type_support),
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -1213,10 +1213,10 @@ _rcl_parse_resource_match_token(rcl_lexer_lookahead2_t * lex_lookahead)
     ret = rcl_lexer_lookahead2_accept(lex_lookahead, NULL, NULL);
   } else if (RCL_LEXEME_WILD_ONE == lexeme) {
     RCL_SET_ERROR_MSG("Wildcard '*' is not implemented");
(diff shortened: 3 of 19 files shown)


clalancette:
Pulls: https://github.com/ros2/rcl/pull/1202
Gist: https://gist.githubusercontent.com/clalancette/e3f7e4a8739d5a2fe4825922491b9561/raw/01d769156a1c9b934554eac77efc9aa4fb529cff/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14995
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22309/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16572/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1840/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23100/)

rcl/src/rcl/publisher.c
```diff
@@ -421,7 +421,11 @@ rcl_publisher_is_valid(const rcl_publisher_t * publisher)
     return false;  // error already set
   }
   if (!rcl_context_is_valid(publisher->impl->context)) {
-    RCL_SET_ERROR_MSG("publisher's context is invalid");
+    if (!rcl_error_is_set()) {
+      // rcl_context_is_valid can return false both in the error case, and when the context
+      // hasn't been initialized.  It will only set the error message in the first case.
+      RCL_SET_ERROR_MSG("publisher's context is invalid");
+    }
```
fujitatomoya, line 428: i would do the following only to use the designed interfaces. (relying on the implementation behavior could be changed, for example we could add the error message if the context id is zero in the future? that breaks this if statement.)

```suggestion
    RCL_SET_ERROR_MSG("publisher's context is either invalid or uninitialized");
```

rcl/src/rcl/subscription.c
```diff
@@ -100,8 +100,11 @@ rcl_subscription_init(
     1, sizeof(rcl_subscription_impl_t), allocator->state);
   RCL_CHECK_FOR_NULL_WITH_MSG(
     subscription->impl, "allocating memory failed", ret = RCL_RET_BAD_ALLOC; goto cleanup);
+
+  // options
+  subscription->impl->options = *options;
```
fujitatomoya, line 105: i think originally impl structure is created after `rmw_handle` is created successfully. if we take this change, maybe the same change can be added to `rcl_publisher_init`?

fujitatomoya (commented):
lgtm, but a couple of comments.

rcl/src/rcl/publisher.c
clalancette: Unfortunately, we can't do this and still get rid of the warning.  That's because `rcl_context_is_valid` can set an error if, for example, the `context` argument is NULL.  In that case, it will return with `false`, but also have set the error message, so setting it again here leads to the "overwritten" warning message.  Unfortunately, `rcl_context_is_valid` can *also* return `false` in the case where the context has already been shutdown.  In that case, it does *not* set the error, and returns `false`.  In both cases we want to fail here, but we don't necessarily want to overwrite the error message.  That's why we have this funky code.

(I'll also note that this is what https://github.com/ros2/rcl/pull/1086 was trying to fix, but that was too difficult to do)

rcl/src/rcl/subscription.c
clalancette: I moved this earlier because in the case that `rmw_create_subscription` fails, we try to run `rcl_subscription_options_fini` on the uninitialized structure, which returns its own warning for an uninitialized structure.  By copying it earlier, we avoid that in cleanup paths.

For the publisher, we have no equivalent of `rcl_subscription_options_fini`; it is a POD type.  So when we copy it doesn't really matter.  I don't have a strong opinion either way, to be honest.

rcl/src/rcl/publisher.c
fujitatomoya: thanks for the explanation.

rcl/src/rcl/subscription.c
fujitatomoya: i see. i think that is okay with current code.

fujitatomoya (approved):

PR #1219, opened 2025-03-23
Fix typo in message header include in doc

christophebedard:


fujitatomoya (approved):

fujitatomoya:
@christophebedard ah sorry, i should have run CI? this is just a doc section removing `s`, i hope that should be okay.

christophebedard:
We could've relied on the Rpr job, but they all seem to be disabled currently.

Should be fine, but, just to be safe, I can run a single Linux job and only test `rcl`: [Build Status](https://ci.ros2.org/job/ci_linux/22847/)

Issue #991, opened 2022-07-04
Make rcl_lifecycle transition errors more readable


alsora:
## Feature request

Currently rcl_lifecycle transition errors are printed as follows in a C++ application
```
[rcl_lifecycle]: No transition matching 3 found for current state unconfigured
[]: Unable to start transition 3 from current state unconfigured: Transition is not registered.
```

It would be very helpful if rather than a number `3` there was the actual name of the transition being invoked, to avoid developers a constant lookup into the message definition (which is not even linked, so this error may seem extremely confusing for not-experienced people)

CursedRock17:
Can I ask what the context of your transition error is? I know there are two methods which return the first message of "No transition matching 3", this is the method `rcl_lifecycle_get_transition_by_id` that takes in a [`uint8_t id`](https://github.com/ros2/rcl/blob/rolling/rcl_lifecycle/src/rcl_lifecycle.c#L327-L346). Then, there is method, `rcl_lifecycle_get_transition_by_label` that takes in[`const char * label`](https://github.com/ros2/rcl/blob/rolling/rcl_lifecycle/src/rcl_lifecycle.c#L306-L325). These methods are carried underneath multiple functions when trying to get the state of any transition, so there's a good chance that it gets lost as an state id somewhere, in your case.

fujitatomoya:
@CursedRock17 
CC: @alsora 

i think @alsora just wants the message like

```bash
[rcl_lifecycle]: No transition matching "activate" (ID:3) found for current state unconfigured
```


i think that this is `nice-to-have`.

label and id map can be referred to `state_machine.transition_map`, but this requires the API change for `rcl_lifecycle_get_transition_by_label` and `rcl_lifecycle_get_transition_by_id` to pass the `state_machine`. if that is okay, we can add `rcl_lifecycle_get_transition_label_by_id` and `rcl_lifecycle_get_transition_id_by_label` from the `state_machine.transition_map`, and print both of them.

but i am not convinced for not major enhancement with breaking API... and it is easy to look it up with https://github.com/ros2/rcl_interfaces/blob/rolling/lifecycle_msgs/msg/Transition.msg#L22

We do not want to move the warning message either for this.

alsora:
> i think @alsora just wants the message like
[rcl_lifecycle]: No transition matching "activate" (ID:3) found for current state unconfigured

yes exactly.

[...]

PR #1229, opened 2025-04-28
introduce rcl_lifecycle_get_transition_label_by_id().

fujitatomoya:
part of https://github.com/ros2/rcl/issues/991

### DIFF
--- a/rcl_lifecycle/include/rcl_lifecycle/rcl_lifecycle.h
+++ b/rcl_lifecycle/include/rcl_lifecycle/rcl_lifecycle.h
@@ -351,6 +351,30 @@ rcl_lifecycle_get_transition_by_label(
   const rcl_lifecycle_state_t * state,
   const char * label);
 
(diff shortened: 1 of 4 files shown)


fujitatomoya (commented):
I was originally going to deprecate `rcl_lifecycle_get_transition_by_id`, and look up the label based on the identification and call `rcl_lifecycle_get_transition_by_label` only. but we cannot do that, because id and label are not mapped identically in the transition map. (there can be same label for different identification)

so what we can do here is that introduce `rcl_lifecycle_get_transition_label_by_id` tries to get the label based on the identification from the transition map, and we still call `rcl_lifecycle_get_transition_by_id` with user specified id. with this, we can at least print the label based on the id as information.

since this bothers me a bit, i just went ahead to patch it for myself.

fujitatomoya:
@alsora can you review this?

alsora (approved):

PR #1238, opened 2025-05-14
remove unnecessary test_with_localhost_only.

fujitatomoya:
i think this test is redundant and `LOCALHOST_ONLY` environmental variable has been removed already. so we can just remove this test case for the efficiency.

### DIFF
--- a/rcl/test/rcl/test_discovery_options.cpp
+++ b/rcl/test/rcl/test_discovery_options.cpp
@@ -275,26 +275,3 @@ TEST(TestDiscoveryInfo, test_get_both) {
   EXPECT_EQ(0u, discovery_options_var.static_peers_count);
   EXPECT_EQ(RCL_RET_OK, rmw_discovery_options_fini(&discovery_options_var));
 }
-
-// localhost_only is deprecated but still honored to prevail discovery_options.
-// see https://github.com/ros2/ros2_documentation/pull/3519#discussion_r1186541935
-// TODO(fujitatomoya): remove localhost_only completely after deprecation period.
-TEST(TestDiscoveryInfo, test_with_localhost_only) {
-  {
-    // No environment variable set (default subnet, no specific peers)
-    check_discovery(RMW_AUTOMATIC_DISCOVERY_RANGE_SUBNET, 0);
-  }
-
-  {
-    // only ROS_AUTOMATIC_DISCOVERY_RANGE and ROS_STATIC_PEERS set
-    ASSERT_TRUE(rcutils_set_env("ROS_AUTOMATIC_DISCOVERY_RANGE", "LOCALHOST"));
-    ASSERT_TRUE(rcutils_set_env("ROS_STATIC_PEERS", "127.0.0.1;localhost.com"));
-    check_discovery(RMW_AUTOMATIC_DISCOVERY_RANGE_LOCALHOST, 2);
-  }
-
-  {
-    ASSERT_TRUE(rcutils_set_env("ROS_AUTOMATIC_DISCOVERY_RANGE", "SUBNET"));
-    ASSERT_TRUE(rcutils_set_env("ROS_STATIC_PEERS", "192.168.0.1;remote.com"));
-    check_discovery(RMW_AUTOMATIC_DISCOVERY_RANGE_SUBNET, 2);
-  }
-}


fujitatomoya:
Pulls: ros2/rcl#1238
Gist: https://gist.githubusercontent.com/fujitatomoya/9ebbc85eae3ae4d0fff7a5cd1c00e9a9/raw/17a405510ccb17e227ad680304b4b0474c27da48/ros2.repos
BUILD args:  --packages-up-to rcl
TEST args:  --packages-select rcl
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15980
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23616/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17724/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3069/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24341/)

fujitatomoya:
@jmachowinski can you review this?

ahcorde (approved):

jmachowinski:
Is there some warning for the end user if LOCALHOST_ONLY is set ?

fujitatomoya:
@jmachowinski it had been until jazzy, but `LOCALHOST_ONLY` had been removed already for kilted, so not any more. (as system behavior, this has been done some time ago.) this PR just removes left-over test.

jmachowinski:
People often jump version. Meaning that people jumping from humble to L whatever will never see a warning that their environment variable has no effect any more...

fujitatomoya:
that suggests we should change the deprecation period then? maybe LTS release basis? I feel like it is kind of long time, we need 4 years to remove the APIs. (2 years deprecation, 2 years to remove)... 🤔

jmachowinski:
I think the code API deprecation is fine as the compiler will direct tell you that stuff is broken.

This case however is different as there is no tooling to tell you that something is broken.
The change is very subtle for the end user, but has a high impact. 
In Foxy / Humble
```
export ROS_LOCALHOST_ONLY=1
``` 
Worked as expected. Google search points to the ancient documentation, and about every
stackoverflow post suggests to do this. 

If there is no warning at all, that this env variable has no effect any more, debugging this will be hell.
I would actually propose to keep the warning for environment variables that don't have any effect any
more forever...

fujitatomoya:
that is good question.
i think that is why there is deprecation period to say that `ROS_LOCALHOST_ONLY` is deprecated and removed, use dynamic discovery range. but once it is gone/removed after deprecation period, if there are still users use `ROS_LOCALHOST_ONLY`, that silently skip the configuration.

maybe what we should have done is to enforce the error that tells user `ROS_LOCALHOST_ONLY` is no valid anymore?

> Google search points to the ancient documentation, and about every stackoverflow post suggests to do this.

true, even ChatGPT recommends `ROS_LOCALHOST_ONLY` instead of https://docs.ros.org/en/rolling/Tutorials/Advanced/Improved-Dynamic-Discovery.html

jmachowinski:
We talked in the PMC meeting about this, and the consensus was to keep warnings about unused env variables forever. We should therefore readd the warning.

fujitatomoya:
@jmachowinski thanks for bringing this to PMC meeting, and sorry that i missed the today's meeting.

i am not sure what was discussed in the meeting from the minute, but i disagree on keeping the deprecated warnings for unused env variables forever... you mean enforcing error instead? i may be missing some context...

I do have several questions.

- I assume keeping the warnings means that deprecated env variables are still effective, correct? at least, that is how warning works and developer expects. so we need to keep the deprecated warnings and env variable(still effective) forever?
- If that is so, why do we even need the warning in the 1st place? it will be just a noise for user application, because we will never remove those? if we really take this path, i do not think we need the warning at all.
- if we keep the deprecated warnings and env variables forever, that could compile up many warnings and environmental variables? could be duplication and similar things, annoying code maintainability... i think it is also gonna be hard to add the environmental variables.

IMO, the problem here is that we did not enforce the user to stop using the removed environmental variables after the deprecation period. so that, once it is removed after deprecation period with warning, if there are still users use ROS_LOCALHOST_ONLY, that silently skip the configuration.

what i do suggest against this problem is,

- we can deprecate / remove the environmental variable. (will not keep them forever...)
- deprecation warning should be given for one distribution cycle, enforcement error should be give for one distribution cycle. (that saves the user who migrate the ROS 2 system from LTS-old to LTS-new.)

What if a user migrates their ROS 2 system across more than two distributions, such as from Humble to Kilted? that i do not have the answer... I would like to have feedback on this. if we cannot accept this case, we might need to keep the enforcement error forever?

jmachowinski:
Okay, I think warning means something different for you.

All I want is an array of deprecated environment variables somewhere in rcl, and a loop checking if they are set.
If any deprecated environment variable is set, a warning message that an outdated variable is set should be shown on the console. 
Making this an error and stopping the node, may be better though.

That's it. The goal here is to give anybody coming from an older ros2 version a fighting chance to figure out why nothing is working any more as expected.

This is the equivalent, of checking, if any of the command line arguments of the node process is bogus.

fujitatomoya:
@jmachowinski thanks for the explanation (including offline discussion). now i see what needs to be done here.

i created the issue to track this bug fix? minor feature? here https://github.com/ros2/rcl/issues/1241.

PR #1260, opened 2025-09-29
Refer to 'the middleware' and not 'the DDS implementation' in doc

christophebedard:
## Description

To be a bit more abstract.


fujitatomoya (approved):

christophebedard:

christophebedard:
I think the Rpr job itself is enough, especially since this makes the line shorter.

christophebedard:
The only Rpr test failure is `test_rcl_get_publishers_subscription_info_by_topic`, and I doubt that's related to this change, so I'll merge this.

PR #1278, opened 2025-12-15
Remove default from switch with enum, so that compiler warns.

fujitatomoya:
## Description

Now the compiler will warn us if new enum values are added to it but aren't handled in this function.

Fixes # (issue)

### Is this user-facing behavior change?

No

### Did you use Generative AI?

No

### Additional Information

> [!NOTE]
> No backports requested, this is just a minor enhancement for maintainability.

PR #1279, opened 2025-12-17
parse_key() should use yaml_map_lvl_t instead of uint_32.

fujitatomoya:
## Description

Follow up of https://github.com/ros2/rcl/pull/1278

https://github.com/ros2/rcl/pull/1278 exposes https://github.com/ros2/rmw_zenoh/pull/871#pullrequestreview-3588307809 failure but we missed to test there. the proper fix is that, it should use enum data type for the switch statement, so that the compiler can always give us warning during build time.

Fixes # (issue)

Technically yes, because this changes the parse_key function argument signature.

No

### Additional Information

> [!NOTE]
> DO NOT BACKPORT THIS.

### DIFF
--- a/rcl_yaml_param_parser/src/impl/parse.h
+++ b/rcl_yaml_param_parser/src/impl/parse.h
@@ -53,7 +53,7 @@ RCL_YAML_PARAM_PARSER_PUBLIC
 RCUTILS_WARN_UNUSED
 rcutils_ret_t parse_key(
   const yaml_event_t event,
-  uint32_t * map_level,
+  yaml_map_lvl_t * map_level,
   bool * is_new_map,
   size_t * node_idx,
   size_t * parameter_idx,
--- a/rcl_yaml_param_parser/src/parse.c
+++ b/rcl_yaml_param_parser/src/parse.c
@@ -747,7 +747,7 @@ _get_float_value(
 ///
 rcutils_ret_t parse_key(
   const yaml_event_t event,
-  uint32_t * map_level,
+  yaml_map_lvl_t * map_level,
   bool * is_new_map,
   size_t * node_idx,
   size_t * parameter_idx,
@@ -901,7 +901,7 @@ rcutils_ret_t parse_file_events(
   bool is_seq = false;
   uint32_t line_num = 0;
   data_types_t seq_data_type = DATA_TYPE_UNKNOWN;
-  uint32_t map_level = 1U;
+  yaml_map_lvl_t map_level = MAP_NODE_NAME_LVL;
   uint32_t map_depth = 0U;
   bool is_new_map = false;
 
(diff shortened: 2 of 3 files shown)


rcl_yaml_param_parser/src/impl/parse.h
```diff
@@ -53,7 +53,7 @@ RCL_YAML_PARAM_PARSER_PUBLIC
 RCUTILS_WARN_UNUSED
 rcutils_ret_t parse_key(
   const yaml_event_t event,
-  uint32_t * map_level,
+  yaml_map_lvl_t * map_level,
```
fujitatomoya, line 56: This does never need to be `uint32_t`, but `yaml_map_lvl_t`. this allows the compiler to give us the warning, and we do not even do the test for the invalid argument.

rcl_yaml_param_parser/test/test_parse.cpp
```diff
@@ -436,16 +436,6 @@ TEST(TestParse, parse_key_bad_args)
   EXPECT_TRUE(rcutils_error_is_set());
   rcutils_reset_error();
 
-  // map_level is not a valid value
-  map_level = 42;
```
fujitatomoya, line 440: Now this can be detected by the compiler during build because it checks the enum data, that is why we do not even need to have this test after all.

fujitatomoya:
Pulls: ros2/rcl#1279
Gist: https://gist.githubusercontent.com/fujitatomoya/9ef5113d3b832a77fd680b3834e1c50f/raw/ca9530b117a7ce4b9fdf72f38d54a17def0263b7/ros2.repos
BUILD args:  --packages-above-and-dependencies rcl rcl_yaml_param_parser
TEST args:  --packages-above rcl rcl_yaml_param_parser
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17821
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/26686/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20260/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6487/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26269/)

rcl_yaml_param_parser/src/parse.c
```diff
@@ -747,7 +747,7 @@ _get_float_value(
 ///
 rcutils_ret_t parse_key(
   const yaml_event_t event,
-  uint32_t * map_level,
+  yaml_map_lvl_t * map_level,
```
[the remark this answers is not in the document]
fujitatomoya: i think this is opposite, if we are adding the default case, the compiler cannot give us warning even if we are missing some cases and handled in default case. that would be more difficult to detect at runtime debug. using proper enum type in switch statement allows the compiler to check this for the user.

fujitatomoya:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/26688/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20261/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6488/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26270/)

ahcorde (approved):

Issue #1277, opened 2025-12-09
rcl_wait() versus rcl_service_set_on_new_request_callback() behaviour and expected pattern


tho-:
Hi, I am noticing an unexpected behaviour when combining rcl_wait() on a service handle and a callback for the same service set with rcl_service_set_on_new_request_callback(). My issue is that rcl_wait() returns immediately with no delay, no error and no event in the wait set once a first request for a service has been received.

Here is a pseudo-code example of what I am doing:
```
void callback(void *data, size_t n)
{
  rmw_service_info_t info;

  rc = rcl_take_request_with_info(&svc, &info, ...);
  ...
  rc = rcl_send_response(&svc, &info.request_id, ...);
}

int main()
{
  ...
  rc = rcl_service_init(&svc, node, ...);
  rc = rcl_service_set_on_new_request_callback(&svc, callback, ...);
  ...
  while(rcl_context_is_valid(ctx)) {
    rc = rcl_wait_set_clear(&wait_set);
    rc = rcl_wait_set_add_service(&svc, ...);
    rc = rcl_wait(&wait_set, timeout_10s);
    /* intially waits correctly until the first request, callback is processed */
    /* then always triggered immediately after the first request, no events returned in the wait_set */
  }
}
```

I could track down the wake up of rcl_wait down to the fastdds implementation (which indeed has some wake up condition true, which should not be the case until another service request is received). And for some reason this wake up condition does not make it to the events in the wait_set (all events are NULL).

If instead I do not add the service in the wait set, using e.g. only some unrelated guard condition, the service callback is still executed properly and the rcl_wait call blocks normally until e.g. the timeout is reached.

Since I use a service callback, there is indeed no need to wait for the service in the wait_set, but  I was wondering if the above coding pattern is expected to fail and thus forbidden (and this maybe should be documented) or if it's some unexpected issue that may be fixed (I tested only with fastdds so far).

Thanks for your insights,
Best, Anthony

kscottz:
We discussed this in our weekly project meeting. Can you bring your issue to the next [Client Library Working Group Meeting on January 9th](https://docs.ros.org/en/jazzy/The-ROS2-Project/Governance.html) so we can discuss? [You have been added to the agenda](https://docs.google.com/document/d/1-tU2AKoHKLCVHPBz0QchYownmZZvYjHmVUQXT2WBMOw/edit?usp=sharing).

jmachowinski:
[...]

PR #1289, opened 2026-02-12
Improved documentation of rcl_XYZ_set_on_new_XYZ_callback

### LINKED ISSUE #1282: Improve documentation of rcl_XYZ_set_on_new_XYZ_callback
The documentation of the rcl_XYZ_set_on_new_XYZ_callback does not state the calling thread context, 
and the need to not do blocking operations in the callbacks. 

Here is an example of the documentaiton from the rclcpp layer for the subscription case:

```c++
  /// Set a callback to be called when each new message is received.
  /**
   * The callback receives a size_t which is the number of messages received
   * since the last time this callback was called.
   * Normally this is 1, but can be > 1 if messages were received before any
   * callback was set.
   *
   * Since this callback is called from the middleware, you should aim to make
   * it fast and not blocking.
   * If you need to do a lot of work or wait for some other event, you should
   * spin it off to another thread, otherwise you risk blocking the middleware.
   *
   * Calling it again will clear any previously set callback.
   *
   * This function is thread-safe.
   *
   * If you want more information available in the callback, like the subscription
   * or other information, you may use a lambda with captures or std::bind.
   *
   * \sa rmw_subscription_set_on_new_message_callback
   * \sa rcl_subscription_set_on_new_message_callback
   *
   * \param[in] callback functor to be called when a new message is received
   */
``` 

The task would be to update the rcl and rmw documenation for the functions.

Relates https://github.com/ros2/rcl/issues/1277

rushsahay:
## Description
Signed off: Rushhaank Sahay rushhaank.sahay@gmail.com
Updates the documentation for rcl_service_set_on_new_request_callback and rcl_subscription_set_on_new_message_callback methods to encourage users to make the process non-blocking and warn them of unexpected behavior if it is blocked

Fixes #1282 
Addresses #1277 

Yes. It changes user behavior as it specifies to users the proper way to use these methods to avoid unexpected behaviors.

No

### DIFF
--- a/rcl/include/rcl/client.h
+++ b/rcl/include/rcl/client.h
@@ -477,14 +477,14 @@ rcl_client_response_subscription_get_actual_qos(const rcl_client_t * client);
  *
  * Since this callback is called from the middleware, you should
  * aim to make it fast and not blocking. This callback is intended to implement an event driven executor and
- * not process data directly.  
- * 
+ * not process data directly.
+ *
  * Doing work in this callback can cause delays,
  * deadlocks, or latency due to cross thread waiting
  * as this process runs on middleware managed threads
- * and is meant only to notify the executor that new data 
+ * and is meant only to notify the executor that new data
  * is available in the middleware queue
- * 
+ *
  * <hr>
  * Attribute          | Adherence
  * ------------------ | -------------
--- a/rcl/include/rcl/client.h
+++ b/rcl/include/rcl/client.h
@@ -479,9 +479,9 @@ rcl_client_response_subscription_get_actual_qos(const rcl_client_t * client);
  * aim to make it fast and not blocking. This callback is intended to implement an event driven executor and
  * not process data directly.  
  * 
- * Using blocking operators, waiting for other syncronized actions, 
- * or sending responses directly using this callback will result in
- * the thread being called from an unexpected/different thread context.
+ * Blocking or performing synchronous work here may cause subsequent callbacks
+ * or responses to execute in executor threads different than the current 
+ * middleware thread.
  * 
  * <hr>
  * Attribute          | Adherence
--- a/rcl/include/rcl/client.h
+++ b/rcl/include/rcl/client.h
@@ -476,14 +476,15 @@ rcl_client_response_subscription_get_actual_qos(const rcl_client_t * client);
  * \sa rmw_client_set_on_new_response_callback for details about this function.
  *
  * Since this callback is called from the middleware, you should
- * aim to make it fast and not blocking. This callback is intended to implement an event driven executor and
+ * aim to make it fast and not blocking. This callback
+ * is intended to implement an event driven executor and
  * not process data directly.
  *
  * Doing work in this callback can cause delays,
  * deadlocks, or latency due to cross thread waiting
  * as this process runs on middleware managed threads
  * and is meant only to notify the executor that new data
- * is available in the middleware queue
+ * is available in the middleware queue.
  *
  * <hr>
  * Attribute          | Adherence
--- a/rcl/include/rcl/client.h
+++ b/rcl/include/rcl/client.h
@@ -479,9 +479,11 @@ rcl_client_response_subscription_get_actual_qos(const rcl_client_t * client);
  * aim to make it fast and not blocking. This callback is intended to implement an event driven executor and
  * not process data directly.  
  * 
- * Blocking or performing synchronous work here may cause subsequent callbacks
- * or responses to execute in executor threads different than the current 
- * middleware thread.
+ * Doing work in this callback can cause delays,
+ * deadlocks, or latency due to cross thread waiting
+ * as this process runs on middleware managed threads
+ * and is meant only to notify the executor that new data 
+ * is available in the middleware queue
  * 
  * <hr>
  * Attribute          | Adherence
--- a/rcl/include/rcl/client.h
+++ b/rcl/include/rcl/client.h
@@ -475,6 +475,14 @@ rcl_client_response_subscription_get_actual_qos(const rcl_client_t * client);
  *
  * \sa rmw_client_set_on_new_response_callback for details about this function.
  *
+ * Since this callback is called from the middleware, you should
+ * aim to make it fast and not blocking. This callback is intended to implement an event driven executor and
+ * not process data directly.  
+ * 
+ * Using blocking operators, waiting for other syncronized actions, 
+ * or sending responses directly using this callback will result in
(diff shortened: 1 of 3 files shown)


fujitatomoya (commented):
what about `rcl_client_set_on_new_response_callback`? i think that needs to be updated accordingly?

@jmachowinski can you review this?

rushsahay:
Thanks for spotting that! Have added the description to that method as well. Is there a reason why the "Rpr__rcl__ubuntu_noble_amd64" check is failing on this pull request? Let me know if I need to do something else to ensure it passes.

rcl/include/rcl/client.h
```diff
@@ -475,6 +475,14 @@ rcl_client_response_subscription_get_actual_qos(const rcl_client_t * client);
  *
  * \sa rmw_client_set_on_new_response_callback for details about this function.
  *
+ * Since this callback is called from the middleware, you should
+ * aim to make it fast and not blocking. This callback is intended to implement an event driven executor and
+ * not process data directly.  
+ * 
+ * Using blocking operators, waiting for other syncronized actions, 
+ * or sending responses directly using this callback will result in
+ * unexpected behavior
```
jmachowinski: I would not call this unexpected behavior, it is pretty known what will happen, you block the middleware threads. It should be mentioned that these callback will be called from a different / unexpected thread context.

rushsahay: Have reworded to make this clearer.

fujitatomoya:
@rushsahay DCO is missing, can you `--signoff` the commit?

jmachowinski:
The new text does not make any sense.

jmachowinski:
The text still does not make any sense, @rushsahay could you explain here in the comments what you think is going on ?
Alternative join the next client workgroup meeting to discuss this.

rushsahay:
In my head, the purpose of the method is to notify the executor of work that is available, so when this is blocked it can lead to latency or cause deadlocks if waiting on work to be done on the executor. Therefore, when this process is blocked then there is executor dependent work being done in the middleware threads that violates the separation between middleware and executor threads leading to a different thread context. Let me know if I'm missing something here and what can be added to make the documentation on this clearer.

jmachowinski:
Rethink this a bit, the middleware spawns its own threads, that block on network sockets and after they receive a full message, they push it into a queue and call the callbacks that are been set here.

Totally independent from this, there is the executor, running in the main thread, that fetches messages from the middleware queue calls the subscription, service etc callbacks with them. 

Currently there are two mechanisms for an executor the notified about new data been available in the middleware queues: 
A polling based one (the waitset approach) and the callback / events one.

With this background information, what would you think is critical for these callbacks regarding thread safety and why does it matter in the first place ?

rushsahay:
With this context, it seems important that the callbacks stay thread-safe because they run on threads managed by the middleware, while the executor processes messages on a different thread. It seems that the callback's job is just to notify the executor that new data is available in the queue rather than actually process the information. Doing work in these callbacks can cause delays and other possible scenarios like deadlocks, where both the executor and middleware threads are waiting on each other, or latency due to cross thread waiting which is why this matters in the first place.

jmachowinski:
With this new context, can you now re-read your added documentation and try to improve it ?

jmachowinski:
Pulls: ros2/rcl#1289
Gist: https://gist.githubusercontent.com/jmachowinski/833c64052b3549c84f4560aaf2876fd2/raw/30f241fa2b38df05de5b049dc0284e93fa452706/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18493
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28056/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21029/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7908/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26999/)

jmachowinski (approved):

fujitatomoya:
@rushsahay this PR has a lot of conflicts, can you rebase and resolve the conflicts?

jmachowinski:
@rushsahay you would also need to fix the uncrustify warnings to move this PR forward

rushsahay:
Have made these changes. Thanks again for the feedback. Let me know if any other changes are needed.

rcl/include/rcl/service.h
```diff
@@ -507,6 +507,16 @@ rcl_service_response_publisher_get_actual_qos(const rcl_service_t * service);
  *
  * \sa rmw_service_set_on_new_request_callback for details about this function.
  *
+ * Since this callback is called from the middleware, you should
+ * aim to make it fast and not blocking. This callback is intended to implement an event driven executor and
```
fujitatomoya: this line has more than 100 characters in a line, guess this will be picked up by checker.

rcl/include/rcl/service.h
```diff
@@ -507,6 +507,16 @@ rcl_service_response_publisher_get_actual_qos(const rcl_service_t * service);
...
+ * not process data directly.
+ *
+ * Doing work in this callback can cause delays,
+ * deadlocks, or latency due to cross thread waiting
+ * as this process runs on middleware managed threads
+ * and is meant only to notify the executor that new data
+ * is available in the middleware queue
```
fujitatomoya: missing dot?

```suggestion
 * is available in the middleware queue.
```

fujitatomoya (changes requested):
the same comments apply to all files.

rushsahay:
@fujitatomoya Have made the changes.

fujitatomoya (approved):

fujitatomoya:
Pulls: ros2/rcl#1289
Gist: https://gist.githubusercontent.com/fujitatomoya/c06bf95215516905abecde039c62a033/raw/08810e4a4019d589126b78109d71c6b19d503e4c/ros2.repos
BUILD args:  --packages-up-to rcl
TEST args:  --packages-select rcl
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18855
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28575/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21420/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8422/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27441/)

PR #1292, opened 2026-02-23
Use the POSIX locale to parse YAML double

### LINKED ISSUE #1288: rcl_yaml_param_parser is locale-dependent
### Generated by Generative AI

_No response_

### Operating System:

```Linux ThinkPad-P53 6.8.0-94-generic #96-Ubuntu SMP PREEMPT_DYNAMIC Fri Jan  9 20:36:55 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux```

jazzy

_No response_

### RMW Configuration (if applicable):

_No response_

### Client library (if applicable):

rcl, rclcpp

### 'ros2 doctor --report' output

_No response_

### Steps to reproduce issue

1. Create a C or a C++ program using rcl, rclcpp or rclpy, set the global LC_NUMERIC locale to something like "fr_FR.UTF-8" or "de_DE.UTF-8".
2. Run the program with something like `--ros-args -p my_param:=1.123` and try to read the double value for my_param

### Expected behavior

The override value for my_param is a double with value 1.123

### Actual behavior

The override value for my_param is a string.

### Additional information

rcl_yaml_param_parser's [`parse_value`](https://github.com/ros2/rcl/blob/834ebb85581f6ee118a11c90cefceaa7a27bd18c/rcl_yaml_param_parser/src/parse.c#L721) is using strtod, which is locale-dependent, and thus fails if the locale use anything else than a dot for the decimal parser. This makes it a non compliant YAML parser, as YAML clearly states the [dot is the decimal separator](https://yaml.org/spec/1.2.2/#10214-floating-point).

Usually, a C/C++ program use the minimum "C" locale, but Qt programs, for example, use the user-preferred locale. That is the case for rviz2.

This caused the following issues:
- https://github.com/moveit/moveit2_tutorials/issues/508
- https://github.com/moveit/moveit2/issues/1049
- https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/723
- And more

I see a few solutions:
1. Use strtod_l, but it is not in the C stantard (it's a GNU extension).
2. strdup the input string, search and replace the decimal separator using `localeconv()`.
3. Use another implementation of double parsing I don't know about.
4. Change https://docs.ros.org/en/rolling/Installation/Ubuntu-Install-Debians.html#set-locale to remove "it should be fine if you’re using a different UTF-8 supported locale." and make it clear we need a english-like LC_NUMERIC. I don't like this approach.

### PR DESCRIPTION
**Pull request by Hugal31**
## Description

Call `strtod` surrounded with `uselocale(POSIX-locale)`, `uselocale(previous-locale)` to parse double following the YAML standard even if the LC_NUMERIC locale do not uses dots as the decimal separator.

Fixes #1288

User using a LC_NUMERIC locale with dot as the decimal separator will not see any changes. Users using other LC_NUMERIC locale will not have issues parsing doubles with `--ros-args`.

No.

### Additional Information

I am not super happy to have to instantiate a new locale and store it in a static variable. I would have preferred a locale-independent parsing function, but I did not find one without using GNU extension, add another dependency or use the C++ library.

Performance-wise, this adds a few nanoseconds to the double parsing. Moreover, in RCL, get_value this is not in the hot-path.

### DIFF
--- a/rcl_yaml_param_parser/src/parse.c
+++ b/rcl_yaml_param_parser/src/parse.c
@@ -14,6 +14,7 @@
 
 #include <ctype.h>
 #include <errno.h>
+#include <locale.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -779,6 +780,20 @@ _get_int_value(
   return RCUTILS_RET_ERROR;
 }
 
+///
+/// Calls strtod with the default "POSIX/C" locale.
+///
+static double strtod_posix_locale(const char *restrict nptr, char **restrict endptr)
+{
+    static locale_t c_locale = 0;
+    if (c_locale == 0)
+        c_locale = newlocale(LC_ALL_MASK, "C", 0);
+    locale_t old_locale = uselocale(c_locale);
+    double result = strtod(nptr, endptr);
+    uselocale(old_locale);
+    return result;
+}
+
 rcutils_ret_t
 _get_float_value(
   const char * const value,
@@ -807,12 +822,12 @@ _get_float_value(
     for (iter_ptr = value; !isalpha(*iter_ptr); ) {
       iter_ptr += 1;
     }
-    dval = strtod(iter_ptr, &endptr);
+    dval = strtod_posix_locale(iter_ptr, &endptr);
     if (*value == '-') {
       dval = -dval;
     }
   } else {
-    dval = strtod(value, &endptr);
+    dval = strtod_posix_locale(value, &endptr);
   }
   if ((0 == errno) && (NULL != endptr)) {
     if ((NULL != endptr) && (endptr != value)) {
--- a/rcl_yaml_param_parser/src/parse.c
+++ b/rcl_yaml_param_parser/src/parse.c
@@ -18,6 +18,12 @@
 #include <stdlib.h>
 #include <string.h>
 
+#ifdef _WIN32
+#include <windows.h>
+#else
+#include <threads.h>
+#endif
+
 #include <yaml.h>
 
 #include "rcutils/allocator.h"
@@ -780,6 +786,32 @@ _get_int_value(
   return RCUTILS_RET_ERROR;
 }
 
+// Initialization of c_locale, used by strtod_locale_independent.
+#ifdef _WIN32
+static _locale_t c_locale = NULL;
+static INIT_ONCE c_locale_once_flag = INIT_ONCE_STATIC_INIT;
+
+BOOL CALLBACK init_c_locale(
+  PINIT_ONCE init_once,
+  PVOID parameter,
+  PVOID *context)
+{
+  (void)init_once;
+  (void)parameter;
+  (void)context;
+  c_locale = _create_locale(LC_NUMERIC, "C");
+  return TRUE;
+}
+#else
+static locale_t c_locale = 0;
+static once_flag c_locale_once_flag = ONCE_FLAG_INIT;
+
+static void init_c_locale()
+{
+  c_locale = newlocale(LC_NUMERIC_MASK, "C", 0);
+}
+#endif
+
 ///
 /// Calls strtod with the default "C" locale.
 /// \param[in] nptr the string to parse.
@@ -791,30 +823,24 @@ _get_int_value(
 static double strtod_locale_independent(const char *restrict nptr, char **restrict endptr)
 {
 #ifdef _WIN32
-  static _locale_t c_locale = NULL;
+  InitOnceExecuteOnce(&c_locale_once_flag, init_c_locale, NULL, NULL);
 
   if (NULL == c_locale) {
-    c_locale = _create_locale(LC_NUMERIC, "C");
-    if (NULL == c_locale) {
-      if (NULL != endptr) {
-        *endptr = (char *)nptr;
-      }
-      return 0.;
+    if (NULL != endptr) {
+      *endptr = (char *)nptr;
     }
+    return 0.;
   }
 
   return _strtod_l(nptr, endptr, c_locale);
 #else
-  static locale_t c_locale = 0;
+  call_once(&c_locale_once_flag, init_c_locale);
 
   if (0 == c_locale) {
-    c_locale = newlocale(LC_NUMERIC_MASK, "C", 0);
-    if (0 == c_locale) {
-      if (NULL != endptr) {
-        *endptr = (char *)nptr;
-      }
-      return 0.;
+    if (NULL != endptr) {
+      *endptr = (char *)nptr;
     }
+    return 0.;
   }
 
   locale_t old_locale = uselocale(c_locale);
--- a/rcl_yaml_param_parser/src/parse.c
+++ b/rcl_yaml_param_parser/src/parse.c
@@ -781,17 +781,26 @@ _get_int_value(
 }
 
 ///
-/// Calls strtod with the default "POSIX/C" locale.
+/// Calls strtod with the default "C" locale.
 ///
-static double strtod_posix_locale(const char *restrict nptr, char **restrict endptr)
+static double strtod_locale_independent(const char *restrict nptr, char **restrict endptr)
 {
-    static locale_t c_locale = 0;
-    if (c_locale == 0)
-        c_locale = newlocale(LC_ALL_MASK, "C", 0);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: rcl_yaml_param_parser/src/parse.c [hunk 1]
```diff
@@ -779,6 +780,20 @@ _get_int_value(
...
+static double strtod_posix_locale(const char *restrict nptr, char **restrict endptr)
+{
+    static locale_t c_locale = 0;
+    if (c_locale == 0)
+        c_locale = newlocale(LC_ALL_MASK, "C", 0);
+    locale_t old_locale = uselocale(c_locale);
+    double result = strtod(nptr, endptr);
+    uselocale(old_locale);
+    return result;
+}
```
fujitatomoya: this does not work on windows, does it? we should probably insert `ifdef` pre-processor for windows with `_create_locale(LC_NUMERIC, "C")` and `_strtod_l()`?

rcl_yaml_param_parser/src/parse.c
```diff
@@ -779,6 +780,20 @@ _get_int_value(
...
 }
 
+///
+/// Calls strtod with the default "POSIX/C" locale.
+///
+static double strtod_posix_locale(const char *restrict nptr, char **restrict endptr)
+{
+    static locale_t c_locale = 0;
+    if (c_locale == 0)
+        c_locale = newlocale(LC_ALL_MASK, "C", 0);
```
fujitatomoya: it would be enough with `LC_NUMERIC_MASK`?

```suggestion
        c_locale = newlocale(LC_NUMERIC_MASK, "C", 0);
```

rcl_yaml_param_parser/test/test_parse.cpp
```diff
@@ -153,6 +155,55 @@ TEST(TestParse, parse_value) {
   event.data.scalar.tag = NULL;
 }
 
+TEST(TestParse, parse_value_locale_independent) {
+  if (!std::setlocale(LC_NUMERIC, "fr_FR.UTF-8") && !std::setlocale(LC_NUMERIC, "de_DE.UTF-8"))
+  {
+    GTEST_SKIP() << "Could not set LC_NUMERIC to FR or DE";
+    return;
```
fujitatomoya: redundant code, no need to return here.

```suggestion
```

fujitatomoya:
@Barry-Xu-2018 can you review this?

rcl_yaml_param_parser/src/parse.c
Hugal31: Technically I could set `0` instead of an `LC_*_MASK`, since the non-specified categories are filled with the "C" locale. But yes, this would better express what we want.

rcl_yaml_param_parser/src/parse.c
Hugal31: I don't know why I assumed POSIX functions would be available on windows.

rcl_yaml_param_parser/src/parse.c
```diff
@@ -779,6 +780,29 @@ _get_int_value(
   return RCUTILS_RET_ERROR;
 }
 
+///
+/// Calls strtod with the default "C" locale.
+///
+static double strtod_locale_independent(const char *restrict nptr, char **restrict endptr)
+{
+#ifdef __WIN32
```
fujitatomoya: should be below? (single underscore)

```suggestion
#ifdef _WIN32
```

rcl_yaml_param_parser/src/parse.c
```diff
@@ -779,6 +780,29 @@ _get_int_value(
...
+  static _locale_t c_locale = 0;
+  if (c_locale == 0) {
+    c_locale = _create_locale(LC_NUMERIC, "C");
+  }
+  return _strtod_l(nptr, endptr, c_locale);
+#else
+  static locale_t c_locale = 0;
+  if (c_locale == 0) {
+    c_locale = newlocale(LC_NUMERIC_MASK, "C", 0);
```
fujitatomoya: on error, newlocale() can return (locale_t) 0. e.g ENOMEM
i think we need to add the check on the return value from newlocale() to avoid silently falling back to the buggy behavior.

fujitatomoya (commented):
almost lgtm, just a couple of minor comments.

fujitatomoya (approved):

fujitatomoya:
Pulls: ros2/rcl#1292
Gist: https://gist.githubusercontent.com/fujitatomoya/e7654aaffd0c22d27490e902aa7178d6/raw/00b06b4accb04ea65bd926ea1a9b4027fe0aa9dc/ros2.repos
BUILD args:  --packages-above-and-dependencies rcl_yaml_param_parser
TEST args:  --packages-above rcl_yaml_param_parser
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18318
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27817/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20837/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7680/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26804/)

fujitatomoya:
@Barry-Xu-2018 i will leave this to you.

rcl_yaml_param_parser/src/parse.c
```diff
@@ -779,6 +780,50 @@ _get_int_value(
   return RCUTILS_RET_ERROR;
 }
 
+///
+/// Calls strtod with the default "C" locale.
+/// \param[in] nptr the string to parse.
+/// \param[out] endptr if not NULL, the location to store the end of the parsed string.
+/// \return The parsed value if something was parsed.
+/// \return 0. if no conversion could be performed.
+/// \return 0. if an error occured, errno will be set, and if not NULL, endptr will be set to nptr.
```
Barry-Xu-2018, line 821: Is it possible that a valid output is 0. ？
It’s better to use RCUTILS_RET_XXX as the return value.

rcl_yaml_param_parser/src/parse.c
```diff
@@ -779,6 +780,50 @@ _get_int_value(
...
+///
+static double strtod_locale_independent(const char *restrict nptr, char **restrict endptr)
+{
+#ifdef _WIN32
+  static _locale_t c_locale = NULL;
```
Barry-Xu-2018: This function may be called simultaneously by multiple threads.  
Multiple threads may compete when accessing it for the first time.

rcl_yaml_param_parser/src/parse.c
Hugal31: Yes, 0. is a possible valid value. I am just mimicking strtod own behavior and [man page](https://man7.org/linux/man-pages/man3/strtod.3.html). I chose to stick to strtod original interface.
Just as with plain strtod, you can distinguish an invalid input string from 0. by looking at the value of endptr after calling strtod, which we are. errors also set errno.

rcl_yaml_param_parser/src/parse.c
Hugal31: I fixed it using once_flag / INIT_ONCE.

Barry-Xu-2018 (approved):
LGTM. 
I will re-run CI.

Barry-Xu-2018:
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18333

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27838/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20852/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7701/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26821/)

Barry-Xu-2018:
- Linux-rhel
    The failure is because of CI internal errors. So I re-run it.
    * [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7726/)
- Windows
   The failure cases are unrelated to this PR.

christophebedard:
@Barry-Xu-2018 looks like the RHEL rerun passed, can we merge this?

fujitatomoya:
@Hugal31 thanks for the patch!

PR #1287, opened 2026-01-21
Add 2 interfaces for configuring action client feedback subscription contents filter

Barry-Xu-2018:
## Description

Use Content Filtering for action feedback topic on action clients #1258  
This new interface will be called in rclcpp/rclpy while action client gets Goal ID.

### Is this user-facing behavior change?

No. 

### Did you use Generative AI?

Use github copilot (GPT-5.1-Codex) for test codes

### Additional Information

PR #1301, opened 2026-03-18
improve rcl_action tests failures non-default RMW impls.

fujitatomoya:
## Description

different approach from https://github.com/ros2/rmw_cyclonedds/pull/565

this PR fixes the following failures with `rmw_cyclonedds_cpp`,

```console
export RMW_IMPLEMENTATION=rmw_cyclonedds_cpp
colcon test --event-handlers console_direct+ --packages-select rcl_action
...
     3 FAILED TESTS
    -- run_test.py: return code 1
    -- run_test.py: inject classname prefix into gtest result file '/root/ros2_ws/colcon_ws/build/rcl_action/test_results/rcl_action/test_wait.gtest.xml'
    -- run_test.py: verify result file '/root/ros2_ws/colcon_ws/build/rcl_action/test_results/rcl_action/test_wait.gtest.xml'
  >>>
build/rcl_action/test_results/rcl_action/test_action_client.gtest.xml: 11 tests, 0 errors, 1 failure, 0 skipped
- rcl_action.TestActionClientBaseFixture test_action_client_init_fini (/root/ros2_ws/colcon_ws/src/ros2/rcl/rcl_action/test/rcl_action/test_action_client.cpp:87)
  <<< failure message
    /root/ros2_ws/colcon_ws/src/ros2/rcl/rcl_action/test/rcl_action/test_action_client.cpp:183
    Value of: rcutils_error_is_set()
      Actual: true
    Expected: false
  >>>
build/rcl_action/test_results/rcl_action/test_action_server.gtest.xml: 22 tests, 0 errors, 1 failure, 0 skipped
- rcl_action.TestActionServer test_set_internal_services_introspection_contents (/root/ros2_ws/colcon_ws/src/ros2/rcl/rcl_action/test/rcl_action/test_action_server.cpp:896)
  <<< failure message
    /root/ros2_ws/colcon_ws/src/ros2/rcl/rcl_action/test/rcl_action/test_action_server.cpp:287
    Value of: get_publisher_count(get_result_service_event_topic_name) == expect_publisher_count
      Actual: false
    Expected: true
  >>>
build/rcl_action/test_results/rcl_action/test_wait.gtest.xml: 6 tests, 0 errors, 3 failures, 0 skipped
- rcl_action.TestActionClientWait test_wait_set_add_action_client (/root/ros2_ws/colcon_ws/src/ros2/rcl/rcl_action/test/rcl_action/test_wait.cpp:153)
  <<< failure message
    /root/ros2_ws/colcon_ws/src/ros2/rcl/rcl_action/test/rcl_action/test_wait.cpp:64
    Value of: rcutils_error_is_set()
      Actual: true
    Expected: false
    rmw_subscription_get_content_filter: unimplemented, at /root/ros2_ws/colcon_ws/src/ros2/rmw_cyclonedds/rmw_cyclonedds_cpp/src/rmw_node.cpp:3255
  >>>
- rcl_action.TestActionClientWait test_client_wait_set_get_num_entities (/root/ros2_ws/colcon_ws/src/ros2/rcl/rcl_action/test/rcl_action/test_wait.cpp:362)
  <<< failure message
    /root/ros2_ws/colcon_ws/src/ros2/rcl/rcl_action/test/rcl_action/test_wait.cpp:64
    Value of: rcutils_error_is_set()
      Actual: true
    Expected: false
    rmw_subscription_get_content_filter: unimplemented, at /root/ros2_ws/colcon_ws/src/ros2/rmw_cyclonedds/rmw_cyclonedds_cpp/src/rmw_node.cpp:3255
  >>>
- rcl_action.TestActionClientWait test_client_wait_set_get_entities_ready (/root/ros2_ws/colcon_ws/src/ros2/rcl/rcl_action/test/rcl_action/test_wait.cpp:527)
  <<< failure message
    /root/ros2_ws/colcon_ws/src/ros2/rcl/rcl_action/test/rcl_action/test_wait.cpp:64
    Value of: rcutils_error_is_set()
      Actual: true
    Expected: false
    rmw_subscription_get_content_filter: unimplemented, at /root/ros2_ws/colcon_ws/src/ros2/rmw_cyclonedds/rmw_cyclonedds_cpp/src/rmw_node.cpp:3255
  >>>
```

this PR just call `rcutils_reset_error` right after `rmw_subscription_get_content_filter` during the construction of subscription. this is being done in the ROS 2 system library, not issued by user application by purpose. so we should reset the error strings to avoid leaking a stale error state out of subscription init.

this can fix the many warnings at once. the thing is that, some unit tests are only running with default rmw implementation which is rmw_fastrtps that supports content filtering feature. so those warnings are not observed before unless specific rmw configuration is set to start the test. 

part of https://github.com/ros2/rmw_cyclonedds/pull/550

No

No

### DIFF
--- a/rcl/src/rcl/subscription.c
+++ b/rcl/src/rcl/subscription.c
@@ -145,6 +145,10 @@ rcl_subscription_init(
   if (rmw_subscription_get_content_filter(
     subscription->impl->rmw_handle, NULL, NULL) == RMW_RET_UNSUPPORTED)
   {
+    // depending on rmw implementation, this call may set an error string (e.g. "unimplemented")
+    // so we must call rcutils_reset_error() in both branches to avoid leaking a stale error
+    // state out of subscription init.
+    rcutils_reset_error();
     subscription->impl->rmw_handle->is_cft_supported = false;
   } else {
     rcutils_reset_error();


rcl/src/rcl/subscription.c
```diff
@@ -145,6 +145,10 @@ rcl_subscription_init(
   if (rmw_subscription_get_content_filter(
     subscription->impl->rmw_handle, NULL, NULL) == RMW_RET_UNSUPPORTED)
   {
+    // depending on rmw implementation, this call may set an error string (e.g. "unimplemented")
+    // so we must call rcutils_reset_error() in both branches to avoid leaking a stale error
+    // state out of subscription init.
+    rcutils_reset_error();
```
fujitatomoya, line 151: i think this makes sense because this is being called in ROS 2 system, not by user application. and eventually construction of subscription succeeds even if rmw implementation does not support content filtering feature. so we should reset the error strings here.

if the user application calls `rmw_subscription_get_content_filter` on purpose, the error strings will stay and user application expects the message `unimplemented...`.

fujitatomoya:
@Barry-Xu-2018 can you review this?

CC: @jmachowinski @mjcarroll

fujitatomoya:
Pulls: ros2/rcl#1301
Gist: https://gist.githubusercontent.com/fujitatomoya/6bb54b81b5fdcbbcee37ff602365fefd/raw/d500630426712eaaa52dfa5f7168334d18a734dd/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18533
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28104/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21075/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7958/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27041/)

fujitatomoya:
Pulls: ros2/rcl#1301
Gist: https://gist.githubusercontent.com/fujitatomoya/9edae6ef089f4083cdcaf311cc386db9/raw/d500630426712eaaa52dfa5f7168334d18a734dd/ros2.repos
BUILD args:  --packages-above-and-dependencies rcl --cmake-args -DRMW_IMPLEMENTATION=rmw_cyclonedds_cpp
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18534
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28105/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21076/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7955/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27042/)

fujitatomoya:
i started 2 CI with/without `--cmake-args -DRMW_IMPLEMENTATION=rmw_cyclonedds_cpp`.

rcl/src/rcl/subscription.c
jmachowinski: uhm, just move it out of the if/else entirely ? 

Anyway, I am confused about what is going on here, why are we calling this in the constructor in the first place and not in rcl_subscription_is_cft_supported ?

Also I can't find any place, were rcl_subscription_is_cft_supported is actually called in the whole stack.

jmachowinski (changes requested):

jmachowinski:
I just went a bit over the code and directly found more stuff that is wrong here.

https://github.com/ros2/rcl/blob/c5fe2bf0a57502462d5717b0fd2e4072e6b55c63/rcl/src/rcl/subscription.c#L488

Never returns true, even if content filtering is actually active.

@iuhilnehc-ynos can you have a look at this ?

rcl/src/rcl/subscription.c
fujitatomoya: i think we need to call cft API to see if that is supported or not at the construction.

> Also I can't find any place, were rcl_subscription_is_cft_supported is actually called in the whole stack.

this is planned to use upper layer for improvement, https://github.com/ros2/rcl/pull/1287 and https://github.com/ros2/rclcpp/pull/3034

fujitatomoya:
@jmachowinski 

> Never returns true, even if content filtering is actually active.

if the content filtering expression and parameters are not set (means using content filtering), this returns false. that is expected behavior.

> @iuhilnehc-ynos can you have a look at this ?

unfortunately he is not working with us anymore... i do not think we can get reply from him...

jmachowinski:
> if the content filtering expression and parameters are not set (means using content filtering), this returns false. that is expected behavior.

There is no place, were is_cft_enabled is ever set to true, this looks strange. Also there is no test that it ever returns true, only tests for false.

fujitatomoya:
hmmm that is strange, true case should be tested with cft supported rmws... @Barry-Xu-2018 can you check on this?

for this PR, i am okay with either way. keeping the caching `is_cft_supported` data via construction so that it can return the true/false on next user call immediately, or as @jmachowinski suggests we can do it lazily on the 1st call of rcl_subscription_is_cft_supported and the cache the result. @Barry-Xu-2018 what do you think?

Barry-Xu-2018:
> Anyway, I am confused about what is going on here, why are we calling this in the constructor in the first place and not in rcl_subscription_is_cft_supported ?

Whether an RMW implementation supports content filters is fixed, so I think it should be determined during subscription initialization. The upper-level rclcpp/rclpy can directly provide an is_cft_support() function, instead of attempting to configure a content filter and only discovering that the RMW implementation does not support it when it returns "unsupported." This approach makes the code clearer. 

The original logic has already been optimized based on this way.
e.g.
https://github.com/ros2/rclcpp/blob/df2ac887ed518e704a3c4150a6041e16542a6d7a/rclcpp/src/rclcpp/parameter_event_handler.cpp#L77-L82
```
bool
ParameterEventHandler::configure_nodes_filter(const std::vector<std::string> & node_names)
{
  if (!event_subscription_->is_cft_supported()) {
    return false;
  }
```

If using the lazy approach, two functions need to be modified. When the RMW implementation does not support content filters and their RMW functions return "unsupported", we need to set is_cft_supported accordingly.
- rcl_subscription_set_content_filter() 
- rcl_subscription_get_content_filter()
So I prefer to set it up during the initialization of the subscription.

rcl/src/rcl/subscription.c
Barry-Xu-2018: > uhm, just move it out of the if/else entirely ?

The same suggestion.

Barry-Xu-2018:
@jmachowinski 

> I just went a bit over the code and directly found more stuff that is wrong here.
> 
> https://github.com/ros2/rcl/blob/c5fe2bf0a57502462d5717b0fd2e4072e6b55c63/rcl/src/rcl/subscription.c#L488
> 
> Never returns true, even if content filtering is actually active.

Here is my understanding.
```c
bool
rcl_subscription_is_cft_enabled(const rcl_subscription_t * subscription)
{
  if (!rcl_subscription_is_valid(subscription)) {
    return false;
  }
  return subscription->impl->rmw_handle->is_cft_enabled;
}
```
`subscription->impl->rmw_handle->is_cft_enabled` is set in rmw-xxx  while call configure content filter function.
In rmw_fastdds_cpp
```c++
rmw_ret_t
rmw_subscription_set_content_filter(
  rmw_subscription_t * subscription,
  const rmw_subscription_content_filter_options_t * options)
{
  RMW_CHECK_ARGUMENT_FOR_NULL(subscription, RMW_RET_INVALID_ARGUMENT);
  RMW_CHECK_ARGUMENT_FOR_NULL(options, RMW_RET_INVALID_ARGUMENT);
  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
    subscription,
    subscription->implementation_identifier,
    eprosima_fastrtps_identifier,
    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
  rmw_ret_t ret = rmw_fastrtps_shared_cpp::__rmw_subscription_set_content_filter(
    subscription, options);
  auto info = static_cast<const CustomSubscriberInfo *>(subscription->data);
  subscription->is_cft_enabled = (info && info->filtered_topic_);   <<=== Here
  return ret;
}
```
Even if `rmw_subscription_set_content_filter()` returns successfully, it does not mean that the CFT has already taken effect. You still need to call rcl_subscription_is_cft_enabled() to determine whether the internal CFT of the RMW implementation is enabled.

Barry-Xu-2018:
@fujitatomoya 

> hmmm that is strange, true case should be tested with cft supported rmws... @Barry-Xu-2018 can you check on this?

Actually, it was tested.  

Use `test_subscription_content_filtered` as an example,

https://github.com/ros2/rcl/blob/c5fe2bf0a57502462d5717b0fd2e4072e6b55c63/rcl/test/rcl/test_subscription.cpp#L937-L974

You’ll notice that the test doesn’t directly check the return value of `rcl_subscription_is_cft_enabled()`, but instead conducts different tests based on its return value. That is, even if the rmw implementation does not support CFT, there is a corresponding branch to handle it.

jmachowinski:
@Barry-Xu-2018 Thanks for the explanation on where and how is_cft_enabled is set and manipulated.

This now opens new question for me :
If the mechanism is, that the rmw_create_subscription creates and fills in the handle correctly, why are we manipulating it in the first place in the rcl layer ?

E.g.:
```c++
  // Check if the subscription supports content filtering
  // If RMW implementation doesn't support content filtering, calling
  // rmw_subscription_get_content_filter() with unavailable parameters will always return
  // RMW_RET_UNSUPPORTED.
  if (rmw_subscription_get_content_filter(
    subscription->impl->rmw_handle, NULL, NULL) == RMW_RET_UNSUPPORTED)
  {
    subscription->impl->rmw_handle->is_cft_supported = false;
  } else {
    rcutils_reset_error();
    subscription->impl->rmw_handle->is_cft_supported = true;
  }
``` 
This code block should not be present at all, 
```c
subscription->impl->rmw_handle->is_cft_supported
``` 
must contain the correct value after rmw_create_subscription. Anything else I would consider a bug in the rmw implementation.

Barry-Xu-2018:
@jmachowinski 

> This now opens new question for me :
> If the mechanism is, that the rmw_create_subscription creates and fills in the handle correctly, why are we manipulating it in the first place in the rcl layer ?

You’re right, this value really should be set in rmw_xxx.

At the time when we designed the RMW CFT interface, we indeed did not consider the existence of the variable subscription->impl->rmw_handle->is_cft_supported (I added this value recently. ros2/rmw#415). Now, it would be quite troublesome to require each rmw_xxx to modify and fill in this value themselves, so I used a workaround: during initialization in rcl, `subscription->impl->rmw_handle->is_cft_supported` is configured based on whether calling the  RMW CFT interface returns “unsupported.” This way, we only need to modify one place.

jmachowinski:
I would argue that the design is now broken. 
You have a variable that is in the RMW handle, that is modified in the RCL layer. 
If it is supposed to be some RCL internal thing / is basically owned by RCL, it should be 
in the rcl subscription handle. Anyway, I also dislike the workaround of calling the rmw_subscription_get_content_filter and guessing the support from the result. It sound 
way more straight forward to me, to patch the RWM layer to set the correct value.

And, as only fastDDS supports this anyway, we only need to do one patch. For the rest it should 
default to false anyway.

fujitatomoya:
@jmachowinski @Barry-Xu-2018 thanks for the comments.

so after all, i agree that we should put that in rmw implementation, not in rcl. with that, we do not even need this PR i believe.
@Barry-Xu-2018 can you address that fix?

> as only fastDDS supports this anyway, we only need to do one patch. For the rest it should default to false anyway.

RTI ConnextDDS also supports this.

Barry-Xu-2018:
@jmachowinski 

> I would argue that the design is now broken.
> You have a variable that is in the RMW handle, that is modified in the RCL layer.
If it is supposed to be some RCL internal thing / is basically owned by RCL, it should be
in the rcl subscription handle. Anyway, I also dislike the workaround of calling the rmw_subscription_get_content_filter and guessing the support from the result. It sound
way more straight forward to me, to patch the RWM layer to set the correct value.
>
> And, as only fastDDS supports this anyway, we only need to do one patch. For the rest it should
default to false anyway.

Thank you for your comments. I will make changes to rmw_fastdds/rmw_connext to remove this workaround at the RCL layer.

Barry-Xu-2018:
> so after all, i agree that we should put that in rmw implementation, not in rcl. with that, we do not even need this PR i believe.

I'll handle it.

Barry-Xu-2018:
@jmachowinski @fujitatomoya 

PRs have been created.

- rmw_fastrtps_cpp https://github.com/ros2/rmw_fastrtps/pull/869 
- rmw_connext https://github.com/ros2/rmw_connextdds/pull/223
- rcl #1304

jmachowinski:
Closing this PR in favor for https://github.com/ros2/rcl/pull/1304

PR #1319, opened 2026-05-14
Add support for building with micro-ROS

Amronos:
## Description

This is an upstreaming (with some changes) of the various patches that have been getting maintained in https://github.com/micro-ROS/rcl for a few years.
This should prevent us from having to resolve conflicts every ROS 2 release, and make it possible to always deliver the latest `rcl` without having to update the fork every so often.
This mainly uses a new CMake option called `RCL_MICROROS` is used with `#ifdef` and `#ifndef` to modify the `rcl` code to work with micro-ROS. 

No

### Additional Information

This has been tested in https://github.com/micro-ROS/micro_ros_espidf_component/pull/323.

cc: @EugenioCollado @Narukara


fujitatomoya (commented):
@Amronos 

this is not one-time API agreement, the question that we are really asking here is that if we support micro-ROS rcl in ROS 2 mainline. i just do not know the current discussion or consensus for that, do we have on-going discussion or agreement about this? IMO this is going to be better maintenance for micro-ROS, but it would be easy enough to break micro-ROS as rcl grows because most of maintainers do not know the micro-ROS requirement. to avoid this situation, CI/CD pipeline to make sure before that. that brings me back to the original question above.

CC: @mjcarroll

Amronos:
Thanks for the quick reply, @fujitatomoya!
Sorry, I should have talked more about this in the original post.
We have nightly CI in the micro-ROS repositories that runs all our tests and builds on the latest ROS 2 sources.
The plan is that whenever something breaks, we will make a PR to this repository to fix it. It shouldn't be a problem for end-users as the breaking changes would usually be on `rolling`.
The overall goal of this PR is that we would no longer need to maintain the fork for newer ROS 2 distributions and fix conflicts regularly.

PR #1323, opened 2026-07-11
Treat bare YAML nan and inf as strings

### LINKED ISSUE #1320: rcl_yaml_param_parser treats bare nan/inf as doubles via fallback (inconsistent with composable node parameter loading)
### Operating System:

`Linux 6.8.0-111-generic #111-Ubuntu` 

Kilted

rmw_cyclonedds_cpp

### Steps to reproduce issue

We produced a minimal working example you can pull and run to see the difference in parsing behaviour between composed/standalone nodes here: https://github.com/botsandus/yaml_nan_type_bug

The bare basics of what the above is doing is as follows: 

1. We ceate a parameter yaml containing a bare nan value:

```yaml
node_name:
  ros__parameters:
    some_param: nan
```

2. We create a node that declares a parameter and reads it as a double:
```cpp
node->declare_parameter("some_param", 0.0,
    rcl_interfaces::msg::ParameterDescriptor().set__dynamic_typing(true));
auto val = node->get_parameter("some_param").as_double();
```

3. Launch it as a standalone node with --params-file:
```bash
ros2 launch yaml_nan_type_bug standalone.launch.py
```
The node starts successfully. The parameter is loaded as a double (NaN).

4. Launch the same node as a composable node via LoadComposableNodes into a component_container:
```bash
ros2 launch yaml_nan_type_bug composed.launch.py
```
The node throws `InvalidParameterTypeException: parameter 'some_param' has invalid type: expected [double] got [string]` and the container crashes.

### Expected behavior

We would expect that both  launch methods should produce the same parameter type for the same yaml input. The nan/inf should be treated consistently, either as a double in both paths or as a string in both paths.

We discovered that in the yaml 1.1 spec, only dot-prefixed forms (like `.nan`, `.NaN`, `.NAN`, `.inf`, `.Inf`, `.INF`) are valid float literals, and bare nan is not a valid yaml float.

### Actual behavior

- **Standalone** node: `rcl_yaml_param_parser` parses bare `nan` as a double via `strtod()` fallback
- **Composable** node: `launch_ros` uses `yaml.safe_load()` (part of pyyaml), which correctly treats bare nan as a string per the yaml 1.1 spec.

This means migrating a node from standalone to composable can silently change parameter types, causing crashes or behavioural differences. This is how we discovered the issue originally 🙀 

### Additional information

**Root cause**: In `rcl_yaml_param_parser/src/parse.c`, the float detecton in `get_value()` explicitly checks for yaml 1.1 dot-prefixed float forms, then falls through to `strtod()` for all other values. This makes the C parser accidentally more permissive than we might expect from the spec

**How the two paths are different**: Standalone nodes load parameters via --params-file, parsed by the C-level `rcl_yaml_param_parser` within the node process. Composable nodes can't do this because the container process is already running, so instead, `launch_ros` reads the yaml using `yaml.safe_load()`, and serialises the the values into the `composition_interfaces/srv/LoadNode` service request, and sends to the container. These two parsers disagree on whether bare nan/inf are floats or strings.

We investigated a bit more on this, and we do have a proposed fix, but I wanted to open this issue first so there could be some feedback, but what we think we can do is this:

**Proposed fix**: Reject strtod() results that are nan or inf, since these can only come from bare nan/inf (the valid dot-prefixed forms are already handled above). Approximately 2 lines added around [here](https://github.com/ros2/rcl/blob/7447e7f637910ad169cf3f602476707584b46576/rcl_yaml_param_parser/src/parse.c#L198) in `rcl_yaml_param_parser` we could do this:
```cpp
dval = strtod(value, &endptr);
if (endptr != value && (isnan(dval) || isinf(dval))) {
    endptr = (char *)value;
}
```

Old-Ding:
## Description

`rcl_yaml_param_parser` explicitly handles the YAML 1.1 dot-prefixed non-finite forms such as `.nan` and `.inf`. Its generic `strtod()` fallback also accepted untagged bare `nan`, `inf`, `+inf`, and `-inf`, even though YAML treats those forms as strings.

Reject non-finite results only during implicit type inference, leaving both the existing dot-prefixed forms and explicitly tagged values such as `!!float nan` unchanged. Extend the special-float integration fixture to verify the four implicit bare forms remain strings and the four explicitly tagged forms remain doubles.

Fixes #1320.

Yes. Untagged bare `nan` and `inf` parameter values are now strings, matching YAML and the composable-node parameter loading path. Valid dot-prefixed and explicit `!!float` non-finite values continue to be parsed as doubles.

Yes. OpenAI Codex assisted with root-cause analysis and preparing the focused code and regression test. I reviewed the diff and verification results.

### Additional Information

Local verification:
- a native C probe verified implicit bare non-finite values are rejected as floats, explicit-tag mode accepts them, dot-prefixed forms remain floats, and finite values are unchanged
- PyYAML verification confirmed the fixture has 11 string entries and four explicitly tagged non-finite doubles
- source invariants confirm only the explicit `YAML_FLOAT_TAG` path enables generic non-finite parsing
- all changed lines are at most 100 characters
- UTF-8/CRLF and `git diff --check`

The full `rcl_yaml_param_parser` gtest requires ROS 2 build dependencies that are unavailable on this Windows host; repository CI covers the integration test.

### DIFF
--- a/rcl_yaml_param_parser/src/parse.c
+++ b/rcl_yaml_param_parser/src/parse.c
@@ -15,6 +15,7 @@
 #include <ctype.h>
 #include <errno.h>
 #include <locale.h>
+#include <math.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -138,6 +139,7 @@ _get_int_value(
 /// \param[in] value the float value to get
 /// \param[out] val_type the value type
 /// \param[out] ret_val the converted value when value is valid
+/// \param[in] allow_nonfinite_fallback whether generic non-finite forms are allowed
 /// \param[in] allocator the allocator to use
 /// \return RCUTILS_RET_OK if value is valid, or
 /// \return RCUTILS_RET_ERROR if value is not valid
@@ -147,6 +149,7 @@ _get_float_value(
   const char * const value,
   data_types_t * val_type,
   void ** ret_val,
+  const bool allow_nonfinite_fallback,
   const rcutils_allocator_t allocator);
 
 ///
@@ -213,7 +216,7 @@ void * get_value(
 
     /// Check for float tag
     if (strcmp(YAML_FLOAT_TAG, (char *)tag) == 0) {
-      if (_get_float_value(value, val_type, &ret_val, allocator) != RCUTILS_RET_ERROR) {
+      if (_get_float_value(value, val_type, &ret_val, true, allocator) != RCUTILS_RET_ERROR) {
         return ret_val;
       } else {
         return NULL;
@@ -247,7 +250,7 @@ void * get_value(
     }
 
     /// Check for float
-    if (_get_float_value(value, val_type, &ret_val, allocator) != RCUTILS_RET_ERROR) {
+    if (_get_float_value(value, val_type, &ret_val, false, allocator) != RCUTILS_RET_ERROR) {
       return ret_val;
     }
   }
@@ -855,6 +858,7 @@ _get_float_value(
   const char * const value,
   data_types_t * val_type,
   void ** ret_val,
+  const bool allow_nonfinite_fallback,
   const rcutils_allocator_t allocator)
 {
   errno = 0;
@@ -884,6 +888,10 @@ _get_float_value(
     }
   } else {
     dval = strtod_locale_independent(value, &endptr);
+    // Implicit YAML floats require the dot-prefixed forms handled above.
+    if (!allow_nonfinite_fallback && (isnan(dval) || isinf(dval))) {
+      endptr = (char *)value;
+    }
   }
   if ((0 == errno) && (NULL != endptr)) {
     if ((NULL != endptr) && (endptr != value)) {
--- a/rcl_yaml_param_parser/test/special_float.yaml
+++ b/rcl_yaml_param_parser/test/special_float.yaml
@@ -1,4 +1,5 @@
 test_node:
     ros__parameters:
-        isstring: [string, .nananan, .nAN, .Nan, .infinf, .INf, .infinity]
+        isstring: [string, .nananan, .nAN, .Nan, .infinf, .INf, .infinity, nan, inf, +inf, -inf]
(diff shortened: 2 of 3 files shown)


fujitatomoya (commented):
@Old-Ding the fix itself looks good to me.

seems like you are creating bunch of PRs in a few days.
please have a close look at https://discourse.openrobotics.org/t/guidelines-for-ai-assisted-contributions-in-the-ros-project/55903.

and can you provide the test result for this? either log or screen capture?

fujitatomoya:
> @Old-Ding the fix itself looks good to me.
> 
> seems like you are creating bunch of PRs in a few days. please have a close look at https://discourse.openrobotics.org/t/guidelines-for-ai-assisted-contributions-in-the-ros-project/55903.
> 
> and can you provide the test result for this? either log or screen capture?

@Old-Ding friendly ping.

Old-Ding:
Verification result for the current head: https://github.com/Old-Ding/rcl/actions/runs/29411533231

I ran the target package in an Ubuntu ROS Rolling container:

- `colcon build --packages-select rcl_yaml_param_parser`
- `colcon test --packages-select rcl_yaml_param_parser`
- `test_parse_yaml` passed, including the updated `special_float_point` coverage.
- `colcon test-result --verbose`: 183 tests, 0 errors, 0 failures, 32 skipped.

The full build and test log is linked above.

Old-Ding:
@fujitatomoya Thank you for the reminder. I have read the AI-assisted contribution guidelines. This PR description discloses the use of Codex, and I reviewed the parser paths and the regression coverage before submitting it.

Verification for the current PR source is here: https://github.com/Old-Ding/rcl/actions/runs/29411533231

- Built and tested `rcl_yaml_param_parser` in an Ubuntu ROS Rolling container.
- `test_parse_yaml` passed, including the updated `special_float_point` case.
- `colcon test-result --verbose`: 183 tests, 0 errors, 0 failures, 32 skipped.

The workflow branch contains only its verification workflow in addition to PR head `0bcf6ec`; the source tree under test is otherwise identical. I will continue to keep submissions narrow and include reproducible focused test evidence with them.

Issue #1001, opened 2022-08-18
uninformative crash and error message for invalid parameter override


christianrauch:
## Bug report

Setting an invalid parameter override crashes the process and gives an uninformative error message.

**Required Info:**

- Operating System: Ubuntu 22.04
- Installation type: binaries
- Version or commit hash: 5.3.1
- DDS implementation: default
- Client library (if applicable): rcl

#### Steps to reproduce issue
1. `sudo apt install ros-humble-demo-nodes-cpp`
2. `ros2 run demo_nodes_cpp talker --ros-args -p bla`

#### Expected behavior

An error message or warning informing about the missing value assignment should be shown. I.e. the message should say that `-p bla` is not a valid parameter override, but `-p bla:=5` is.

Even better, the parameter `bla` should be set to uninitialised (`ParameterType::PARAMETER_NOT_SET`).

#### Actual behavior

Crash with:
```
[ERROR] [1660855233.229710370] [rcl]: Failed to parse global arguments
terminate called after throwing an instance of 'rclcpp::exceptions::RCLInvalidROSArgsError'
  what():  failed to initialize rcl: Couldn't parse parameter override rule: '-p bla'. Error: Expected lexeme type (22) not found, search ended at index 3, at ./src/rcl/lexer_lookahead.c:239, at ./src/rcl/arguments.c:343
[ros2run]: Aborted
```

fujitatomoya:
I think that this is because application does not catch the exception from `rclcpp::init` (`rcl_init`)

```bash
# ros2 run demo_nodes_cpp add_two_ints_client 1 2 --ros-args -p bla
[ERROR] [1660931840.357026857] [rcl]: Failed to parse global arguments
terminate called after throwing an instance of 'rclcpp::exceptions::RCLInvalidROSArgsError'
  what():  failed to initialize rcl: Couldn't parse parameter override rule: '-p bla'. Error: Expected lexeme type (22) not found, search ended at index 3, at /root/ros2_ws/colcon_ws/src/ros2/rcl/rcl/src/rcl/lexer_lookahead.c:239, at /root/ros2_ws/colcon_ws/src/ros2/rcl/rcl/src/rcl/arguments.c:343
[ros2run]: Aborted
```

this can be with following patch,

```patch
diff --git a/demo_nodes_cpp/src/services/add_two_ints_client.cpp b/demo_nodes_cpp/src/services/add_two_ints_client.cpp
index 8442801..b0cdf0d 100644
[...]

PR #1324, opened 2026-07-11
Improve missing parameter override error

Old-Ding:
## Description

A parameter override such as `-p bla` currently exposes the internal lexer diagnostic `Expected lexeme type (22)`, which does not tell the user that the assignment syntax is missing.

Translate the separator parse failure at the parameter-rule parser layer into the expected `name:=value` form. Keep the existing outer argument context, return codes, and all other invalid-rule diagnostics unchanged. Add a regression test through the public `rcl_parse_arguments()` entry point.

Addresses the diagnostic portion of #1001.

Yes. Invalid parameter overrides that omit `:=value` now report the expected rule format instead of a lexer token number.

Yes. OpenAI Codex assisted with root-cause analysis and preparing the focused code and regression test. I reviewed the diff and verification results.

### Additional Information

Local verification:
- source-level control-flow assertion verified that only the missing-separator branch replaces the lexer error and retains `RCL_RET_INVALID_PARAM_RULE`
- UTF-8/CRLF validation for both changed files
- `git diff --check`

The native `test_arguments` gtest requires the ROS 2 build dependencies and CMake/colcon tooling, which are not installed on this Windows host; repository CI covers the regression test.

### DIFF
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -1980,6 +1980,8 @@ _rcl_parse_param_rule(
 
   ret = rcl_lexer_lookahead2_expect(&lex_lookahead, RCL_LEXEME_SEPARATOR, NULL, NULL);
   if (RCL_RET_WRONG_LEXEME == ret) {
+    rcl_reset_error();
+    RCL_SET_ERROR_MSG("Parameter override rule must have the format 'name:=value'");
     ret = RCL_RET_INVALID_PARAM_RULE;
     goto cleanup;
   }
--- a/rcl/src/rcl/arguments.c
+++ b/rcl/src/rcl/arguments.c
@@ -1971,21 +1971,18 @@ _rcl_parse_param_rule(
   // TODO(hidmic): switch to _rcl_parse_resource_match() when parameter names
   //               are standardized to use slashes in lieu of dots.
   ret = _rcl_parse_param_name(&lex_lookahead, params->allocator, &param_name);
-  if (RCL_RET_OK != ret) {
-    if (RCL_RET_WRONG_LEXEME == ret) {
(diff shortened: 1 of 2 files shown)


rcl/src/rcl/arguments.c
```diff
@@ -1980,6 +1980,8 @@ _rcl_parse_param_rule(
 
   ret = rcl_lexer_lookahead2_expect(&lex_lookahead, RCL_LEXEME_SEPARATOR, NULL, NULL);
   if (RCL_RET_WRONG_LEXEME == ret) {
+    rcl_reset_error();
```
fujitatomoya, line 1979: this would (acutally does) delete the error messages provided by underlying implementation.
i think what we should do here is capture the error, reset the buffer, and then set a new formatted message that includes the previous error as context.

fujitatomoya (changes requested):

rcl/src/rcl/arguments.c
Old-Ding: Addressed in dd4b6e7. The parser now captures the lexer error before resetting it, then includes that diagnostic as context in the parameter override format message. The regression test checks for both the user-facing guidance and the underlying lexer error.

Old-Ding: Follow-up: the initial revision only covered the separator path. The updated change also routes errors reported by `_rcl_parse_param_name()` through the same context-preserving wrapper.

I verified the updated commit (`ff95ccd`) in a Rolling container: [`test_arguments` passed, and all 52 CTest entries passed with 0 failures](https://github.com/Old-Ding/rcl/actions/runs/29412935004). The test binary was confirmed to load the locally built `/tmp/colcon_ws/build/rcl/librcl.so`.

PR #1335, opened 2026-09-14
Allow specifying an initial timer trigger time

thomasmoore-torc:
# rcl: Allow specifying an initial timer trigger time

**Branch:** `timer-initial-call-time` (this repo: `ros2/rcl`)
**Companion PR:** ros2/rclcpp#3270 — "Allow specifying an initial timer trigger time in `create_timer`/`create_wall_timer`" (built on top of this change)

## Summary

`rcl_timer_init2` (and the deprecated `rcl_timer_init`) always compute a timer's first trigger time as `now() + period`. There's no way to ask for a timer that fires immediately, after a custom one-off delay, or at a specific absolute point in time — callers have to fake it with extra bookkeeping (e.g. a zero-period timer that gets manually reset after the first callback).

This adds `rcl_timer_init3`, which takes an explicit `initial_call_time` (absolute `rcl_time_point_value_t`) alongside `period`. An initial time in the past makes the timer ready immediately, matching the existing "past next-call-time" semantics timers already have after their first trigger.

## Changes

- `rcl_timer_init3(timer, clock, context, initial_call_time, period, callback, allocator, autostart)` — new entry point; `initial_call_time` seeds `next_call_time` directly instead of deriving it from `now() + period`.
- `rcl_timer_init2` is reimplemented as a thin wrapper: it computes `now() + period` and forwards to `rcl_timer_init3`, so its behavior and signature are unchanged for existing callers.
- New tests in `test/rcl/test_timer.cpp`:
  - `test_timer_init3_with_invalid_arguments`
  - `test_timer_init3_honors_future_initial_call_time`
  - `test_timer_init3_with_past_initial_call_time_is_ready`

## Compatibility

Purely additive — no existing signatures change behavior. `rcl_timer_init2`'s outputs are bit-for-bit identical to before for every existing caller.

## Testing

Built and tested against `rolling` in a `ros:rolling-ros-base` container via `colcon build`/`colcon test` for the `rcl` package, plus `uncrustify`/`cpplint` lint targets. All green.

## Landing order

This is a prerequisite for the companion `rclcpp` PR, which calls `rcl_timer_init3` from `rclcpp::TimerBase`/`GenericTimer`/`WallTimer`. That PR can't build without this one merged (or at least available) first — please land this one first, or coordinate merges.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rcl/include/rcl/timer.h
```diff
@@ -180,6 +180,133 @@ rcl_timer_init2(
...
+RCL_PUBLIC
+RCUTILS_DEPRECATED_WITH_MSG("Call rcl_timer_init2 instead")
+rcl_ret_t
+rcl_timer_init(
+  rcl_timer_t * timer,
+  rcl_clock_t * clock,
+  rcl_context_t * context,
+  int64_t period,
+  const rcl_timer_callback_t callback,
+  rcl_allocator_t allocator);
```
fujitatomoya: why do we need to keep this? adding new deprecated function does not really make sense to me.

and another question is that, we would want to TICKTOC to deprecate rcl_timer_init2, makes rcl_timer_init official? tick-tocking init2 out and reinstating rcl_timer_init as the official function is viable now that the old symbol is long gone.

rcl/src/rcl/timer.c
```diff
@@ -109,6 +109,28 @@ rcl_timer_init2(
   const rcl_timer_callback_t callback,
   rcl_allocator_t allocator,
   bool autostart)
+{
+  rcl_time_point_value_t now;
+  rcl_ret_t now_ret = rcl_clock_get_now(clock, &now);
+  if (now_ret != RCL_RET_OK) {
+    return now_ret;  // rcl error state should already be set.
+  }
+  rcl_time_point_value_t initial_call_time = now + period;
```
fujitatomoya: next_call_time can be earlier than last_call_time + period by the delta between the two reads.
normally this can be some nanoseconds, but for an RCL_ROS_TIME clock a sim-time jump between the two reads makes the skew unbounded.

rcl/src/rcl/timer.c
```diff
@@ -234,6 +258,30 @@ rcl_timer_clock(const rcl_timer_t * timer, rcl_clock_t ** clock)
...
 }
 
+/// Advance next_call_time by whole periods, if necessary, until it is after now.
+/**
+ * If next_call_time is already after now, it is returned unchanged.
+ * A period of zero is considered always ready, and now is returned in that case
+ * if next_call_time is not already after now.
+ */
+static int64_t
+_rcl_timer_next_call_time_after(int64_t next_call_time, int64_t period, rcl_time_point_value_t now)
```
fujitatomoya, line 323: if next_call_time is maximum number such as INT64_MAX, this will overflow and UB?

rcl/include/rcl/timer.h
thomasmoore-torc: This was a carry over from porting this change from our Jazzy-based environment to rolling. It has been removed.

rcl/src/rcl/timer.c
thomasmoore-torc: This has been addressed

rcl/src/rcl/timer.c
thomasmoore-torc: This has been addressed

PR #1337, opened 2026-09-18
Fix status message dropped when received before goal response

### LINKED ISSUE #2782: Status is Never Set to Executing if ros2 action rclcpp_action
### Operating System:

Linux tony-laptop 6.11.0-21-generic #21~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Mon Feb 24 16:52:15 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

rolling, jazzy

_No response_

### RMW Configuration (if applicable):

Can be reproduced in a `colcon test` `gtest` where all nodes are in the same process.

### Client library (if applicable):

rclcpp

### 'ros2 doctor --report' output

```Formatted

```

### Steps to reproduce issue

Run the tests in [this PR](https://github.com/ros2/rclcpp/pull/2781) to deterministically reproduce the issue.

Alternatively:
1. Create an action server and action client, ideally with a very short delay before calling `execute()` on the server goal handle. Threading (as in the examples) can lead to faster send of the status message, leading to a higher risk of status arriving before the send goal callback response.
3. Have the action client wait for `Executing` status (e.g. by looping in `get_status` `on the goal handle in a timer callback).
4. The action client will not receive the `Executing` status.

Note that debug message of the form `[DEBUG] [1743106794.846124580] [rclcpp_action.test.client.fibonacci_action_test_client.rclcpp_action]: Received status for unknown goal. Ignoring...` will be printed when the status is received before the callback.

### Expected behavior

The tests in [this PR](https://github.com/ros2/rclcpp/pull/2781) should pass. The aforementioned debug log should not be printed, and the `Executing` status should have been received correctly.

### Actual behavior

The tests fail, `Executing` status is never set. A debug message of the form `[DEBUG] [1743106794.846124580] [rclcpp_action.test.client.fibonacci_action_test_client.rclcpp_action]: Received status for unknown goal. Ignoring...` will be printed when the status is received before the callback.

### Additional information

I believe that the following code can drop status messages in some cases, in `rclcpp_action/include/rclcpp_action/client.hpp`:
```  std::shared_future<typename GoalHandle::SharedPtr>
  async_send_goal(const Goal & goal, const SendGoalOptions & options = SendGoalOptions())
  {
    // Put promise in the heap to move it around.
    auto promise = std::make_shared<std::promise<typename GoalHandle::SharedPtr>>();
    std::shared_future<typename GoalHandle::SharedPtr> future(promise->get_future());
    using GoalRequest = typename ActionT::Impl::SendGoalService::Request;
    auto goal_request = std::make_shared<GoalRequest>();
    goal_request->goal_id.uuid = this->generate_goal_id();
    goal_request->goal = goal;
    this->send_goal_request(
      std::static_pointer_cast<void>(goal_request),
      [this, goal_request, options, promise](std::shared_ptr<void> response) mutable
      {
<snip some code for readability>
        std::shared_ptr<GoalHandle> goal_handle(
          new GoalHandle(goal_info, options.feedback_callback, options.result_callback));
        {
          std::lock_guard<std::recursive_mutex> guard(goal_handles_mutex_);
          goal_handles_[goal_handle->get_goal_id()] = goal_handle;
        }
```

This code adds the `goal_handle` to `goal_handles_` only after success of the `send_goal` service. The code in the `status` callback ignores updates for goals not in `goal_handles_`:
```  void
  handle_status_message(std::shared_ptr<void> message) override
  {
    std::lock_guard<std::recursive_mutex> guard(goal_handles_mutex_);
    using GoalStatusMessage = typename ActionT::Impl::GoalStatusMessage;
    auto status_message = std::static_pointer_cast<GoalStatusMessage>(message);
    for (const GoalStatus & status : status_message->status_list) {
      const GoalUUID & goal_id = status.goal_info.goal_id.uuid;
      if (goal_handles_.count(goal_id) == 0) {
        RCLCPP_DEBUG(
          this->get_logger(),
          "Received status for unknown goal. Ignoring...");
        continue;
      }
```

In the case that the status update for a goal is published before the `send_goal` callback completes, the status update is dropped. I suspect that feedback messages arriving before callback completion are also dropped through a similar mechanism, but I have not tested it yet. I would be happy to help fix this issue and work on additional tests and test cases.

Aaravanand00:
## Description

This PR fixes an issue where action status messages can be dropped if they are received by the client before the goal response has been fully processed. 

The `rcl_action_client_wait_set_get_entities_ready` function previously evaluated the readiness of all client events independently. If a goal response and a status update were both available on the same iteration, they could be processed out of order, leading to the status update being dropped or handled inappropriately before the goal was fully registered by the client.

To fix this, this PR enforces a strict processing priority hierarchy via if-else branching in `rcl_action_client_wait_set_get_entities_ready`. Goal responses are now given the highest priority. If a goal response is ready, it will be the only event marked as ready, intentionally delaying the status message processing until the next executor tick. This ensures the client correctly processes the goal acceptance before ever seeing its status.

This provides the correct underlying fix in `rcl` for the race condition discussed in `ros2/rclcpp#3275` and directly resolves the behavior described in `ros2/rclcpp#2782`.

Fixes ros2/rclcpp#2782

Yes, this improves the reliability of ROS 2 action clients. Users will no longer encounter race conditions where immediate action status updates (such as a rapid transition to the "Executing" state) are skipped or dropped by the client right after sending a goal.

Yes, this pull request was generated with the help of ai agent. The AI generated the logic modifications in `rcl_action_client_wait_set_get_entities_ready` (`rcl_action/src/rcl_action/action_client.c`) to implement the event priority hierarchy.

### Additional Information

This fix was implemented based on the maintainer feedback provided by `@jmachowinski` on `ros2/rclcpp#3275`, who highlighted that the `rcl` layer needs to give out only the correct next ready event to properly reconstruct the event order and avoid workarounds in the `rclcpp` layer.


fujitatomoya (changes requested):
i believe this is incomplete, and not really the fix to address the original issue.
the race is only handled when both events are ready in the same wait iteration, and this is hard requirement... that is said, static priority is not sent order.

Aaravanand00:
> i believe this is incomplete, and not really the fix to address the original issue. the race is only handled when both events are ready in the same wait iteration, and this is hard requirement... that is said, static priority is not sent order.

@fujitatomoya I see your point and I completely agree. Since wait_set only processes events available in the current iteration, setting a static priority here doesn't help if the Status message arrives in a completely different wait iteration before the GoalResponse (which happens due to DDS out-of-order delivery across different topics).

Since rcl cannot peek into the GoalStatusArray without take-ing it, we can't block the Status entity from becoming ready without potentially delaying valid status updates for other already accepted goals. Reconstructing the actual sent order at this layer would likely require modifying action_msgs to include sequence numbers which feels like a significant API change.

IMO, since we can't reliably solve the network out-of-order delivery at the rcl layer without breaking changes, the most practical approach might be to cache the premature status messages in rclcpp_action::Client (as originally proposed in ros2/rclcpp#3275). This ensures no statuses are dropped while waiting for the goal response.

Please correct me if I'm wrong, or let me know if there's a different architectural approach you'd prefer....

